PR #206, opened 2020-08-12
Trotter decomposition for local Hamiltonians

stavros11:
This is a first prototype of how Trotterization could be implemented. The implementation is based on Section 4.1 from [arXiv:1901.05824](https://arxiv.org/abs/1901.05824) where we assume that the Hamiltonian is a sum of two-body terms which we split to odd and even. 

This PR implements a `LocalHamiltonian` class that represents such Hamiltonian and has a `circuit` method which returns a Qibo circuit that is equivalent to a Trotter time step (Eq. (51) from the paper). The circuit contains `gates.Unitary` gates that implement the exponentials.

There might be a few ways to extend the `LocalHamiltonian` class further and I will probably update with a more concrete TODO list later. Once I finalize the implementation I will improve testing and documentation. I already added an example for demonstration.


scarrazza (approved):
Thanks, I agree for all points.

scarrazza:
@stavros11 shall we merge now?

stavros11:
I think we can merge both this and #222 but depending on which will be merged first some conflicts may be created. Also I would add some tests of the gap callback using `TrotterHamiltonian`, which I didn't do because I didn't know if we were going to merge both.
I think we could merge this first since it is earlier and then modify #222 before merging it.

scarrazza:
Perfect, lets do that, so we can have a 0.0.1b3 beta release today.

PR #216, opened 2020-08-21
Adiabatic evolution for the Exact Cover problem.

igres26:
Initial commit for the example of the Exact Cover problem using adiabatic evolution.


stavros11:
@igres26, following our email conversation let me explain a bit more specifically how I would construct the `TrotterHamiltonian`. I managed to construct the problem Hamiltonian successfully, however I am still having some trouble with the easy one (you can find bellow why). I will update when I fix my issue, but until then some detailed explanations:

Let's start with the problem Hamiltonian:
This is a sum of the following term 0.5 * (Z1Z2 + Z2Z3 + Z3Z1 - Z1 - Z2 - Z3) + 1 for every clause. This consists of the two-qubit term 0.5 * (Z1Z2 - Z1) which for a clause `c` appears one for each of the following pairs: `(c[0], c[1]), (c[1], c[2]), (c[2], c[0])`. The corresponding `TrotterHamiltonian` can be defined as follows:
```Python
# Define the two-qubit term as Hamiltonian
m1 = 0.5 * (np.kron(matrices.Z, matrices.Z) - np.kron(matrices.Z, matrices.I))
ham1 = hamiltonians.Hamiltonian(2, m1, numpy=True)

# count the number of times a pair (i, j) appears in the clauses in order to properly normalize
pairs = collections.Counter()
for q0, q1, q2 in clauses:
    pairs[(q0 - 1, q1 - 1)] += 1
    pairs[(q1 - 1, q2 - 1)] += 1
    pairs[(q2 - 1, q0 - 1)] += 1

# define a single part dictionary mapping the term `ham1` to all terms with proper normalization
part1 = {pair: ham1 if n == 1 else n * ham1 for pair, n in pairs.items()}
# define TrotterHamiltonian using a single dictionary and 
# add the constant term that comes for summing all the 1s
H1T = hamiltonians.TrotterHamiltonian(part1) + len(clauses)
```
Some notes regarding this:
* In the email I wrote that a qubit id cannot appear twice in the same dictionary which is (fortunately) **wrong**. The terms appearing in a single dictionary should indeed be commuting but Qibo is not actually checking for this. In this problem a single dictionary should be sufficient because the Hamiltonian contains Z terms only which are commuting anyway. Therefore you also don't need to do any odd/even partition like the example.
* The reason I am counting how many times each pair appears in the clauses is that Python dictionaries have distinct keys. If a pair appears twice or more then the number of times it appears should be multiplied to the corresponding term otherwise it will be treated as if it appeared only once.
* I am using `numpy=True` when defining the two-qubit terms because numpy is faster than Tensorflow for small object operations.
* It is possible to use three (or more) qubit terms in the dictionary, eg. `{(q0, q1, q2): ham}`, however this is not recommended for two reasons: i) The Hamiltonian terms in this dictionary should be exponentiated which is preferrable to do to smaller matrices (such as two-qubit which are 4x4), ii) Custom operator gates (which are fast) support up to two-qubit gates, so if you would like to use three-qubit gates you would have to switch to a Tensorflow backend which are slower.

Regarding the easy Hamiltonian:
This is just a sum of one-qubit terms which in principle should be easy, however there is the following issue: During adiabatic evolution the sum "easy" + "hard" Hamiltonian is constructed and currently the addition of TrotterHamiltonians is implemented only if their dictionaries agree (have the same keys). This means that H0 should be defined using a single dictionary that has the keys of `part1` from above. The way I attempted to do this is the following:
```Python
# construct two-qubit term for h0 (where X acts on qubit 0)
m0 = 0.5 * np.kron(matrices.I - matrices.X, matrices.I)
ham0 = hamiltonians.Hamiltonian(2, m0, numpy=True)
# construct TrotterHamiltonian using a single dictionary which has the same keys as `part1`
part0 = {pair: ham0 for pair in part1.keys()}
H0T = hamiltonians.TrotterHamiltonian(part0)
```
I also think that by doing this you get the `times` normalization automatically because qubit `q` appears in `times[q]` pairs of the `part1` dictionary.

Finally you can check whether the `TrotterHamiltonian` you constructed agrees with the original one by taking its *dense* form and comparing the matrices:
```Python
np.testing.assert_allclose(H0.matrix, H0T.dense.matrix)
np.testing.assert_allclose(H1.matrix, H1T.dense.matrix)
```
If you do this for the above Hamiltonians you will see that `H1` is correct but for `H0` some values disagree (not sure why yet).

stavros11:
I just realized that while the Z terms in the problem Hamiltonian are commuting, when you start mixing it with H0 you get non-commuting terms because of the Xs. Qibo won't have a problem with this, however the Trotter expansion used will be wrong. I think the solution to this is simple though, we just have to split the single dictionary I created above to mutliple ones so that no pairs with the same first qubit appear in the same dictionary.

igres26:
> Regarding the easy Hamiltonian:
> 
> ```python
> # construct two-qubit term for h0 (where X acts on qubit 0)
> m0 = 0.5 * np.kron(matrices.I - matrices.X, matrices.I)
> ham0 = hamiltonians.Hamiltonian(2, m0, numpy=True)
> # construct TrotterHamiltonian using a single dictionary which has the same keys as `part1`
> part0 = {pair: ham0 for pair in part1.keys()}
> H0T = hamiltonians.TrotterHamiltonian(part0)
> ```
> 
> I also think that by doing this you get the `times` normalization automatically because qubit `q` appears in `times[q]` pairs of the `part1` dictionary.
> 
> Finally you can check whether the `TrotterHamiltonian` you constructed agrees with the original one by taking its _dense_ form and comparing the matrices:
> 
> ```python
> np.testing.assert_allclose(H0.matrix, H0T.dense.matrix)
> np.testing.assert_allclose(H1.matrix, H1T.dense.matrix)
> ```
> 
> If you do this for the above Hamiltonians you will see that `H1` is correct but for `H0` some values disagree (not sure why yet).

using:

```python
part0 = {pair: ham0 if n == 1 else n * ham0 for pair, n in pairs.items()}
```

for the easy Hamiltonian seem to solve the issue that the resulting matrices are not the same and solves the `times` issue.

stavros11:
> using:
> 
> ```python
> part0 = {pair: ham0 if n == 1 else n * ham0 for pair, n in pairs.items()}
> ```
> 
> for the easy Hamiltonian seem to solve the issue that the resulting matrices are not the same and solves the `times` issue.

Thanks for finding this, indeed this solves the issue. Just to summarize the Hamiltonians can be constructed as:
```Python
# Construct two-qubit Hamiltonian terms
m0 = 0.5 * np.kron((matrices.I - matrices.X), matrices.I)
m1 = 0.5 * (np.kron(matrices.Z, matrices.Z) - np.kron(matrices.Z, matrices.I))
ham0 = hamiltonians.Hamiltonian(2, m0, numpy=True)
ham1 = hamiltonians.Hamiltonian(2, m1, numpy=True)

# Count the number of times each pair appears in the clauses
pairs = collections.Counter()
for q0, q1, q2 in clauses:
    pairs[(q0 - 1, q1 - 1)] += 1
    pairs[(q1 - 1, q2 - 1)] += 1
    pairs[(q2 - 1, q0 - 1)] += 1

# Define TrotterHamiltonian with proper normalization
part0 = {pair: ham0 if n == 1 else n * ham0 for pair, n in pairs.items()}
part1 = {pair: ham1 if n == 1 else n * ham1 for pair, n in pairs.items()}
H0T = hamiltonians.TrotterHamiltonian(part0)
H1T = hamiltonians.TrotterHamiltonian(part1) + len(clauses)
```
If you use these Hamiltonians in the place of `H0` and `H1` in your code Trotter decomposition will be used automatically for evolution and should be faster. Just make sure to merge the latest master first.

However, as I wrote above I think that the above decomposition is not correct because once H0 and H1 are mixed there will be non-commuting terms in the same dictionary. Qibo will **not** raise any errors for this but the Trotter decomposition used in the evolution will most likely be wrong and will not agree with your current version. In order to fix this we should split the `pairs` dictionary to multiple dictionaries so that qubits are not repeated. This can be done as:

```Python
# Construct two-qubit Hamiltonian terms
m0, m1, ham0, ham1 = ... # same as above

# Count the number of times each pair appears in the clauses
pairs = ... # same as above

# Split `pairs` to multiple dictionaries so that qubits
# in a single dictionary are not repeated
multi_pairs, multi_sets = [], []
for pair, n in pairs.items():
    pair_set = set(pair)
    not_added = True
    for p, s in zip(multi_pairs, multi_sets):
        if not pair_set & s:
            p[pair] = n
            s |= pair_set
            not_added = False
            break
    if not_added:
        multi_pairs.append({pair: n})
        multi_sets.append(set(pair))

# Define TrotterHamiltonian with proper normalization
parts0 = ({pair: ham0 if n == 1 else n * ham0 for pair, n in p.items()}
          for p in multi_pairs)
parts1 = ({pair: ham1 if n == 1 else n * ham1 for pair, n in p.items()}
          for p in multi_pairs)
H0T = hamiltonians.TrotterHamiltonian(*parts0)
H1T = hamiltonians.TrotterHamiltonian(*parts1) + len(clauses)
```
The split done in `multi_pairs` is equivalent to the even/odd split in TFIM, just a bit more complicated because here interactions are not limited to nearest neighbors.

If you like, you can try both the above constructions and check if evolution works as expected. It is easy to switch between them anyway (just remove the`multi_pairs` part). I have not tried but I would guess that the first approach will not work while the second will fix the issue.

One last comment is that when using Trotter Hamiltonians it is useful to define the initial state of the evolution. This way you will not need to diagonalize H0 to find its ground state which is the default initial state and the code will be able to work for more qubits. I see that you define the initial state in `main.py` anyway so you will not have this issue.

igres26:
After some tests, there seems to **not** be a noticeable difference in the final result of the evolution using the different dictionaries, I have tried for up to 16 qubits and the results are really similar.

As for `callbacks` during a Trotter evolution, if the system is too big to carry the full Hamiltonian in memory, the kernel seems to just die after a while. Do we deem the `callbacks.Gap` not possible when a Trotter evolution of more than 16 qubits is performed? How does the gap calculation work when dealing with Trotterized Hamiltonians?

There also seems to be an issue when performing:

```python
np.testing.assert_allclose(H0.matrix, H0T.dense.matrix)
np.testing.assert_allclose(H1.matrix, H1T.dense.matrix)
```

for any number of qubits that is not 8, the H1 does not match exactly with H1T for what seems to be only 1 number. Even though it seems to not affect the final evolution result.

For 10 qubits:

```python
AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 1 / 1048576 (9.54e-05%)
Max absolute difference: 5.32907052e-15
Max relative difference: 1.
 x: array([[ 8.+0.j,  0.+0.j,  0.+0.j, ...,  0.+0.j,  0.+0.j,  0.+0.j],
       [ 0.+0.j,  6.+0.j,  0.+0.j, ...,  0.+0.j,  0.+0.j,  0.+0.j],
       [ 0.+0.j,  0.+0.j,  7.+0.j, ...,  0.+0.j,  0.+0.j,  0.+0.j],...
 y: array([[ 8.+0.j,  0.+0.j,  0.+0.j, ...,  0.+0.j,  0.+0.j,  0.+0.j],
       [ 0.+0.j,  6.+0.j,  0.+0.j, ...,  0.+0.j,  0.+0.j,  0.+0.j],
       [ 0.+0.j,  0.+0.j,  7.+0.j, ...,  0.+0.j,  0.+0.j,  0.+0.j],...
```

For 4 qubits:
```python
AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 1 / 256 (0.391%)
Max absolute difference: 8.8817842e-16
Max relative difference: 1.
 x: array([[ 3.+0.j,  0.+0.j,  0.+0.j,  0.+0.j,  0.+0.j,  0.+0.j,  0.+0.j,
         0.+0.j,  0.+0.j,  0.+0.j,  0.+0.j,  0.+0.j,  0.+0.j,  0.+0.j,
         0.+0.j,  0.+0.j],...
 y: array([[3.000000e+00+0.j, 0.000000e+00+0.j, 0.000000e+00+0.j,
        0.000000e+00+0.j, 0.000000e+00+0.j, 0.000000e+00+0.j,
        0.000000e+00+0.j, 0.000000e+00+0.j, 0.000000e+00+0.j,...
```

stavros11:
> After some tests, there seems to **not** be a noticeable difference in the final result of the evolution using the different dictionaries, I have tried for up to 16 qubits and the results are really similar.

The difference between the two is how exp(-1j dt H) is decomposed to multiple exponentials using the Trotter formula. Both decompositions have an error that depends on dt, however it is supposed to be smaller if the commutation conditions are satisfied. I guess the difference for this example and choice of dt happens to be small.


> As for `callbacks` during a Trotter evolution, if the system is too big to carry the full Hamiltonian in memory, the kernel seems to just die after a while. Do we deem the `callbacks.Gap` not possible when a Trotter evolution of more than 16 qubits is performed? How does the gap calculation work when dealing with Trotterized Hamiltonians?

The gap callback currently diagonalizes the full 2^N x 2^N Hamiltonian matrix to calculate the gap. For Trotter Hamiltonians where the matrix is not available by definition, it automatically creates it by calling `hamiltonian.dense`. This is a temporary solution to make things working for the first release until we implement smarter diagonalization methods that do not calculate all the eigenvalues and/or do not need to construct the full matrix. Therefore what you observe is expected but it will be fixed in the future without requiring any changes in your code. If you don't use callbacks you should be able to do 16 qubits with Trotter as it is using standard Qibo circuits.


> There also seems to be an issue when performing:
> 
> ```python
> np.testing.assert_allclose(H0.matrix, H0T.dense.matrix)
> np.testing.assert_allclose(H1.matrix, H1T.dense.matrix)
> ```
> 
> for any number of qubits that is not 8, the H1 does not match exactly with H1T for what seems to be only 1 number.

This is just precision issue, not important. It happens when using `assert_allclose` for values that are 0. If you specify and absolute tolerance, eg:
`np.testing.assert_allclose(H1.matrix, H1T.dense.matrix, atol=1e-12)`
it should be okay.

It generally sounds that your code is working as expected so whenever you are ready you can push this.

examples/adiabatic-3SAT/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+Adiabatic quantum computation aims to reach the ground state of a problem Hamiltonian encoding the solution of a hard problem, by adiabatically evolving a system from the ground state a known, easy to prepare, Hamiltonian to the problem Hamiltonian.
+
+An Exact Cover instance of a 3SAT problem is characterized by a set of clauses containing 3 bits that are considered satisfied if one of them is in position 1, while the other remain at 0. The solution of this instance is bitstring that fulfills all the clauses at the same time.
+
+## Adiabatic evolution
+
+Adiabatic compution deals with finding the ground state of a Hamiltonian that encodes de solution of a computational problem. This has to be mapped to the time dependent Hamiltonian in the form
+
+![hamiltonian](images/H.png)
```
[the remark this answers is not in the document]
scarrazza: You can use gimp, and perform a resize fixing the aspect ratio. 
I think something around 50px for the weight should work.

scarrazza:
@igres26 concerning test, I think the error we are getting in this last commits is due `solver=rk45`, not sure this is supposed to work.

igres26:
About `rk45`, the error could be in that when `--trotter` is activated, only `solver=exp` is left as an option. I can maybe add a flag that does not let you use trotter and rk45 at the same time?

scarrazza:
Instead of a flag I would suggest a check that raises an error message with the problematic configuration is tested.

stavros11:
Indeed Trotter is only working with the `'exp'` solver, not the RK ones. You can either raise an error in your code or let it raise the error it currently raises (although this may be a bit obscure for someone who doesn't have prior experience with our code).
Regarding the test, we can either not test the problematic configuration by doing
`@pytest.mark.parametrize(("solver", "trotter"), [("exp", True), ("exp", False), ("rk45", False)])`
or change the test to
```Python
if trotter and solver != "exp":
        run_script(args)
else:
    run_script(args)
```

@scarrazza: a note not relevant to this example: I need to investigate why RK fails with the Trotter Hamiltonian here. RK methods are not able to use the Trotter decomposition by definition, however the solver should have worked (and `__matmul__` shouldn't have failed).

examples/adiabatic-3SAT/main.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--nqubits", default=8, type=int)
+    parser.add_argument("--T", default=10, type=float)
+    parser.add_argument("--dt", default=1e-2, type=float)
+    parser.add_argument("--solver", default="exp", type=str)
+    parser.add_argument("--plot", default=True, type=functions.str2bool)
+    parser.add_argument("--trotter", default=True, type=functions.str2bool)
```
stavros11, line 68: If you like to avoid defining this `str2bool` function, for boolean parser arguments you can use `parser.add_argument("--trotter", action="store_true")`. 
If you do so, note that the default value will be `False`. Particularly:
* `python main.py` will not use Trotter
* `python main.py --trotter` will use Trotter.

You can use `"store_false"` if you like the opposite behavior.

examples/adiabatic-3SAT/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+The initial Hamiltonian used in this example is the following, 
+
+![initial-hamiltonian](images/h0.png)
+
+which has a ground state of the equal superposition of all possible computational states.
+
+As for the problem Hamiltonian that encodes the solution of an Exact cover problem, first we define the operator
+
+![z-operator](images/z.png)
```
stavros11, line 27: This image appears broken. I think it should be `z-matrix.png` instead of `z.png`.

stavros11 (commented):
> @scarrazza: a note not relevant to this example: I need to investigate why RK fails with the Trotter Hamiltonian here. RK methods are not able to use the Trotter decomposition by definition, however the solver should have worked (and `__matmul__` shouldn't have failed).

Update: I just fixed this issue and now Trotter Hamiltonians should work with all solvers. No need to do any changes in the code or tests. I think all tests should pass now.

@igres26 thanks for all the updates. Almost everything looks good to me and is working. I would still suggest the following changes:
* Rename this to `adiabatic3sat` to be similar to `grover3sat`.
* Use the same data folder for both examples. You can also upload more instances if you have. This would be useful to benchmark Trotter on more qubits.

stavros11:
Here are some benchmarks of the latest code in this branch (evolution only - no callbacks) using the DGX:

nqubits  | GPU Exp | GPU Trotter | CPU Exp | CPU Trotter
-- | -- | -- | -- | --
4 | 9.689867973 | 2.83823514 | 5.844917297 | 1.973899126
8 | 10.98952818 | 4.748253107 | 80.36308408 | 7.861931086
10 | 51.3461349 | 5.696549177 | 2479.807066 | 9.891137838
12 | 1710.442897 | 6.994450808 | Terminated | 11.62688708
14 | OOM | 7.448305607 |   | 12.6099093
16 |   | 9.139055252 |   | 17.10409141

The default options (dt = 0.01, T=10, complex128) are used in all cases.

scarrazza:
Shall we perform a final review here?

igres26:
I have changed the name of the folders to match the `grover3sat` and `data3sat` in the examples.

For the rest, we could unify this pull request with the optimized schedule one if everything else is in order here.

stavros11 (approved):
Great, thanks for the updates. This looks good to me. Don't worry about the tests failing, I will fix this later today because right now I don't have computer access.

Regarding optimization, I will fix the conflicts (if any) and will try to produce some plots using what you proposed. I will update in the other PR.

scarrazza (approved):
Thanks, looks good also for me.

PR #222, opened 2020-08-23
Gap callback for adiabatic evolution

stavros11:
This implements two features related to the `AdiabaticEvolution` model, which could be used to simplify the code of #216 and potentially other application examples.

1. A `AdiabaticEvolution.hamiltonian` method which can be used to access the time-dependent Hamiltonian of the adiabatic evolution. Currently this is only possible by doing `evolution_model.solver.hamiltonian(t)` which is not very user-friendly and could also lead to errors if it was called before or after the model's execution (due to the way the Hamiltonian is set). Now the `evolution_model.hamiltonian` method raises all the required errors to avoid bad usage.

2. A `callbacks.Gap` callback which can be used to get the eigen-energies and gap during an adiabatic evolution as a function of time. Usage is explained in docs. Note that this uses the `hamiltonian.eigenvalues()` methods so performance will depend on how diagonalization is applied there. In the future we could replace these with sparse methods or methods that find only the first few eigenvalues and are faster than complete diagonalization.

Note that the gap callback will not work directly with the `TrotterHamiltonian` of #206 (assuming we merge both), because currently the `TrotterHamiltonian` does not have the `eigenvalues`/`eigenvectors`. I am planning to temporarily implement these for `TrotterHamiltonian` using just the full matrix and in the future we can think of more efficient ways we can do this.


scarrazza (approved):
Looks good to me.

scarrazza:
@stavros11 shall we merge?

stavros11:
I fixed the conflicts with master and added a test of the gap callback using `TrotterHamiltonian`. I believe this is ready now.

scarrazza:
Thanks, tests are passing on CPU and GPU, lets merge.

PR #226, opened 2020-08-28
Optimizing adiabatic evolution scheduling for the Exact Cover

stavros11:
As we discussed on Wednesday, this extends the 3SAT adiabatic evolution example to include some optimization of the scheduling. I added the following options in the existing `main.py`:
* Possibility to use a polynomial scheduling function by passing a `--params` argument that defines the polynomial coefficients. If this argument is not passed the default linear scheduling is used.
* Possibility to optimize the total evolution time `T` and/or the polynomial coefficients (if `--params` is used) by passing a `--method` argument that specifies the optimization method.

I updated the README to document the above changes properly and also added some example plots for N=10 qubits that show the gap and the optimized scheduling. 
I also plotted the probability to find the correct solution as a function of the total evolution time T to show that the optimized scheduling converges to the solution faster. Note however that for this example the optimized scheduling reaches 99% probability at T=50 while linear takes T=400, so it does not follow the sqrt rule.


igres26:
Everything looks good. The changes in the `README.md` also are very good. 

An optimization of the schedule is a great example. But following this [paper](https://arxiv.org/pdf/quant-ph/0311017.pdf) by Román Orús and José Ignacio Latorre we see that the minimum gap for an Exact Cover problem happens always around the `s=0.7` part. So we could also provide schedule `--params` that achieve a slow-down of the schedule around that part even before applying optimization.

It would be interesting to check how much gain optimized parameters give when compared to this first schedule choice.

As for the rest, we could merge it to the other branch.

scarrazza (approved):
Looks good to me, let's wait for the tests and merge.

PR #227, opened 2020-08-28
QAOA

DiegoGM91:
This PR implements the QAOA in the models.py file.

In order to do so, the evolution.py file is modified so as to allow for negative time evolution.

WARNING: Lacking documentation for the webpage.


DiegoGM91:
Additional comments:

I use the scipy.optimize.minimize instead of the qibo optimizers because, if I understand properly, the latter does not allow to have additional (fixed) arguments in the loss function.

I tried the code out with the XXZ Hamiltonian with a small number of qubits, and it is able to get to the right global minimum.

When using the default exact exponentiation solver, 'exp', combined with the 'Powell' minimization method, instead of the 'rk4' or 'rk45', the QAOA needs a deeper circuit to get to the right minimum. Any clue on why??

stavros11 (commented):
Many thanks for implementing this, it will be a useful addition to the library. Here are a few comments that I think will make this a bit easier to use. If you disagree with something let me know.

Firstly, I would change a few things in the structure of functions and their inputs. More particularly:

* Remove `nqubits` argument because it can be read from the given Hamiltonian as `hamiltonian.nqubits`.

* Move the `solver` argument in `__init__` and define a `self.solver = solver` property there. This way you won't have to pass any fixed arguments to the `scipy.optimize.minimize` call and you could use `qibo.optimize` as we do in our other variational models. This would also allow using the SGD and CMA optimizers (which are not available in scipy).

* Make the list/array of variational parameters an attribute of the QAOA. This will allow the user to execute the model (for example after training) without having to set the parameters every time.

* Implement a `__call__` method that executes the model on a given initial state (similar to what we have for circuits).

* Rename `train` to `minimize` to match what we have in the VQE and the adiabatic evolution (not that `train` is wrong, it is just better to use the same naming conventions everywhere).

To be more explicit, the code after the above changes would look like this (**not tested**):
```Python
class QAOA(object):
    """ This class implements the QAOA algorithm."""
    import numpy as np
    from qibo import hamiltonians

    def __init__(self, hamiltonian, mixer=None, solver="exp"):
        """
        Initializes an instance of the QAOA.

        Args:
            hamiltonian (qibo.hamiltonians.Hamiltonian): problem Hamiltonian whose ground state is sought.
            mixer (qibo.hamiltonians.Hamiltonian): mixer Hamiltonian. If None, it is taken as -sum(sigma_x).
            solver (string): solver employed for StateEvolution.
        """
        # list of QAOA variational parameters (angles)
        self.params = None
        # problem hamiltonian
        self.nqubits = hamiltonian.qubits
        # evolution solver
        self.solver = solver
        # mixer hamiltonian (default = -sum(sigma_x))
        if mixer == None:
            self.mixer = self.hamiltonians.X(nqubits)
        else:
            self.mixer = mixer

    def set_parameters(p):
        self.params = p

    def __call__(self, initial_state=None):
        """Applies the QAOA exponentials to a state."""
        # We define the evolution operators for the mixer and problem hamiltonians
        evolve = []
        for i in range(len(self.params)//2):
            evolve.append(StateEvolution(self.hamiltonian, self.params[2*i], solver=self.solver))
            evolve.append(StateEvolution(self.mixer, self.params[2*i+1], solver=self.solver))
        # We evolve the state
        state = self.get_initial_state(initial_state)
        for i, evolution_operator in enumerate(evolve):
            state = evolution_operator(final_time=self.params[i], initial_state=state)
        return state

    def get_initial_state(self, state=None):
        # FIXME: Cast states to Tensorflow properly
        if state is None:
            return self.np.ones(2**nqubits) / self.np.sqrt(2**nqubits)
        else:
            return initial_state

    def minimize(self, initial_p, initial_state=None, method='Powell'):
        """Optimizes the variational parameters of the QAOA.

        Args:
            initial_p (numpy.array or list): initial guess for the parameters of the QAOA.
            initial_state (np.ndarray): initial state of the QAOA.

        Returns:
            (float, numpy.array) with (minimum found for the problem Hamiltonian, optimal angles)
        """
        # FIXME: Use Qibo optimizers in this method
        from scipy.optimize import minimize
        def loss(p):
            state = self(initial_state)
            return self.np.float64(self.hamiltonian.expectation(state))

        if len(initial_p)%2 != 0:
            raise ValueError('Initial guess for the parameters must contain an even number of values')

        print('Optimizing QAOA...')
        result = minimize(loss, initial_p, method=method)

        return result.fun, result.x
```
Here I removed the `self.initial_state` attribute and made this an argument to `__call__` and `minimize`. If you think it is more useful to fix the initial state using an attribute that's fine by me.

Apart from these structural changes, it seems that the `__call__` method uses the `StateEvolution` models to perform a single time step of size `p[i]`. I think it would be simple if we implemented this using the solvers directly, without having to define `StateEvolution` models. This would look like:
```Python
from qibo import solvers
def __call__(self, initial_state=None):
    state = self.get_initial_state(initial_state)
    for i in range(len(self.params) // 2):
        solver = self.solvers.factory[self.solver](p[2 * i], self.hamiltonian)
        state = solver(state)
        solver = self.solvers.factory[self.solver](p[2 * i + 1], self.mixer)
        state = solver(state)
    # FIXME: Remember to normalize when using RK solvers!
```
This could be refactored further to avoid re-defining the solvers over and over inside the loop. I think this requires some changes in the `solvers.py` file, so I can take care of it.

The last point to remember is that once we finalize the structure of functions, we should add a `test_qaoa.py` file that tests all the methods (not only optimization, but also single execution with fixed parameters) in the `QAOA` class for different Hamiltonians (including Trotter Hamiltonians).

If you like and you have time you can have a go for any of these points. Otherwise let me know and I can also push some changes here directly (if you agree).

> WARNING: Lacking documentation for the webpage.

Don't worry about this, it is actually done for the most part because the webpage automatically copies the docstrings you already wrote in each function. We will just add a short how-to example when we are done with the code.

> When using the default exact exponentiation solver, 'exp', combined with the 'Powell' minimization method, instead of the 'rk4' or 'rk45', the QAOA needs a deeper circuit to get to the right minimum. Any clue on why??

I have not done any numerical tests with the code yet, however I am slightly confused about whether it makes sense to use RK methods here. The 'exp' solver should be fine because it essentially applies the operators that appear in the original QAOA ansatz. However RK methods apply an approximation of this operator as if integrating Schodinger's equation. The two should agree as `dt -> 0`, however this is not the case here since `dt = p[i]` and `p` can be whatever, right?

DiegoGM91:
Hi @stavros11, thanks for the comments and suggestions. I agree with your proposals. 
If you want, you can push the changes. About the FIXME comments in the code, I think I can take care of the ``FIXME: Use Qibo optimizers in this method``, but for the other ones it is probably better and faster if you implement them.

I'm confused by this line in the loss function: ``state = self(initial_state)`` 
Shouldn't it be ``state = self.get_initial_state(initial_state)`` ?

Regarding the  ``rk`` solvers, you are right in that in the QAOA ``dt`` does not need to be ``dt-->0``, so using them would result in a slight modification of the original algorithm (as a matter of fact, I would say it is like cheating, because the evolution now is not unitary). The small trials that I did with these methods seemed to work pretty well, but I would not worry about this at the moment. We might take a look into it in the future.
Maybe we can fix the solver and not let it as an argument for the time being (not really sure about this)? Mi original motivation to include it as an argument was the memory-inefficiency of the actual "exp". I think for the future a good idea would be to use quantum circuits directly for time evolution. This is straightforward when the Hamiltonian is given as a weighted sum of Pauli operators (as it happens often, e.g. in combinatorial optimization), and we would avoid the memory overhead of the matrix exponetiation.

stavros11:
Thanks for the reply. I agree that it will probably be easier if I implement the changes. To avoid confusion with your branch, I will open a new PR with this as base where I implement the suggestions I wrote above. In the meantime, if you like to change anything such as the `minimize` method, feel free to push here.

> Regarding the `rk` solvers, you are right in that in the QAOA `dt` does not need to be `dt-->0`, so using them would result in a slight modification of the original algorithm (as a matter of fact, I would say it is like cheating, because the evolution now is not unitary). The small trials that I did with these methods seemed to work pretty well, but I would not worry about this at the moment. We might take a look into it in the future.

Indeed, let's focus on some basic QAOA functionality for the first verison and later we can experiment with which methods work better and what is more motivated in terms of physics.

> I think for the future a good idea would be to use quantum circuits directly for time evolution. This is straightforward when the Hamiltonian is given as a weighted sum of Pauli operators (as it happens often, e.g. in combinatorial optimization), and we would avoid the memory overhead of the matrix exponetiation.

Currently this is possible if you use a `TrotterHamiltonian` instead of the `Hamiltonian` and it should work with the QAOA code you already have (with minor changes). This will Trotterize the exponentials automatically and will apply them as two-qubit gates via a circuit, so it will is possible to go up to 30+ qubits. However this `TrotterHamiltonian` functionality is probably not very user-friendly and perhaps not very general, so for the future it may be good to develop other ways that the user can pass Hamiltonians which can be decomposed to gates when they are exponentiated.

scarrazza:
Concerning the exp solver, are we sure that the formalism is able to deal with negative dt?

stavros11:
> Concerning the exp solver, are we sure that the formalism is able to deal with negative dt?

I believe so, yes. I cannot think of any problem with negative dt. All the exp solver does is to calculate exp(-1j * dt * H) and multiply it to the state vector. I also updated `test_qaoa_execute` to include negative values in parameters and tests still pass. 
In fact I would expect exp solver to be the only exact solver, as both Trotter and RK require small dt to be accurate (which is not always the case in QAOA).

scarrazza:
Ok, here some observation after playing with this code:
- I printed the final set of parameters for exp (bad final state) and rk4 (good final state), the "best" parameters are orders of magnitude different between each other, e.g. `[ 2.80794122e+13  3.60001258e-04 -6.60718587e-01 -4.79363348e-02]`, while the exp solver never goes beyond the `[-1,1]` range for each parameter. If I am not wrong these parameters are time steps so 1e+13 looks strange.
- Now, I tried to minimize from scratch starting from the rk4 best solution. The rk4 solver finds again the best ground state, however the exp does not, this sounds like some bug or a strong numerical instability. 
- I have tried to perform manually the last minimizer steps, i.e. use the best set of parameters from rk4 and then evaluate the hamiltonian expectation:
  ```python
  qaoa.set_parameters(best_from_rk4)
  state = qaoa()
  print(hamiltonian.expectation(state)
  ```
When using rk4, I get consistently the good state, however when using the same parameters the exp solution gives me a totally different answer. Excluding a possible bug, this may confirms that there are numerical instabilities in the system (given the solution location of the best parameters in the space of solutions) that makes extremely difficult to control if we think about the exp procedure (large matrix, then tiny, etc), while the rk4 performs the same operation "adiabatically". 

Please try to reproduce my steps, and to check the instability hypothesis we can simply try to enlarge the system and monitor the size of the final parameters.

scarrazza:
> When using the default exact exponentiation solver, 'exp', combined with the 'Powell' minimization method, instead of the 'rk4' or 'rk45', the QAOA needs a deeper circuit to get to the right minimum. Any clue on why??

Which circuit configuration produces a good exp solution?

scarrazza:
@stavros11, the segfault due to the forced timeout is still there. To solve this issue definitively, I believe we should increase the timeout to 100 but make sure that the examples have light configurations. I believe github actions machines have different specifications, thus we may also decide to reduce the tests to [linux, osx] with just python 3.8, let me know your preference.

stavros11:
Thanks for these observations, they are very helpful.

> * I printed the final set of parameters for exp (bad final state) and rk4 (good final state), the "best" parameters are orders of magnitude different between each other, e.g. `[ 2.80794122e+13  3.60001258e-04 -6.60718587e-01 -4.79363348e-02]`, while the exp solver never goes beyond the `[-1,1]` range for each parameter. If I am not wrong these parameters are time steps so 1e+13 looks strange.

Yes, I agree that the 1e+13 do not look right. I think it is fine if the parameters go beyond `[-1, 1]`, however I would not expect order of magnitude increase. Also, I do not believe the RK evolution is physical for such large parameters because it has an O(dt^n) error (for some positive integer n) which is acceptable only if dt < 1, right?
However when I optimize using rk45 and Powell (that is `python qaoa.py --solver rk45`) the parameters I get are:
`[-2.54691253 -0.02578692 -0.1452979   0.06131017]` so not extreme values. However since there is a parameter with absolute value >1 I would still not expect the RK solver to give a physical state.

> * I have tried to perform manually the last minimizer steps, i.e. use the best set of parameters from rk4 and then evaluate the hamiltonian expectation:
>   ```python
>   qaoa.set_parameters(best_from_rk4)
>   state = qaoa()
>   print(hamiltonian.expectation(state)
>   ```

I have done this with various configurations and generally I observe the same behavior as you. For example when I optimize using rk45 I find the parameters I wrote above (`[-2.54691253 -0.02578692 -0.1452979   0.06131017]`). When I use these parameters to just evolve the QAOA and calculate the hamiltonian expectation I get `-6.744470` using rk45 and `1.0487516` using exp. Physically speaking I believe the exp result is the correct one because I don't think it makes sense to use RK with dt > 1. This could be the numerical instability you are referring to?

> Which circuit configuration produces a good exp solution?

After some quick tests for 4 qubits, I did not manage to find a good exp solution with up to 20 parameters in terms of energy, however with 10 parameters I could find a good solution using Trotter exp. The final parameters are:
`[1.56174121,  0.45216744,  0.93687333,  0.83813211,  0.38926526,  0.23464604,
   -0.9764948,  0.72818006,  1.05367271,  1.65009944]`
and I get the following energies:
* 3.749714818 using exp,
* -6.74421788 (correct) using Trotter exp,
* 4.000255287 using rk45.

So again there is disagreement between the different solvers when using the same parameters, which is expected because both Trotter and RK have non-negligible deviation from exp when dt > 1.

I may be wrong, but overally I believe the exp is the only solver that gives (physically) correct values for the ansatz we are using. The other solvers are changing the original QAOA ansatz using approximations. The Trotter approximation can be mapped to circuit so we could actually run it on a quantum computer. The RK ones probably no.
This doesn't explain why it is not possible to converge to the right energy with the exp solver, though. I guess that the approximated ansatz using Trotter/RK is easier to optimize compared to the exact one.

scarrazza:
Thanks for the cross check. I think at this point we are agree, the QAOA model is correct, however the optimization procedure requires particular precaution to avoid unphysical solutions. I suggest to merge this PR as soon as the checks are completed and keep the discussion about optimization when solving specific problems that require the QAOA model.

scarrazza (approved):

stavros11:
Thanks. Yes, I agree with merging for and we can keep in mind the optimization issues for the future.

PR #228, opened 2020-08-31
Refactor QAOA

stavros11:
Implements the changes on the QAOA code I described in #227.

TODO:
* add callback functionality (may be useful?),
* check distributed execution when `TrotterHamiltonian` is used,
* finalize and check `minimize` method,
* add example in docs.

### DIFF
--- a/src/qibo/models.py
+++ b/src/qibo/models.py
@@ -201,67 +201,83 @@ class QAOA(object):
     """ This class implements the QAOA algorithm."""
     import numpy as np
     from qibo import hamiltonians
-        
-    def __init__(self, nqubits, hamiltonian, mixer=None, initial_state=None):
+
+    def __init__(self, hamiltonian, mixer=None, solver="exp"):
         """
         Initializes an instance of the QAOA.
-        
+
         Args:
-            nqubits (int): number of qubits.
             hamiltonian (qibo.hamiltonians.Hamiltonian): problem Hamiltonian whose ground state is sought.
             mixer (qibo.hamiltonians.Hamiltonian): mixer Hamiltonian. If None, it is taken as -sum(sigma_x).
-            initial state: initial state of the algorithm. If None, it is taken as the uniform
-                           superposition over computational-basis states.
+            solver (string): solver employed for StateEvolution.
         """
+        # list of QAOA variational parameters (angles)
+        self.params = None
         # problem hamiltonian
         self.hamiltonian = hamiltonian
+        self.nqubits = hamiltonian.qubits
+        # evolution solver
+        self.solver = solver
         # mixer hamiltonian (default = -sum(sigma_x))
         if mixer == None:
             self.mixer = self.hamiltonians.X(nqubits)
         else:
             self.mixer = mixer
-        # initial state
-        if initial_state == None:
-            self.initial_state = self.np.ones(2**nqubits) / self.np.sqrt(2**nqubits)
+
+    def set_parameters(p):
+        self.params = p
+
+    def __call__(self, initial_state=None):
+        """Applies the QAOA exponentials to a state."""
+        # We define the evolution operators for the mixer and problem hamiltonians
+        evolve = []
+        for i in range(len(self.params)//2):
+            evolve.append(StateEvolution(self.hamiltonian, self.params[2*i], solver=self.solver))
+            evolve.append(StateEvolution(self.mixer, self.params[2*i+1], solver=self.solver))
+        # We evolve the state
+        state = self.get_initial_state(initial_state)
+        for i, evolution_operator in enumerate(evolve):
+            state = evolution_operator(final_time=self.params[i], initial_state=state)
+        return state
+
+    def get_initial_state(self, state=None):
+        # FIXME: Cast states to Tensorflow properly
+        if state is None:
+            return self.np.ones(2**nqubits) / self.np.sqrt(2**nqubits)
         else:
-            self.initial_state = initial_state
-                   
-    def loss(self, p, solver='exp'):
-        """
-        Args: 
-            p (numpy.array or list): parameters of the QAOA.
-            solver (string): solver employed for StateEvolution.
+            return initial_state
+
+    def _loss(self, initial_state=None):
+        """Calculates the QAOA loss (energy).
+
+        Args:
+            initial_state (np.ndarray):
+
         Returns:
             (numpy.float64) with the value of the loss function.
         """
-        # We define the evolution operators for the mixer and problem hamiltonians 
-        evolve = []
-        for i in range(len(p)//2):
-            evolve.append(StateEvolution(self.hamiltonian, p[2*i], solver=solver))
-            evolve.append(StateEvolution(self.mixer, p[2*i+1], solver=solver))
-        # We evolve the state
-        state = self.initial_state
-        for i,evolution_operator in enumerate(evolve):
-            state = evolution_operator(final_time=p[i], initial_state=state)
-            
+        state = self(initial_state)
         return self.np.float64(self.hamiltonian.expectation(state))
-        
-    def train(self, initial_p, method='Powell', solver='exp'):
-        """
+
+    def minimize(self, initial_p, initial_state=None, method='Powell'):
+        """Optimizes the variational parameters of the QAOA.
+
         Args:
             initial_p (numpy.array or list): initial guess for the parameters of the QAOA.
-            method (string): minimization method.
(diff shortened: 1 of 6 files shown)


stavros11:
I believe this is in good state for a first review. I implemented the changes described in #227 and added an example script (in `benchmars/qaoa.py`) that performs the QAOA optimization for the XXZ Hamiltonian. 

The only issue that remains unsolved is that, as Diego mentioned in #227, optimization seems to converge to the right energy when RK solvers are used but fails when the "exp" solver is used. This is a bit strange because "exp" is supposed to be the exact one.

src/qibo/models.py
```diff
@@ -198,70 +199,178 @@ def loss(params):
...
+    from qibo import hamiltonians, optimizers
+    from qibo.config import K, DTYPES, log
+    from qibo.callbacks import Norm
+
+    def __init__(self, hamiltonian, mixer=None, solver="exp", callbacks=[],
+                 accelerators=None, memory_device="/CPU:0"):
+        # list of QAOA variational parameters (angles)
+        self.params = None
         # problem hamiltonian
         self.hamiltonian = hamiltonian
```
scarrazza, line 244: Shall we check that hamiltonian is really a hamiltonian object?

src/qibo/models.py
```diff
@@ -198,70 +199,178 @@ def loss(params):
...
-            self.initial_state = self.np.ones(2**nqubits) / self.np.sqrt(2**nqubits)
-        else:
-            self.initial_state = initial_state
-                   
-    def loss(self, p, solver='exp'):
+
+        # create circuits for Trotter Hamiltonians
+        if (accelerators is not None and (
+            not isinstance(self.hamiltonian, self.hamiltonians.TrotterHamiltonian)
+            or solver != "exp")):
```
scarrazza, line 262: Ok, this means that rk is not supported, is there a technical motivation for that?

DiegoGM91:
> I believe this is in good state for a first review. I implemented the changes described in #227 and added an example script (in `benchmars/qaoa.py`) that performs the QAOA optimization for the XXZ Hamiltonian.
> 
> The only issue that remains unsolved is that, as Diego mentioned in #227, optimization seems to converge to the right energy when RK solvers are used but fails when the "exp" solver is used. This is a bit strange because "exp" is supposed to be the exact one.

The`exp` method also converges eventually, but needs a much higher value of `len(p)`, i.e. a much deeper circuit. No clue on why.

src/qibo/models.py
DiegoGM91: Maybe that since `rk` is not unitary, then you couldn't really map it to an actual quantum circuit executed on QPU (at least in a direct and straightforward manner)?

stavros11: RK is not supported only when distributed execution is used, this `NotImplemetedError` is raised only when RK is used and `accelerators` is not `None`. For single device execution RK is supported and working.
The reason is what Diego said, distributed execution uses the `DistributedCircuit` class, so it requires the evolution/QAOA to be mapped to a circuit. In the current implementation this is possible only when `TrotterHamiltonian` + "exp" solver is used. RK is indeed not unitary so it is not possible to map it to a circuit anyway (regardless of implementation).

Note that it is possible to use RK + `TrotterHamiltonian`. This will calculate the dot products between Hamiltonian and state term-by-term using two-qubit gates, instead of creating the full Hamiltonian matrix. This is still not unitary because the various terms are summed (and also requires copies of the state vector).

src/qibo/models.py
stavros11: I added a check for this.

src/qibo/models.py
scarrazza: Ok, that's clear, thanks.

scarrazza:
Is this behaviour optimizer dependent?

src/qibo/models.py
```diff
@@ -198,70 +199,182 @@ def loss(params):
...
+        if mixer is None:
+            trotter = isinstance(
+                self.hamiltonian, self.hamiltonians.TrotterHamiltonian)
+            self.mixer = self.hamiltonians.X(self.nqubits, trotter=trotter)
         else:
+            if type(mixer) != type(hamiltonian):
+                  raise_error(TypeError, "Given Hamiltonian is of type {} "
+                                         "while mixer is of type {}."
+                                         "".format(type(hamiltonian),
+                                                   type(mixer)))
```
DiegoGM91, line 259: Maybe it is simpler to just check that mixer is of the right type? (because if hamiltonian is right, but not mixer, this message will appear and the user might not know which of the two is wrong)

stavros11:
> The`exp` method also converges eventually, but needs a much higher value of `len(p)`, i.e. a much deeper circuit. No clue on why.

I can also confirm that using the `benchmarks/qaoa.py` script. With RK methods it gets the right energy with four parameters. With "exp" I could get the right energy with 20 parameters.

> Is this behaviour optimizer dependent?

I was actually doing some experiments right now. A quick observation is that for 4 qubits with 4 parameters, it converges to the right energy only with RK + Powell. When exp + Powell or any solver + BFGS the final energy is wrong.

src/qibo/models.py
stavros11: Thanks for the review.

The reason I check it this way is that the hamiltonians may be either `Hamiltonian` or `TrotterHamiltonian`. Both types are correct as long as both `hamiltonian` and `mixer` are of the same type (any of the two). I believe there is no confusion because I already added a check that the `hamiltonian` is of proper type earlier (following Stefano's suggestion).

I fixed the rest points you suggested.

DiegoGM91: >The reason I check it this way is that the hamiltonians may be either Hamiltonian or TrotterHamiltonian. Both types are correct as long as both hamiltonian and mixer are of the same type (any of the two). I believe there is no confusion because I already added a check that the hamiltonian is of proper type earlier (following Stefano's suggestion).

I see. Thanks!

scarrazza:
Btw, there is some example which is using too many resources or raising segfauts, we have to investigate about that.

stavros11:
I agree. I think segfaults are happening due to the timeout, because I noticed that sometimes when I use ctrl+c to terminate a script that is running I may get a segfault. This is probably due to the custom operators, I am not sure if there is a way to fix it on the C++ level (eg. better exceptions).

The examples tests pass when I run them locally but take some time. I guess that it is a resource issue in the workflows.

scarrazza:
Yeah, another hypothesis is that the segfault come from the scipy, infact some of the error messages derive from optimizers.py.

scarrazza:
On my laptop I get test issues for adiabatic3sat, in particular this one:
```
FAILED examples/test_examples.py::test_adiabatic3sat[BFGS-None-True-exp-8-1-10-0.1] - SystemError: <built-in method __array__...
```
which I believe is one of the first configurations, some of the other configurations are requiring lots of memory, and this is probably why we get the crashes.

stavros11:
> which I believe is one of the first configurations, some of the other configurations are requiring lots of memory, and this is probably why we get the crashes.

Thanks for looking into this. When I run this test alone on my local desktop I do not get any memory problems (in fact memory usage is <1GB) but it takes a lot of time. Two easy solutions I can think of are:
* test using 4 qubits instead of 8, or better
* pass the `maxiter` argument (or better add and pass a `maxfev` argument) to limit the scipy optimization time. We are not testing for convergence in this test anyway.

scarrazza:
Yeah, here another configuration which on my pc with tf2.3 consumes more than 10GB:
```
nqubits=8 
instnce=1 
T=10 
dt=0.1 
solver=rk45 
Trotter=False 
params=None 
method=BFGS
```

scarrazza:
I can confirm that 4 qubits is working fine. However I like your second suggestion, could you please give a try and see if the CI pass?

stavros11:
> I can confirm that 4 qubits is working fine. However I like your second suggestion, could you please give a try and see if the CI pass?

I pushed with nqubits=8, maxiter=1 to check the CI. 
However, I can also confirm that the configuration you wrote above uses a lot of memory. My computer crashes when I run `main.py` using the parameters you wrote. We may want to investigate this issue regardless of the tests, because 8 qubits shouldn't consume that much memory anyway.

scarrazza:
Better than before but probably still too much resource intensive. Could you please reduce to 4 qubits?

scarrazza:
Thanks, somehow the memory issue is something that is hard to reproduce, the standalone code seems to work fine.
So, let's merge this PR and then look at the "exp" business.

PR #230, opened 2020-09-01
New images for the AE readme

igres26:


scarrazza (approved):

stavros11:
New images look good. I also made a small modification in the plot arrangement in the "optimization example" section. I did this because I noticed that the plots were not appearing next to each other in the sphinx docs (unlike the README), so the new arrangement is more consistent with the text.

PR #233, opened 2020-09-04
Tutorials sort

scarrazza:


stavros11 (approved):
Looks good, thanks.

Something minor: In `applications.rst` we say "quantum circuits applied to specific physics problems.". We could remove "physics" as there are applications such as the hash, 3SAT, classification, etc. which are not physics problems. On the other hand, since quantum computing is a physics field it shouldn't be a problem to leave it as it is. I am fine either way.

scarrazza:
Thanks!

PR #238, opened 2020-09-17
Typos documentation

DiegoGM91:
I have corrected a few typos in code snippets in the docs. Also made minimal `import` changes and things like that so that if one is just looking at a particular section but has skipped the previous ones, everything is self-consistent. The latter are not strictly necessary, so as you prefer.

I have reviewed the "Basic examples" section and the "Advanced examples" up to QAOA (inclusive).

Two extra comments are:

- I think that `circuit depth` in the `summary` attribute of `Circuit` is actually referring to number of gates, rather than depth of the circuit, isn't it?

- This code snippet, extracted from the docs:

```
from qibo import gates

c = Circuit(3)
g0 = gates.RX(0, theta=0)
g1 = gates.RY(1, theta=0)
g2 = gates.fSim(0, 2, theta=0, phi=0)
c.add([g0, g1, gates.CZ(1, 2), g2, gates.H(2)])

# or with a flat list as
params = [0.123, 0.456, 0.789, 0.321]
c.set_parameters(params)`
```

gives me the following error: `TypeError: unsupported operand type(s) for /: 'list' and 'float'`

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ from qibo.models import QFT
 circuit = QFT(15)
 
 # Simulate final state wavefunction default initial state is |00>
-final_state = c()
+final_state = circuit()
 ```
 
 Here another example with more gates and shots simulation:
--- a/doc/source/advancedexamples.rst
+++ b/doc/source/advancedexamples.rst
@@ -284,8 +284,13 @@ The following gates support parameter setting:
 
 .. code-block:: python
 
-    c = Circuit(5)
-    pairs = list((i, i + 1) for i in range(0, 4, 2))
+    import numpy as np
+    from qibo.models import Circuit
+    from qibo import gates
+
+    nqubits = 5
+    c = Circuit(nqubits)
+    pairs = [(i, i + 1) for i in range(0, 4, 2)]
     c.add(gates.VariationalLayer(range(nqubits), pairs,
                                  gates.RY, gates.CZ,
                                  params=np.zeros(5)))
@@ -340,7 +345,7 @@ Here is a simple example using the Heisenberg XXZ model Hamiltonian:
     # Optimize starting from a random guess for the variational parameters
     initial_parameters = np.random.uniform(0, 2*np.pi,
                                             2*nqubits*nlayers + nqubits)
-    best, params = vqe.minimize(initial_parameters, method='BFGS')
+    best, params = vqe.minimize(initial_parameters, method='BFGS', compile=False)
 
 
 For more information on the available options of the ``vqe.minimize`` call we
@@ -350,7 +355,7 @@ has to use a backend based on tensorflow primitives and not the default custom
 backend, as custom operators currently do not support automatic differentiation.
 To switch the backend one can do ``qibo.set_backend("matmuleinsum")``.
 Check the :ref:`How to use automatic differentiation? <autodiff-example>`
-for more details.
+section for more details.
 
 A useful gate for defining the ansatz of the VQE is :class:`qibo.base.gates.VariationalLayer`.
 This optimizes performance by fusing the layer of one-qubit parametrized gates with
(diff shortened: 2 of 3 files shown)


scarrazza:
Thanks Diego for the typos.

> I think that circuit depth in the summary attribute of Circuit is actually referring to number of gates, rather than depth of the circuit, isn't it?

yes, it is the number of gates/operations. I think we should review this method, and include the real circuit depth, to avoid confusion.

> gives me the following error: TypeError: unsupported operand type(s) for /: 'list' and 'float'

Yes, we have to cast the list as a numpy array to avoid this issue. @stavros11 , we probably have to manage this cast/check operation internally to the `set_parameters` method.

stavros11 (approved):
Thank you for all these fixes Diego. All changes look good to me. Regarding the imports, personally I do not have a preference. I think making each section self-consistent as you say sounds better.

> I have reviewed the "Basic examples" section and the "Advanced examples" up to QAOA (inclusive).

Do you plan to continue with the rest sections? If no let me know and I can also look at this, because as we can see there are a few typos due to code changes and it would be useful to review.

> > I think that circuit depth in the summary attribute of Circuit is actually referring to number of gates, rather than depth of the circuit, isn't it?
> 
> yes, it is the number of gates/operations. I think we should review this method, and include the real circuit depth, to avoid confusion.

Indeed, this was pointed out in a previous meeting but we forgot to fix it and I noted it again when I was updating the tutorial. I think for now we should change the printed message from "Circuit depth" to "Total number of gates", this can probably be done in this PR. I will open a new PR that calculates the actual circuit depth because I think this is slightly more complicated (it cannot be done directly using just the list that stores all gates that we have now).

> > gives me the following error: TypeError: unsupported operand type(s) for /: 'list' and 'float'
> 
> Yes, we have to cast the list as a numpy array to avoid this issue. @stavros11 , we probably have to manage this cast/check operation internally to the `set_parameters` method.

Thanks for finding this! Yes, casting the list to `np.array` (that is doing `params = np.array([0.123, 0.456, 0.789, 0.321])`) should fix this issue. I just noticed that I have been using array in the test, this is why we did not catch this. I agree that we should leave the snippet as it is and make the change internally in `set_parameters`. Shall we do this in this PR?

scarrazza:
> Shall we do this in this PR?

I think would better to create a new PR where the depth string and the numpy conversion are done.

This PR is ready to merge.

stavros11:
> > Shall we do this in this PR?
> 
> I think would better to create a new PR where the depth string and the numpy conversion are done.
> 
> This PR is ready to merge.

Great, I will open a new PR for the depth and the numpy casting. I will also do the "Circuit depth" -> "Number of gates" change there so that we don't modify any `.py` files in this PR.

I agree with merging this, unless Diego wants to complete the review of "Advanced examples" first.

stavros11:
Another question I forgot to write: 
Currently `circuit.summary` is a `@property`. Do you think that it would look more natural if this was just a normal method accessed as `circuit.summary()`? Looking at some of the changes the latter option looks nicer to me but let me know if you agree and I can change it accordingly in the other PR as well.

scarrazza:
> Do you think that it would look more natural if this was just a normal method accessed as circuit.summary()? Looking at some of the changes the latter option looks nicer to me but let me know if you agree and I can change it accordingly in the other PR as well.

I agree, in particular to preserve symmetry with other methods.

DiegoGM91:
> Do you plan to continue with the rest sections? If no let me know and I can also look at this, because as we can see there are a few typos due to code changes and it would be useful to review.

>I agree with merging this, unless Diego wants to complete the review of "Advanced examples" first

I can review the rest of the sections, yes. But probably not today. So I would suggest it is better to merge this PR so that the fixes in the docs are incorporated for newcomers, and I will open a new PR with the remaining fixes (if any), if you agree.

scarrazza:
Thanks, merging this PR now.

PR #240, opened 2020-09-18
Entanglement spectrum

### LINKED ISSUE #239: Entanglement spectrum callback
Germán suggested including the entanglement spectrum, i.e. `-np.log([eigenvalue for eigenvalue in density matrix])`, as a useful feature.

I think this should be really straightforward as these eigenvalues are already calculated for the `EntanglementEntropy` callback.

scarrazza:
Implements and closes #239.

### DIFF
--- a/src/qibo/tensorflow/callbacks.py
+++ b/src/qibo/tensorflow/callbacks.py
@@ -167,7 +167,7 @@ class EntanglementEntropy(PartialTrace):
         eigvals = tf.math.real(tf.linalg.eigvalsh(rho))
         # Treating zero and negative eigenvalues
         masked_eigvals = tf.gather(eigvals, tf.where(eigvals > EIGVAL_CUTOFF))[:, 0]
-        spectrum = - tf.math.log(masked_eigvals)
+        spectrum = -1 * tf.math.log(masked_eigvals)
         if self.compute_spectrum:
             self.spectrum.append(spectrum)
         entropy = - tf.reduce_sum(masked_eigvals * - spectrum)
--- a/src/qibo/tensorflow/callbacks.py
+++ b/src/qibo/tensorflow/callbacks.py
@@ -181,7 +181,7 @@ class EntanglementEntropy(PartialTrace):
         # Calculate entropy of reduced density matrix
         if self.compute_spectrum:
             entropy, eigvals = self._entropy(rho, compute_eigvals=True)
-            spectrum = - tf.math.log(eigvals)
+            spectrum = -1 * tf.math.log(eigvals)
             self.spectrum.append(spectrum)
         else:
             entropy = self._entropy(rho)
--- a/src/qibo/tensorflow/callbacks.py
+++ b/src/qibo/tensorflow/callbacks.py
@@ -170,7 +170,7 @@ class EntanglementEntropy(PartialTrace):
         spectrum = -1 * tf.math.log(masked_eigvals)
         if self.compute_spectrum:
             self.spectrum.append(spectrum)
-        entropy = - tf.reduce_sum(masked_eigvals * - spectrum)
+        entropy = tf.reduce_sum(masked_eigvals * spectrum)
         return entropy / self._log2
 
     def __call__(self, state: tf.Tensor, is_density_matrix: bool = False
--- a/src/qibo/tensorflow/callbacks.py
+++ b/src/qibo/tensorflow/callbacks.py
@@ -130,6 +130,7 @@ class EntanglementEntropy(PartialTrace):
             for the entropy calculation.
             If `partition` is not given then the first subsystem is the first
             half of the qubits.
+        compute_spectrum (bool): Compute the entanglement spectrum. Default is False.
 
     Example:
         ::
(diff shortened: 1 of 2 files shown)


src/qibo/tensorflow/callbacks.py
```diff
@@ -150,25 +151,41 @@ class EntanglementEntropy(PartialTrace):
             print(entropy[:])
             # Should print [0, 0, 1] which is the entanglement entropy
             # after every gate in the calculation.
+            print(entropy.spectrum)
```
stavros11, line 154: I think in this example this will print an empty list because `compute_spectrum = False` by default. We have to pass `compute_spectrum = True` when we define the entropy callback in line 140.

src/qibo/tensorflow/callbacks.py
```diff
@@ -150,25 +151,41 @@ class EntanglementEntropy(PartialTrace):
...
 
     def __call__(self, state: tf.Tensor, is_density_matrix: bool = False
                  ) -> tf.Tensor:
         # Construct reduced density matrix
         rho = super(EntanglementEntropy, self).__call__(state, is_density_matrix)
         # Calculate entropy of reduced density matrix
-        return self._entropy(rho)
+        if self.compute_spectrum:
+            entropy, eigvals = self._entropy(rho, compute_eigvals=True)
+            spectrum = -1 * tf.math.log(eigvals)
```
stavros11, line 184: Isn't it preferable to do this calculation inside the `_entropy` function and return the spectrum instead of `eigvals` directly? The reason is that currently we are taking the log twice, once here and once in `_entropy` line 171.

src/qibo/tensorflow/callbacks.py
```diff
@@ -150,25 +151,41 @@ class EntanglementEntropy(PartialTrace):
...
     @classmethod
-    def _entropy(cls, rho: tf.Tensor) -> tf.Tensor:
-      """Calculates entropy by diagonalizing the density matrix."""
-      # Diagonalize
-      eigvals = tf.math.real(tf.linalg.eigvalsh(rho))
-      # Treating zero and negative eigenvalues
-      masked_eigvals = tf.gather(eigvals, tf.where(eigvals > EIGVAL_CUTOFF))[:, 0]
-      entropy = - tf.reduce_sum(masked_eigvals * tf.math.log(masked_eigvals))
-      return entropy / cls._log2
+    def _entropy(cls, rho: tf.Tensor, compute_eigvals: bool = False):
```
stavros11, line 165: I do not remember exactly why this function is a `@classmethod` and not a normal method, but I tried removing `@classmethod` and tests pass (except two callback tests but these can be fixed). 
I believe that if we make this a normal method we can save a few lines of code for the spectrum as we would be able to do
```Python
spectrum = - tf.math.log(masked_eigvals)
if self.compute_eigvals:
    self.spectrum.append(spectrum)
entropy = tf.reduce_sum(masked_eigvals * spectrum)
return entropy / self._log2
```
directly here and avoid return two values and having the `compute_eigvals` argument.

This is just a suggestion, it is okay as it is now, let me know what you prefer.

stavros11 (commented):
Thanks for implementing this. Some comments below:

src/qibo/tensorflow/callbacks.py
scarrazza: Thanks well spotted!

src/qibo/tensorflow/callbacks.py
scarrazza: Yes, I fully agree, I would prefer to remove the `@classmethod` and do everything in one go.
I will propose the changes here.

scarrazza:
@stavros11 please have another look.

src/qibo/tests/test_callbacks.py
```diff
@@ -5,14 +5,16 @@
...
+
 
 # Absolute testing tolerance for the cases of zero entanglement entropy
 _atol = 1e-8
 
 
 def test_entropy_product_state():
     """Check that the |++> state has zero entropy."""
-    entropy = callbacks.EntanglementEntropy()
+    entropy = callbacks.EntanglementEntropy(compute_spectrum=True)
```
stavros11, line 17: I believe this is not need unless we also test for `entropy.spectrum` in this test.

src/qibo/tensorflow/callbacks.py
```diff
@@ -150,18 +151,27 @@ class EntanglementEntropy(PartialTrace):
...
+    def _entropy(self, rho: tf.Tensor) -> tf.Tensor:
+        """Calculates entropy by diagonalizing the density matrix."""
+        # Diagonalize
+        eigvals = tf.math.real(tf.linalg.eigvalsh(rho))
+        # Treating zero and negative eigenvalues
+        masked_eigvals = tf.gather(eigvals, tf.where(eigvals > EIGVAL_CUTOFF))[:, 0]
+        spectrum = -1 * tf.math.log(masked_eigvals)
+        if self.compute_spectrum:
+            self.spectrum.append(spectrum)
+        entropy = - tf.reduce_sum(masked_eigvals * - spectrum)
```
stavros11, line 173: We can cancel the two minus signs here, right?

stavros11 (approved):
Thank you for the updates, it looks good to me now.

stavros11 (approved):
Thanks for the update, looks good.

PR #241, opened 2020-09-20
Calculate correct circuit depth

stavros11:
Implements the changes discussed in #238, particularly the following:
* Cast parameters given as a flat list to `circuit.set_parameters()` to `np.array` so that the snippet from #238 works.
* Change `circuit.depth` property to `circuit.ngates` as it returns the total number of gates in the circuit and not the depth.
* Create a new `circuit.depth` property that returns the actual depth, assuming that each gate is placed in the earliest possible position in the circuit according to the qubits it acts on. In order to do this calculation I extended the queue data structure slightly to hold a list of "moments" (in Cirq terminology).
* Change `circuit.summary` from `@property` to simple method so that it can be accessed as `circuit.summary()`.
* Update `circuit.summary()` to print both the (corrected) circuit depth and total number of gates.

I also updated the docs regarding the `circuit.summary()`, however it might make sense to revert these changes because the public docs will be updated while the pip Qibo version will not. However we should remember to update the docs for all changes when we release. Let me know what you think.

### DIFF
--- a/doc/source/examples.rst
+++ b/doc/source/examples.rst
@@ -58,7 +58,7 @@ backend which uses custom tensorflow operators to apply gates.
 How to print a circuit summary?
 -------------------------------
 
-It is possible to print a summary of the circuit using ``circuit.summary``.
+It is possible to print a summary of the circuit using ``circuit.summary()``.
 This will print basic information about the circuit, including its depth, the
 total number of qubits and all gates in order of the number of times they appear.
 The QASM name is used as identifier of gates.
@@ -76,10 +76,11 @@ For example
     c.add(gates.CNOT(1, 2))
     c.add(gates.H(2))
     c.add(gates.TOFFOLI(0, 1, 2))
-    print(c.summary)
+    print(c.summary())
     # Prints
     '''
-    Circuit depth = 6
+    Circuit depth = 5
+    Total number of gates = 6
     Number of qubits = 3
     Most common gates:
     h: 3
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -5,11 +5,16 @@
 Welcome to Qibo!
 ================
 
-Qibo is a Python library for classical simulation of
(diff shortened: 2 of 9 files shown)


scarrazza (approved):
Thanks, the changes are good.

>  however it might make sense to revert these changes because the public docs will be updated while the pip Qibo version will not. However we should remember to update the docs for all changes when we release. Let me know what you think.

I think we should:
- change the documentation urls to "stable" instead of "latest" in all places (README.md, landing page, etc).
- show the qibo version in the docs (I think there is an option for that in sphinx)
- from now on, set the master version to 0.1.1-master or something similar following these lines.

stavros11:
> * change the documentation urls to "stable" instead of "latest" in all places (README.md, landing page, etc).

Thanks for the reply. If I understand correctly the "stable" docs are updated only when a release is made, so I agree that this should be the page that appears by default, because "latest" will be updated as we add new features. The only issue is that using "stable" means that we will temporarily loose the fixes from #238.

> * show the qibo version in the docs (I think there is an option for that in sphinx)
We can show the version number using `|release|` in the `.rst` files but this would appear only in a specific place in the docs (eg. in `index.rst`).

> * from now on, set the master version to 0.1.1-master or something similar following these lines.

For this do we just need to update `__init__.py`?

scarrazza:
Yes, exactly, we can solve this impasse by releasing a 0.1.1 at the end of this month with all changes, including documentation.

> We can show the version number using |release| in the .rst files but this would appear only in a specific place in the docs (eg. in index.rst).

Sounds good, adding the version at least in the home page is be fine.

> For this do we just need to update __init__.py?

yes, exactly. We can append the `-master` or `-dev` suffix for the master version.

stavros11:
> I would probably set it to 0.1.1-dev, so the transition from 0.1.1-dev to 0.1.1 looks more natural.

Thanks, yes I forgot to change 0.1.1.

scarrazza:
Great, thanks. Now, concerning the urls, the only places where we hardcode latest are:
- examples/benchmarks/results/PRECISION.md
- examples/benchmarks/results/HARDWARE.md
- examples/benchmarks/results/ADIABATIC.md
- examples/adiabatic3sat/README.md
- README.md
We can update latest->stable before the 0.1.1 release, so people reading the current latest may benefit from the corrections from Diego.

Issue #231, opened 2020-09-03
General unitary gates


AdrianPerezSalinas:
Sometimes the algorithms, in particular the variational ones, require the implementation of general unitary gates. Unitary gates must always be decomposed in terms of basis gates to be implemented in the quantum computer, but for simulation it might be more efficient if the general gates are directly applied

Would it be possible to add some function that allows the user to create user-defined gates for calling them in a particular program?

AdrianPerezSalinas:
I think so
I assume that this function does not accept parameters and any parametrization must be done outside the function, isn't it? 
And for 2 x 2 matrix it is as simple as `gates.Unitary(np.random.random((2,2)), 0)`

This function can be used with all backends??

stavros11:
> I assume that this function does not accept parameters and any parametrization must be done outside the function, isn't it?
> And for 2 x 2 matrix it is as simple as `gates.Unitary(np.random.random((2,2)), 0)`

Yes, the 2x2 matrix works like this. The matrix that you pass can be variational and you can also update it normally using `circuit.set_parameters` (if it is inside a circuit). If you need to do a calculation using some variational parameters to obtain the final matrix, indeed this has to be done outside the function and then gate should be created using the final matrix.

> Are there any further issues to be aware of with respect to the backends?

Up to two-qubit gates (that is either 2x2 or 4x4 matrices) should work with any backend. Three-qubit gates or more (8x8 or higher) should work only with "defaulteinsum" and "matmuleinsum" (not "custom").

AdrianPerezSalinas:
Okay, backends are understood.

Even though I might be repetitive, would it be easy (for the average user) to create a gate with the same functionality and usage as, for instance, RX, but being an arbitrary unitary gate? 

I am trying to write a circuit with the gate Rz() Ry() Rz() and would like to fuse them all in a single U3 gate

AdrianPerezSalinas:
Thank you very much for your replies, they are being very useful!!

stavros11:
> I am trying to write a circuit with the gate Rz() Ry() Rz() and would like to fuse them all in a single U3 gate

There are two ways to implement this with what we have currently in Qibo, but each may have some complications:

1. Use `circuit.fuse()`, for example
```Python
c.add([RZ(0, theta=1), RY(0, theta=1), RZ(0, theta=1)])
fused_c = c.fuse()
print(c.queue) # has len = 3
print(fused_c.queue) # has len = 1
```
[...]

PR #242, opened 2020-09-22
Add general unitary gates (U1, U2, U3)

stavros11:
This adds the general unitary gates discussed in #231 to Qibo. Note that the I followed [Qiskit's conventions](https://qiskit.org/documentation/stubs/qiskit.circuit.library.U3Gate.html) for all matrices and symbol names, which is slightly different from what Adrian wrote in #231. @AdrianPerezSalinas let me know what you think about this. I believe Qiskit's API is simpler in terms of angle names and they also use `theta` (instead of `lambda`) for U1 which agrees with what we had so far in ZPow.

In particular this PR implements the following changes:
* Rename `ZPow` and `CZPow` to `U1` and `CU1` respectively. I believe I also updated all docs/examples where `ZPow` was appearing.
* Add `U2`, `U3`, `CU2`, `CU3` gates following Qiskit's conventions.
* Add `CRX`, `CRY`, `CRZ` which are one-qubit controlled of the rotations we already had. These could be used in the past (for example `CRY(0, 1)` is `RY(1).controlled_by(0)`, however adding them as explicit gates allows using them with `to_qasm` and `from_qasm`, which was not possible before. Doing `RY(1).controlled_by(0)` now will fall back to `CRY(0, 1)` and will be compatible with Qasm.


AdrianPerezSalinas:
I like this implementation, I think it makes things easier.
With regard to qiskit, I prefer my convention because it is a product of what we consider the standard matrices (Ry, Rz, Rx), but at the end of the day they are equivalent since there is only a relative complex phase between both cases. 

How are the parameters implemented in these gates? Shall we use U3(qubit, theta, phi, lambda) or U3(qubit, (theta, phi, lambda))? And, whatever the case, if we want to set the parameters for variational circuits, can they be set as up to now?

scarrazza:
Thanks for the implementation. We have many places where ZPow and CZPow are used, including the paper, instead of removing these gates, I would suggest to create aliases in order to maintain back compatibility.

stavros11:
Thank you both for the comments.

> Thanks for the implementation. We have many places where ZPow and CZPow are used, including the paper, instead of removing these gates, I would suggest to create aliases in order to maintain back compatibility.

That's a good point. I added the ZPow and CZPow gates back but now invoking them will create a U1/CU1 gate respectively, so Qasm is also properly supported.

> How are the parameters implemented in these gates? Shall we use U3(qubit, theta, phi, lambda) or U3(qubit, (theta, phi, lambda))? And, whatever the case, if we want to set the parameters for variational circuits, can they be set as up to now?

It is U3(qubit, theta, phi, lam), seperate arguments, no need to use tuple, similarly to the fSim we already had with two parameters. By the way you can still use tuple via Python's `U3(qubit, *params)`, just make sure the order of parameters is consistent. When this is merged the documentation will also be updated so these conventions will also appear there.

`circuit.set_parameters` also works normally, using either a list of tuples or a flat list. For example all the cases bellow should work:
```Python
circuit = models.Circuit(2)
circuit.add(gates.RX(0, theta=0))
circuit.add(gates.U3(1, theta=0, phi=0, lam=0))

# option 1, list of tuples
circuit.set_parameters([0.1, (0.2, 0.3, 0.4)])
# option 2, flat list
circuit.set_parameters([0.1, 0.2, 0.3, 0.4])
# option 3, np.array
circuit.set_parameters(np.random.random(4))
```

> With regard to qiskit, I prefer my convention because it is a product of what we consider the standard matrices (Ry, Rz, Rx), but at the end of the day they are equivalent since there is only a relative complex phase between both cases.

Conventions are something we should probably discuss in more detail (for all gates). I believe that because of the `to_qasm` and `from_qasm` methods in the circuit, the best option is to use OpenQASM's conventions for all gates supported by this. However not that this may require to change some conventions we already have and it may create compatibility problems.

AdrianPerezSalinas:
Thanks for your reply, Stavros, I see it is a very good implementation, and it follows the standards of Qibo, so I am happy with that.

The only thing I see a little bit sloppy is the point on conventions. If you think the standard qasm convention is the best one, I have nothing to argue against that. You know much more about this than I do. When you talk about compatibility problems, are you talking about the code itself or about the examples?? In the case of the examples, maybe we could just re-do them, but it is possible that the code itself is harder to fix.

scarrazza:
LGTM, coverage however if failing.

AdrianPerezSalinas (commented):
Is the change U1 -> CPow done? I see plenty of lines where ZPow has been substituted by U1, but as Stefano mentioned, this brings compatibility problems

stavros11:
> LGTM, coverage however if failing.

Indeed there was an issue and the defaulteinsum/matmuleinsum backends weren't using the proper gates for ZPow/CZPow but this should be fixed now.

> Is the change U1 -> CPow done? I see plenty of lines where ZPow has been substituted by U1, but as Stefano mentioned, this brings compatibility problems

Yes, I changed U1 -> ZPow in most places but both gates are available and equivalent (doing `U1(0, theta=0.5)` is exactly the same with `ZPow(0, theta=0.5)`) so it would work even if I did not change. I just propose to use U1 in any future applications to be consistent but ZPow is also available and any older code should also work.

AdrianPerezSalinas (approved):

AdrianPerezSalinas:
Then I am happy with the changes

scarrazza:
Concerning the OpenQasm convention, in https://arxiv.org/pdf/1707.03429.pdf they define the U and CX, while other gates are compositions of U and CX, and many of them defined in the quantum experience header (qelib1.inc). @stavros11 is there the definition are you looking for, maybe eqs. 2-5?

stavros11:
> Concerning the OpenQasm convention, in https://arxiv.org/pdf/1707.03429.pdf they define the U and CX, while other gates are compositions of U and CX, and many of them defined in the quantum experience header (qelib1.inc). @stavros11 is there the definition are you looking for, maybe eqs. 2-5?

Thanks for the reference. I updated the U2 and U3 convention to follow Eq. (2) from the reference, since this also agrees with what Adrian wrote in #231. I believe our RX, RY, RZ rotations and U1 gate already agree with the paper, so all our gates follow the OpenQasm conventions.

src/qibo/tensorflow/gates.py
```diff
@@ -211,17 +211,38 @@ def construct_unitary(self) -> tf.Tensor:
...
+
+class U3(TensorflowGate, base_gates.U3):
+
+    def __init__(self, q, theta, phi, lam):
+        base_gates.U3.__init__(self, q, theta, phi, lam)
+        TensorflowGate.__init__(self)
+
+    def construct_unitary(self):
+        from qibo.tensorflow import cgates
+        return cgates.U3.construct_unitary(self)
```
scarrazza, line 245: Ok, here and in similar places I have a question/problem. I was trying to put Qibo inside a `tf.py_function` so I could use it inside a `tf.function` pipeline, as expected this is possible when using some tf backend different from custom, however if we have numpy calls in tensorflow/gates.py the py_function will crash, so I may suggest to have a "redundancy" and repeat the instructions using pure tf primitives.

src/qibo/base/gates.py
```diff
@@ -2,13 +2,17 @@
...
-              "cx": "CNOT", "swap": "SWAP",
-              "crz": "CZPow", "ccx": "TOFFOLI"}
-PARAMETRIZED_GATES = {"rx", "ry", "rz", "crz"}
+              "u1": "U1", "u2": "U2", "u3": "U3",
+              "cx": "CNOT", "swap": "SWAP", "cz": "CZ",
+              "crx": "CRX", "cry": "CRY", "crz": "CRZ",
+              "cu1": "CU1", "cu3": "CU3",
+              "ccx": "TOFFOLI"}
+PARAMETRIZED_GATES = {"rx", "ry", "rz", "u1", "u2", "u3",
+                      "crx", "cry", "crz", "cu1", "cu3"}
```
scarrazza, line 15: I think cu2 is missing here and above, right?

src/qibo/tensorflow/gates.py
stavros11: That's a good point, however could you find a `tf.function` script that actually crashes? For example I just tried:
```Python
@tf.function
def apply(state):
    gate = gates.U3(0, theta=0.1, phi=0.2, lam=0.3)
    return gate(state)

state = tf.cast(np.random.random(nqubits * (2,)), dtype=tf.complex128)
fstate = apply(state)
```
and it works both with and without `@tf.function`. Also `circuit.compile()` seems to work as well, for example:
```Python
c = Circuit(1)
c.add(gates.U3(0, theta=0.1, phi=0.2, lam=0.3))
c.compile()
fstate = c()
```
Regardless of that, it is probably better if all `construct_unitary` for Tensorflow gates return `tf.Tensor` instead of numpy arrays for consistency. This can be done either with repeating code (as you say) or just with casting `tf.cast(cgates.U3.construct_unitary(self), dtype=DTYPES.get('DTYPECPX'))`. Let me know what you think.

src/qibo/base/gates.py
stavros11: That's correct but it is actually on purpose. The reason is that this dictionary and set are only used by the `to_qasm`/`from_qasm` methods and I think that cu2 doesn't exist in Qasm. For example in the Qasm paper ([arXiv:1707.03429](https://arxiv.org/pdf/1707.03429.pdf)) both cu1 and cu3 are defined on page 12, however there is no reference to cu2 anywhere in the paper.
Also, when I export the Qasm code to the IBM experience online (for example to plot a circuit), it understands cu1 and cu3 but it doesn't understand cu2.

There are two options: If I add cu2 in the set then our Qasm exports may contain cu2's but these may not be understood by other libraries (like the IBM experience). If I leave it as it is and the user tries to do `to_qasm` in a circuit that contains CU2 it will raise an error that cu2 is not supported by Qasm. I am fine either way.

scarrazza: Ok, let's keep the qasm interface compatible with IBM for the time being, we can discuss about that later for the client-server configuration.

src/qibo/tensorflow/gates.py
scarrazza: Right, I think these functions are fine because we return a np.array. I think we should keep the initialization full numpy, to reduce the overhead that we have observed with tf primitives. I will open a PR if I get the `py_function` error again.

PR #245, opened 2020-09-24
Circuit inversion method

### LINKED ISSUE #237: Invert/reverse circuits
As discussed, a useful tool to implement in Qibo would be an easy way to invert, or reverse existing circuits.

In several occasions quantum circuits require the inverse of a particular part of the circuit in order to complete the algorithm. This is true, for example, for Shor's algorithm as well as most Grover based algorithms including Amplitude Estimation.

A quick way of getting a circuit you have already created and get its inverse would help circuit design. This would require applying the dagger of all gates of the initial circuit in reverse order.

stavros11:
Closes #237 by implementing the following changes:

* Add `.dagger()` method to all gates which creates a new gate object equivalent to the dagger of the original one.
* Add `.invert()` method to circuits which creates a new circuit object that is the inverse of the original one as described in #237 (dagger of all gates in reverse order).
* Add example of `circuit.invert()` in the docs ("How to invert a circuit?").
* Move some of the tests in `test_gates.py` to a new `test_circuits.py` file because the former was getting too large. Now `test_gates.py` contains tests that check specific gates, while `test_circuits.py` tests some general circuit functionality (compile, invert, set initial state, get final state, etc.).


scarrazza (approved):
LGTM.

scarrazza:
@igres26 is that sufficient for your applications?

igres26 (approved):
Everything looks fine.

PR #248, opened 2020-09-29
Defining Hamiltonians using sympy

stavros11:
This implements a new way for defining `Hamiltonian` and `TrotterHamiltonian` objects using the analytic form of the Hamiltonian written with `sympy` symbols. For more details on how this works I refer to the example I added in the docs, copied here for convenience. This shows how to construct the TFIM Hamiltonian using symbols:
```Python
import sympy
import numpy as np
from qibo import hamiltonians, matrices

# Define symbols for X and Z operators
x_symbols = sympy.symbols(" ".join((f"X{i}" for i in range(nqubits))))
z_symbols = sympy.symbols(" ".join((f"Z{i}" for i in range(nqubits))))

# Define Hamiltonian using these symbols
symbolic_ham = sum(z_symbols[i] * z_symbols[i + 1] for i in range(nqubits - 1))
symbolic_ham += z_symbols[0] * z_symbols[-1]
symbolic_ham += sum(x_symbols)

# Define a map from symbols to the matrices they represent
symbol_map = {x: (i, matrices.X) for i, x in enumerate(x_symbols)}
symbol_map.update({x: (i, matrices.Z) for i, x in enumerate(z_symbols)})

# Define ``Hamiltonian`` or ``TrotterHamiltonian``
ham = hamiltonians.Hamiltonian.from_symbolic(symbolic_ham, symbol_map)
trotter_ham = hamiltonians.TrotterHamiltonian.from_symbolic(symbolic_ham, symbol_map)
``` 

Because the symbolic form is the way one would write the Hamiltonian on paper, I believe the new methods will be very useful for defining custom Hamiltonians that we do not have pre-coded in Qibo. Moreover, now Qibo essentially does the Trotterization automatically as the user can define the `TrotterHamiltonian` using symbols, without worrying about how to split the terms, etc..

Particularly this PR implements the following:
* Adds `from_symbolic` method in both `Hamiltonian` and `TrotterHamiltonian` which allows constructing these Hamiltonians using symbolic forms.
* Removes the `TrotterHamiltonian.from_twoqubit_term` constructor which was not very useful and in fact was not working properly for odd qubit numbers. This also fixed an inconsistency we had in the accuracy of evolution for odd qubits. Now both even and odd qubit numbers have an O(dt^2) error when using Trotter.
* Adds `TrotterHamiltonian.from_dictionary` method which allows the user to define a Trotter Hamiltonian using a single dictionary of terms instead of manually splitting to parts in order to respect the commutation relations. The splitting is now done automatically by `TrotterHamiltonian._split_terms`. Note that this method splits the terms by looking at the target qubits only and not whether the corresponding matrices commute.
* Adds `TrotterHamiltonian` tests for Hamiltonian with 3-qubit interactions. These work only with Tensorflow backends because the custom operators support up to two-qubit gates.
* Adds `is_compatible` and `make_compatible` methods to `TrotterHamiltonian`. Two Trotter Hamiltonians are compatible when they have the same "part structure" and this is a requirement in order to add them. The `make_compatible` method is useful for `AdiabaticEvolution` where `h0` and `h1` should be added. Previously the user had to make sure the Hamiltonians were compatible but now, if the Hamiltonians given are not compatible, this is fixed automatically during the evolution model initialization.


doc/source/advancedexamples.rst
```diff
@@ -950,3 +950,89 @@ Note that the ``minimize`` method optimizes both the free parameters ``p`` of
...
+qubits ``nqubits`` could be constructed as:
+
+.. code-block::  python
+
+    import sympy
+    import numpy as np
+    from qibo import hamiltonians, matrices
+
+    # Define symbols for X and Z operators
+    x_symbols = sympy.symbols(" ".join((f"X{i}" for i in range(nqubits))))
```
scarrazza, line 1013: The nqubits definition is missing.

doc/source/advancedexamples.rst
```diff
@@ -950,3 +950,89 @@ Note that the ``minimize`` method optimizes both the free parameters ``p`` of
...
 for the total evolution time should be the last value of the given
 ``initial_guess`` array.
+
+
+.. _symbolicham-example:
+
+How to define custom Hamiltonians using symbols?
+------------------------------------------------
+
+In order to use the VQE, QAOA and time evolution models in Qibo the use has to
```
scarrazza, line 960:
```suggestion
In order to use the VQE, QAOA and time evolution models in Qibo the user has to
```

doc/source/advancedexamples.rst
```diff
@@ -950,3 +950,89 @@ Note that the ``minimize`` method optimizes both the free parameters ``p`` of
...
+define Hamiltonians based on :class:`qibo.base.hamiltonians.Hamiltonian`, or
+:class:`qibo.base.hamiltonians.TrotterHamiltonian` when the Trotter
+decomposition is to be used. Qibo provides pre-coded Hamiltonians for some
+common physics models, such as the transverse field Ising model (TFIM) and the
+Heisenberg model. In order to explore other problems the user needs to define
+the Hamiltonian objects from scratch.
```
scarrazza, line 966: I would add a reference to the precoded hamiltonians.

scarrazza:
I did some tests and seems that sympy is able to generate hamiltonians up to 15 qubits (my memory limit) in a reasonable time.

src/qibo/base/hamiltonians.py
```diff
@@ -38,6 +39,29 @@ def __init__(self, nqubits, matrix, numpy=False):
...
 
+    @classmethod
+    def from_symbolic(cls, symbolic_hamiltonian, symbol_map, numpy=False):  # pragma: no cover
+        """Creates a ``Hamiltonian`` from a symbolic Hamiltonian.
+
+        We refer to the :ref:`How to define custom Hamiltonians using symbols? <symbolicham-example>`
+        example for more details.
+
+        Args:
+            symbolic_hamiltonian: The full Hamiltonian written with symbols.
```
scarrazza, line 50: I would include a `sympy.symbol` here.

src/qibo/base/hamiltonians.py
```diff
@@ -187,6 +211,179 @@ def __matmul__(self, o):
                                              "implemented.".format(type(o)))
 
 
+class SymbolicHamiltonian:
+    """Parses symbolic Hamiltonians defined using ``sympy``.
+
+    This class should not be used by users.
```
scarrazza, line 217: Shall we rename this class as `_SymbolicHamiltonian`?

src/qibo/base/hamiltonians.py
```diff
@@ -187,6 +211,179 @@ def __matmul__(self, o):
...
+    Args:
+        symbolic_hamiltonian: The full Hamiltonian written with symbols.
+        symbol_map (dict): Dictionary that maps each symbol to a pair of
+            (target, matrix).
+    """
+    import sympy
+    from qibo import matrices
+
+    def __init__(self, hamiltonian, symbol_map):
+        self.symbolic = self.sympy.expand(hamiltonian)
```
scarrazza, line 234: Where is this `self.sympy` defined?
Shall we include checks for the hamiltonian and symbol_map types?

src/qibo/base/hamiltonians.py
```diff
@@ -187,6 +211,179 @@ def __matmul__(self, o):
...
+            if term.args:
+                expression = term.args
+            else:
+                expression = (term,)
+
+            symbols = [x for x in expression if x.is_symbol]
+            numbers = [x for x in expression if not x.is_symbol]
+            if numbers:
+                assert len(numbers) == 1
+                const = float(numbers[0])
```
scarrazza, line 250: Is float save or shall we prefer DTYPE (numpy)?

scarrazza (approved):

src/qibo/base/hamiltonians.py
stavros11: It is not appearing in the docs anyway, but I renamed it using the `_` since it is not supposed to be used by the user.

src/qibo/base/hamiltonians.py
stavros11: It was working with `float` but I changed that to `self.matrices.dtype` since this number is multiplied/added to matrices. I am not really sure which choice is generally safer/better.

doc/source/advancedexamples.rst
stavros11: Following our discussion yesterday, I changed this example to be specific for four qubits. This avoids the strange `join` manipulations of strings and should make it more accessible to beginners. I also changed notations slightly and now the `sympy` construction is pretty close to the "paper" form of the Hamiltonian.

stavros11:
Thanks for the comments. I fixed the suggestions and I will be adding some checks in the `_SymbolicHamiltonian` initialization now.

> I did some tests and seems that sympy is able to generate hamiltonians up to 15 qubits (my memory limit) in a reasonable time.

Some numbers regarding this on the TFIM example:

* Constructing the full Hamiltonian (`hamiltonians.Hamiltonian`)

nqubits | default `np.kron` | using `sympy`
-- | -- | --
3 | 0.00053 | 0.02102
4 | 0.00087 | 0.01970
5 | 0.00141 | 0.02109
8 | 0.02612 | 0.04758
9 | 0.14513 | 0.16033
10 | 0.61667 | 0.66452
12 | 16.56261 | 17.40988
13 | 79.23401 | 82.57881

* Constructing the Trotterized Hamiltonian (`hamiltonians.TrotterHamiltonian`)

nqubits | default construction | using `sympy`
-- | -- | --
5 | 0.00018 | 0.02195
6 | 0.00017 | 0.02308
7 | 0.00022 | 0.02582
10 | 0.00016 | 0.02652
11 | 0.00017 | 0.02768
12 | 0.00016 | 0.02873
15 | 0.00018 | 0.03142
20 | 0.00019 | 0.03672
25 | 0.00019 | 0.04436
28 | 0.00020 | 0.04784
29 | 0.00019 | 0.04924
30 | 0.00020 | 0.05185

I agree that the `sympy` overhead is small and it is certainly justified by the simplification it provides in defining Hamiltonians.

src/qibo/base/hamiltonians.py
stavros11: > Where is this `self.sympy` defined?

`sympy` is imported within the `_SymbolicHamiltonian` class.

> Shall we include checks for the hamiltonian and symbol_map types?

I added some checks that the given arguments have the proper types and format. Let's wait for the coverage check too.

scarrazza:
Thanks, looks very good now. Let's merge this.

PR #249, opened 2020-10-01
Fix issue with `TrotterHamiltonian.make_compatible`

stavros11:
After playing with an example application I realized that the `TrotterHamiltonian.make_compatible` method implemented in #248 was not working as expected in a specific case. This may lead to problems when using Trotterized Hamiltonians for adiabatic evolution. This PR fixes this issue and adds a test that captures the problematic case I found.

Tomorrow I will attempt to update our adiabatic3sat example in order to use the new Hamiltonian functionalities (symbolic & `make_compatible`) implemented in #248. This may help in identifying any other issues if they exist, ideally before the 0.1.1 release.


scarrazza (approved):

stavros11:
Thanks for the review. 

Just a small update: I was updating the adiabatic3sat example to use `from_symbolic` and found some other small issues. Particularly I think it crashes when the Hamiltonian contain power symbols (eg. `z**2`). I also think that it is possible to improve performance of Trotterization by re-using objects in cases where all term matrices are the same. I am currently looking at these issues.

scarrazza:
Shall we merge?

stavros11:
> Shall we merge?

Yes, I think this is good as a first implementation of Hamiltonian creation from symbols.

PR #250, opened 2020-10-03
Update adiabatic 3SAT example and fix `from_symbolic` issues

stavros11:
Following #249, this updates our adiabatic3sat example to use the new symbolic Hamiltonian functionality. It also changes the parsing of symbolic Hamiltonians because there were a few issues with the previous implementation, particularly it was failing when the symbolic Hamiltonian contained powers, eg. `H = Z1 ** 2 + Z2 ** 2`. The current version seems to work in all cases I tried, when the Hamiltonian is a polynomial function of tensor products of one-qubit matrices. I believe it is almost always the case.

As you can see in the 3SAT example code (in `functions.py`) using `from_symbolic` significantly simplifies the construction of custom Hamiltonians since we avoid many of the complicated functions we previously had there. However there is a performance issue that remains. Here are some benchmarks of the `adiabatic3sat/main.py` using Trotterized Hamiltonian on CPU:

nqubits | master | symbolic (this PR) |
-- | -- | --
4 | 1.89493 | 2.81755
8 | 9.10203 | 15.76547
10 | 11.64693 | 20.20957
12 | 13.97034 | 23.07952
14 | 14.59212 | 23.76182
16 | 18.73169 | 30.16241
20 | 51.17262 | 73.60861

I thought that this is because `from_symbolic` recreates some Hamiltonian objects during the `TrotterHamiltonian` construction, however a fix that I had in mind did not help much, so I am still investigating how to improve on this.


scarrazza:
Thanks. The performance deterioration is dominated just by the hamiltonian/sympy construction, right?

stavros11:
> Thanks. The performance deterioration is dominated just by the hamiltonian/sympy construction, right?

This what I believe as well, however I would like to do a few more tests to confirm this. I am not sure if at these sizes sympy alone can justify this difference because the Hamiltonians do not have that many terms (eg. the 20 qubit one is just 17 clauses).

Another reason for the difference could be that in the master we exploit the fact that the same Hamiltonian term is applied to all qubits. To demonstrate this better, for example for the TFIM there are two ways to construct the `TrotterHamiltonian`:
```Python
matrix = np.kron(matrices.Z, matrices.Z) + np.kron(matrices.X, matrices.I)
# Use the same term for all qubits
term = Hamiltonian(2, matrix, numpy=True)
term_dict = {(i, i + 1): term for i in range(nqubits - 1)}
term_dict[(nqubits - 1, 0)] = term
trotter_ham = TrotterHamiltonian.from_dictionary(term_dict)

# Use different terms
term_dict = {(i, i + 1): Hamiltonian(2, matrix, numpy=True) for i in range(nqubits - 1)}
term_dict[(nqubits - 1, 0)] = Hamiltonian(2, matrix, numpy=True)
trotter_ham = TrotterHamiltonian.from_dictionary(term_dict)
```
Clearly the first approach is more efficient because it calculates the `exp(-1j * dt * matrix)` only once and uses it for all pairs, while the second will repeat the same calculation many times. This is just a 4x4 matrix so it is not a huge exponentiation but there will still be a difference. I think that `from_symbolic` currently follows the second method. I fixed this in another branch but there was no improvement in the 3SAT performance.

scarrazza:
Ok, you could try to profile the symbolic to numeric transformation in the `_SymbolicHamiltonian`, there might be some specific symbolic operation or combination of operations which might be inefficient.

stavros11:
> Ok, you could try to profile the symbolic to numeric transformation in the `_SymbolicHamiltonian`, there might be some specific symbolic operation or combination of operations which might be inefficient.

I did some profiling and the problem seems to not be related to sympy. Particularly I observe that the matrix exponentiation (`expm`) function has cumtime 2.6sec for 10 qubits on my local machine when using master while this is 16sec when using this branch. Therefore I believe the main performance issue is what I wrote above. I pushed a fix that reduces the `expm` time to 8sec for this example but I will have a more detailed look a bit later.

scarrazza:
Great, let us know if you find further evidences.

scarrazza (approved):
Thanks, with these changes, the example looks really much easier to understand and follow.

stavros11:
I reverted the last commit because it seemed to break the example in terms of finding the right bitstring solution. At the current state this PR should work okay modulo the performance issue mentioned in the first post. If you like we can merge this to #249 to avoid confusion and I can open a new PR with the performance fix once we have that. Otherwise I will push it directly here.

scarrazza:
Ok, lets keep this open until the performance is sorted out.

stavros11:
I fixed the issue and now the 3SAT solution agrees with master, so there are no issues in terms of result. 

In terms of performance, I played a bit with reducing the duplicate Hamiltonian objects created by `make_compatible` and could improve performance slightly, but still not reach the level of master. Here is the updated version of the table from the first post:

nqubits | master | symbolic (first post) | symbolic (latest push)
-- | -- | -- | --
4 | 1.89493 | 2.81755 | 1.87313
8 | 9.10203 | 15.76547 | 12.55959
10 | 11.64693 | 20.20957 | 15.52563
12 | 13.97034 | 23.07952 | 21.11225
14 | 14.59212 | 23.76182 | 21.23374
16 | 18.73169 | 30.16241 | 23.90576
20 | 51.17262 | 73.60861 | 66.17026

According to the profiling, now `hamiltonian.exp` is called the same number of times from this branch and master, however `expm` is called ~6x more from this branch for (for 12 qubits). This fully explains the performance difference shown in the table. It happens because the custom functions used to create the Hamiltonians on master take better advantage of caching compared to the "automatic" creator in `make_compatible`. 

Note that the performance difference is problem dependent. For example I tried adiabatic evolution with TFIM as H1 and there is no difference between this PR and master, because in this case it happens that the automatic Hamiltonian agrees with the one from master.

scarrazza:
Ok, so in summary the symbolic approach introduces a tiny overhead due to extra calls / missed caching. I think that is an acceptable compromise, however I would suggest to highlight this issue in the docs, saying that the numerical approach may be faster than the symbolic one.

stavros11:
> Ok, so in summary the symbolic approach introduces a tiny overhead due to extra calls / missed caching. I think that is an acceptable compromise, however I would suggest to highlight this issue in the docs, saying that the numerical approach may be faster than the symbolic one.

Just for a check, I tried using the normal sum(X) Hamiltonian in the 3SAT example (instead of normalizing using `times`) and performance is improved by a few seconds however it is still not as fast as master. Therefore what I wrote in the previous post is not really true.

In any case, I updated the docs in several places (in `make_compatible` docstring, in the adiabatic evolution how-to example and in the symbolic Hamiltonians how-to example) mentioning that making the Hamiltonians compatible manually when using Trotter decomposition for adiabatic evolution may lead to better performance than relying on this automatic algorithm. I will keep doing some tests to find other cases where these performance differences exist, however if you believe the difference is acceptable then this PR should be okay now.

scarrazza:
Great, thanks. I think we can merge this now.

igres26 (approved):
Yes, I think this is a good addition.

PR #251, opened 2020-10-03
changing documentation urls from latest to stable

### LINKED ISSUE #246: Adjust docs urls
Before releasing 0.1.1 we should update all urls to point to the stable readthedocs pages.

scarrazza:
Closes #246.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,11 +17,11 @@ Some of the key features of Qibo are:
 
 Qibo documentation is available at [qibo.readthedocs.io](https://qibo.readthedocs.io).
 
-- [Installation](https://qibo.readthedocs.io/en/latest/installation.html)
+- [Installation](https://qibo.readthedocs.io/en/stable/installation.html)
 - [Documentation](https://qibo.readthedocs.io/)
-- [Components](https://qibo.readthedocs.io/en/latest/qibo.html)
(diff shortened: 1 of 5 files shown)


stavros11 (approved):
Thanks for the update and the typo fixes. I could not find any other "latest" references in our docs. 

Just two questions:
* The documentation badge in ` README.md` is still using the latest. Is this what we want?
* When accessing the `https://qibo.readthedocs.io/` webpage without specifying latest or stable I am automatically redirected to latest. Is there a way to switch that to stable?

scarrazza:
> The documentation badge in README.md is still using the latest. Is this what we want?

yes, because we can quickly see if there troubles with the latest master documentation creation.

> When accessing the https://qibo.readthedocs.io/ webpage without specifying latest or stable I am automatically redirected to latest. Is there a way to switch that to stable?

yes, I will change the default behaviour of readthedocs to show stable instead of latest by default.

stavros11:
Just a reminder that in #238 @DiegoGM91 checked the how-to examples in our documentation up to the QAOA section. Shall I make a pass over the remaining sections before the v0.1.1 release or has this already been done?

scarrazza:
Yes, please, that's a good point.

Issue #244, opened 2020-09-24
Check openqasm is cross platform


scarrazza:
Following our discussion, we should try to check that the qasm generated by Qibo is compatible with other providers.
I believe would be great to have tests for the conversion between frameworks.

scarrazza:
@stavros11 I have performed some tests and indeed the qasm generated from Qibo seems fully compatible with Qiskit. I have tested the qasm implementation in Cirq, however it does not support all gates, see table [here](https://github.com/quantumlib/Cirq/issues/1548#issuecomment-485039043).

If you agree, I will include minimal tests for Cirq and more extensive tests using Qiskit (creating an extra tests dependency).

stavros11:
> If you agree, I will include minimal tests for Cirq and more extensive tests using Qiskit (creating an extra tests dependency).

Thanks for checking this, it would be good to have some tests. Looking at Cirq's table, I should also note that we also do not support all the Qasm gates. For example all commands denoted as NOT supported by Cirq (gate, opaque, reset, if and barrier) are also not supported from Qibo. Let me know if you believe that it would be useful to add support for some of these operations.

PR #252, opened 2020-10-03
adding qibo to cirq qasm tests

scarrazza:
Implements a very basic test for #244 using cirq.
This import qasm generated from Qibo in Cirq and compare the final wavefunction.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -80,7 +80,7 @@ setup(
     install_requires=requirements,
     extras_require={
         "docs": ["sphinx", "sphinx_rtd_theme", "recommonmark", "sphinxcontrib-bibtex", "sphinx_markdown_tables"],
-        "tests": ["cirq"],
+        "tests": ["cirq", "ply"],
     },
     python_requires=">=3.6.0",
     long_description=long_description,
--- a/src/qibo/tests/test_qasm.py
+++ b/src/qibo/tests/test_qasm.py
@@ -1,6 +1,9 @@
 import pytest
+import numpy as np
 from qibo.models import Circuit
 from qibo import gates, __version__
+import cirq
+from cirq.contrib.qasm_import import circuit_from_qasm, exception
 
 
 def assert_strings_equal(a, b):
@@ -30,6 +33,17 @@ qreg q[2];
 h q[0];
 h q[1];"""
     assert_strings_equal(c.to_qasm(), target)
+
+
+def test_simple_cirq_qiskit(atol=1e-7):
+    c1 = Circuit(2)
+    c1.add(gates.H(0))
+    c1.add(gates.H(1))
+
+    c2 = circuit_from_qasm(c1.to_qasm())
+    final_state_c1 = c1()
+    final_state_c2 = cirq.Simulator().simulate(c2).final_state
+    np.testing.assert_allclose(final_state_c1, final_state_c2, atol=atol)
 
 
 def test_unknown_gate_error():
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -80,7 +80,7 @@ def finalize_options(self):
     install_requires=requirements,
     extras_require={
         "docs": ["sphinx", "sphinx_rtd_theme", "recommonmark", "sphinxcontrib-bibtex", "sphinx_markdown_tables"],
-        "tests": ["cirq"],
+        "tests": ["cirq", "ply"],
```
stavros11, line 83: I noticed that it is required to install `ply` in order to run these tests but this is due to a `cirq` bug right? Shouldn't this be in Cirq's requirements?

src/qibo/tests/test_qasm.py
```diff
@@ -90,6 +118,20 @@ def test_toffoli():
     assert_strings_equal(c.to_qasm(), target)
 
 
+def test_toffoli_cirq(atol=1e-7):
```
stavros11, line 121: When the same `atol` is required for several tests in a file, I usually define a global `_atol = 1e-7` variable and use `atol=_atol` in each function. This is in order to try different `atol` values without having to change each test separately.

src/qibo/tests/test_qasm.py
```diff
@@ -32,6 +35,17 @@ def test_simple():
     assert_strings_equal(c.to_qasm(), target)
 
 
+def test_simple_cirq_qiskit(atol=1e-7):
```
stavros11, line 38: Is there any reason we mention `qiskit` here? We are not testing anything with qiskit right?

src/qibo/tests/test_qasm.py
```diff
@@ -69,6 +83,20 @@ def test_multiqubit_gates():
     assert_strings_equal(c.to_qasm(), target)
 
 
+def test_multiqubit_gates_cirq_qiskit(atol=1e-7):
```
stavros11, line 86: Same comment about `qiskit`.

src/qibo/tests/test_qasm.py
```diff
@@ -90,6 +118,20 @@ def test_toffoli():
...
+    c1 = Circuit(3)
+    c1.add(gates.Y(0))
+    c1.add(gates.TOFFOLI(0, 1, 2))
+    c1.add(gates.X(1))
+    c1.add(gates.TOFFOLI(0, 2, 1))
+    c1.add(gates.Z(2))
+    c1.add(gates.TOFFOLI(1, 2, 0))
+    c2 = circuit_from_qasm(c1.to_qasm())
+    final_state_c1 = c1()
+    final_state_c2 = cirq.Simulator().simulate(c2).final_state
```
stavros11, line 131: Just noting that in this circuit only the last Toffoli will act. This is fine because we test both cases of active/inactive controls, I am just mentioning it because sometimes test circuits are trivial and they end up not testing the gate we intend for. A simple way around this is to use a random initial state but this is not required here.

stavros11 (approved):
Thanks for adding these tests. Looks good to me, the comments are just for naming conventions.

setup.py
scarrazza: I think so, on my laptop installing cirq seems to be sufficient, however the github actions are failing without this adding this dependency explicitly.

src/qibo/tests/test_qasm.py
scarrazza: Sure, good point, thanks.

scarrazza:
Thanks for the review. I have include the inverse test, namely take qasm from Cirq and import in Qibo.
I had to include a parsing for strings like ['pi*0.38'] which are understood by Cirq and Qiskit.
Please let me know if you agree, I believe we can consider this implementation acceptable for a basic qasm check.

stavros11:
Thanks for the updates. Yes, I agree that this is acceptable as a qasm cross-platform check with Cirq.

Regarding the `pi` parsing, it looks good. In principle we could add an explicit test for this as well but since it is tested from the Cirq tests, if you think this is sufficient that's fine with me.
I noticed that Cirq's `to_qasm` prefers to use this `gate(pi * float)` convention, while our `to_qasm` just prints the value of the parameter without dividing by `pi`. Do you think we should follow the same as convention in `to_qasm` as well? Interestingly (since tests pass) our current `to_qasm` convention works with Cirq.

scarrazza:
Thanks, I think the current convention is ok, as far as the tests are passing.

PR #254, opened 2020-10-05
Allow adding circuits to larger ones

### LINKED ISSUE #253: Adding circuits to larger ones
When building large circuits with a lot of parts, it is sometimes useful to add up existing circuits to the larger one. A common occurrence might be adding a QFT to a subset of the qubits in the circuit. 

As of right now, there seems to only be the possibility of adding circuits with the same amount of set qubits.

A library of useful and repeating circuits, like QFT, a VQE layer, the Diffusion transform in Grover with an easy way to append them into your bigger circuit will be a useful addition to Qibo. Adding also the inverse of said circuits is also really common.

stavros11:
Closes #253 by implementing a `circuit.on_qubits(*q)` method on circuits that returns a generator of gates that exist in `circuit` acting on the specified `qubits`.

Example:
```Python

# Create a large circuit on 8 qubits
circuit = models.Circuit(8)
# Add a QFT on even qubits
circuit.add(models.QFT(4).on_qubits(*range(0, 8, 2)))
# Add rotations to all 8 qubits
circuit.add((gates.RX(i, theta=0) for i in range(8)))
# Add an inverse QFT on odd qubits
circuit.add(models.QFT(4).invert().on_qubits(*range(1, 8, 2)))
```

Regarding the performance issue I mentioned in #253, I did some quick benchmarks on this on my local laptop. Particularly, I compare the above script with creating the QFT and inverse QFT gates with a custom generator from scratch (instead of using the QFT model and `on_qubits`). I see no difference in execution time, but as expected there is some difference in circuit creation time.

Creation time
nqubits | custom creation | using `on_qubits` |
-- | -- | --
10 | 0.01617 | 0.02304
12 | 0.00878 | 0.01711
14 | 0.01048 | 0.01804
16 | 0.01176 | 0.02001
18 | 0.02110 | 0.02792
20 | 0.01309 | 0.03663
22 | 0.02233 | 0.03643
24 | 0.02970 | 0.04299
26 | 0.03967 | 0.07076

Execution time
nqubits | custom creation | using `on_qubits` |
-- | -- | --
10 | 0.00409 | 0.00301
12 | 0.00456 | 0.00292
14 | 0.00444 | 0.00381
16 | 0.00871 | 0.00688
18 | 0.02110 | 0.03392
20 | 0.17917 | 0.17245
22 | 0.93621 | 0.99574
24 | 4.46404 | 4.38530
26 | 19.7839 | 18.7598

Given that that the difference is small and typically creation time << execution time and also that Qibo provides several functions to avoid circuit re-creation whenever possible (such as `set_parameters`), I believe using `on_qubits` should be fine in terms of performance.

### DIFF
--- a/doc/source/advancedexamples.rst
+++ b/doc/source/advancedexamples.rst
@@ -335,7 +335,33 @@ for example:
 
     # Create the total circuit as subroutine + middle + subroutine^{-1}
     circuit = subroutine + middle + subroutine.invert()
+    
 
+Note that circuit addition works only between circuits that act on the same number
+of qubits. It is often useful to add subroutines only on a subset of qubits of the 
+large circuit. This is possible using the :meth:`qibo.base.circuit.BaseCircuit.on_qubits`
+method. For example:
+
+.. code-block:: python
+
+    import numpy as np
+    from qibo.models import Circuit
+    from qibo import gates
+    
+    # Create a small circuit of 4 qubits
+    smallc = Circuit(4)
+    smallc.add((gates.RX(i, theta=0.1) for i in range(4)))
+    smallc.add((gates.CNOT(0, 1), CNOT(2, 3)))
+
+    # Create a large circuit on 8 qubits
+    largec = Circuit(8)
+    # Add the small circuit on even qubits
+    largec.add(smallc.on_qubits(*range(0, 8, 2)))
+    # Add a QFT on odd qubits
+    largec.add(models.QFT(4).on_qubits(*range(1, 8, 2)))
+    # Add an inverse QFT on first 6 qubits
+    largec.add(models.QFT(6).invert().on_qubits(*range(6)))
+    
 
 .. _vqe-example:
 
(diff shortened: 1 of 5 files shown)


scarrazza (approved):
Thank you, this looks very clean and I agree that from a performance point of view this is quite good.

igres26 (approved):
This does look good to me as well. I assume that the given qubits have to be in order of appearence within the circuit creation right? Qubit '0' in the circuit will be the first one you give to the `.on_qubits()` module?

stavros11:
Thanks for the reviews.

> This does look good to me as well. I assume that the given qubits have to be in order of appearence within the circuit creation right? Qubit '0' in the circuit will be the first one you give to the `.on_qubits()` module?

Yes, if the small `circuit` acts on `n` qubits then you can think of `circuit.on_qubits()` as an n-qubit gate so you have to specify n numbers when you call it. The first of these numbers will be treated as the qubit-0 of the small circuit, the second as the qubit-1, etc. until qubit-n.

PR #259, opened 2020-10-12
Fix issues with `TrotterHamiltonian.from_symbolic` and `make_compatible`

stavros11:
This PR fixes some issues related to the construction of Trotter Hamiltonians using symbols and the `make_compatible` function. Particularly:

* `TrotterHamiltonian.from_symbolic` has a mechanism that merges one- and two-qubit terms to two-qubit ones for efficiency (in order to apply less unitary gates). This is implemented in `_SymbolicHamiltonian._merge_one_qubit` but it was not working properly for some Hamiltonians because it was not flipping some two-qubit terms in order to merge all one-qubit terms. Here I used a recursive approach to implement this mechanism from scratch which I believe will do it properly. It works for the Hamiltonians we tried.
* In some cases, the `make_compatible` method was skipping terms that did not exist in the one-qubit Hamiltonian and as a result the returned Hamiltonian was not actually compatible with the two-qubit one. I changed the way `make_compatible` works and this issue should be fixed.

Many thanks to @igres26 for pointing out these issues.


scarrazza:
Thanks @stavros11 and @igres26 for spotting this issue. 
Are the new tests covering the bad configurations for both points you have quoted?

stavros11:
> Thanks @stavros11 and @igres26 for spotting this issue.
> Are the new tests covering the bad configurations for both points you have quoted?

I have been checking the Hamiltonians of interest with some local scripts, however it is indeed good to have it in our tests, so I just pushed some related tests.

src/qibo/tests/regressions/trotter_adiabatic_bfgs.out
```diff
@@ -1,2 +1,2 @@
-5.145435393136968960e-01
-2.009895284405383897e+00
+5.149684200950115365e-01
```
scarrazza, line 1: Any special motivation to update this file?

stavros11: This PR changes how the `TrotterHamiltonian` is constructed from symbols and the `make_compatible` so the grouping of terms in the Trotter decomposition may be slightly different and thus lead to slightly different evolution (with the difference being smaller than the expected Trotter error). This also changes slightly the optimal parameters found by this optimization test. The difference between old and new values is small (<1e-3) so I believe it can be explained from this.

stavros11:
Small note not relevant to this PR: Sometimes coverage results are not correct. For example a few commits above (in commit `e9fc0af8d8f212157933571495d3cd2804c1e3bf`) there was the following piece in the code
```Python
if len(pair_sets[target]) == 1:
    if not pair_sets[target]:
        return None
```
where obviously the line under `if` cannot run, however coverage passed somehow.

PR #261, opened 2020-10-12
Fix circuit depth calculation

stavros11:
Many thanks to @igres26 again for pointing out that the depth calculation was not correct. This PR fixes the issue and adds comparisons to depths calculated by Cirq in both `test_cirq.py` and `test_qasm.py` tests. 

If anyone has circuits available in either Qibo, Qasm or Cirq, we can cross-check that the depth calculated by Qibo agrees with Cirq in these cases as well.


scarrazza (approved):
Thanks, the changes look good.

PR #262, opened 2020-10-13
State collapse gate

### LINKED ISSUE #258: Collapse state during measurements
Currently the measurement gate works by sampling the state vector and returning the sampled bit strings, however it does not have any effect on the state vector. This means that the final state of a circuit that contains measurements (which can be accessed as `circuit.final_state` after execution) is the same with the final state that we would get if the circuit contained no measurements. 

The state collapse is not done for two reasons:
1. It is not uniquely defined if `nshots > 1`. We would need to collapse to `nshots` different states and this would be memory inefficient. 
2. Not collapsing the state allows re-sampling it in case more measurement shots are needed.

On the other hand collapsing the state when `nshots = 1` is useful for some applications. I have an implementation of the collapse mechanism based on Tensorflow primitives that I will probably extend to custom operators for efficiency. The main question is what would be a good interface for this. I can think of the following possibilities:

* Add a new `gates.Collapse(*q, result)` gate that collapses qubits `*q` to the given `result` bit-string. This could be added in any place in the circuit as any other gates, but not after measurements on `*q` because currently we do not allow gates on measured qubits (after the measurement). This would be a bit counter-intuitive, given that the collapse typically happens after measurement, but it could still be a useful interface due to its flexibility.
* Integrate the collapse mechanism in the existing measurement gate `gates.M`, for example via a flag: `gates.M(*q, collapse=True)`. The collapse would only be possible when `nshots = 1` and would collapse the state to the measured result (which is selected probabilistically). This is less flexible than the previous interface which allowed the user to select the bit string to collapse to.

The second approach is also slightly more complicated in terms of implementation, because currently when `gates.M` is called on a state vector it returns a measurement object (and not a new state vector). All other gates return a state. If we do the collapse we would need to modify `gates.M` so that it returns the collapsed state and access the actual measurement result in a different way.

I am open to any other suggestions regarding an interface of this that you think it would be useful for applications.

stavros11:
Closes #258 by implementing a `Collapse(*q, result)` gate. A new custom operator is used for the `"custom"` backend and an implementation based on Tensorflow primmitives is used by the `"defaulteinsum"`, `"matmuleinsum"` backends. As explained in #258 I follow an approach where the Collapse gate is independent of measurements and can be used anywhere in the circuit. Depending on feedback from potential use in applications we may change this in the future and integrate the state collapse within measurements.

I am opening this as a draft in case anyone wants to check the custom operator code. I still need to implement the equivalent CUDA kernel and add some documentation for the new gate. I will mark this as ready for review once these are done.


stavros11:
@scarrazza, a small question regarding the GPU implementation. The new kernel takes the following arguments:
```C++
REGISTER_OP("CollapseState")            \
    .Attr("T: {complex64, complex128}") \
    .Input("state: T")                  \
    .Input("qubits: int32")             \
    .Input("result: int64")             \
    .Attr("nqubits: int")               \
    .Output("out: T")                   \
    .SetShapeFn(::tensorflow::shape_inference::UnchangedShape);
```
so it is very similar to the gate kernels we already have. However, when executing on GPU it seems that the `state` tensor is not read properly. For example, if I try doing:
```C++
std::cout << state.flat<T>().data()[0];
```
inside the [`Compute`](https://github.com/Quantum-TII/qibo/blob/3958c672058222bb4a3c3d56a497ad32fed9e2ca/src/qibo/tensorflow/custom_operators/cc/kernels/apply_gate_kernels.cc#L380) function of the `CollapseStateOp` I get a segmentation fault. The same code works without any problem when executing on CPU.

I thought that this was an issue with my CUDA kernel (which is not pushed here yet), however note that this segmentation fault happens in the Op, before calling the actual kernel. This code is shared between CPU and GPU and CPU works, so I am not sure what exactly is the issue. Do you have any idea regarding this?

scarrazza:
Yes, this happens because this data array is allocated with the cuda malloc, directly on GPU, so if you try to access data outside the cuda kernel the code crashes with segfault . If I remember correctly, for the current operators we tried to have/pass local CPU copies to avoid this issue.

stavros11:
> Yes, this happens because this data array is allocated with the cuda malloc, directly on GPU, so if you try to access data outside the cuda kernel the code crashes with segfault . If I remember correctly, for the current operators we tried to have/pass local CPU copies to avoid this issue.

Thanks for the response. Yes, I finally realized this some time ago and fixed the issue. A problem that remains is how to calculate the norm in order to normalize the collapsed state on GPU. On CPU it worked by defining a `norm` variable outside the parallel-for, however it is not possible to do exactly the same for GPU. As a temporary solution I do the normalization using tf when using GPU and I was planning to do some benchmarks to see how this affects speed/memory. I will mark this ready once I have these results.

Regarding the tests, there still seems to be an issue with the examples.

stavros11:
I did some quick benchmarks on both CPU and GPU and interestingly the implementation based on tf primitives is faster than the custom operator for this particular gate.

On CPU:
nqubits | custom op | native tf
-- | -- | --
27 | 2.31419 | 0.9235
28 | 4.48814 | 1.84209
29 | 8.74110 | 3.75925
30 | 18.68336 | 7.48180

On GPU:

nqubits | custom op (no norm) | custom op (tf norm) | native tf
-- | -- | -- | --
27 | 0.00789 | 0.00928 | 0.00653
28 | 0.01527 | 0.01637 | 0.01259
29 | 0.03058 | 0.03246 | 0.02802
30 | 0.06118 | OOM | OOM

As expected the custom operator thakes less memory due to the in-place update but its performance seems to be slower. I guess there are two options, we either keep the implementation based on tf primitives for all backends (like we do for the measurement gate) or we try to optimize the custom operator to beat native tf performance.

scarrazza:
Concerning the `norm` you could probably create a custom cuda kernel launched with thread and block (1,1) at the end of the computation. The performance looks surprising, what happens if you use taskset to 1 core, does the performance matches the tf primitives?

stavros11:
> The performance looks surprising, what happens if you use taskset to 1 core, does the performance matches the tf primitives?

I did a different benchmark where I use the actual `Collapse` gate in a Circuit in both cases, native and custom operator, as this is what the user would do. Previously I was testing using the custom operator directly but I believe it is more consistent if we follow the gate/circuit approach.

As you suspected, for CPU there is indeed an issue with parallelization, I guess because there is a global variable `norm` that is shared among the parallelized for loops and this probably breaks parallelization. I get the following times with taskset

nqubits | native (40 threads) | custom (40 threads) | native (1 thread) | custom (1 thread)
-- | -- | -- | -- | --
27 | 1.13144 | 2.33522 | 5.23967 | 2.33921
28 | 2.20509 | 6.03257 | 10.43789 | 4.67633
29 | 4.42454 | 8.73213 | 20.80805 | 9.48913
30 | 8.88325 | 18.85916 | 41.55814 | 18.65041

Multi-threading clearly doesn't work properly for the custom operator. I also noticed that the 40-thread times vary a lot if I rerun for the same number of qubits.

On GPU both approaches are similar in terms of total time when benchmarking on the gate level. Using tf primitives has slightly faster circuit creation time faster but slower execution time. The custom execution can probably go to 30 qubits if we do the norm as you propose.

nqubits | native creation | native simulation | custom creation | custom execution
-- | -- | -- | -- | -- 
27 | 0.00072 | 0.01074 | 0.00113 | 0.00951 
28 | 0.00073 | 0.01746 | 0.00113 | 0.01603
29 | 0.00075 | 0.03047 | 0.00114 | 0.02920

On CPU execution times are similar between both approaches and << execution.

scarrazza:
Ok, good. So, for GPU the benefit is memory, and maybe performance if you move the norm calculation inside the kernel. For CPU, I agree there is some blocking operation, you can try to remove the norm and see if the performance is comparable to native.

stavros11:
> Ok, good. So, for GPU the benefit is memory, and maybe performance if you move the norm calculation inside the kernel. For CPU, I agree there is some blocking operation, you can try to remove the norm and see if the performance is comparable to native.

Yes, this is what I was doing now and it indeed fixes the problem. I added the results in the table above for easier comparison. `custom2` is using the same custom operator to zero the collapsed components without calculating any norm and it does the normalization on tensorflow aftewards as `state / tf.math.sqrt(tf.reduce_sum(tf.square(tf.abs(state)))` (like the GPU case).

stavros11:
I added a test of using the Collapse gate on distributed circuits and fixed some issues related to that. Particularly care should be taken with the normalization step because naive implementation would normalize each piece in each device which is not correct as normalization should use the total norm (sum of norms^2 of all pieces).

Regarding the custom operator, I pushed the version that does the norm on Tensorflow for both CPU and GPU. The benchmarks are in the above table ("custom2" column for CPU) so with the current version we have about x2 improvement compared to native ops on multi-threading CPU and no significant difference on GPU. 

I will mark this as ready so that we don't delay the review process it further, however I am noting that the possibility of having a custom operator for norm remains and I will try to look at this a bit more. A short summary on some approaches that I tried yesterday:

On CPU I tried two custom norm approaches:
* Calculating the norm on a single (non-parallelized) for loop and then applying it to the state with a parallel for. This works and performance compared to the current version is better when using one thread but worse when using 40 threads. I guess that performance is better on single thread because we update less values in the collapsed state (we only need to update the non-zero valus), however due to the non-parallelized norm calculation it is still slower when using multi-threading. Since multi-threading is the most common configuration in modern machines I preferred to leave the current version.
* Parallelizing the loop that calculates the norm by defining a `norms` vector (of size `ncores`) instead of a single `norm` variable. The total norm would then be `sum(norms)`. This approach did not work neither with an `std::vector<double>` for `norms` nor with a C++ array. In both cases the sum was not giving the correct norm when using parallel-for. I guess that calculating the sum of a vector is a quite common subroutine and there should be ways to be parallelized. For example, do you happen to have a quick reference of the `tf.reduce_sum` kernel? We may be able to follow the approach they used there.

On GPU, I tried doing the norm calculation on single block but did not succeed so far (due to inexperience with CUDA), so I cannot say anything about performance. I don't expect significant speed-up but we will probably be able do one more qubit if we do a custom norm (unless it requires copies). I can confirm that even a single `tf.reduce_sum(state)` fails for 30 qubits on GPU.

scarrazza:
@stavros11 thanks for report and effort. I will review the code and check the points you have raised.

doc/source/examples.rst
```diff
@@ -180,6 +180,59 @@ Unmeasured qubits are ignored by the measurement objects. Also, the
 order that qubits appear in the results is defined by the order the user added
 the measurements and not the qubit ids.
 
+The measurement gate is only used to sample the measured bitstrings and does not
+have  any effect on the state vector. If the user wishes to access the full
+state vector of a circuit that was measured, this is possible using the
+:py:attr:`qibo.base.circuit.BaseCircuit.final_state` property of circuits.
+There reasons that measurement gates do not affect the state are first that
```
scarrazza, line 187: I think this sentence is hard to understand.

doc/source/examples.rst
```diff
@@ -180,6 +180,59 @@ Unmeasured qubits are ignored by the measurement objects. Also, the
...
+    c2 = Circuit(5)
+    # take the measured bitstring from ``result``
+    bitstring = result.samples(binary=True)[0]
+    # collapse the state according to this bitstring
+    c2.add(gates.Collapse(0, 2, 3, result=bitstring))
+    # do more gates
+    c2.add((gates.H(i) for i in range(5)))
+    # execute using the measured final state of ``c1`` as the initial state
+    final_state = c2(initial_state=c1.final_state)
+
```
scarrazza, line 235: The mechanism looks ok, and I believe also memory is preserved, right?

scarrazza:
@stavros11 concerning the `tf.reduce_sum`, tf implements that [here](https://github.com/tensorflow/tensorflow/blob/a1bc56203f21a5a4995311825ffaba7a670d7747/tensorflow/core/kernels/sparse_reduce_op.cc#L135). In particular if we follow the `SparseReduceOp`, we can see that the function delegates the reduction to `TTypes<T>::UnalignedVec::sum()` which is a method provided by Eigen. So, I think your proposal about storing single pieces in a shared Tensor object and then using `.sum()` may be a proper solution, hopefully for both CPU and GPU.

scarrazza (approved):
The implementation looks pretty good. If you think is possible to perform the norm using the TType then lets go for it, otherwise we can merge this now and open another PR. Concerning tests, I see that setting the timeout to 1000s, the CI takes ~2h to complete, setting to 200s it takes 40min to complete. An intermediate solution could be setting to 100 and reducing the number of configurations.

stavros11:
Thanks a lot for the review and the reference regarding the norm.

> @stavros11 concerning the `tf.reduce_sum`, tf implements that [here](https://github.com/tensorflow/tensorflow/blob/a1bc56203f21a5a4995311825ffaba7a670d7747/tensorflow/core/kernels/sparse_reduce_op.cc#L135). In particular if we follow the `SparseReduceOp`, we can see that the function delegates the reduction to `TTypes<T>::UnalignedVec::sum()` which is a method provided by Eigen. So, I think your proposal about storing single pieces in a shared Tensor object and then using `.sum()` may be a proper solution, hopefully for both CPU and GPU.

I had a quick look at this and if I understand correctly the Eigen `sum()` method can be used to calculate the sum of elements of the vector and it supports multi-threading automatically. My main issue with this is that we do not have the vector we would like to sum, due to two reasons:
1. We have the `state` tensor, however we would like to sum `abs(state)^2`. Would this mean that we have to create a new vector that is the absolute square of `state` and then use `.sum()` on this vector? Because this would create a copy of the state vector and we would probably lose any memory (and perhaps performance) advantage over the tf op.
2. Specifically for the collapse gate case, we do not need to sum all the components of the state because some of them are set to 0 and do not contribute. Particularly we would like to do the following sum:
```C++
for (auto g = 0; g < nstates; g++) {
  norm += abs_square(state[GetIndex(g, result)]);
}
```
where the `GetIndex` method calculates the indices of non-zero components (we already have that method). Point 1 is more important, since point 2 is not taken into account even in the current implementation. When we do `tf.reduce_sum` we actually sum all the 0 values too.

> Concerning tests, I see that setting the timeout to 1000s, the CI takes ~2h to complete, setting to 200s it takes 40min to complete. An intermediate solution could be setting to 100 and reducing the number of configurations.

Thanks for looking into this. In terms of development, the faster the tests (with complete coverage) the more convenient it is. Personally I always run the tests locally before pushing so in principle the time taken by the examples test is not important as I can skip this test when I need quicker results. However I usually rely on GitHub actions as a second check and also for the coverage and in that regard anything beyond 10-15 minutes is a bit annoying. On the other hand, I also do not see any simple and stable solution. Between stability and time, I think I would prefer something stable (without random seg faults) even if it takes a bit more time.

scarrazza:
Right, concerning point 1, if I understand correctly, I think we can prepare an extra object, say `norm[nthread]` which will be send to the DoWork function and filled by it, then we can proceed with .sum or similar approaches. I believe this approach will not create another large vector, but simply use a double array with nthread entries. For GPU, the equivalent is document [here](https://developer.download.nvidia.com/assets/cuda/files/reduction.pdf), the idea is quite similar the idea is similar. If for some reason my previous point does not work, then we could just avoid using threadpool and move this custom operator openmp, it is particular annoying that documentation about the ThreadPool, including the original Eigen::ThreadPool does not talk about reductions at all.

Concerning the examples tests, I will running now a unlimited walltime so we can check how long it really takes.

doc/source/examples.rst
stavros11: I changed this slightly by splitting the hard sentence to smaller ones. Let me know if you believe it is still hard.

doc/source/examples.rst
stavros11: We discussed that yesterday, but just so that we also have it written for future reference. There are two points:
* Yes, the piece of code given in this example does not create any copies of the state vector. If custom operators are used, the `c1.final_state` and the state that will be propagated in `c2` should point to the same Tensor in memory.
* The `gates.Collapse` gate, currently creates copies of the state internally because it uses `tf.reduce_sum` to calculate the norm. If we successfully move this to custom operators then we will probably manage to preserve memory completely.

stavros11:
> Right, concerning point 1, if I understand correctly, I think we can prepare an extra object, say `norm[nthread]` which will be send to the DoWork function and filled by it, then we can proceed with .sum or similar approaches. I believe this approach will not create another large vector, but simply use a double array with nthread entries.

I could make the normalization work on multi-threading CPU with the following code (up to some adjustments):
```C++
auto AbsSquare = [&](T& x) {
  return x.real() * x.real() + x.imag() * x.imag();
};

Eigen::VectorXf norms(ncores + 1);
norms.setZero();
auto CalculateNorm = [&](int64 t, int64 w) {
  int n = ((int) t / nreps);
  for (auto g = t; g < w; g++) {
    norms[n] += AbsSquare(state[GetIndex(g, res)]);
  }
};
thread_pool->ParallelFor(nstates, p, CalculateNorm);
```
The difference in performance compared to the current version (tf normalization) is not very large
nqubits | native tf (40) | current (tf norm) (40) | new (custom norm) (40) |
-- | -- | -- | -- |
27 | 1.1314 | 0.5572 | 0.4621
28 | 2.2051 | 1.0570 | 1.0056
29 | 4.4245 | 2.0616 | 1.9099
30 | 8.8832 | 4.0698 | 3.8785

I haven't tried single-thread with taskset yet, however I can confirm that the new version (custom norm) uses exactly the memory required for a single state vector, while the current version (tf norm) creates copies and uses more memory. So the only significant advantage of the custom norm is that we gain one more qubit.

If you believe this is worth having, I will try to fix some segfaults that I am getting in some cases and then try to duplicate it to GPU following your reference. I assume that we will also not gain much in terms of performance but we will be able to simulate one more qubit.

scarrazza:
Yes, this looks great, and indeed doesn't improve much in speed but preserves memory, so I think it is a good idea to have this implementation here.

Issue #263, opened 2020-10-14
Wrong defined cost function in autoencoder tutorial


bpcarlos:
I am now reproducing the results of the autoencoder tutorial, and I saw that someone changed the cost function, that is, the "encoder hamiltonian". The code looks better now, but the cost function has to be wrong defined. Why? Because it cannot go below 0 and it does. What we have now:

`encoder = 0.5 * (compress + 2*hamiltonians.Z(nqubits))`

What we had:

```
def encoder_hamiltonian(nqubits, ncompress):
        encoder = Hamiltonian(nqubits)
        encoder.hamiltonian = 0
        eye = matrices.I
        sz = matrices.Z
        for i in range(ncompress):
            h = 1
            for j in range(nqubits):
                if i == j:
                    h = np.kron(sz, h)
                else:
                    h = np.kron(eye, h)
            encoder.hamiltonian -= h
        
        h = 1
        for i in range(nqubits):
            h = np.kron(eye, h)        
        encoder.hamiltonian += ncompress*h
        encoder.hamiltonian /= 2
        return encoder
```

I believe the problem is that hamiltonians.Z(nqubits) has to be something like hamiltonians.Z(compress), and then some tensor product between that and identities to have the same dimension as identity(nqubits)*compress (?)

Note: in the definitions of hamiltonians.Y(nqubits) and hamiltonians.Z(nqubits) there is a typo where it says pauli-X in the description.

stavros11:
Thanks for finding this. I must have changed that during some of the updates / attempts to simplify the tutorials as we were implementing new features in Qibo, but indeed the `0.5 * (compress + 2*hamiltonians.Z(nqubits))` does not agree with the one created by the function you write.

> I believe the problem is that hamiltonians.Z(nqubits) has to be something like hamiltonians.Z(compress), and then some tensor product between that and identities to have the same dimension as identity(nqubits)*compress (?)

Yes, the simplest way I could find to reproduce the correct Hamiltonian using Qibo's `hamiltonians.Z` follows this approach and is the following:
```Python
def encoder_hamiltonian_simple(nqubits, ncompress):
    m0 = hamiltonians.Z(ncompress, numpy=True).matrix
    m1 = np.eye(2 ** (nqubits - ncompress), dtype=m0.dtype)
    ham = hamiltonians.Hamiltonian(nqubits, np.kron(m1, m0))
    return 0.5 * (ham + ncompress)
```
[...]

PR #265, opened 2020-10-14
New tutorial

bpcarlos:
Added new tutorial (EF-QAE). The tutorial implements the code for the paper results, for a fixed number of qubits (6) and compressed qubits (2). Layers, type of autoencoder and example can be chosen by the user.

Fixed the cost function of the previous autoencoder tutorial as well.


examples/EF_QAE/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+![diagram](images/diagram.png)
+
+A figure of merit for the wrong answer when training is simply the total amount of non-zero measurement outcomes on the discarded qubits, which shall be minimized. In order to design the cost function to be local, different outcomes may be penalized by their Hamming distance to the |0> state, which is just the number of symbols that are different in the binary representation. Notice that this cost function has a value of zero if and only if the compression is successfully completed.
+
+To implement the quantum autoencoder model on a quantum computer we must define the form of the parametrized unitary, decomposing it into a quantum circuit suitable for optimization. This can be done using a similar structure to that in the following figure.
+
+
+<img src="images/ansatz.png" width="70%">
```
scarrazza, line 28: This calls to html doesn't work on sphinx, could you please use the markdown statement, and scale the plot to 70% width manually, i.e. using GIMP?

bpcarlos: Sure, give me a second.

examples/EF_QAE/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+where *i*,*j* simply indicates a component of the vector, and **θ** are the parameters adjusted in the optimization loop.
+
+## How to run an example?
+
+To run a particular instance of the problem we have to set up the initial
+arguments:
+- `layers` (int): number of ansatz layers.
+- `autoencoder` (int): 0 to run the EF-QAE or 1 to run the QAE.
+- `example` (int): 0 to run Ising model example or 1 to run the Handwritten digits example.
```
scarrazza, line 41: Could you please report this info in the main function, through argparse using the `help` option?

examples/EF_QAE/main.py
```diff
@@ -0,0 +1,246 @@
...
+            circuit.add(gates.CZ(5, 1)); circuit.add(gates.CZ(4, 2)); circuit.add(gates.CZ(4, 0))
+            for q in range(nqubits):
+                circuit.add(gates.RY(q, theta[index]))
+                index+=1
+            circuit.add(gates.CZ(5, 4)); circuit.add(gates.CZ(5, 2))
+            circuit.add(gates.CZ(4, 3)); circuit.add(gates.CZ(5, 0)); circuit.add(gates.CZ(4, 1))
+        for q in range(nqubits-compress, nqubits, 1):
+            circuit.add(gates.RY(q, theta[index]))
+            index+=1
+        return circuit
```
scarrazza, line 53: As @stavros11 was mentioning, I also believe is much faster to allocate the ansatz circuits once and then use the `set_parameter` method during the optimization.

examples/EF_QAE/main.py
```diff
@@ -0,0 +1,246 @@
...
+            initial_params = np.random.uniform(0, 2*np.pi, nparams)
+        
+            result = minimize(cost_function_EF_QAE_Digits, initial_params,
+                              args=(count), method='BFGS', options={'maxiter': 5.0e4})           
+            
+        else:
+            sys.exit("You have to introduce a value of 0 or 1 in the autoencoder argument.")
+            
+    else:
+        sys.exit("You have to introduce a value of 0 or 1 in the example argument.")
```
scarrazza, line 235: Please use `raise` or `raise_qibo` instead of exit.

examples/EF_QAE/main.py
bpcarlos: How will this work on the other ansatz, that has two parameters on each Ry? I know how to do it in the one that you mentioned, but not in the other case.

scarrazza: In principle you can write a function that takes `theta` from the optimizer, and rotate to the circuit basis, then you send this object with set_parameters.

examples/EF_QAE/main.py
```diff
@@ -0,0 +1,246 @@
...
+            Quantum circuit.
+        """
+        circuit = models.Circuit(nqubits)
+        index = 0
+        for l in range(layers):
+            for q in range(nqubits):
+                circuit.add(gates.RY(q, theta[index]))
+                index+=1
+            circuit.add(gates.CZ(5, 4)); circuit.add(gates.CZ(5, 3))
+            circuit.add(gates.CZ(5, 1)); circuit.add(gates.CZ(4, 2)); circuit.add(gates.CZ(4, 0))
```
scarrazza, line 44: Please do not use multi statements per line.

examples/EF_QAE/main.py
scarrazza: You can do something like this:
```python
def rotate(theta, x):
    new_theta = []
    for l in range(layers):
        for q in range(nqubits):
           new_theta.append(theta[i]*x + theta[index+1])
           index+=2
    ...
   return new_theta
```
and then the cost function should do
```python
newparams = rotate(params)
circuit.set_parameters(newparams)
final_state = circuit()
```

examples/EF_QAE/main.py
```diff
@@ -0,0 +1,265 @@
...
+                    circuit.add(gates.CZ(4, 0))
+                    for q in range(nqubits):
+                        circuit.add(gates.RY(q, theta=0))
+                    circuit.add(gates.CZ(5, 4))
+                    circuit.add(gates.CZ(5, 2))
+                    circuit.add(gates.CZ(4, 3))
+                    circuit.add(gates.CZ(5, 0))
+                    circuit.add(gates.CZ(4, 1))
+                for q in range(nqubits-compress, nqubits, 1):
+                    circuit.add(gates.RY(q, theta=0))
```
stavros11, line 83: Here the circuit is defined inside the cost function, so using `set_parameters` won't really make any difference. The circuit will again be recreated every time the cost function is called during optimization, so it is essentially the same without using `set_parameters`. 
If transforming this to the `set_parameters` format is complicated and this should be merged asap then we could leave it as it was before and postpone the update for later.

examples/EF_QAE/main.py
```diff
@@ -0,0 +1,265 @@
...
+        
+            
+            nparams = 4 * nqubits * layers + 2 * compress
+            initial_params = np.random.uniform(0, 2*np.pi, nparams)
+        
+            result = minimize(cost_function_EF_QAE_Digits, initial_params,
+                              args=(count), method='BFGS', options={'maxiter': 5.0e4})           
+            
+        else:
+            raise("You have to introduce a value of 0 or 1 in the autoencoder argument.")
```
stavros11, line 277: I get `TypeError: exceptions must derive from BaseException` from this line so I think the correct way to do it would be 
```Python
raise ValueError("You have to introduce a value of 0 or 1 in the autoencoder argument.")
```

bpcarlos:
Thanks @scarrazza. Everything should be updated now.

stavros11 (commented):
I fixed the documentation typo mentioned in #263 for the Pauli Hamiltonians. Regarding the code:

examples/EF_QAE/main.py
bpcarlos: I believe the last updated should be fine regarding to that, I manage to do it with Stefano's help.

examples/EF_QAE/main.py
stavros11: I know that this is fixed now, but just for future reference note that it is possible to do it in a single line/command as:
```Python
circuit.add([gates.CZ(5, 1), gates.CZ(4, 2), gates.CZ(4, 0)])
```

examples/EF_QAE/main.py
bpcarlos: Thanks Stavros. I just fixed it.

examples/EF_QAE/main.py
bpcarlos: Fine, good to know. Thanks.

examples/EF_QAE/main.py
stavros11: Thanks for the update, it seems to be working well now. The only thing I think we need to change is to move the four circuit definitions outside of the cost functions. This way we will avoid recreating the circuit again and again during optimization. You can just move it on top, outside of each function, eg:
```Python
for l in range(layers):
    for q in range(nqubits):
        circuit.add(gates.RY(q, theta=0))
     circuit.add(gates.CZ(5, 4))
     ...

def cost_function_EF_QAE_Digits(params, count):
    """..."""
    cost = 0
    newparams = rotate(params, 1)
    circuit.set_parameters(newparams)
    ...
```
I tested this and it seems to work as expected (in terms of the printed losses).

Regarding the test that fails, I think these segfaults are due to the time-out so there is probably nothing you can do. We need to look at this.

bpcarlos: Done. Thanks Stavros.

Is not possible then to approve the PR?

stavros11 (approved):
Thanks, looks okay to me, but I am not sure if Stefano has any other comments. I will also try to look at the test issue in some time.

scarrazza (approved):
Looks good, lets merge and monitor the test failure.

PR #266, opened 2020-10-14
Fixing tests for examples

scarrazza:
Here I fix some missing if/elif/else for the new ef example.


bpcarlos:
Thanks, Stefano. I updated the readme as well with the arXiv number.

bpcarlos (approved):

PR #267, opened 2020-10-23
Increasing examples timeout

scarrazza:
Here we increase the `--examples-timeout` flag when testing examples. I have switched off these tests for macos because this can takes up to 4h, while the virtual machines for linux takes ~1h50min.

### DIFF
--- a/.github/workflows/rules.yml
+++ b/.github/workflows/rules.yml
@@ -26,9 +26,12 @@ jobs:
     - name: Test with pylint
       run: |
         pylint src -E -d E1123,E1120
-    - name: Test with pytest
+    - name: Test with pytest core
       run: |
         pytest --cov=qibo --cov-report=xml --pyargs qibo
+    - name: Test examples
+      if: startsWith(matrix.os, 'ubuntu')
+      run: |
         pytest examples/
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v1
--- a/.github/workflows/rules.yml
+++ b/.github/workflows/rules.yml
@@ -34,6 +34,7 @@ jobs:
       run: |
         pytest examples/
     - name: Upload coverage to Codecov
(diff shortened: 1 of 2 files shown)


stavros11 (approved):
Thanks for fixing this issue. The solution sounds good to me as long as the new timeout is sufficient for all tests to always pass without problem. I guess we can only confirm this as the actions are re-executed a couple of times (eg. in other PRs). Is there any cost/monthly limit for GitHub actions? If it is free, I guess it is not a big issue to take 2 hours.

Regarding the coverage, do you think it would make sense to disable the examples test for one of the ubuntu tests and use that one for coverage? The examples are not included in the coverage check anyway and this would allow us to get faster coverage results. Otherwise, I can just check the coverage locally for new PRs. In principle we could also use the mac tests for coverage, where the example tests are already disabled, however I see that the memory error is not tested in Mac due to the Tensorflow's issue.

Side note: regarding the core tests, I believe `test_fusion.py` takes more time than the rest. I may review the configurations in that file in another PR at some point because it may be possible to reduce the time to less than 10mins.

scarrazza:
@stavros11, thanks, there is a hard limit for jobs but it beyond 2h. I like your suggestion about coverage, so let me propose to run the full pipeline only the latest ubuntu-python configuration.

stavros11:
> I like your suggestion about coverage, so let me propose to run the full pipeline only the latest ubuntu-python configuration.

Great, thanks. I believe that is a good solution.

> @stavros11, thanks, there is a hard limit for jobs but it beyond 2h.

I see, then I guess we may have a problem at some point because some of the previous runs in this PR's commits exceeded this limit. For example there is an Ubuntu - Python 3.8 run above that took 2h19m.

scarrazza:
Sorry, it was a typo 6h instead of 2h, so we are fine. Let's see if my previous commit passes.

PR #269, opened 2020-10-25
Custom operator for normalization in Collapse gate

stavros11:
As discussed in #262 this implements the state normalization required in the Collapse gate within the custom kernel instead of using Tensorflow primitives. Performance is very close to using the tf norm:

nqubits | Collapse with tf norm | Collapse with custom norm 
-- | -- | --
27 | 0.56978 | 0.51419
28 | 1.06392 | 0.85471
29 | 2.01795 | 1.91229
30 | 4.01938 | 3.68727
31 | 7.89229 | 7.67427
32 | 15.8202 | 14.4606

however the custom norm is more stable in terms of RAM usage and uses the memory required for a single state copy. For example at 32 qubits double precision the custom norm uses 64GB, while the tf norm goes up to 120GB.

Currently the custom norm is implemented for CPU only and works with multi-threading. There are two issues to resolve (decide how to approach):

* Regarding GPU, after a look at the [reduction slides](https://developer.download.nvidia.com/assets/cuda/files/reduction.pdf), it seems to me that we need a copy of the vector in order to calculate its sum using reduction methods. The reason is that half of the elements of the original array are modified in order to calculate the sum, so this operation cannot be done in-place. With this in mind, I am not sure if we would get over the current `tf.reduce_sum` implementation both in terms of memory and performance. In fact this may be one of the reasons that Tensorflow (which is GPU oriented) avoids in-place updates?

* Regarding the distributed implementation, as I mentioned in #262, we cannot do the normalization in parallel in each device because the total norm is the sum of "norms" of each device piece. For this I have a `normalize` flag inside the `Collapse` gate which is always `True` except for the case that this gate is used in a distributed circuit where normalization is handled separately. Currently this normalization happens on CPU using `tf.reduce_sum`. There are various ways this could be improved, either by doing this on GPU (I am not sure about the communication overhead) or writing a separate custom norm operator for CPU, I am just not sure if this is sufficiently important to complicate at this stage to complicate our code-base with more custom operators.

### DIFF
--- a/src/qibo/tensorflow/custom_operators/cc/kernels/apply_gate_kernels.cc
+++ b/src/qibo/tensorflow/custom_operators/cc/kernels/apply_gate_kernels.cc
@@ -269,8 +269,7 @@ struct CollapseStateFunctor<CPUDevice, T, NormType> {
     if (nreps > 0 && nstates / nreps > nnorms) {
       nnorms = nstates / nreps;
     }
-    NormType normstorage[nnorms];
-    typename TTypes<NormType>::UnalignedVec norms(normstorage, nnorms);
+    Eigen::Matrix<NormType, Eigen::Dynamic, 1> norms(nnorms);
     norms.setZero();
 
     auto AbsSquare = [&](T x) {
@@ -286,7 +285,7 @@ struct CollapseStateFunctor<CPUDevice, T, NormType> {
         for (auto h = 0; h < res; h++) {
           state[GetIndex(g, h)] = 0;
         }
-        norms(n) += AbsSquare(state[GetIndex(g, res)]);
+        norms[n] += AbsSquare(state[GetIndex(g, res)]);
         for (auto h = res + 1; h < nsubstates; h++) {
           state[GetIndex(g, h)] = 0;
         }
@@ -295,8 +294,7 @@ struct CollapseStateFunctor<CPUDevice, T, NormType> {
     thread_pool->ParallelFor(nstates, p, ZeroState);
 
     if (normalize) {
-      const Eigen::Tensor<NormType, 0, 1, int> normsum = norms.sum();
-      auto norm = std::sqrt(normsum(0));
+      auto norm = std::sqrt(norms.sum());
       auto NormalizeComponent = [&](T& x) {
         x = T(x.real() / norm, x.imag() / norm);
       };
--- a/src/qibo/tensorflow/custom_operators/cc/kernels/apply_gate_kernels.cc
+++ b/src/qibo/tensorflow/custom_operators/cc/kernels/apply_gate_kernels.cc
@@ -267,41 +267,36 @@ struct CollapseStateFunctor<CPUDevice, T> {
       return x.real() * x.real() + x.imag() * x.imag();
     };
 
-    // Calculate true norm for comparison
-    double tnorm = 0;
-    for (auto g = 0; g < nstates; g++) {
-      tnorm += AbsSquare(state[GetIndex(g, res)]);
-    }
-
-    //typename TTypes<double>::UnalignedVec norms;
-    //norms.setZero();
     Eigen::VectorXf norms(ncores + 1);
     norms.setZero();
-    auto CalculateNorm = [&](int64 t, int64 w) {
-      int n = ((int) t / nreps);
-      for (auto g = t; g < w; g++) {
-        norms[n] += AbsSquare(state[GetIndex(g, res)]);
-      }
-    };
-    thread_pool->ParallelFor(nstates, p, CalculateNorm);
-
-    for (int i = 0; i < ncores + 1; i++) {
-      std::cout << norms[i] << " ";
-    }
-    std::cout << "\nCalculated norm: " << norms.sum() << std::endl;
-    std::cout << "True norm: " << tnorm << std::endl;
-
     auto ZeroState = [&](int64 t, int64 w) {
+      int n = 0;
+      if (nreps > 0) {
+        n = ((int) t / nreps);
+      }
       for (auto g = t; g < w; g++) {
         for (auto h = 0; h < res; h++) {
           state[GetIndex(g, h)] = 0;
         }
+        norms[n] += AbsSquare(state[GetIndex(g, res)]);
         for (auto h = res + 1; h < nsubstates; h++) {
           state[GetIndex(g, h)] = 0;
         }
       }
     };
     thread_pool->ParallelFor(nstates, p, ZeroState);
+
+    auto norm = std::sqrt(norms.sum());
+    auto NormalizeComponent = [&](T& x) {
+      x = T(x.real() / norm, x.imag() / norm);
+    };
+
+    auto NormalizeState = [&](int64 t, int64 w) {
+      for (auto g = t; g < w; g++) {
+        NormalizeComponent(state[GetIndex(g, res)]);
+      }
+    };
+    thread_pool->ParallelFor(nstates, p, NormalizeState);
   }
 };
 
--- a/src/qibo/tensorflow/custom_operators/cc/kernels/apply_gate_kernels.cc
+++ b/src/qibo/tensorflow/custom_operators/cc/kernels/apply_gate_kernels.cc
@@ -237,6 +237,7 @@ struct CollapseStateFunctor<CPUDevice, T> {
                   const int64* result) const {
     int64 nstates = (int64)1 << (nqubits - ntargets);
     int64 nsubstates = (int64)1 << ntargets;
+    const int64 res = result[0];
 
     // Set multi-threading
     auto thread_pool =
@@ -262,12 +263,32 @@ struct CollapseStateFunctor<CPUDevice, T> {
       return i;
     };
 
+    //typename TTypes<double>::UnalignedVec norms;
+    //norms.setZero();
+    Eigen::VectorXf norms(ncores);
+    norms.setZero();
+    for (int i = 0; i < ncores; i++) {
+      std::cout << i << "," << norms[i] << " ";
+    }
+    std::cout << std::endl;
+    double tnorm = 0;
+    for (auto g = 0; g < nstates; g++) {
+      auto i = GetIndex(g, res);
+      norms[((int)g % ncores)] += state[i].real();
+      tnorm += state[i].real();
+    }
+    for (int i = 0; i < ncores; i++) {
+      std::cout << norms[i] << " ";
+    }
+    std::cout << "\nCalculated norm: " << norms.sum() << std::endl;
+    std::cout << "Total norm: " << tnorm << std::endl;
+
     auto ZeroState = [&](int64 t, int64 w) {
       for (auto g = t; g < w; g++) {
-        for (auto h = 0; h < result[0]; h++) {
+        for (auto h = 0; h < res; h++) {
           state[GetIndex(g, h)] = 0;
         }
-        for (auto h = result[0] + 1; h < nsubstates; h++) {
+        for (auto h = res + 1; h < nsubstates; h++) {
           state[GetIndex(g, h)] = 0;
         }
       }
--- a/src/qibo/tensorflow/custom_operators/cc/kernels/apply_gate_kernels.cc
+++ b/src/qibo/tensorflow/custom_operators/cc/kernels/apply_gate_kernels.cc
@@ -14,6 +14,24 @@ namespace functor {
 
 using thread::ThreadPool;
 
+// Helper methods for complex numbers
+template <typename T>
+T cmult(T a, T b) {
+  return T(a.real() * b.real() - a.imag() * b.imag(),
+           a.real() * b.imag() + a.imag() * b.real());
+}
+
+template <typename T>
+T cadd(T a, T b) {
+  return T(a.real() + b.real(), a.imag() + b.imag());
+}
+
+template <typename T>
+double AbsSquare(T x) {
+  return x.real() * x.real() + x.imag() * x.imag();
+}
+
+
 template <typename T>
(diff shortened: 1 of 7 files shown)


src/qibo/tensorflow/custom_operators/cc/kernels/apply_gate_kernels.cc
```diff
@@ -262,17 +270,43 @@ struct CollapseStateFunctor<CPUDevice, T> {
       return i;
     };
 
+    // Define vector that holds norms during parallel calculation
+    int nnorms = ncores + 1;
+    if (nreps > 0 && nstates / nreps > nnorms) {
+      nnorms = nstates / nreps;
+    }
+    Eigen::VectorXf norms(nnorms);
```
scarrazza, line 278: The code looks fine, but I think here we should use the tensorflow TType<T> object instead of calling directly Eigen.

scarrazza:
@stavros11, thanks, the CPU performance and memory footprint are really great with this approach. For the distributed approach, I think the current behaviour is quite acceptable, it will consume more memory but that's an acceptable trade-off.

Concerning GPU, as you can see from the document, there are several approaches, but the simplest is when you consider a support array with size matching the blocks size (in our case 1024) and then use that object as shared array for the sums over threads.

src/qibo/tensorflow/custom_operators/cc/kernels/apply_gate_kernels.cc
stavros11: As far as I can see in [tensor_types.h](https://github.com/tensorflow/tensorflow/blob/a1bc56203f21a5a4995311825ffaba7a670d7747/tensorflow/core/framework/tensor_types.h), which I believe does the mapping between Eigen objects and TTypes, there is no TTypes equivalent to `Eigen::VectorXf`, only to `Eigen::TensorMap`. Is this the type we would like to use here? I guess that there is a way to do it with `TensorMap` but we would have to change the code slightly.

By the way, I think this works even with a `std::vector` in the place of `Eigen::VectorXf`. We just need to change `.sum()` to `std::accumulate`. I haven't done any benchmarks to see if there is any performance difference.

scarrazza: Ok, ideally I believe we need a `UnalignedVec`. I am fine with VectorXf, however you are hardcoding the type, I think there is a VectorXd, right?

stavros11: One way I could make it work with `UnalignedVec` is the following:
```C++
double storage[nnorms];
typename TTypes<double>::UnalignedVec norms(storage, nnorms);
norms.setZero();
...
    norms(n) += AbsSquare(state[GetIndex(g, res)]);
...
auto norm = std::sqrt(((Eigen::Tensor<double, 0, 1, long int>)norms.sum())(0));
```
There might be a more Tensorflow-y way to convert `norms.sum()` to tensor to avoid the `Eigen::Tensor`. As far as I understand `UnalignedVec` and all other `TTypes` are `Eigen::TensorMap` and cannot allocate memory themeselves, they just point to existing objects. So I am not sure if it is possible to avoid the `storage` variable (or something equivalent).

Regarding the type, this is a good point and in fact `Eigen::VectorXd` is probably a better choice indeed. However in some sense this also hardcodes the type to `double` as does the `UnalignedVec` method above. Ideally we would like something that follows the real part of `T`, which I am not sure how to get at the type level. Another solution is to just use the complex `T` type for the norm but we would have to carry around a 0 as the imaginary part which would be suboptimal (perhaps without significant difference).

scarrazza: Yes, this can be quite annoying, so I may suggest to use the double norm with VectorXd. For sure we can extract the type of the real/imag part of T, but then we need another mechanism to select the appropriate type from Eigen, or simply use C primitives with a template class.

stavros11: After some experimentation I could implement this with variable type as follows: 

The `norms` vector is declared as:
```C++
decltype(state[0].real()) norm;
decltype(norm) normstorage[nnorms];
typename TTypes<decltype(norm)>::UnalignedVec norms(normstorage, nnorms);
```
and the final norm is calculated as:
```C++
const Eigen::Tensor<decltype(norm), 0, 1, int> normsum = norms.sum();
norm = std::sqrt(normsum(0));
```
This seems to work, however it shows a strange variation in performance. For example for 30 qubits I get times ranging from 2.5 to 7 seconds, while the current version (with `VectorXf`) is stable to around 3.5 seconds.

Another strange observation is that `VectorXd` is actually the fastest. Specifically I get the following times and are pretty stable:
nqubits | VectorXd | VectorXf
-- | -- | --
28 | 0.71076| 0.85471
29 | 1.38268 | 1.91229
30 | 2.81394 | 3.68727
31 | 5.49702 | 7.67427
32 | 11.3545 | 14.4606

All methods take the same memory. Let me know which approach you think is the best.

scarrazza: Thanks for the investigation. I think the decltype+typename are slowing down the process, so I fully agree in using a VectorXd instead.

stavros11:
I just realized that the [`AbsSquare`](https://github.com/Quantum-TII/qibo/blob/add0e2938d13999ddbd766ca274e4cba52e5f590/src/qibo/tensorflow/custom_operators/cc/kernels/apply_gate_kernels.cc#L30) method is hardcoded as `double` which may explain why `VectorXd` is faster than `VectorXf` (I guess the latter does more conversions between floats and doubles). I will do a few more checks with that in mind and if the situation is similar to what I wrote above I will change to `VectorXd` as you propose. I am not sure if there exists a way to do the variable type without slowing it down.

> Concerning GPU, as you can see from the document, there are several approaches, but the simplest is when you consider a support array with size matching the blocks size (in our case 1024) and then use that object as shared array for the sums over threads.

I think I managed to do the sum on GPU following what you write here. Thanks for the explanation! I do it with two calls, the first creates a vector of size 1024 with the norms of each thread and the second does the reduction of this vector using the code from the slides. I have done a few checks and it seems to calculate the correct norm, but I still need to fix a few issues in the code before pushing. I also plan to do some benchmarks to see how this compares to the native tf solution and whether we finally gain a qubit in terms of memory. I will push the code here once I have all this.

scarrazza:
> I am not sure if there exists a way to do the variable type without slowing it down.

Probably if you write explicitly the template for each case complex64 and complex128, so you can hardcode and precompile both approaches before execution.

> I think I managed to do the sum on GPU following what you write here. 

Perfect.

stavros11:
I pushed the CUDA kernel that does the normalization and changed to `Eigen::VectorXd` for the CPU. The CUDA kernel works, performance for single runs is similar to what we had before and as expected we gain one qubit in memory, so we can do 30 qubits in double precision instead of 29 which was the limit when using `tf.reduce_sum` for the norm.

Regarding performance, notice that I wrote for single run. I noticed that if I run a circuit repeatedly as
```Python
circuit.add(gates.Collapse(0, result=0))

for _ in range(nreps):
    state = circuit()
    del(state) # have to do this deletion otherwise GPU will run OOM
```
time does not increase linearly with `nreps` as I would expect. For example for 28 qubits I get the following times:

| nreps | simulation time (sec) |
| -- | -- |
| 50 | 0.02346062660217285 |
| 100 | 0.03359484672546387 |
| 200 | 9.528245449066162 |
| 300 | 42.064990520477295 |

I will try to pinpoint this issue but I suspect it has to do with the memory allocation, because I am using `cudaMalloc` to create the `norm` vectors on GPU and I suspect that Tensorflow doesn't like this. As commented in the code, normally one would do `cudaFree` in the end however if I do that then repeated execution of Collapse does not work properly and tests don't pass.

TODO summary to keep track of this PR's progress:
- [ ] Fix the issue with repeated execution on GPU.
- [ ] Attempt to parallelize the GPU norm calculation to multiple blocks (currently I use one block). This may increase performance.
- [ ] Use a template variable for the norm vector types both on CPU and GPU.
- [ ] Benchmark the final version on both CPU and GPU.

stavros11:
> Probably if you write explicitly the template for each case complex64 and complex128, so you can hardcode and precompile both approaches before execution.

I implemented the norm type with a template in the last set of commits, both for CPU and GPU. Here is a comparison of the latest version of this PR (custom norm for both CPU and GPU) and the version from #262 that uses `tf.reduce_sum` for the norm:

**CPU (40-threads)**
nqubits | Collapse tf norm | Collapse custom norm |
-- | -- | -- 
28 | 1.09692 | 0.7224
29 | 2.08101 | 1.46323
30 | 4.06606 | 2.76722
31 | 8.12420 | 5.40362
32 | 15.79736 | 11.29282

**GPU**
nqubits | Collapse tf norm | Collapse custom norm |
-- | -- | -- 
27 | 0.00956 | 0.00971
28 | 0.01592 | 0.01468
29 | 0.02778 | 0.02933
30 | OOM | 0.05553

I believe performance and memory is acceptable in both cases, however (mostly out of curiosity) I will try to play a little with the block number for the GPU case because currently the computation is using only one block.

stavros11:
I pushed some small modifications in the GPU kernel, particularly used `numBlocks` to parallelize the state normalization kernel and moved the norm square root in the sum kernel so that it only happens once (instead of applying the square root repeatedly when normalizing each component). There is no significant change in performance from these changes.

I could not find a way to use `numBlocks > 1` in the kernels that calculate the sum. If I understand the mechanism correctly, in order to do this we would need to add more components to the norm vector and I guess eventually we would just need to fine tune the number and size of blocks to get the best performance. However I would expect the differences to be tiny for our particular application, so I am not sure if it is worth going through that.

Since normalization works in both CPU and GPU and I believe that performance (tables in previous post) is acceptable, I would suggest to merge this to #262 and have a final review of the collapse gate.

scarrazza:
Thanks, I agree. Do you still observe the memory increase on CPU/GPU?

stavros11:
> Thanks, I agree. Do you still observe the memory increase on CPU/GPU?

The custom operators for the Collapse gate take the expected memory on both CPU and GPU, same as all other gates. The 32GB GPU can run up to 30 qubits in double precision. This is one qubit more compared to using the native Tensorflow Collapse gate. On CPU memory usage is similar, for example 32 qubits take 64GB.

scarrazza:
OK, and what about the `for _` loop you have quoted some time ago?

stavros11:
> OK, and what about the `for _` loop you have quoted some time ago?

If you mean the repeated circuit executions on a single script I have actually been doing some more benchmarks about this right now and was planning to open an issue quoting the exact script and numbers. I prefer to keep it separate from this PR because it is not related to the Collapse gate, I can reproduce the same behavior using other gates and the master branch. I am not sure if it is really an issue or it is expected but we can discuss this once I have all numbers.

src/qibo/tensorflow/custom_operators/cc/kernels/apply_gate_kernels.cu.cc
```diff
@@ -562,7 +609,25 @@ struct CollapseStateFunctor<GPUDevice, T> {
...
+        state, block_norms, qubits, result, nstates, ntargets);
+      VectorReductionKernel<NormType><<<1, blockSize, 0, d.stream()>>>(
+        block_norms, norms);
+      NormalizeCollapsedStateKernel<T, NormType><<<numBlocks, blockSize, 0, d.stream()>>>(
+        state, norms, qubits, result, nstates, ntargets);
+
+      // if we use ``cudaFree`` here the custom operator fails if it is used
+      // twice in the same script
+      //cudaFree(block_norms);
+      //cudaFree(norms);
```
scarrazza, line 629: This looks strange. You are creating pointers and not cleaning the memory so this may generate a segfault, right?

stavros11: Based on the CUDA examples that I read in order to do this, I also thought that I should do `cudaFree` in the end in order to clean the memory. When I did so, a stand-alone script that I had for checking the norm was working properly, however `test_custom_operators.py` was failing. After playing a little I realized that when running the Collapse gate twice in the same script, the norm calculation is wrong in the second time if `cudaFree` is used. This explains why tests were failing, since tests run as a single script. If I remove `cudaFree` both the stand-alone script and the tests work properly.

I just left it as a comment because I could not really explain why this happens. Is this because of Tensorflow's automatic memory management?

scarrazza: The problem derives from the missing initialization of the arrays to 0. 
You have to initialize the arrays with memset:
```c++
      NormType *block_norms, *norms;
      cudaMalloc((void**)&block_norms, sizeof(NormType) * blockSize);
      cudaMemset(block_norms, 0, sizeof(NormType) * blockSize);
      cudaMalloc((void**)&norms, sizeof(NormType));
      cudaMemset(norms, 0, sizeof(NormType));
      ...
      cudaFree(block_norms);
      cudaFree(norms);
```
Anyway, I will open another PR where we replace the cuda* calls with a TF approach.

stavros11: Many thanks for this clarification. Indeed I checked using `cudaMemset` and everything works as expected with `cudaFree` enabled. It is a bit strange that it worked and tests were passing previously without explicitly initializing the arrays to 0.

I am not pushing the change since we will probably merge #275 here which fixes these issues anyway, but thanks for pointing this out.

PR #270, opened 2020-10-26
Noise simulation with repeated execution

### LINKED ISSUE #91: Errors without density matrices
As discussed in the previous meeting, an error simulation that does not require more memory is a good alternative in order to test noise effects on larger quantum circuits.

This approach is not fast, as one will have to compute the noisy circuit every time a measurement is made, but might be the only alternative in cases where there is not enough memory to store the density matrices.

stavros11:
This is a first attempt to simulate noise without density matrices, in order to avoid doubling the effective qubit number. Instead circuit execution is repeated many times and different X, Y, Z noise gates are applied randomly each time. If my method is acceptable this closes #91.

The idea is quite simple. A new gate called `ProbabilisticNoiseChannel` is added which is physically equivalent to the `NoiseChannel` we already have and applies X (bit-flip), Y and Z (phase-flip) errors with given probabilities px, py, pz. The difference is that in the new channel, instead of applying it as a real channel on the density matrix, we sample the gates and apply them to the state vector. I think it is easier to understand looking at the code that defines the action on a `state`:
```Python
if np.random.random() < px:
    state = gates.X(state)
if np.random.random() < py:
    state = gates.Y(state)
if np.random.random() < pz:
    state = gates.Z(state)
```
Obviously this will have different behavior every time it is executed so the idea is to execute the circuit many times and get the results collectively. This can be done using `circuit.execute()`/`circuit()` method and passing `nshots` as the number of repetitions to run.

Note that this is in contrast with the current behavior of `nshots`, where we executed the circuit only once and then sampled the measurements. Now the circuit will actually be executed `nshots` times. However this only happens when a `ProbabilisticNoiseChannel` exists in the circuit (Qibo decides that automatically), otherwise the old behavior is used since it is much faster.

**Method**
This method I used is probably the most straightforward one could think of, however I am not sure if it is the optimal in terms of getting reliable noisy samples with the smallest number of repetitions. I am open to suggestions for better methods if they exist.

**Performance and parallelization**
Currently the circuit is executed `nshots` (the number of repetitions) times sequentially so execution time will increase linearly with `nshots`. As usual with MC methods, it is possible to have multiple workers executing the same circuit in parallel for different random instances. This can be done both in one device (assuming it has sufficient memory to hold multiple copies of the state), in multiple devices or a combination of both. In terms of API we could have something like:

```result = circuit(nshots={"/GPU:0": (4, 200), "/GPU:1": (4, 200)})```

which would execute 4 workers on each GPU for 200 repetitions each, for a total of 2 GPUs x 4 workers x 200 samples = 1600 samples. The main issue with that is determining the optimal parallelization configuration for each problem because in some cases the overhead of recreating the circuit in the same and/or new devices may be large and is better to just perform sequential executions of the same circuit. 

Note that the intra-device parallelization could be related to #255, however the noise case is slightly easier because it is exactly the same circuit that we would run multiple times in parallel (just with different random instances for the noise channel).

### CODE REVIEW

File: doc/source/advancedexamples.rst [hunk 1]
```diff
@@ -579,10 +579,65 @@ and using the `Sequential model API <https://www.tensorflow.org/api_docs/python/
...
 
+.. _noisy-example:
+
 How to perform noisy simulation?
 --------------------------------
 
-Qibo can perform noisy simulation using density matrices.
+Qibo can perform noisy simulation with two different methods: by repeating the
+circuit execution multiple times and applying noise channels probabilistically
+or by using density density matrices.
```
scarrazza, line 589:
```suggestion
or by using density matrices.
```

stavros11:
@scarrazza I am not sure if you have already reviewed large part of this, but I believe there is a cleaner and more general approach to implement noise with repetition (without density matrices) starting from #276. Instead of defining a new `ProbabilisticNoiseChannel` gate, we could use the same `NoiseChannel` gate and just implement the `_state_vector_call` and `_density_matrix_call` methods for each case (with/without density matrices). This would be simpler for the user (less gates to chose from) and if implemented properly would allow to simulate with repetition (without density matrices) any channel of the form 

E[\rho ] = (1 - sum(p_i) )\rho *  + sum( p_i U_i \rho U_i^\dagger ) 

which could be useful when extending the noise models beyond the `NoiseChannel` we currently have. For example [thermal relaxation](https://qiskit.org/documentation/stubs/qiskit.providers.aer.noise.thermal_relaxation_error.html#qiskit.providers.aer.noise.thermal_relaxation_error) takes this form for some choice of the parameters.

Based on this, my plan is to push the required changes directly on #276 and close the current PR. I will do this once I get everything working.

scarrazza:
Thanks for letting me know, and yes, I realized the other PR.

PR #271, opened 2020-10-27
Add parallel version of L-BFGS-B minimizer

DiegoGM91:
This PR adds a parallel implementation of the L-BFGS-B minimization algorithm from the package ``optimparallel`` to qibo optimizers. I believe this may be useful to accelerate the training of variational quantum circuits in some cases.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,4 @@ cma
 joblib
 matplotlib
 blessings
+optimparallel
--- a/src/qibo/optimizers.py
+++ b/src/qibo/optimizers.py
@@ -52,12 +52,19 @@ def newtonian(loss, initial_parameters, method='Powell', options=None):
             optimized.
         initial_parameters (np.ndarray): Initial guess for the variational
             parameters.
-        method (str): Name of method supported by ``scipy.optimize.minimize``.
+        method (str): Name of method supported by ``scipy.optimize.minimize`` or "parallel_lbfgsb". The later uses  
+                ``optimparallel.minimize_parallel``, which is a parallel implementation of scipy's "l-bfgs-b" algorithm.
         options (dict): Dictionary with options accepted by
             ``scipy.optimize.minimize``.
     """
-    from scipy.optimize import minimize
-    m = minimize(loss, initial_parameters, method=method, options=options)
+    if method != 'parallel_lbfgsb':
+        from scipy.optimize import minimize
+        m = minimize(loss, initial_parameters, method=method, options=options)
+        
+    else:
+        from optimparallel import minimize_parallel
+        m = minimize_parallel(loss, initial_parameters, options=options)
+
     return m.fun, m.x
 
 


scarrazza:
@DiegoGM91 thanks for your effort. I think there are important 2 points we have to address here:
- Qibo's circuit evaluation is thread-safe only if you allocate a new initial_state for each parallel circuit evaluation (because we do in place updates). So this optimizer will work if you allocate the initial state directly inside the cost function, or if you use the default initial state (|0>).
- Performance: we did some tests and we know that tensorflow "doesn't like" parallel evaluations on the same device in eager mode. We have observed that running 2 or more instances of circuit in parallel doesn't improve performance, in comparison to a sequential evaluation. Do you observed some consistent speed up?

DiegoGM91:
Thanks for the comments @scarrazza.

Regarding the first point, I just tried it using the default initial state |0...0> (this will occur in many variational algorithms)

Regarding the second point, I didn't do any proper benchmark, I just observed how cost functions evaluations were printed out on the screen, and I would say that in my test case, there was an  improvement in speed. I can make however some proper benchmarks to see whether this is actually the case or not, and under which circumstances.

scarrazza:
@DiegoGM91 following our discussion last week, let me summarize the steps to validate this PR:
- consider a specific case study where we can demonstrate performance improvements
- check how to extend the approach to other applications

DiegoGM91:
Here are some benchmarks comparing the performance of the `l-bfgs-b` method from `scipy` and `optimparallel`, for the quantum text classifier. The numbers in the table indicate the time it takes to complete one iteration of the optimization procedure (which includes multiple cost function evaluations), in seconds. I show mean results obtained using `timeit`. Results were obtained on an Intel i7-9750H CPU with six cores and 16GB of memory. Single thread was set for the quantum circuit execution using ``tf.config.threading.set_inter_op_parallelism_threads(1)`` and ``tf.config.threading.set_intra_op_parallelism_threads(1)``.

|                                                       |optimparallel | scipy | ratio |
|:--------------------------------------------:|:---------------:|:--------:|:------:|
|4 qubits, 3 layers, 80 data points   |        16.6      |   70.1|  4.22 |
|4 qubits, 3 layers, 400 data points |         85        |   357 |  4.20 |
|4 qubits, 4 layers, 80 data points   |        15.8      |   68   |  4.30 |
|4 qubits, 4 layers, 400 data points |         78        |   344 |  4.41 |
|8 qubits, 4 layers, 80 data points |        65           |   264  |  4.06 |
|8 qubits, 4 layers, 400 data points |         319       |  1353 |  5.12 |

I can extend these benchmarks to the `VQE` and `QAOA` models and also to higher qubit numbers, and see how it goes, if you think it is worth it.

scarrazza:
@DiegoGM91 thanks for this benchmark, just a quick question, what happens if you run both codes using `taskset -c 0 <your script>`?

DiegoGM91:
@scarrazza when restricting to a single CPU, then I get 358 seconds for 4 qubits, 4 layers, 400 data points with `optimparallel`.

scarrazza:
Great, and the `scipy` numbers are same?

DiegoGM91:
Yes, I get 364 seconds for 4 qubits, 4 layers, 400 data points with `scipy`.

scarrazza:
@DiegoGM91 I am playing with this PR, and I need probably more information about your runs.
If I take the vqe example in `qibo/examples/benchmarks/vqe.py` and run:
```
time python vqe.py --nqubits 3 --nlayers 4 --method l-bfgs-b --maxiter 15000
```
it runs and takes 2.8s to complete, while if I do:
```
time python vqe.py --nqubits 3 --nlayers 4 --method parallel_lbfgsb --maxiter 15000
```
the code freezes at the the first iteration (probably due to the concurrent futures in optimparallel).

Could you please try this example and let me know if this freeze happens for you too?

DiegoGM91:
@scarrazza yeah, the same happens for me too. It just freezes out.

scarrazza:
Then could you please share the layout of your code? Do you use set_parameter or reallocate the circuit at each loss function call?

DiegoGM91:
I use the `set_parameters` method. This is the code:

```python
    def minimize(self, init_theta, data=None, labels=None, nshots=None):
        """
        Args: 
            init_theta: iterable with the initial angles to be used in the circuit.        
            data: numpy.ndarray data[page][feature]  (this is an array of feature vectors). 
            labels: iterable with the labels of the feature vectors to be classified.                      
            nshots: int number of runs of the circuit during the sampling process (default=None).         
        
        Returns: 
            numpy.float64 with value of the minimum found, numpy.ndarray with the optimal angles        
        """        
        from optimparallel import minimize_parallel
        result = minimize_parallel(self.cost_function, init_theta, args=(data,labels,loss,nshots), options={'maxiter':10})
        loss = result.fun
        optimal_angles = result.x     
        
        return loss, optimal_angles

    def cost_function(self, theta, data=None, labels=None, nshots=None):
        """    
        Args: 
            theta: iterable with the biases and the angles to be used in the circuit.
            data: numpy.ndarray data[page][feature]  (this is an array of feature vectors).       
            labels: iterable with the labels of the feature vectors to be classified.                 
            nshots: int number of runs of the circuit during the sampling process (default=None).
    
        Returns: 
            numpy.float32 with the value of the square-loss function
        """
        Predictions = np.zeros(shape=(len(data),self.measured_qubits))
    
        for i, feature_vector in enumerate(data):
            Predictions[i] = self.predictions(theta, feature_vector, nshots)

         def square_loss(labels, predictions):
                """    
                Args: 
                    labels: list or numpy.array with the qubit labels of the quantum states to be classified                    
                    predictions: list or numpy.array with the qubit predictions for the quantum states to be classified
    
                Returns: 
                    numpy.float32 with the value of the square-loss function
                """
                loss = 0
                for l, p in zip(labels, predictions):    
                    for qubit in range(self.measured_qubits):
                        loss += (l[qubit] - p[qubit]) ** 2
    
                return loss / len(labels)

        return square_loss(labels, Predictions)

    def predictions(self, theta, feature_vector, nshots=None):
            """    
            Args: 
                theta: iterable with the biases and angles to be used in the circuit.                
                feature_vector: numpy.ndarray with the feature_vector to be classified.
                nshots: int number of runs of the circuit during the sampling process (default=None).
    
           Returns: 
                numpy.ndarray() with predictions for each qubit, for the initial state
            """  
            bias = np.array(theta[0:self.measured_qubits])
            self.circuit(theta, feature_vector)
            result = self._circuit(nshots)
            prediction = np.zeros(self.measured_qubits)

            if nshots is None:
                for qubit in range(self.measured_qubits):
                    for clase in range(self.nclasses):           
                        binary = bin(clase)[2:].zfill(self.measured_qubits)
                        prediction[qubit] += result[clase]**2 * (1-2*int(binary[-qubit-1]))
            else:
                result = result.frequencies(binary=False)   
                for qubit in range(self.measured_qubits):      
                    for clase in range(self.nclasses):           
                        binary = bin(clase)[2:].zfill(self.measured_qubits)
                        prediction[qubit] += result[clase] * (1-2*int(binary[-qubit-1]))
                    
                prediction /= nshots
            
            return prediction + bias

    def circuit(self, theta, feature_vector):
            """    
            Args: 
                theta: iterable with the biases and angles to be used in the circuit.
                feature_vector: numpy.ndarray with the feature_vector to be classified.
    
            Returns: 
                qibo.tensorflow.circuit.TensorflowCircuit with the variational circuit for angles "theta"
            """
            bias = np.array(theta[0:self.measured_qubits])
            circuit_params = np.array(theta[self.measured_qubits:])
            circuit_params = [circuit_params[i:i+2] for i in range(0,len(circuit_params),2)]
            angles = [x*omega[0]+omega[1] for l in range(self.nlayers) for x,omega in zip(2*list(feature_vector),circuit_params[2*l*self.nqubits:2*(l+1)*self.nqubits])]
            for x,omega in zip(feature_vector,circuit_params[2*self.nlayers*self.nqubits:]):
                angles.append(x*omega[0]+omega[1])
        
            self._circuit.set_parameters(np.array(angles))
        
            return self._circuit
```

where `self._circuit = self.ansatz(rotations)` is just a qibo circuit with some parameterized gates.

scarrazza:
Are you using the custom backend or something else? (defauleinsum, matmuleinsum)

DiegoGM91:
Custom

scarrazza:
And if you set the intra/inter number to > 1 it code freezes, right?

DiegoGM91:
I hadn't tried that because of the low number of qubits I use in the classifier (it turns out you can classify complicated data using very few qubits).

But right, it freezes when I set the intra/inter number to > 1.

scarrazza:
Closing in favour of #281.

PR #273, opened 2020-11-02
fix docs

### LINKED ISSUE #268: NVCC path when building for CUDA
I finally attempted to install Tensorflow and Qibo in a machine with CUDA GPU and had a CUDA path related problem when building the custom operators. More specifically, while my CUDA path is `/usr/lib/cuda/`, nvcc is not located in the same folder but rather in `/usr/local/bin/`. Therefore when I follow the instructions from our docs the build fails because it cannot locate nvcc. I could fix this issue by manually changing the [`NVCC` definition in our Makefile](https://github.com/Quantum-TII/qibo/blob/a8f1d3df9dbe4c3e96c129c7c7c152a1e583b3e9/src/qibo/tensorflow/custom_operators/Makefile#L7) to `NVCC = nvcc`, which worked in my case because `nvcc` was included in the system's path. 

Installation worked properly after this change and could confirm that the GPU was used using benchmark times. I thought of mentioning the issue here as it may be relevant for others. In our docs it seems that CUDA_PATH should direct to CUDA. Shall we change this to NVCC_PATH? I am not sure if for my case these two paths are different because something went wrong during the CUDA installation.

scarrazza:
Closes #268.

### DIFF
--- a/doc/source/installation.rst
+++ b/doc/source/installation.rst
@@ -72,7 +72,7 @@ variable sets the python interpreter path.
 
       2. exporting the ``CUDA_PATH`` variable with the CUDA installation path containing the cuda compiler.
 
-      3. make sure the NVCC compiler is available from ``CUDA_PATH/bin/nvcc``, otherwise the compilation may fail.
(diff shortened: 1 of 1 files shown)


stavros11 (approved):
Looks good, thanks.

Perhaps we could add that if NVCC is installed but is not located in the desired path the user can locate it using `whereis nvcc` and copy it from there to `CUDA_PATH/bin`, since this is a very simple solution. But if you think that writing this could confuse average users it is okay to leave it out.

PR #274, opened 2020-11-02
Remove references to state vector to avoid OOM errors

stavros11:
As discussed this removes the references to the state tensor before executing a circuit to avoid OOM errors when a circuit is executed multiple times on the same script. Note that if the user uses an additional reference to the final state (eg. `state = circuit()`, then `state` should also be deleted manually (`del(state)`) to avoid OOMs.

I tested this and it works when the default initial state is used. The situation is a bit more complicated when using a custom initial state. For example
```Python
initial_state = tf.zeros(2 ** nqubits, dtype=tf.complex128)
for i in range(nreps):
    initial_state = tf.zeros(2 ** nqubits, dtype=tf.complex128)
    state = circuit(initial_state)
    del(state)
```
works, however
```Python
for i in range(nreps):
    initial_state = tf.zeros(2 ** nqubits, dtype=tf.complex128)
    state = circuit(initial_state)
    del(state)
```
runs out of memory for nreps > 1 when `nqubits` is the maximum the GPU can run in single rep. 

I also noticed that using numpy array for initialization leads to OOM if the numpy dtype is different than the tf dtype used in Qibo. I added an `.astype` in the casting function to fix this issue.

### DIFF
--- a/src/qibo/base/measurements.py
+++ b/src/qibo/base/measurements.py
@@ -14,9 +14,6 @@ class GateResult:
 
     Args:
         qubits: Sorted tuple of qubit ids that the measurement gate acts on.
-        state: Reference to the tensor that holds the state that was sampled.
-            The state should have shape ``nqubits * (2,)`` if it is a state vector
-            or ``2 * nqubits * (2,)`` if it is a density matrix.
         decimal_samples: Tensor holding the measured samples in decimal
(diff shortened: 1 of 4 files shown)


scarrazza (approved):
Thanks, looks good!

PR #275, opened 2020-11-02
Use tf primitives for GPU allocation

scarrazza:
Replaces cudaMalloc/cudaFree with temporary Tensors.


stavros11 (approved):
Thank you very much for implementing this, it looks much better than my solution. I checked and performance is stable and the same with what is reported in the last table of #269 so this should be good to merge.

Issue #129, opened 2020-06-17
Custom operators for density matrices


stavros11:
Currently density matrices are only supported by the slow gates that use tensorflow primitives. If we want a faster density matrix simulator we have to extend our custom operators to density matrices. This would require to write new specialized kernels both for CPU and CUDA.

PR #276, opened 2020-11-04
Density matrix simulation using custom operators

stavros11:
This PR allows using the "custom" backend to simulate density matrices. So far density matrices were only supported by the native tensorflow backends. Unlike what is described in #129, it is not required to write new kernels because we can use the state vector kernels and treat the density matrix as a vector in a space with twice as many qubits. I note that for the future we may consider a more efficient representation that exploits the fact that the positivity and symmetry of the density matrix (this may require new kernels).

This PR:
* Changes the density matrix API to completely separate state vector and density matrix simulation. Currently the circuit object automatically determines whether it uses density matrices or state vectors and particularly density matrices are only used when a matrix is given as initial state or a channel is found in the circuit's gates. I believe this may be confusing for the user so I added a `density_matrix=False/True` flag during circuit intialization that specifies explicitly the type of simulation. I will add examples in the docs once this PR is ready.
* Adds support for density matrices for all gates of the custom backend and implements the channels we currently have (`NoiseChannel` and `GeneralChannel`) for the custom backend.

### Benchmarks

**A. Using standard gates**
```Python
circuit = models.Circuit(nqubits, density_matrix=True)
thetas = np.random.random((2, nqubits))
circuit.add((gates.RX(i, theta=t) for i, t in enumerate(thetas[0])))
circuit.add((gates.CNOT(i, i + 1) for i in range(0, nqubits - 1, 2)))
circuit.add((gates.RX(i, theta=t) for i, t in enumerate(thetas[1])))
circuit.add((gates.CZ(i, i + 1) for i in range(1, nqubits - 2, 2)))
```

* V100 32GB - complex128

nqubits | custom | defaulteinsum | matmuleinsum
-- | -- | -- | --
5 | 0.00717 | 0.45055 | 0.47832 
6 | 0.00794 | 0.44400 | 0.53755 
7 | 0.00906 | 0.45308 | 0.47071 
8 | 0.00960 | 0.44869 | 0.47097 
9 | 0.01083 | 0.45920 | 0.47642 
10 | 0.01262 | 0.45497 | 0.49003 
11 | 0.01994 | 0.45386 | 0.48579 
12 | 0.05696 | 0.46579 | 0.48984 
13 | 0.21595 | 0.93557 | 1.02846 
14 | 0.90314 | 3.63308 | 6.45652 
15 | 3.78549 | OOM | OOM 

* CPU (40 threads) - complex128

nqubits | custom | defaulteinsum | matmuleinsum
-- | -- | -- | --
5 | 0.00669 | 0.01260 | 0.01703 
6 | 0.00881 | 0.01748 | 0.02591 
7 | 0.00825 | 0.03395 | 0.04116 
8 | 0.00903 | 0.08050 | 0.10828 
9 | 0.01502 | 0.31283 | 0.15011 
10 | 0.02858 | 1.33489 | 0.43454 
11 | 0.12693 | 7.13705 | 4.27585 
12 | 0.62657 | 33.09538 | 17.30354 
13 | 2.61106 | 153.42283 | 68.80942 
14 | 11.11244 | 660.79871 | 292.31645 

**B. Using a noise channel**
```Python
circuit = models.Circuit(nqubits, density_matrix=True)
circuit.add((gates.NoiseChannel(i, px=0.2, py=0.05, pz=0.1) for i in range(nqubits)))
```

* V100 32GB - complex128

nqubits | custom | defaulteinsum | matmuleinsum
-- | -- | -- | --
5 | 0.01139 | 0.59474 | 0.47961 
6 | 0.01311 | 0.44951 | 0.48937 
7 | 0.01511 | 0.55261 | 1.06373 
8 | 0.04441 | 0.47595 | 0.47363 
9 | 0.01886 | 0.47205 | 0.47568 
10 | 0.02357 | 0.46394 | 0.60038 
11 | 0.06382 | 0.99325 | 0.50095 
12 | 0.25231 | 0.49612 | 1.30994 
13 | 1.07725 | 1.65317 | 1.92001 
14 | 4.66518 | 6.38950 | 9.77977 
14 | OOM | OOM | OOM

* CPU (40 threads) - complex128

nqubits | custom | defaulteinsum | matmuleinsum
-- | -- | -- | --
5 | 0.01796 | 0.01502 | 0.02108 
6 | 0.02328 | 0.02149 | 0.03117 
7 | 0.03982 | 0.04069 | 0.06335 
8 | 0.04453 | 0.09192 | 0.11126 
9 | 0.05688 | 0.32826 | 0.18424 
10 | 0.14869 | 1.38686 | 0.46737 
11 | 1.03981 | 7.34684 | 4.90600 
12 | 4.40796 | 33.44202 | 19.96818 
13 | 17.69451 | 155.81472 | 79.19668 
14 | 73.99172 | 710.74249 | 344.16950 

### TODO

- [x] Add density matrix tests for all available gates.
- [x] Update examples that use density matrices for the new API and to use the custom backend (I believe it is only the 3_tangle)
- [x] Fix coverage.
- [x] Update docs.


stavros11:
I completed the TODO tasks noted above so this should be ready for review regarding the new cleaner density matrix API and adding density matrix support to the custom backend.

After discussing with Sergi, I believe the following points are remaining to have a more complete first implementation of noise:

1. *(feature)* Bitflip noise to measurement samples.
2. *(feature)* Decoherence / T1, T2 errors.
3. *(optimization)* Memory-efficient density matrix representation using 2^(n-1) * (2^n + 1) elements (since it is a symmetric matrix) instead of the 2^(2n) elements we are currently using.

I will address these with seperate PRs to simplify the review process and keep better track of feature history. I would prioritize the first two so that we have a more complete list of features (the first is very simple anyway) and when these are fixed, depending on other tasks/priorities we may also look at point 3.

stavros11:
As mentioned in the now closed #270, now this PR implements both the density matrix and sampling methods for noise simulation in a more compact way. To summarize the functionality:

First the user specifies if density matrices are to be used during circuit creation using the `density_matrix` flag (which is `False` by default). So 
```Python
c = models.Circuit(nqubits)
```
creates a circuit that always uses state vectors, while 
```Python
c = models.Circuit(nqubits, density_matrix = True)
```
creates a circuit that uses density matrices. All Qibo gates and backends can be used on both types of circuits. We no longer need to change to a Tensorflow backend to use density matrices.

In addition to the existing unitary gates, this PR implements the following channels (substituting the channels we had):
1. `KrausChannel` (old `GeneralChannel`): The general channel defined by arbitrary Kraus operators.
2. `UnitaryChannel`: Special case of (1) that can be written as E[\rho] = (1 - sum(p)) \rho + sum( p U \rho U^\dagger ) where U are unitary operators and p are probabilities.
3. `PauliNoiseChannel` (old `NoiseChannel`): Special case of (2) where the U's are the three Pauli matrices.

The first channel can only be used with density matrices and adding it to a circuit that uses state vectors (was not created with `density_matrix = True`) will raise error. The other two channels can be used in both types of circuits. If they are used in a circuit that uses state vectors, they will act probabilistically using the repeated execution functionality that was implemented in #270.

Regarding the points mentioned in the above post, point 1 (bitflips to measurements) is already implemented in #278, while 2 (T1/T2 errors) should now be easy to implement by adding some new channels (will open a new PR).

doc/source/advancedexamples.rst
```diff
@@ -579,86 +579,149 @@ and using the `Sequential model API <https://www.tensorflow.org/api_docs/python/
...
+as follows:
+
+.. code-block:: python
+
+    import numpy as np
+    from qibo import models, gates
+
+    # Define circuit
+    c = models.Circuit(5)
+    c.add((gates.RX(i, theta=t) for i, t in enumerate(np.random.random())))
```
scarrazza, line 672: I think you have to set the random size explicitly otherwise the loop will fail.
```suggestion
    c.add((gates.RX(i, theta=t) for i, t in enumerate(np.random.random(1))))
```

stavros11: That's correct, thanks for noting. I fixed this together with the typo noted in #278 since it was the same file. It should be okay once we merge #278 here.

src/qibo/base/circuit.py
```diff
@@ -678,7 +678,11 @@ def final_state(self): # pragma: no cover
 
     @abstractmethod
     def execute(self, *args): # pragma: no cover
-        """Executes the circuit. Exact implementation depends on the backend."""
+        """Executes the circuit. Exact implementation depends on the backend.
+
+        See :meth:`qibo.tensorflow.circuit.TensorflowCircuit.execute` for more
+        detaisl.
```
scarrazza, line 684:
```suggestion
        details.
```

src/qibo/tensorflow/gates.py
```diff
@@ -183,6 +188,7 @@ def __init__(self, *q: int, result: List[int] = 0):
         TensorflowGate.__init__(self)
         self.order = None
         self.ids = None
+        self.dm_result = None
```
scarrazza, line 191: Please rename this attribute with the full name, so it is easier to understand.

src/qibo/base/circuit.py
stavros11: Fixed, thanks.

src/qibo/tensorflow/gates.py
stavros11: I renamed this to `density_matrix_result`. In the last commit I also added some checks that probabilities are between 0 and 1 in the `UnitaryChannel` gate (I noted to do this when I was running the tutorial).

scarrazza (approved):
Very good, the code looks clear, and the tests are passing on CPU and GPU. Let's merge this.

PR #277, opened 2020-11-04
Shor (2n+3) example using Qibo

igres26:
Here is the implementation of Shor's factoring algorithm using Qibo.

It follows the implementation introduced by Beauregard in [arxiv.0205095](https://arxiv.org/abs/quant-ph/0205095) that requires only 2n+3 bits if the iQFT is done semiclassically.

Feel free to check the code and tell me if there is anything that I should add, either in the implementation or the `README.md` file.


scarrazza:
@igres26 could you please check if the example test is failing for some reasong/bad configuration of your example?

igres26:
Okay. I think a problem might be that in order for the algorithm to work properly, the minimum number for N that Shor can factorize is 15. We could probably raise an error for that. 

Also, it might be useful to add a primality check to make sure that the number you are trying to factorize is not prime already, since that will probably also make it run given times or might raise some errors within. 

I am not sure that testing for different values of A is necessary, as this is only an option if you know of a good A, or want to try a specific A value for reproducibility.

igres26:
There seems to be another example at the end of the checks report on ubuntu-latest 3.8, is that what should happen?

scarrazza:
@igres26 thanks for checking. The tests for examples are running only on the ubuntu-latest 3.8, to avoid long running times.

Following your comments we may change the setup to:
```python
@pytest.mark.parametrize("N", [15, 20])
@pytest.mark.parametrize("times", [10, 15])
@pytest.mark.parametrize("A", [None])
@pytest.mark.parametrize("semiclassical", [True, False])
@pytest.mark.parametrize("enhance", [True, False])
```

Please let me know if you agree?

igres26:
@scarrazza I would go for something more like:

```python
@pytest.mark.parametrize("N", [15, 21])
@pytest.mark.parametrize("times", [2, 10])
@pytest.mark.parametrize("A", [None])
@pytest.mark.parametrize("semiclassical", [True, False])
@pytest.mark.parametrize("enhance", [True, False])
```

To not have even numbers in N, as they raise a Value Error as even numbers are always factorized by 2. We could consider testing for higher N, but it will start taking too long for the fully quantum iQFT.

The times parameter at 10 is put so that it will more than likely always resolve, so we could try with a smaller times number. 

The rest seems fine.

scarrazza:
Ok, this last CI configuration seems to work well.

examples/shor/README.md
```diff
@@ -0,0 +1,76 @@
+# Shor's factorization algorithm
+
+Code at: [https://github.com/Quantum-TII/qibo/tree/shor/examples/shor](https://github.com/Quantum-TII/qibo/tree/shor/examples/shor)
+
+The original paper by Shor [arXiv:9508027](https://arxiv.org/abs/quant-ph/9508027) stated that problems such as the discrete logarithm problem and factorization could be reduced to order finding. Then, Shor proposed an order finding algorithm using a quantum computer that would run in polynomial time. 
+
+One of the first circuit proposals for Shor's algorithm, mainly on the construction of a quantum modular exponentiation, followed shortly [arXiv:9511018](https://arxiv.org/abs/quant-ph/9511018) by Verdal Barenco and Ekert. This implementation required 7n+2 qubits, where n refers to the number of bits needed to encode the number to factorize.
```
scarrazza, line 7: Vedral, Barenco and Ekert.

examples/shor/README.md
```diff
@@ -0,0 +1,76 @@
...
+#### Quantum order finding algorithm
+
+The aim of this algorithm is to find the least integer r>0 such that x^r = 1 mod N.
+
+- Start from state |00...00>|00...01>.
+- Create full superposition on the first register.
+- Apply modular exponentiation to second register such as ~|j>|x^j mod N>.
+- Apply inverse Quantum Fourier tranform on the first register.
+- Measure the first quantum register. Ther result divided by the maximum possible value can be interpreted as ~(s/r)
+- Use the continued fractions algorithm to estimate the value of r
```
scarrazza, line 20: Full stop, commas missing.

examples/shor/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+The order finding algorithm on a quantum circuit looks like the following.
+
+![order-finding](images/order-finding.png)
+
+#### Quantum factorization of number N
+
+The order finding algorithm can be used as a subroutine to factorize a number N.
+
+To find a noon-trivial factor of N:
```
scarrazza, line 30:
```suggestion
To find a non-trivial factor of N:
```

examples/shor/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+- If N is even, return the factor 2.
+- Check if N = a^b for integers a > 1 and b > 2.
+- Choose x randomly in range(1, N-1). If gcd(x, N) > 1 return it as a factor.
+- Use the quantum order finding algorithm to find order r of x modulo N.
+- If r is even and x^(r/2) is not equal to -1 mod N. Compute gcd(x^(r/2) - 1, N) and gcd(x^(r/2) + 1, N) and check if they are non-trivial factors of N.
+
+This algorith has a chance to fail, as not all random values of x will output an even value for r. In that case the algorithm has to be re-run. It also might be the case that the output of the quantum computer is close, but not exactly the peak we are looking for, in that case, the algorithm could also fail. Regardless, Shor argues that repeating the experiment O(log(log(N))) times is enough to find the solution with high probability.
```
scarrazza, line 38:
```suggestion
This algorithm has a chance to fail, as not all random values of x will output an even value for r. In that case the algorithm has to be re-run. It also might be the case that the output of the quantum computer is close, but not exactly the peak we are looking for, in that case, the algorithm could also fail. Regardless, Shor argues that repeating the experiment O(log(log(N))) times is enough to find the solution with high probability.
```

scarrazza (approved):
Thanks for this great example! 
The docs and code look clear to me.

igres26:
I have applied the changes to the `README.md` file. Thanks for the review.

scarrazza:
@igres26 this is almost ready for merge if @stavros11 agrees.

examples/shor/functions.py
```diff
@@ -0,0 +1,399 @@
...
+    print(f'  - Total number of qubits used: {4*n+2}.\n')
+
+    # Building the quantum circuit
+    for i in range(len(q_reg)):
+        circuit.add(gates.H(q_reg[i]))
+    circuit.add(gates.X(x[len(x)-1]))
+    for i in (range(len(q_reg))):
+        a_i = (a**(2**(i)))
+        circuit.add(c_U(q_reg[i], x, b, a_i, N, ancilla, n))
+    circuit.add(i_qft(q_reg))
```
stavros11, line 336: Small simplifications:
```Python
circuit.add((gates.H(q) for q in q_reg))
circuit.add(gates.X(x[-1]))
circuit.add((c_U(q, x, b, a**(2**i), N, ancilla, n) for i, q in enumerate(q_reg)))
circuit.add(i_qft(q_reg))
```
Similar things can probably be applied in other places but it is minor/optional, just keep whatever looks best for you.

examples/shor/main.py
```diff
@@ -0,0 +1,163 @@
...
+                print('-'*60+'\n')
+            elif f1 != 1 and f1 != N:
+                print(f'Found as factors for {N}:  {f1}  and  {N//f1}.\n')
+                return f1, N//f1
+                break
+            elif f2 != 1 and f2 != N:
+                print(f'Found as factors for {N}:  {f2}  and  {N//f2}.\n')
+                return f2, N//f2
+                break
+        if enhance:
```
stavros11, line 66: I believe it is possible to avoid some code repetition here by defining appropriate functions. For example the first two for loops look identical except for the `ss = s + c` line that becomes `ss = s - c`. So we could do something like `def enhance_func(s, c, ...)` and just call `enhance_func(s, c, ...)` and `enhance_func(s, -c, ...)`. 

I haven't looked at it in much detail but perhaps we could also merge some other cases together and save a few lines of code. For example, the
```Python
f1 = np.gcd((a**(r//2))-1, N)
f2 = np.gcd((a**(r//2))+1, N)
if ...
```
pattern is repeated four times, so it could just be written as a single function called four times.

If you have time you can play with this, otherwise I can proposing something later today and we can check how it looks.

stavros11 (approved):
> @igres26 this is almost ready for merge if @stavros11 agrees.

It looks good to me as well. The only thing I would probably change would be to try to make the classical post-processing part a bit more compact in terms of code, because there are several lines that are repeated 2-4 times. If @igres26 agrees I can have a go and propose something for this later today.

Other than that, I am also planning to do some CPU, GPU and perhaps multi-GPU (if I can find relevant configurations) benchmarks of this. This would require to isolate the quantum part which is the interesting one in terms of performance. I would prefer to finalize the noise PRs before jumping into this so it may take some time, but it can be done after this PR is merged anyway.

examples/shor/main.py
```diff
@@ -0,0 +1,163 @@
...
+            continue
+        print(f"Using Shor's algorithm to factorize N = {N} with a = {a}.\n")
+        if semiclassical:
+            s = functions.quantum_order_finding_semiclassical(N, a)
+        else:
+            s = functions.quantum_order_finding_full(N, a)
+        sr = s/2**(2*n)
+        r = fractions.Fraction.from_float(sr).limit_denominator(N).denominator
+        print(f'Quantum circuit outputs r = {r}.\n')
+        for i in range(1):
```
stavros11, line 44: What is the purpose of this loop? Also, you are using the name `i` here and in the main loop of line 26, which may create some inconsistencies (unless it is on purpose).

stavros11 (commented):
I had a more detailed look and a way to avoid some code repetition is by defining:
```Python
def find_factors(r, a, N):
    if r % 2 != 0:
        print('The value found for r is not even. Trying again.\n')
        print('-'*60+'\n')
        return None
    if a**(r//2) == -1%N:
        print('Unusable value for r found. Trying again.\n')
        print('-'*60+'\n')
        return None
    f1 = np.gcd((a**(r//2))-1, N)
    f2 = np.gcd((a**(r//2))+1, N)
    if (f1 == N or f1 == 1) and (f2 == N or f2 == 1):
        print(f'Trivial factors 1 and {N} found. Trying again.\n')
        print('-'*60+'\n')
        return None
    if f1 != 1 and f1 != N:
        f2 = N // f1
    elif f2 != 1 and f2 != N:
        f1 = N // f2
    print(f'Found as factors for {N}:  {f1}  and  {f2}.\n')
    return f1, f2
```
and changing the code after line 43 to
```Python
        print(f'Quantum circuit outputs r = {r}.\n')
        factors = functions.find_factors(r, a, N)
        if factors:
            return factors
        if enhance:
            for c in range(-4, 5):
                sr = (s + c) / 2**(2*n)
                rr = fractions.Fraction.from_float(sr).limit_denominator(N).denominator
                if rr == r:
                    continue
                print(f'Checking for near values outputs r = {rr}.\n')
                factors = functions.find_factors(rr, a, N)
                if factors:
                    return factors
            print('Checking multiples.\n')
            for c in range(2, n):
                rr = c*r
                factors = functions.find_factors(rr, a, N)
                if factors:
                    return factors
```
I believe this has the same behavior with the current code but saves several lines.

examples/shor/main.py
igres26: The idea was so that a continue would exit the 'loop' and start the next instance. This `i` repetition was probably a mistake on my part then.

examples/shor/functions.py
```diff
@@ -0,0 +1,420 @@
...
+    circuit.add(gates.H(q_reg))
+    circuit.add(gates.M(q_reg))
+    result = circuit(nshots=1)
+    r.append(result.frequencies(binary=False).most_common()[0][0])
+    initial_state = circuit.final_state
+    
+    # Using multiple measurements for the semiclassical QFT.
+    for i in range(1, 2*n):
+        circuit = Circuit(2*n+3)
+        circuit.add(gates.Collapse(q_reg, r[-1]))
```
stavros11, line 378: I believe the collapse gate here does not work as intended and shoud be changed as:
```suggestion
        circuit.add(gates.Collapse(q_reg, result=[r[-1]]))
```

The reason is that the collapse gate definition is `gates.Collapse(*q, result=0)` which means that it accepts multiple qubits as targets. Therefore when you do `Collapse(q_reg, r[-1])` you create a collapse gate acting on qubits `(q_reg, r[-1])` with `result=0` (the default), while what you really want is to collapse `q_reg` to `result=r[-1]`, right?

There are a few other issues I found with the Collapse gate which I will fix in another PR. For example, here ideally you should be able to do `result=r[-1]` (without the list `[r[-1]]`), however if you try this now it will fail (I will fix this). The `result=` keyword should definitely be used when you pass a result, because otherwise he will treat the result number as an additional qubit id.

What I still don't understand is how the algorithm works despite this mistake in the implementation.

stavros11:
Here are some preliminary benchmarks of this:

N | nqubits full | GPU full (sec) | CPU full (sec)  | nqubits semiclassical | GPU semiclassical (sec) | CPU semiclassical (sec)
 -- | -- | -- | -- | -- | -- | -- 
15 | 18 | 1.62925 | 2.07996 | 11 | 1.61887 | 2.03413
21 | 22 | 4.34616 | 13.05518 | 13 | 3.27355 | 4.50020
35 | 26 | 37.48722 | 408.41841 | 15 | 5.82065 | 8.18849
57 | 26 | 37.31165 | 408.94449 | 15 | 5.70549 | 8.36659
69 | 30 | 843.42691 | 10404.33510 | 17 | 9.40674 | 13.82651
87 | 30 | 842.00402 | 10406.81778 | 17 | 9.77131 | 13.88058
95 | 30 | 842.00479 | 10412.59089 | 17 | 9.67036 | 13.80262
1007 | 42 | - | - | 23 | 60.39379 | 
1037 | 46 | - | - | 25 | 175.79384 | 
1703 | 46 | - | - | 25 | 173.76473 | 
2047 | 46 | - | - | 25 | 173.09988 | 
2603 | 50 | - | - | 27 | 743.12006 | 
4083 | 50 | - | - | 27 | 750.63834 |

Times refer only to the execution of the quantum circuits, not the whole factoring algorithm. The quantum circuit may need to be executed more than once in order to succesfully find the factors. The time in the table is calculated by averaging the execution times (note that the number of executions is random and different for each N). I give the modified `main.py` I used for the benchmarks below in case anyone wants to check exactly how I did the logging:

<details>
  <summary>Benchmark script</summary>
  
  ```Python
def exit(data, factors=None, filename=None):
    if factors is not None:
        # log factors found with Shor's quantum algorithm
        data[-1]["factor1"] = int(factors[0])
        data[-1]["factor2"] = int(factors[1])
    # write logs to file
    if filename is not None:
        with open(filename, "w") as file:
            json.dump(data, file)
    return factors


def main(N, times, A, semiclassical, enhance, filename):
    if N % 2 == 0:
        raise ValueError(f'Input number {N} needs to be odd for it to be a hard problem.')
    f = functions.find_factor_of_prime_power(N)
    if f is not None:
        raise ValueError(f'Input number {N} is a prime power with factor {f}.')
    n = int(np.ceil(np.log2(N)))

    # open log file if it exists
    if filename is not None and os.path.isfile(filename):
        with open(filename, "r") as file:
            data = json.load(file)
    else:
        data = []

    new_data = False
    for i in range(times):
        if A:
            a = A
        else:
            a = np.random.randint(2, N - 1)

        # append run data to logs
        data.append({"date": datetime.now().strftime("%d/%m/%Y %H:%M:%S"),
                     "cuda_visible_devices": os.environ["CUDA_VISIBLE_DEVICES"],
                     "N": N, "a": int(a), "s": 0, "iteration": i,
                     "factor1": None, "factor2": None,
                     "nqubits": 0,
                     "semiclassical": semiclassical,
                     "device": qibo.get_device(),
                     "backend": qibo.get_backend(),
                     "precision": qibo.get_precision(),
                     "total_time": 0})

        f = np.gcd(a, N)
        if 1 < f < N:
            print(f"N = {N} and a = {a} share factor {f}. Trying again for quantum instance.\n")
            print('-'*60+'\n')
            # log factors found classically
            data[-1]["factor1"] = int(f)
            data[-1]["factor2"] = int(N // f)
            continue
        print(f"Using Shor's algorithm to factorize N = {N} with a = {a}.\n")

        start_time = time.time()
        if semiclassical:
            s, nqubits = functions.quantum_order_finding_semiclassical(N, a)
        else:
            s, nqubits = functions.quantum_order_finding_full(N, a)
        # log quantum run
        data[-1]["total_time"] =  time.time() - start_time
        data[-1]["nqubits"] = int(nqubits)
        data[-1]["s"] = int(s)

        sr = s/2**(2*n)
        r = fractions.Fraction.from_float(sr).limit_denominator(N).denominator
        print(f'Quantum circuit outputs r = {r}.\n')
        factors = functions.find_factors(r, a, N)
        if factors:
            return exit(data, factors, filename)
        if enhance:
            for c in range(-4, 5):
                sr = (s + c) / 2**(2*n)
                rr = fractions.Fraction.from_float(sr).limit_denominator(N).denominator
                if rr == r:
                    continue
                print(f'Checking for near values outputs r = {rr}.\n')
                factors = functions.find_factors(rr, a, N)
                if factors:
                    return exit(data, factors, filename)
            print('Checking multiples.\n')
            for c in range(2, n):
                rr = c*r
                factors = functions.find_factors(rr, a, N)
                if factors:
                    return exit(data, factors, filename)
    return exit(data, None, filename)
  ```
</details>


Some things I noticed:
* As mentioned in the comment above there are some issues with the Collapse gate both regarding its use in this PR but also in its general implementation (I will open a PR tomorrow for the latter).
* When executing the semiclassical for large numbers (that need ~30 qubits) on GPU, it seems to get stuck somewhere with the GPU at 0% utilization. I plan to investigate more why this happens.

examples/shor/functions.py
igres26: I see, was it like this when we first implemented Collapse? I think I did not change much from our first implementation. I will fix this. 

If I have to guess why it still seems to work is because the collapse fixes the measured qubit to either 1 or 0, so the ones that coincide with 0 still work. But that could explain why the probability of success seemed lower than I expected.

examples/shor/functions.py
```diff
@@ -0,0 +1,420 @@
...
+    ancilla = 2*n + 1
+    q_reg = 2*n + 2
+    circuit = Circuit(2*n+3)
+    print(f'  - Total number of qubits used: {2*n+3}.\n')
+    r = []
+    
+    # Building the quantum circuit
+    circuit.add(gates.H(q_reg))
+    circuit.add(gates.X(x[len(x)-1]))
+    a_i = (a**(2**(2*n - 1)))
```
stavros11, line 367: I explored a bit more why the semiclassical approach crashes for 29 qubits and the problem seems to be this line. @igres26, if you like to confirm this you can simply print something right before calculating `a_i` and something right after and try to run this for a large number (eg. `python main.py --semiclassical --N 7781`) in a machine with sufficient memory (eg. DGX).

The issue is that `a_i` is a huge number which presumably Python cannot handle. For example if we have 29 qubits semiclassical, this means that `n=13` so `2**(2*n-1)=2**25` and then you attempt to calculate `a_i = a ** (2 ** 25)` which is huge if a > 1. This number has log10(a_i) ~ 2 ** 25 log10(a) > 30 million digits so it is probably not a good idea to work with it. I cannot think of any easy trick around this. Since this number is probably a phase of some gate, is there any possibility that it could be simplified using periodicity? (usually in phases you can subtract multiples of 2pi)

igres26: Umm... that makes sense. The thing here is that we want to perform the modular exponentiation of `a**x` with `x` up to `N**2`, that is why it goes as much as `a**(2**(2*n-1))`. I suppose that when a is close to N, this number gets too big for python to keep track of correctly. 

This is not a phase of a gate, but it is part of the modular exponentiation, so it might be the case that we can apply some modulo N before calculating it. A modulo N operation is done just before calculating the angles of the needed gates.

igres26:
I think the problem with big numbers should be solved now. We can now push for larger numbers if needed.

stavros11:
> I think the problem with big numbers should be solved now. We can now push for larger numbers if needed.

I believe it is indeed fixed, thanks! Here are some benchmarks including larger numbers on the DGX GPU:

N | nqubits full | GPU full (sec) | nqubits semiclassical | GPU semiclassical (sec)
 -- | -- | -- | -- | --
15 | 18 | 1.60387 | 11 | 1.56673
21 | 22 | 4.29408 | 13 | 3.18129
35 | 26 | 37.93513 | 15 | 5.64433
55 | 26 | 37.77242 | 15 | 5.66065
77 | 30 | 842.60244 | 17 | 9.78004
91 | 30 | 841.48022 | 17 | 9.63527
95 | 30 | 841.57430 | 17 | 9.82748
1007 | 42 | - | 23 | 61.37579
1763 | 46 | - | 25 | 167.97211
1927 | 46 | - | 25 | 168.97477
2021 | 46 | - | 25 | 168.21349
3953 | 50 | - | 27 | 679.87418
4087 | 50 | - | 27 | 680.13621
7387 | 54 | - | 29 | 3272.63385
8051 | 54 | - | 29 | 3265.72749

Following what JI suggested, I used numbers with large prime factors (eg. 8051 = 83 * 97). As pointed out in the previous post, the times refer to a single execution of the quantum subroutine, but more than one executions may be needed to get the factors. I guess the number of executions depends on the choice of a which is random. For example 2021 required 10 executions, while 8051 required only one. When >1 executions are needed I calculate the average simulation time. I can confirm that the script succesfully found the factors for all the numbers mentioned above.

I am now running for N=15707 and N=15481 (31 qubits) which is the largest we can do on single GPU with single precision. I will update the table once I have the results.

**Distributed**

I performed a distributed run using all 4x DGX GPUs and single precision to factor N=32399=179*181 which requires 33 qubits. This took **13864.3sec** (a little less than 4 hours) for a single execution but did not find the factors. I guess that if we keep trying for different a it should find them, it will just need 4 hours for each a. I just using our distributed circuit out-of-the-box without any optimizations or re-ordering of gates, which is very simple to add to the code. It may be possible to improve this time with more work, though.

scarrazza:
@stavros11 thanks for this great benchmark.
@igres26 @stavros11 shall we merge this PR?

igres26:
Seems to be working fine with me. Maybe the larger examples can be found with less iterations using the `--enhance` module. But everything seems fine on my end.

stavros11:
> @stavros11 thanks for this great benchmark.
> @igres26 @stavros11 shall we merge this PR?

It looks good to me and now works for large numbers and finds the factors after a few iterations. It found the factors for all numbers in the benchmark table except the largest one (32399) but this is because I did only one repetition for that one.

I am not sure if Sergi has done any checks on whether the probabilities of success are at the expected levels. In my benchmark logging, I keep track of the periods and factors found in each iteration but nothing regarding probabilities.

igres26:
The probability of finding a suitable a is 1/2 and then there is a bit of extra error probability from not measuring the peak but instead a number close to it, or measuring the 0 peak which is always there.

Without the `--enhance` flag, the number of iterations should be around ~log(log(N)) according to the original Shor paper.

PR #278, opened 2020-11-06
Post-measurement bit-flip errors

stavros11:
Following the [last post](https://github.com/Quantum-TII/qibo/pull/276#issuecomment-722297367) from #276, this adds some functionality to add bit-flip noise to measured samples (point 1). More specifically this implements:

* A `CircuitResult.apply_bitflips(error_map)` method that can be used to add bitflip noise on existing measured samples.
* The possibility to pass a bit-flip error map in measurement gates:

```Python
c = models.Circuit(5)
c.add(...)
c.add(gates.M(0, 1, bitflips=0.2))
c.add(gates.M(3, 4, bitflips={3: 0.1, 4: 0.3}))
result = c(nshots=100)
# measured samples will have some bit-flip errors according to the given probabilities
```

Note that in both cases the errors are added in the measurement results after simulation/sampling is finished. The user could easily do this manually using just a numpy/Tensorflow random generator however I thought it may be useful to have this functionality built-in.

PR #279, opened 2020-11-13
Thermal relaxation error channel

### LINKED ISSUE #63: Thermal relaxation and measurement errors
Hi all, 

I am including some issues about errors.

First of all: measurements error
As far as we know, some others simulation packages include measurement errors in two differnt ways. One way is to apply a channel before applying the gate. The other way is to perform a measurement and then flip the result with certain probability. This probability does not have to be the same for 0-> 1 and 1-> 0, so it allows non-symmetric errors. Is it possible to implement this kind of errors?

Second: thermal relaxation errors
These errors are designed to simulate the interaction between the qubit and the environment. It is inherited from NMR: https://en.wikipedia.org/wiki/Relaxation_%28NMR%29, https://arxiv.org/pdf/1609.04485.pdf . It can be understood as "how the qubit loses its quantum behaviour", however I am not an expert of this file. Take a look at that and tell me if this makes sense to you.

stavros11:
This implements two new channels: `ResetChannel` and `ThermalRelaxationChannel`, both following the corresponding [Qiskit implementation](https://qiskit.org/documentation/_modules/qiskit/providers/aer/noise/errors/standard_errors.html#thermal_relaxation_error). Note that the thermal relaxation channel, at least as it is implemented in Qiskit, is different depending on which of T1, T2 is larger. Specifically:

* If T1 >= T2 (which I think is the common case experimentally), the channel is a probabilistic mixture of unitaries and collapse gate so it can be implemented as a special case of `UnitaryChannel`. This case can also be simulated with state vectors and sampling.

* If T1 < T2 the channel is general (not a unitary mixture) and can only be used with density matrices. Qiskit uses the Choi matrix representation to define the channel in this case but since we don't have an implementation of this, I just worked out the transformation on paper and coded the functions that implement it both for the custom and tf backends.

All transformation equations are written in detail in our docs. If @AdrianPerezSalinas agrees, this PR together with #278 (already merged) should close #63.

### DIFF
--- a/src/qibo/base/gates.py
+++ b/src/qibo/base/gates.py
@@ -1843,3 +1843,58 @@ class PauliNoiseChannel(UnitaryChannel):
 
         self.init_args = [q]
         self.init_kwargs = {"px": px, "py": py, "pz": pz, "seed": seed}
+
+
+class ThermalRelaxationChannel(UnitaryChannel):
+
+    def __init__(self, q, t1, t2, time, excited_population=0, seed=None):
+        probs = self._calculate_probs(t1, t2, time, excited_population)
+        gates = [self.module.Z(q), self.module.Collapse(q),
+                 self.module.X(q)]
+
+        super(ThermalRelaxationChannel, self).__init__(probs, gates, seed=seed)
+        self.name = "ThermalRelaxationChannel"
+        assert self.target_qubits == (q,)
+
+        self.init_args = [q, t1, t2, time]
+        self.init_kwargs = {"excited_population": excited_population,
+                            "seed": seed}
+
+    @staticmethod
+    def _calculate_probs(t1, t2, time, excited_population):
+        import numpy as np
+        if excited_population < 0 or excited_population > 1:
+            raise_error(ValueError, "Invalid excited state population {}."
+                                    "".format(excited_population))
+        if time < 0:
+            raise_error(ValueError, "Invalid gate_time ({} < 0)".format(time))
+        if t1 <= 0:
+            raise_error(ValueError, "Invalid T_1 relaxation time parameter: T_1 <= 0.")
+        if t2 <= 0:
+            raise_error(ValueError, "Invalid T_2 relaxation time parameter: T_2 <= 0.")
+        if t2 - 2 * t1 > 0:
+            raise_error(ValueError, "Invalid T_2 relaxation time parameter: "
+                                    "T_2 greater than 2 * T_1.")
+        # T1 relaxation rate
+        if t1 == np.inf:
+            rate1 = 0
+            p_reset = 0
+        else:
+            rate1 = 1 / t1
+            p_reset = 1 - np.exp(-time * rate1)
+        # T2 dephasing rate
+        if t2 == np.inf:
+            rate2 = 0
+            exp_t2 = 1
+        else:
+            rate2 = 1 / t2
+            exp_t2 = np.exp(-time * rate2)
+
+        if t2 > t1:
+            # TODO: Implement this case
+            raise_error(NotImplementedError)
+        else:
+            probs = ((1 - p_reset) * (1 - np.exp(-time * (rate2 - rate1))) / 2,
+                     p_reset * (1 - excited_population),
+                     p_reset * excited_population)
+        return probs
--- a/src/qibo/base/gates.py
+++ b/src/qibo/base/gates.py
@@ -1826,3 +1826,20 @@ class PauliNoiseChannel(UnitaryChannel):
 
         self.init_args = [q]
         self.init_kwargs = {"px": px, "py": py, "pz": pz, "seed": seed}
+
+
+class ResetChannel(UnitaryChannel):
+
+    def __init__(self, *q, p0=0.0, p1=0.0, seed=None):
+        probs, gates = [], []
+        for r, p in enumerate([p0, p1]):
+            if p > 0:
+                probs.append(p)
+                gates.append(gates.Collapse(*q, result=r))
+
+        super(ResetChannel, self).__init__(probs, gates, seed=seed)
+        self.name = "ResetChannel"
+        assert self.target_qubits == tuple(q)
+
+        self.init_args = [q]
+        self.init_kwargs = {"p0": p0, "p1": p1, "seed": seed}
--- a/src/qibo/base/gates.py
+++ b/src/qibo/base/gates.py
@@ -1669,23 +1669,6 @@ class CallbackGate(Gate):
         self.callback.nqubits = n
 
 
-class ResetChannel(Gate):
-
-    def __init__(self, q, p0=0.0, p1=0.0, seed=None):
-        super(ResetChannel, self).__init__()
-        self.name = "ResetChannel"
-        self.target_qubits == (q,)
-        self.seed = seed
-
-        self.p0, self.p1 = p0, p1
-        self.psum = p0 + p1
-        self.collapse = self.module.Collapse(q)
-        self.flip = self.module.X(q)
-
-        self.init_args = [q]
-        self.init_kwargs = {"p0": p0, "p1": p1, "seed": seed}
-
-
 class KrausChannel(Gate):
     """General channel defined by arbitrary Krauss operators.
 
@@ -1845,6 +1828,19 @@ class PauliNoiseChannel(UnitaryChannel):
         self.init_kwargs = {"px": px, "py": py, "pz": pz, "seed": seed}
 
 
+class ResetChannel(UnitaryChannel):
+
+    def __init__(self, q, p0=0.0, p1=0.0, seed=None):
+        probs = [p0, p1]
+        gates = [self.module.Collapse(q), self.module.X(q)]
+        super(ResetChannel, self).__init__(probs, gates, seed=seed)
+        self.name = "ResetChannel"
+        assert self.target_qubits == (q,)
+
+        self.init_args = [q]
+        self.init_kwargs = {"p0": p0, "p1": p1, "seed": seed}
+
+
 class ThermalRelaxationChannel(UnitaryChannel):
 
     def __init__(self, q, t1, t2, time, excited_population=0, seed=None):
@@ -1869,9 +1865,11 @@ class ThermalRelaxationChannel(UnitaryChannel):
         if time < 0:
             raise_error(ValueError, "Invalid gate_time ({} < 0)".format(time))
         if t1 <= 0:
-            raise_error(ValueError, "Invalid T_1 relaxation time parameter: T_1 <= 0.")
+            raise_error(ValueError, "Invalid T_1 relaxation time parameter: "
+                                    "T_1 <= 0.")
         if t2 <= 0:
-            raise_error(ValueError, "Invalid T_2 relaxation time parameter: T_2 <= 0.")
+            raise_error(ValueError, "Invalid T_2 relaxation time parameter: "
+                                    "T_2 <= 0.")
         if t2 - 2 * t1 > 0:
             raise_error(ValueError, "Invalid T_2 relaxation time parameter: "
                                     "T_2 greater than 2 * T_1.")
--- a/src/qibo/base/gates.py
+++ b/src/qibo/base/gates.py
@@ -1934,10 +1934,6 @@ class ThermalRelaxationChannel:
             instead of density matrices is used to simulate this gate.
     """
 
-    def __new__(cls, q, t1, t2, time, excited_population=0, seed=None):
-        # abstract method
-        raise_error(NotImplementedError)
-
     def __init__(self, q, t1, t2, time, excited_population=0, seed=None):
         self.name = "ThermalRelaxationChannel"
         self.init_args = [q, t1, t2, time]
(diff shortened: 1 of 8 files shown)


src/qibo/base/gates.py
```diff
@@ -1863,3 +1868,158 @@ def __init__(self, q, px=0, py=0, pz=0, seed=None):
...
         self.init_kwargs = {"px": px, "py": py, "pz": pz, "seed": seed}
+
+
+class ResetChannel(UnitaryChannel):
+    """Single-qubit reset channel.
+
+    Implements the following transformation:
+
+    .. math::
+        \\mathcal{E}(\\rho ) = (1 - p_0 - p_1) \\rho +
```
scarrazza, line 1879:
```suggestion
        \\mathcal{E}(\\rho ) = (1 - p_0 - p_1) \\rho
```

src/qibo/base/gates.py
```diff
@@ -1863,3 +1868,158 @@ def __init__(self, q, px=0, py=0, pz=0, seed=None):
...
+    and :math:`p_z = 1 - e^{-t / T_1} + e^{-t / T_2} - e^{t / T_1 - t / T_2}`.
+    Here :math:`\\eta` is the ``excited_population``
+    and :math:`t` is the ``time``, both controlled by the user.
+    This gate is based on
+    `Qiskit's thermal relaxation error channel <https://qiskit.org/documentation/stubs/qiskit.providers.aer.noise.thermal_relaxation_error.html#qiskit.providers.aer.noise.thermal_relaxation_error>`_.
+
+    Args:
+        q (int): Qubit id that the noise channel acts on.
+        t1 (float): T1 relaxation time.
+        t2 (float): T2 dephasing time.
```
scarrazza, line 1951: Maybe you should say something about:
- t1, t2 > 0
- time >= 0
- t2 > 2 * t1

src/qibo/base/gates.py
```diff
@@ -1863,3 +1868,158 @@ def __init__(self, q, px=0, py=0, pz=0, seed=None):
...
+                                    "T_2 greater than 2 * T_1.")
+
+        p_reset = 1 - np.exp(-time / t1)
+        p0 = p_reset * (1 - excited_population)
+        p1 = p_reset * excited_population
+        if t1 < t2:
+            exp = np.exp(-time / t2)
+        else:
+            rate1, rate2 = 1 / t1, 1 / t2
+            exp = (1 - p_reset) * (1 - np.exp(-time * (rate2 - rate1))) / 2
```
scarrazza, line 1990: I'm trying to find this eq. is this supposed to be `(1 - pz - p0 - p1)`?

src/qibo/tensorflow/cgates.py
```diff
@@ -737,10 +737,12 @@ def _prepare(self):
             inv_gate = self._invert(gate)
             # use a ``set`` for this loop because it may be ``inv_gate == gate``
             for g in {gate, inv_gate}:
-                g.density_matrix = self.density_matrix
-                g.device = self.device
-                g.nqubits = self.nqubits
+                if g is not None:
```
scarrazza, line 740: Is it possible to have `None` here?

src/qibo/tensorflow/cgates.py
```diff
@@ -797,5 +802,100 @@ def __init__(self, q: int, px: float = 0, py: float = 0, pz: float = 0,
 
     @staticmethod
     def _invert(gate):
-        """For Pauli gates we can use same gate for state inversion for efficiency."""
+        # for Pauli gates we can use same gate as inverse for efficiency
         return gate
+
+
+class ResetChannel(UnitaryChannel, base_gates.ResetChannel):
```
scarrazza, line 809: I have a couple of questions here:
- I see that the inheritance from MatrixGate is used in H, however not for X, Y, Z, is that correct?
- this class inherits from UnitaryChannel, but you don't initialize its constructor because you are planning to use just the `_density_matrix_call`, correct? (and not the _prepare which needs the KrausChannel)

src/qibo/tensorflow/cgates.py
```diff
@@ -797,5 +802,100 @@ def __init__(self, q: int, px: float = 0, py: float = 0, pz: float = 0,
...
+
+    def __new__(cls, q, t1, t2, time, excited_population=0, seed=None):
+        if BACKEND.get('GATES') == "custom":
+            cls_a = _ThermalRelaxationChannelA
+            cls_b = _ThermalRelaxationChannelB
+        else:
+            from qibo.tensorflow import gates
+            cls_a = gates._ThermalRelaxationChannelA
+            cls_b = gates._ThermalRelaxationChannelB
+        if t2 > t1:
```
scarrazza, line 846: Ok, I suppose checking if t1,t2 are doubles in in `base_gates.ThermalRelaxationChannel` is probably useless at this stage.

src/qibo/tensorflow/gates.py
```diff
@@ -35,6 +36,22 @@ def _prepare(self):
         rank = int(np.log2(int(matrix.shape[0])))
         self.matrix = tf.reshape(matrix, 2 * rank * (2,))
 
+    def _calculate_einsum_cache(self):
+        if self.is_controlled_by:
+            if self.density_matrix:
+                from qibo.tensorflow import einsum
+                self.einsum = einsum.DefaultEinsum()
```
scarrazza, line 43: Could you please remember me why? (maybe we need some comments here)

src/qibo/tensorflow/gates.py
```diff
@@ -508,11 +492,11 @@ def _calculate_unitaries(self):
         return matrices, additional_matrix
 
     def _prepare(self):
-        self.cgates.VariationalLayer._prepare(self)
+        cgates.VariationalLayer._prepare(self)
```
scarrazza, line 495: That's a good idea, avoids code duplication. The only downside (which we don't have so far) is if the attributes used by the external method do are not allocated/named from the original class, i.e. in this case you suppose `self.is_dagger` exists, because it is obviously allocated in the base class.

scarrazza (approved):
LGTM, I have placed some comments, but in general looks clean and good to merge.
I have verified that tests are passing on GPU.

src/qibo/base/gates.py
stavros11: In this case (T1 > T2) `exp` refers to the first probability in the `probs` list so it is ``pz``. If you expand the product `(1 - p_reset) * (1 - np.exp(...))` you will find that it is equivalent to the equation for pz given in the docs. This is the "original" equation from Qiskit for pz, I just wrote the expanded form in the docs because it was looking better with the symbols I used there.

src/qibo/tensorflow/cgates.py
stavros11: Yes, this may be a bit obscure as it is implemented now. I think the only case it happens is when the gate is the Collapse gate which has no inverse, so I use `None` for the inverse gate (see the `_invert` method definition in `ResetChannel` - line 818 of `cgates.py`). The Collapse gate is non-invertible mathematically speaking anyway, since it is not really a "gate" but the result of measuring. Both `ResetChannel` and `ThermalRelaxationChannel` use the `Collapse` gate.

By the way, it is generally not required to invert gates in order to apply the channels. The reason I do this here is that custom operators work with in-place updates, while when applying channels we need to apply several gates to the same state and then sum the terms. There are two ways to do this:
1. Create a copy of the state (can be done for example with a trick `temp_state = state + tf.zeros_like(state)`) and use this in the custom operators, or
2. Apply each channel gate to the state but then apply its inverse to reset to to the original state before applying the next channel gate (what I do now).

The second method is a bit more complicated in terms of code but I used it because it was slightly faster in some tests I did.

src/qibo/tensorflow/cgates.py
stavros11: > * I see that the inheritance from MatrixGate is used in H, however not for X, Y, Z, is that correct?

Yes, `MatrixGate` is essentially the "abstract" class for gates that use the `op.apply_gate` or `op.apply_two_qubit_gate` custom operators which take a gate `matrix` argument. X, Y and Z do not inherit this because they are not applied via matrix multiplication but via the `op.apply_x`, `op.apply_y`, `op.apply_z` which have the "matrices" of each gate hard-coded in the C++ code.

> * this class inherits from UnitaryChannel, but you don't initialize its constructor because you are planning to use just the _density_matrix_call, correct? (and not the _prepare which needs the KrausChannel)

Inheritance is a bit complicated to follow with channels, but I am not sure if there is any way to simplify things. For example, you may notice that I never call the constructor of the parent channel in the backend modules (inside the `tensorflow` folder) because the parent constructor is called from the base constructor.

Regarding the `ResetChannel`, it inherits both the `_prepare` and `_density_matrix_call` methods from the `UnitaryChannel`. Also, channels that inherit the `UnitaryChannel` are the ones that can be used with state vectors / sampling. The others can only be used with density matrices.

src/qibo/tensorflow/gates.py
stavros11: After expanding the density matrix tests in the two previous PRs, I noticed that `'matmuleinsum'` was not working properly for the case where `controlled_by` gates were used on density matrices. This is a temporary solution that falls back to the `'defaulteinsum'` automatically so that everything works properly. The long term solution would be to actually fix matmuleinsum. I just did not want to delay these PRs for this because I believe the combination density matrices + controlled by + matmuleinsum would be very rare (if it ever appears) in practice. Perhaps I should open an issue about that though.

src/qibo/tensorflow/gates.py
stavros11: Thanks for the review and for merging. I was actually going to reply to this, I just did not have time before. Yes, this is done to avoid code duplication, however I generally don't like it as a practice to call other class methods without strictly respecting inheritance. 

Generally, I am not really sure if the way we implement gates for different backends is the best one and what problems could arise if we attempt to add a new backend in the future. This is because the initial design did not consider that we would have the custom backend (this was added as Qibo evolved) and admittedly the focus was more on benchmarks/performance rather than code simplicity and having a well-defined inheritance structure. I was a bit loose on the latter part for the sake of getting things to work and with acceptable performance. At some point, depending on future developments, we may want to have a second look on the backends and gate inheritance and have a more well-defined structure of what is defined in base and what in each backend.

The main reason I am not spending much time on this now is that I am quite unsure about the future of Tensorflow backends. Since we added density matrix support on the custom backend the only two things that could justify using a native backend now are 1. compilation and 2. automatic differentiation, which I am not sure how useful are in applications (point 2 could be useful, however it is unclear if this approach is really useful for optimization of variational circuits). With this in mind I am not sure for how long it would be worth to maintain + complicate the code & docs for keeping these backends. On the other hand, we cannot really drop them as Qibo was officially released this way.

scarrazza: Thanks for the reply, and yes, that's a good point. Indeed, we could simplify the backends and limit the implementation to custom however we still in a warm-up phase. I would like to have only the custom backend, however the sgd approach will not work out of the box. We may investigate more about https://github.com/tensorflow/tensorflow/issues/40006 or eventually replace tf with [jax](https://github.com/google/jax).

AdrianPerezSalinas:
> This implements two new channels: `ResetChannel` and `ThermalRelaxationChannel`, both following the corresponding [Qiskit implementation](https://qiskit.org/documentation/_modules/qiskit/providers/aer/noise/errors/standard_errors.html#thermal_relaxation_error). Note that the thermal relaxation channel, at least as it is implemented in Qiskit, is different depending on which of T1, T2 is larger. Specifically:
> 
>     * If T1 >= T2 (which I think is the common case experimentally), the channel is a probabilistic mixture of unitaries and collapse gate so it can be implemented as a special case of `UnitaryChannel`. This case can also be simulated with state vectors and sampling.
> 
>     * If T1 < T2 the channel is general (not a unitary mixture) and can only be used with density matrices. Qiskit uses the Choi matrix representation to define the channel in this case but since we don't have an implementation of this, I just worked out the transformation on paper and coded the functions that implement it both for the custom and tf backends.
> 
> 
> All transformation equations are written in detail in our docs. If @AdrianPerezSalinas agrees, this PR together with #278 (already merged) should close #63.



> This implements two new channels: `ResetChannel` and `ThermalRelaxationChannel`, both following the corresponding [Qiskit implementation](https://qiskit.org/documentation/_modules/qiskit/providers/aer/noise/errors/standard_errors.html#thermal_relaxation_error). Note that the thermal relaxation channel, at least as it is implemented in Qiskit, is different depending on which of T1, T2 is larger. Specifically:
> 
>     * If T1 >= T2 (which I think is the common case experimentally), the channel is a probabilistic mixture of unitaries and collapse gate so it can be implemented as a special case of `UnitaryChannel`. This case can also be simulated with state vectors and sampling.
> 
>     * If T1 < T2 the channel is general (not a unitary mixture) and can only be used with density matrices. Qiskit uses the Choi matrix representation to define the channel in this case but since we don't have an implementation of this, I just worked out the transformation on paper and coded the functions that implement it both for the custom and tf backends.
> 
> 
> All transformation equations are written in detail in our docs. If @AdrianPerezSalinas agrees, this PR together with #278 (already merged) should close #63.

Hi all, I know I am a bit late, but I have checked this PR and the previous issues on thermal errors, and everything seems fine to me. Thank you for the work!!

PR #280, opened 2020-11-14
Upgrading tf to 2.4.0

scarrazza:


stavros11:
Have you tried our tests with tf2.4? I just tried and I get the following failure 
```
AttributeError: module 'tensorflow' has no attribute 'python'
```
because we are using `oom_error = tf.python.framework.errors_impl.ResourceExhaustedError` in some exceptions in order to raise our own OOM error. If this is the only issue, I guess we could just find what is the corresponding error for tf2.4 and update accordingly, but we should do it in a way that does not break compatibility with older tf versions.

Note that pip still installs tf2.3.1 by default (unless we explicitly do `pip install tensorflow==2.4.0rc1`), which means that CI is not testing for tf2.4. We may want to update that as well.

scarrazza:
Yes, sorry for not being more explicit, this should be a draft PR to fix this https://github.com/Quantum-TII/qibo/network/alert/requirements.txt/tensorflow/open. As soon as pypi uploads the final tf release we can go ahead and fix/test the changes.

scarrazza:
Yes, sorry for not being more explicit, this should be a draft PR to fix this https://github.com/Quantum-TII/qibo/network/alert/requirements.txt/tensorflow/open. As soon as pypi uploads the final tf release we can go ahead and fix/test the changes.

scarrazza:
@stavros11 this should be ready to go.

stavros11 (approved):
Thanks for the updates, for me it works with tf2.4 on CPU. I could not test on GPU yet due to the CUDA issue (tf2.4 requires CUDA 11 while I had 10.1). 

I would expect this to be fine on GPU but if you already have a tf2.4 GPU configuration installed in any machine it would be great to confirm this by running the tests there.

scarrazza:
@stavros11 I have checked using anaconda with python38, tf2.4.0 cuda11. Tests are passing on GPU. I will retry on other systems and then merge.

scarrazza:
I confirm that tests are passing even with python3.9 and cuda 11.1.

PR #281, opened 2020-11-15
Parallel L-BFGS-B

scarrazza:
This PR should replace #271 by:
- making the circuit evaluation thread-safe
- fixing the computation freeze when using custom operators

@stavros11 given that multiprocessing requires changes that from a design point of view, in order to achieve non negligible performance improvements, forces the code to work with singletons and global states, this PR is a draft to be discussed before going into all implementation details, tests, etc.

At the current stage, the VQE code in `examples/benchmarks/vqe.py` is working with:
```bash
python vqe.py --nqubits 4 --nlayers 4 --method parallel_lbfgsb --maxiter 15000
```

Could you please have a look and let me know your opinion? Personally, I totally dislike approaches with singletons/global states, however in this particular case seems to be the only effective approach after testing several alternatives (passing circuits as args, doing in place copies, passing class methods).


scarrazza:
I have checked the details of the optimparallel in order to understand how it is implemented. The L-BFGS-B method is by construction sequential, and each step requires the evaluation of the function and its gradient. Thus, the term "parallelize"  here refers to computing the function value and its gradient in parallel for a given set of parameters. This may be particularly efficient for the numerical gradients evaluation if the function evaluation is costly.

That being said, the parallel optimization is controlled by the sequential `scipy.minimize` which instead of calling the native function, takes an object which parallelizes the function and gradient evaluations. 

@stavros11 if the code in this PR is too obscure, we can possibly implement a parallel wrapper for scipy.minimize which deals with circuits using multiprocessing.

src/qibo/optimizers.py
```diff
@@ -56,8 +56,16 @@ def newtonian(loss, initial_parameters, method='Powell', options=None):
         options (dict): Dictionary with options accepted by
             ``scipy.optimize.minimize``.
     """
-    from scipy.optimize import minimize
-    m = minimize(loss, initial_parameters, method=method, options=options)
+    if method == 'parallel_lbfgsb':
```
stavros11, line 59: That's okay, but the fact that it is lower case may create some confusion with the equivalent pure scipy method which is `"L-BFGS-B"` with capitals and dashes. Perhaps something like `"parallel_LBFGSB"` or `"parallel_L-BFGS-B"` would be better?

src/qibo/optimizers.py
```diff
@@ -56,8 +56,16 @@ def newtonian(loss, initial_parameters, method='Powell', options=None):
...
+        from qibo.config import raise_error, get_backend, THREADS
+        if get_backend() == 'custom' and THREADS.get("intra") != 1:
+            raise_error(RuntimeError, "Parallel L-BFGS-B with custom operator requires to"
+                                    " set the QIBO_NUM_THREADS=1 environment variable.")
+        from optimparallel import minimize_parallel
```
stavros11, line 64: I would add an additional check
```Python
if "GPU" in get_device():
    raise_error(RuntimeError, "Parallel L-BFGS-B cannot be used with GPU.")
```
The reason is that when I first tested this on my machine, I accidentally forgot the GPU enabled and noticed that the code just gets stuck and does nothing in this case. In addition to this, if the script is killed with Ctrl+C then the GPU memory is not emptied (same problem we observed on DGX) unless the machine is rebooted (which can be annoying and may even not be possible for some servers).

src/qibo/optimizers.py
```diff
@@ -56,8 +56,16 @@ def newtonian(loss, initial_parameters, method='Powell', options=None):
         options (dict): Dictionary with options accepted by
             ``scipy.optimize.minimize``.
     """
-    from scipy.optimize import minimize
-    m = minimize(loss, initial_parameters, method=method, options=options)
+    if method == 'parallel_lbfgsb':
+        from qibo.config import raise_error, get_backend, THREADS
+        if get_backend() == 'custom' and THREADS.get("intra") != 1:
```
stavros11, line 61: In principle we could set the intra threads to 1 automatically here instead of raising an error, right? On the other hand, it may not be a good idea to modify things under-the-hood without the user knowing.

src/qibo/models.py
```diff
@@ -169,7 +193,7 @@ def __init__(self, circuit, hamiltonian):
         self.circuit = circuit
         self.hamiltonian = hamiltonian
 
-    def minimize(self, initial_state, method='Powell', options=None, compile=True):
+    def minimize(self, initial_state, method='Powell', options=None, parallel=None, compile=True):
```
stavros11, line 196: If I understood this correctly, here `parallel` has to be a dictionary that contains the `"max_workers"` key and maps it to an integer. Wouldn't be simpler in terms of usage if `parallel` was just an integer and perhaps was renamed to `max_workers` to be more explicit? Unless this dictionary also controls something else that I am missing.

Either way, I guess that once we add documentation it will be clear what `parallel` is from the user perspective.

src/qibo/models.py
```diff
@@ -138,6 +139,29 @@ def _DistributedQFT(nqubits: int,
...
+
+    def setup(self, circuit, hamiltonian, parallel):
+        self.hamiltonian = hamiltonian
+        max_workers = self.get_max_workers(parallel)
+        self.circuits = [circuit.copy(deep=True) for _ in range(max_workers)]
+
+    def loss(self, params, *args):
+        import multiprocessing as mp
+        thread_id = int(mp.current_process().name[-1])-1
+        circuit = self.circuits[thread_id]
```
stavros11, line 159: We could also use a dictionary for `self.circuits` to avoid this `int` conversion of `thread_id`, however I don't think it is much of a difference.

src/qibo/config.py
```diff
@@ -37,9 +37,21 @@ def raise_error(exception, message=None, args=None):
...
+    # Set the intra number of threads to the environment flag
+    if "QIBO_NUM_THREADS" in os.environ:
+        nthreads = int(os.environ["QIBO_NUM_THREADS"])
+        tf.config.threading.set_intra_op_parallelism_threads(nthreads)
+
+
+    THREADS = {
+        'inter': tf.config.threading.get_inter_op_parallelism_threads(),
+        'intra': tf.config.threading.get_intra_op_parallelism_threads()
+    }
```
stavros11, line 53: If you think that this functionality would be useful in other applications apart from `optimparallel` (which is not very interesting case sinc the NUM_THREADS value is constrained to be 1), we could possibly add `set_threads`/`get_threads` methods similar to the `set_device`/`get_device` functionality.

stavros11 (commented):
Thanks for implementing this. I left some comments below, but these are mostly regarding the usage/API and not really the code structure or performance.

> @stavros11 if the code in this PR is too obscure, we can possibly implement a parallel wrapper for scipy.minimize which deals with circuits using multiprocessing.

Personally, I did not find this obscure, definitely not as much as I expected after our discussion. It was not very hard to follow what is happening even though I do not have any prior experience with similar implementations and I guess this is a good sign. I also did some quick performance checks using the `examples/benchmarks/vqe.py` script and indeed this provides a significant speed-up compared to pure scipy (by the way, when we finalize this PR we should remember to support it in the benchmark script by adding a `parallel` flag there). 

In summary, since Qibo focuses mostly on performance, personally I am fine with including this as it is. Now if you think that implementing our own "parallelizer" is cleaner and can achieve similar performance then it sounds like a better solution, in the sense that we avoid the `optimparallel` dependence and usually the less external dependencies the better.

src/qibo/config.py
scarrazza: Unfortunately we can't change the inter/intra after allocating tf objects, this should be the first thing to do after importing tf. However we could add to the custom tf operators an argument with the number of allowed cores for the thread pooling objects (but this will only for our operators). 

This is connected to your previous comment about changing automatically the number of threads to 1.

scarrazza:
@stavros11 this PR is ready for review.

src/qibo/optimizers.py
```diff
@@ -109,4 +124,144 @@ def opt_step():
...
+        """Creates singleton instance."""
+        if cls._instance is None:
+            cls._instance = super(ParallelBFGSResources, cls).__new__(cls, *args, **kwargs)
+        return cls._instance
+
+    def loss(self, params):
+        """Computes loss (custom_loss) for a specific set of parameters.
+        This class performs the lock mechanism to duplicate objects
+        for each process.
+        """
```
stavros11, line 158: If I understand correctly, this class is fairly general in the sense that it can be used to parallelize the calculation of any function assigned as the `custom_loss` variable, right? Perhaps it can be generalized slightly more by making the `params` argument optional.

The reason is that this could be useful in repeated circuit evaluations, such as these we implemented for noise. In this case we just need to execute the same circuit (without changing any parameters) several times and collect the results. So in this case we would have something like `custom_loss = lambda c: c()`. We would also need to collect the results from all processes somehow when calling using `self.pool`.

stavros11 (approved):
Thanks for implementing this, it looks good to me. Comments are for some typos only. The only problem I see is regarding testing, but we already discussed the difficulties associated with this. In some small VQE examples I ran optimization seems to work as expected (simlar to scipy).

Also performance at least in my local computer is pretty good when comparing with scipy's L-BFGS-B. I will try to do some more extensive benchmarks tomorrow. I am particularly curious how time scales with `processes` in a machine with many threads (eg. whether it saturates or if it peaks around the half number of threads).

scarrazza:
@stavros11, I have fixed the points you have raised and the documentation.
Please let me know if agree, so we can merge.

src/qibo/optimizers.py
```diff
@@ -36,32 +38,56 @@ def cma(loss, initial_parameters, options=None):
...
         options (dict): Dictionary with options accepted by
             ``scipy.optimize.minimize``.
+        processes (int): number of processes when using the parallel BFGS method.
+        args (tuple): optional arguments for the loss function.
     """
-    from scipy.optimize import minimize
-    m = minimize(loss, initial_parameters, method=method, options=options)
+    if method == 'parallel_L-BFGS-B':
+        from qibo.config import raise_error, get_device, get_threads, log
+        if processes is None and get_threads() != 1: # pragma: no cover
```
stavros11, line 77: I would change this check to

```suggestion
        if ((processes is not None and processes * get_threads() > psutil.cpu_count())
            or processes is None and get_threads() != 1): # pragma: no cover
```
to also capture cases where both `processes` and threads are set by the user but their product is higher than the available number of logical threads. For example someone may do `qibo.set_threads(8)` and call the optimizer with `processes=10` which will most likely lead to poor performance if the machine has less than 80 threads, but it will not be captured by the check as it is now.

Also, although not very important, I would move this check after the GPU check/error. Now if someone uses this with a GPU it will most likely raise both the warning and the GPU RuntimeError, however the warning will not be relevant in this case.

stavros11 (commented):
Thank you for the update. The documentation note looks good. I have a small comment regarding the warning check, let me know if it makes sense.

src/qibo/optimizers.py
scarrazza: Thanks, that's a quite good remark.

PR #282, opened 2020-11-16
Fix issues with Collapse gate

stavros11:
Fixes some Collapse gate issues that I found while playing with the Shor example. Particularly the two following issues:

* Constructing the collapse gate fails when `result` is a `numpy` integer (eg. `np.int32`, `np.int64`) instead of pure Python `int`. I fixed this using the `NUMERIC_TYPES` we had in the Hamiltonians and added a relevant test.
* The C++ norm calculation is not working on a particular configuration of number of qubits and number of threads. For example using the collapse on 10 qubits on >35 threads is failing. This is because of a missing `+ 1` in the `nnorms` definition for the CPU custom operator. Things like that are a bit harder to capture in tests because most common machines and presumably the ones used by CI do not have that many threads.

### DIFF
--- a/src/qibo/base/gates.py
+++ b/src/qibo/base/gates.py
@@ -552,7 +552,7 @@ class Collapse(Gate):
 
     @result.setter
     def result(self, res):
-        res = self._result_to_list(res)
+        res = self._result_to_list(res) # pylint: disable=E1111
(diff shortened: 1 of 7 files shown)


scarrazza (approved):
Thanks, loos good!

PR #285, opened 2020-11-22
adding scipy requirement

scarrazza:
Scipy is not a requirement of tf>=2.2 so we have to include it to the requirements list.
This issue is already visible in Qibo 0.1.1, however we didn't catch it because sphinx requires scipy.


stavros11 (approved):
Good catch, thanks for fixing this!

PR #286, opened 2020-11-23
Revisiting gate inheritance

stavros11:
This updates how the gate classes inherit the corresponding base gates and the abstract backend gate. As we discussed, currently each backend gate inherits both the corresponding base gate of the same type and the abstract backend gate. For example `tensorflow.cgates.H` inherits both `base.gates.H` (which inherits `base.gates.Gate`) and `tensorflow.cgates.TensorflowGate` (which also inherits `base.gates.Gate`) and it is a bit complicated to keep track of which methods are defined in the general base and which methods are specific to the tensorflow backend. This PR attempts to fix this with the addition of the `base.abstract_gates` file which contains all the different abstract gates, particularly:

* `Gate`: The main Qibo gate defining the basic functionality such as holding and setting target/control qubits, gate decomposition. All base gates should inherit this. This gate does **NOT** have the functionality to be applied to states.
* `ParametrizedGate`: Extends `Gate` for gates with variational parameters by implementing a parameter getter/setter functionality.
* `SpecialGate`: Extends `Gate` for some special Qibo gates which do not correspond to real quantum operations. Currently these are only the `CallbackGate` and `Flatten`. This replaces the `self.is_special_gate` we currently have.
* `BackendGate`: Abstract class that defines the methods required to apply gates to states.

The idea now is that in order to define a new backend one has to just inherit the `BackendGate` and implement all of its abstract methods. 

This PR also fixes some other smaller things, such as simplifying method/variable names and using the `@abstractmethod` decorator when appropriate. I believe it would be useful to do a similar cleaning pass to the rest of Qibo functionality (circuits, callbacks and hamiltonian).


scarrazza:
@stavros11 could please fix the conflict?

stavros11:
> @stavros11 could please fix the conflict?

Thanks for the reminder, it should be okay now.

scarrazza (approved):
LGTM, I have performed checked on GPU and they are passing.

PR #288, opened 2020-11-28
qPDF code and tutorial

scarrazza:
This PR implements the final version of the `qPDF` model, documentation and tutorial.


scarrazza:
@AdrianPerezSalinas @stavros11 this PR collects all the pieces for the qPDF paper.
However I am not sure if we should keep all the pretrained models here, it is also true that the file sizes are not big.

AdrianPerezSalinas:
Hi @scarrazza @stavros11 , 

with respect to the pre-trained models, maybe we could keep just the `*results.pkl` files, whose size is very small and may save a lot of time. If we decide to do so, I have to take care of introducing the last results in the folder.

scarrazza:
@AdrianPerezSalinas yes, please remove/update the pickle files.

AdrianPerezSalinas:
Done!

scarrazza:
@AdrianPerezSalinas looks good to me.
@stavros11 could you please have a quick look at this? 

The paper will be out tomorrow, so would be great to have this PR merged asap (and eventually fix potential issues later).

src/qibo/hep/qpdf.py
```diff
@@ -0,0 +1,254 @@
...
+            raise_error(RuntimeError, "Layers must be positive and integer.")
+        if not isinstance(nqubits, int) or nqubits < 1: # pragma: no cover
+            raise_error(RuntimeError, "Number of qubits must be positive and integer.")
+        if not isinstance(multi_output, bool): # pragma: no cover
+            raise_error(TypeError, "multi-output must be a boolean.")
+
+        # load ansatz
+        try:
+            self.circuit, self.rotation, self.nparams = globals(
+            )[f"ansatz_{ansatz}"](layers, nqubits)
```
stavros11, line 29: I am not sure if there is a reason that the ansatz functions (`ansatz_Fourier` and `ansatz_Weighed`) are defined as globals (perhaps to be used outside the `qPDF` model?), however if they are moved within the class it would be possible to access them using `getattr(self, f"ansatz_{ansatz}")` which is usually safer.

src/qibo/tests/test_hep.py
```diff
@@ -0,0 +1,22 @@
...
+def test_qpdf(ansatz, layers, nqubits, multi_output):
+    """Performs a qPDF circuit minimization test."""
+    model = qPDF(ansatz, layers, nqubits, multi_output)
+    params = np.random.rand(model.nparams)
+    result = model.predict(params, [0.1])
+    if multi_output:
+        assert result.shape[1] == nqubits
+        assert result.all() > 0
+    else:
+        assert result > 0
```
stavros11, line 22: Not necessary to do this now, however at some point it may be useful to add a test that checks that the result is the expected one. If I understand correctly this just checks positivity but not if the exact numbers are correct.

doc/source/hep.rst
```diff
@@ -0,0 +1,28 @@
...
+Quantum PDFs
+------------
+
+Qibo provides a variational circuit model for parton distribution functions,
+named qPDF. This model is based on :class:`qibo.models.Circuit` and provides a
+simple API to evaluate PDF flavours at specific values of the momentum fraction
+x. Further details and references about this model are presented in the
+``examples/qPDF`` tutorial.
+
+.. _generalpurpose:
```
stavros11, line 21: I think this is leftover from a different file.

doc/source/tutorials/qPDF/qPDF.ipynb
```diff
@@ -0,0 +1 @@
+../../../../examples/qPDF/qPDF.ipynb
```
stavros11, line 1: I noticed that some equations in the notebook are images and some are latex equations. It would be great to update this at some point so that everything is the same. I am not sure what appears better in the sphinx docs.

stavros11 (commented):
I had a quick first pass and it looks good to merge for now. The notebook works for me.

I left some comments below but we can consider these later. Perhaps the only thing I would fix would be to remove the `.. _generalpurpose:` typo from `hep.rst` because it may create issues with the documentation references (it is a reference to a different section).

Note that when I try to generate the docs locally I get a
```
PandocMissing in tutorials/qPDF/qPDF.ipynb:
Pandoc wasn't found.
```
error. I am not sure if that is expected.

scarrazza:
@stavros11 thanks you very much, I have implemented all points you have raised.

PR #290, opened 2020-12-02
testing wheels

scarrazza:
This PR introduces a check for generated whl files. This also fixes the libomp dependency, now the user does not need to install libomp with homebrew before using Qibo.

There is an issue that has to be solved for linux https://github.com/pypa/manylinux/issues/836

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -41,13 +41,24 @@ jobs:
     - name: Build macos wheels
       if: startsWith(matrix.os, 'macos')
       run: |
+        python -m pip install --upgrade pip
         pip install -r requirements.txt
         python setup.py bdist_wheel
+        pip install delocate
+        cd dist
+        delocate-wheel -v `ls`
     - name: Store wheels as artifacts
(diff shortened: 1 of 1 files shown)


.github/workflows/wheels.yml
```diff
@@ -41,13 +41,24 @@ jobs:
     - name: Build macos wheels
       if: startsWith(matrix.os, 'macos')
       run: |
+        python -m pip install --upgrade pip
         pip install -r requirements.txt
         python setup.py bdist_wheel
+        pip install delocate
+        cd dist
+        delocate-wheel -v `ls`
```
stavros11, line 49: Probably not relevant for the GitHub actions, but when I try to run this in my local machine it fails with the following message:
```
FileNotFoundError: [Errno 2] No such file or directory: 'otool'
```

stavros11 (approved):
Looks good to me, thanks.

.github/workflows/wheels.yml
scarrazza: Yes, this is a macos specific tool, on linux we have `ldd` which does the same job.

PR #293, opened 2020-12-09
[WIP] Unit test all functions

### LINKED ISSUE #284: Improve tests
When we implemented some of the latest new features, such as the `TrotterHamiltonian` from symbols or the circuit depth calculation, there were several bugs in the code that were not captured from our tests but were indentified by some application examples. This is not how it should normally happen and the tests should discover as many issues as possible.

With that in mind, I would like to review our tests and modify them so that they follow the "unit testing" idea more closely. That is have tests for each small function that exists in the code. Some features such as the circuit fusion or how the gates are "compiled" for distributed execution are currently only tested globally and not function-by-function.

Along with the test review, I think it would also make the code better if we had look at #148. These issues may be considered lower priority compared to adding features needed for applications but we should keep them in mind.

stavros11:
Fixes #284. The main idea is to write a test for every function that exists in the core qibo code and try to maintain a one-to-one correspondence between tests files and qibo modules. This will ensure 100% coverage, avoid bugs as much as possible and will also make testing different backends simpler. For example, currently all our tests are using the Tensorflow circuits/gates while in principle we should be able to test at least the base functionality without requiring Tensorflow.

My plan is to create a `test_` file for each of the following modules:

- [x] `abstractions.abstract_gates` & `base.gates`
- [x] `abstractions.circuit`
- [x] `abstractions.callbacks`
- [ ] `abstractions.hamiltonians`
- [x] `backends.__init__`
- [x] `backends.numpy`
- [x] `backends.tensorflow`
- [x] `backends.matrices`
- [ ] `core.cgates` & `core.gates`
- [ ] `core.circuit`
- [x] `core.measurements`
- [ ] `core.callbacks`
- [ ] `core.fusion`
- [ ] `core.hamiltonians`
- [ ] `core.einsum`
- [ ] `tensorflow.distcircuit`
- [ ] `tensorflow.distutils`
- [ ] `hep.qpdf`
- [ ] `config`
- [ ] `evolution`
- [ ] `hamiltonians`
- [ ] `models`
- [ ] `optimizers`
- [ ] `parallel`
- [ ] `solvers`

Our current tests will be moved to the appropriate files. We may also keep some general tests such as `test_cirq.py` seperately.


stavros11:
Just noting that I would postpone this refactoring until we have a clearer picture on how the backend mechanism will work (issue to be opened). The reason is that I would like to have the possibility to run part of the tests on environments without all Qibo requirements. 

For example we would have `test_base_*.py` and `test_tensorflow_*.py` tests that test the corresponding `base` and `tensorflow` folders. We should be able to do `pytest test_base_*.py` in environment without Tensorflow. This will make sure that the everything in Qibo base is really backend independent (which is probably not the case right now).

PR #294, opened 2020-12-10
Trainable flag for parametrized gates

stavros11:
This PR adds the `trainable` flag in all parametrized gates which allows the user to hide them from `circuit.get_parameters()` and `circuit.set_parameters()`. For example
```Python
c = Circuit(3)
c.add(gates.RX(0, theta=0))
c.add(gates.RY(1, theta=0.5, trainable=False))
c.add(gates.RZ(2, theta=0.8))

print(c.get_parameters()) 
# will print [0, 0.8] ignoring the RY

c.set_parameters([0.2, 0.4])
# will set RX(0.2) and RZ(0.4) and ignore RY
```

@scarrazza, @igres26 I used the name `trainable` for this flag following Tensorflow/Keras but I am open to alternative suggestions.

### DIFF
--- a/doc/source/advancedexamples.rst
+++ b/doc/source/advancedexamples.rst
@@ -318,10 +318,31 @@ The following gates support parameter setting:
     # rotations and five additional RX rotations are added afterwards.
 
 Note that a ``np.ndarray`` or a ``tf.Tensor`` may also be used in the place of
-a flat list.
+a flat list. Using :meth:`qibo.base.circuit.BaseCircuit.set_parameters` is more
+efficient than recreating a new circuit with new parameter values. The inverse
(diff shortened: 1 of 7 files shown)


scarrazza (approved):
Looks good and follows our discussion. I have just one comment about `get_parameters`, in some context would be useful to have something like `get_parameters(include_not_trainable=True)` to retrieve all parameters for a later circuit reallocation, e.g. when you store the best model parameters and then reload from file.

stavros11:
> Looks good and follows our discussion. I have just one comment about `get_parameters`, in some context would be useful to have something like `get_parameters(include_not_trainable=True)` to retrieve all parameters for a later circuit reallocation, e.g. when you store the best model parameters and then reload from file.

Thanks for the review. I added this option to `get_parameters`.

PR #296, opened 2020-12-14
Create base callbacks

stavros11:
As part of the cleanup related to the backends this PR creates a `qibo.base.callbacks` module. So far we only had `qibo.tensorflow.callbacks` which was not consistent with the rest of the models. Also several methods from this file were independent of Tensorflow and thus can be moved to base.


scarrazza (approved):
This cleanup looks very good, thanks!

PR #297, opened 2020-12-15
Fix test failure on GPU for parallel lbfgs

scarrazza:
Fixes `test_vqe` crash on GPU when running the parallel l-bfgs-b.


stavros11 (approved):
If we don't want to skip this we can also use `with pytest.raises(RuntimeError)`, but either approach should be fine.

PR #298, opened 2020-12-15
Documenting custom VQC optimizations

scarrazza:
Following our discussion last week, I have tested different approaches to make the VQE more general to accept a generic variational quantum circuit optimization. After some tests, I am not fully convinced that a new object is really needed, so in this PR I document how to write a custom VQC and some minimal code cleanup.


doc/source/advancedexamples.rst
```diff
@@ -445,6 +466,51 @@ be written using :class:`qibo.base.gates.VariationalLayer` as follows:
         circuit.add(gates.CZ(0, nqubits - 1))
     circuit.add((gates.RY(i, theta) for i in range(nqubits)))
 
+.. _vqc-example:
+
+How to write a custom VQC optimization?
```
stavros11, line 471: I may be wrong in this but I was a bit confused about the VQC acronym for a few moments, so I would change this to
```suggestion
How to write a custom variational circuit optimization?
```
because I think the VQC acronym is not as commonly used as the VQE.

In the text we could write explicitly "Variational Quantum Circuit (VQC)" for the first time and just use VQC after that.

doc/source/advancedexamples.rst
```diff
@@ -445,6 +466,51 @@ be written using :class:`qibo.base.gates.VariationalLayer` as follows:
...
+.. code-block:: python
+
+    import numpy as np
+    from qibo import models, gates
+    from qibo.optimizers import optimize
+
+    # custom loss function which returns numpy object
+    def myloss(parameters, circuit, target):
+        circuit.set_parameters(parameters)
+        return np.square(np.sum(circuit()-target))
```
stavros11, line 489: Note that this will return a complex number for the loss because `circuit()` will be complex. A more appropriate loss would be 
```Python
return np.sum(np.square(np.abs(circuit() - target)))
```
Even better we could use the fidelity which is the common measure of distance between quantum states:
```Python
return 1 - np.abs(np.conj(target).dot(circuit()))
```

stavros11 (commented):
Thanks for implementing this. I fully agree with the proposed approach of not modifying the VQE model but rather importing the optimizers module and using the custom loss directly there. It is very clean from the user perspective and it is better to keep the VQE as in most cases it refers to Hamiltonian optimization which is how we implement it.

One comment is that I would add a test for the proposed VQC optimization script. We can use the same code you added in the examples, just add it to pytest and use a regressions `.out` file to check the results, like we do in the VQE and QAOA tests. In terms of structure, we could probably merge `test_vqe.py` and `test_qaoa.py` to a single `test_variational.py` and add the VQC test there. I would probably update this in the unit test refactoring anyway.

PR #301, opened 2020-12-16
Updating wheels with cuda 11

scarrazza:
Here we update the wheels build procedure to include cuda 11. For simplicity I am launching the wheels action for all branches.


stavros11 (approved):
Thanks for the update. Note that I still don't have CUDA 11 on my local machine so I could not test the wheel on GPU, but it works on CPU. By the way I tried installing the python38 wheel on dom that has CUDA 11 but installation fails probably because it has Python 3.9.

scarrazza:
Thanks for checking, I have tested the latest wheel package on dom inside a conda environment with tf2.4.0 and cuda 11 and all tests are passing on GPU.

Issue #300, opened 2020-12-16
State abstraction


scarrazza:
Following our discussion, one of the asymmetries in the code is the state representation. We should provide an abstraction layer for the state and math/algebra operations.

stavros11:
Regarding the backend abstraction we could possibly follow what they do in [google/TensorNetwork](https://github.com/google/TensorNetwork/tree/master/tensornetwork/backends). I may be biased because I contributed in this in the past, however I believe their backend scheme is very clean. In our case we could have the following:
```Python
class AbstractBackend:

    def __init__(self):
        self.backend = None
        self.name = "abstract"

    @abstractmethod
    def matmul(self, x, y):
        """Matrix multiplication of two rank-2 tensors."""
        raise_error(NotImplementedError)

    @abstractmethod
    def sum(self, x, axis=0):
        """Exact diagonalization of matrices."""
        raise_error(NotImplementedError)

    @abstractmethod
    def einsum(self, *args):
        """Einsum of arbitrary rank tensors."""
        raise_error(NotImplementedError)

    @abstractmethod
    def eigh(self, x):
        """Exact diagonalization of matrices."""
        raise_error(NotImplementedError)

    ...


class NumpyBackend(AbstractBackend):

    def __init__(self):
        import numpy as np
        self.backend = np
        self.name = "numpy"

    def matmul(self, x, y):
        return self.backend.matmul(x, y)

    def sum(self, x, axis=0):
       return self.backend.sum(x, axis=axis)

    def einsum(self, *args):
        return self.backend.einsum(*args)

    def eigh(self, x):
        return self.backend.linalg.eigh(x)

    ...


class TensorflowBackend(NumpyBackend):

    def __init__(self):
        import tensorflow as tf
        self.backend = tf
        self.name = "tensorflow"

    def sum(self, x, axis=0):
        return self.backend.reduce_sum(x, axis=axis)

    ...

qibo_numpy = NumpyBackend()
qibo_tensorflow = TensorflowBackend()
```
and then use the backends in other modules as
```Python
import qibo_numpy as K
z = K.sum(K.matmul(x, y))
```

This approach may seem a bit redundant because it redefines many methods (it would seem simpler to just do `import tensorflow as K`), however it has several advantages:
* It solves the issue of methods that do not exist in all backends or have different names in each backend. As demonstrated in the example `np.sum` is equivalent to `tf.reduce_sum`.
[...]

PR #303, opened 2020-12-22
Backend abstraction

stavros11:
This is a toy implementation of what is discussed in #300 regarding the numpy/tensorflow backend abstraction and switcher. @scarrazza I used the idea you proposed for the switcher with a few simplifications. I tested with the following script:
```Python
import numpy as np
from qibo import set_computation_backend, K

x = np.random.random((10, 5))
y = np.random.random((5, 8))

set_computation_backend("numpy")
z = K.einsum("ab,bc->ac", np.copy(x), np.copy(y))
w = K.sum(np.copy(y), axis=1)
print()
print(type(z), z.shape) # should be numpy arrays
print(w)

set_computation_backend("tensorflow")
z = K.einsum("ab,bc->ac", np.copy(x), np.copy(y))
w = K.sum(np.copy(y), axis=1)
print()
print(type(z), z.shape) # should be tensorflow Tensors
print(w)


set_computation_backend("numpy")
z = K.einsum("ab,bc->ac", np.copy(x), np.copy(y))
w = K.sum(np.copy(y), axis=1)
print()
print(type(z), z.shape) # should be numpy arrays
print(w)
```
and the switcher seems to work as expected.

Please have a look and if you agree I can add the rest of methods we need in the backends and then think how to integrate it with the rest of Qibo. Ideally we would like almost the whole Qibo code (perhaps except gates) to be free of `tf` calls and use `K` instead.


scarrazza:
@stavros11 thanks for this. It implements exactly what I had in mind and I hope it simplifies the code.
Moving the implementation to the `backend/interface.py` sounds reasonable, however shall we move the `BaseBackend` to `base`?

stavros11:
> Moving the implementation to the `backend/interface.py` sounds reasonable, however shall we move the `BaseBackend` to `base`?

That's a good suggestion. My idea is to first try and refactor the Qibo code (circuits, models, Hamiltonians, etc.) to use `K` instead of `tf`. Depending on how this works it may be possible to simplify our current structure where we have a different folder for each backend.

Right now I am not sure what is the best approach in terms of module structure. I think this should be built in terms of dependencies. One possibility is to have the base independent of all external libraries and have a different folder for the models that actually do the calculation (similar to what we have now, although currently base depends on numpy and I am not sure if it is a good idea to remove that dependency).

stavros11:
After working a bit on this I believe a flexible structure in terms of the backends is the following:

* backends: 
  - __ init __.py
  - base.py
  - numpy.py
  - tensorflow.py
* base:
  - circuits.py
  - gates.py
  - etc.
* core: 
  - circuits.py
  - gates.py
  - etc.

Base defines the structure of Qibo objects via abstract methods without using any backend functionality. The files included in base do not have any numpy or Tensorflow dependencies but cannot perform any calculation. In most cases it is not possible to instantiate base objects because they contain abstract methods.

Core defines the objects used for simulation by implementing the abstract methods that exist in base. The files included in core still do not have any dependencies on numpy or Tensorflow but instead they use a generic backend object denoted as `K`. The structure of `K` is defined in `backends.base`. This can be either the numpy or the Tensorflow backend (default) but more simulation backends (eg. Jax) can be added easily by adding a new file in the backends folder and implementing the required methods. 

It is possible to switch between different backends on the fly using the `set_backend` method which accepts the following options:
* `custom`, `defaulteinsum`, `matmuleinsum`: The usual tensorflow backends.
* `numpy_defaulteinsum`, `numpy_matmuleinsum`: Pure numpy backends based on the same einsum/matmul implementation as tf but using numpy primitives (NOT tested yet)

It is also possible to switch the default backend from tensorflow to numpy by doing `export QIBO_BACKEND=numpy`. This requires a bit more work but once finalized it shoud allow to run Qibo in an environment without tensorflow installed.

Almost everything described above is already implemented in this PR apart from the following details which are still TODO:
- [ ] Remove all tf dependencies from core to allow running on a pure numpy environment.
- [ ] Rename `tensorflow` folder to `core` since it will be independent of tensorflow.
- [ ] Test the numpy backends (will probably be done as part of #293 instead of this PR).

@scarrazza, if you agree with this clean up I will continue with the points noted above.

scarrazza:
Thanks for this great summary. I like pretty much the structure. For the numpy, I think we can just consider 1 implementation, I don't think there will be a relevant difference on CPU.

stavros11:
I believe this PR is ready in terms of functionality, I just need to refactor the tests but I plan to do this as part of #293. We can either review and merge this separately, or merge both together once #293 is finalized. The main features implemented here are the following:
* Code clean up to follow the scheme summarized in the post above.
* Two new backends: `numpy_defaulteinsum` and `numpy_matmuleinsum` which use numpy instead of tf primitives to perform the simulation. These are accessible via the `set_backend` method. Note that the numpy are currently tested in `test_gates.py` but I plan to test them more extensively once I refactor the tests in #293. Ideally we could have a smart testing mechanism that could test numpy even without tf installed (currently our tests use tf in many places).
* The numpy backends can used in an environment without tensorflow installed. If Qibo is installed in such an environment then it will automatically fall back to the `numpy_defaulteinsum` backend. As I said this functionality is not explicitly tested yet because our tests rely on tf but I checked importing Qibo locally in an environment without tf and simulation works as expected using just `np.einsum`.

> For the numpy, I think we can just consider 1 implementation, I don't think there will be a relevant difference on CPU.

With the current implementation we get both "defaulteinsum" and "matmuleinsum" automatically on numpy, since we are just using the tf implementation with an abstract K (which is either np or tf). We could drop "matmuleinsum" completely but this would remove it both from tf and numpy.

Note that the einsum/matmul implementations may not be the most efficient we can do with numpy. Eg. if numpy's indexing is better than tf we could do a different implementation that could be faster. That we have to try writing some new code from gates.

scarrazza:
@stavros11 thanks for this, I review the code.

scarrazza:
@stavros11 could you please fix the conflicts here?

src/qibo/backends/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+        backend (str): possible options are 'custom' for the gates that use
+            custom tensorflow operator and 'defaulteinsum' or 'matmuleinsum'
+            for the gates that use tensorflow primitives (``tf.einsum`` or
+            ``tf.matmul`` respectively).
+    """
+    if not config.ALLOW_SWITCHERS and backend != K.gates:
+        warnings.warn("Backend should not be changed after allocating gates.",
+                      category=RuntimeWarning)
+
+    gate_backend = backend.split("_")
```
scarrazza, line 40: I am not sure to understand why there is a distinction between say `defaulteinsum` and `numpy_defaulteinsum`, if I understand correctly, it falls back to ['defaulteinsum', 'matmuleinsum', 'custom'], right?

Another question, concerning `custom` with the `numpy` backend, I think the `set_backend` should prevent this operation.

src/qibo/backends/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+    return CONSTRUCTED_BACKENDS.get(name)
+
+numpy_backend = construct_backend("numpy")
+numpy_matrices = numpy_backend.matrices
+
+if BACKEND_NAME == "tensorflow":
+    K = TensorflowBackend()
+else: # pragma: no cover
+    # CI uses tensorflow as default backend
+    K = NumpyBackend()
```
scarrazza, line 25: Could you have a backend switcher function between tf and numpy?
We may rename the `set_backend` function below with `set_computational_engine` or `set_computational_model`.

src/qibo/backends/__init__.py
```diff
@@ -0,0 +1,113 @@
+from qibo import config
+from qibo.config import raise_error, log, warnings, BACKEND_NAME
+from qibo.backends.numpy import NumpyBackend
+from qibo.backends.tensorflow import TensorflowBackend
+
+
+CONSTRUCTED_BACKENDS = {}
+def construct_backend(name):
```
scarrazza, line 8: For code convention consistency, shall we make this function "private"?

src/qibo/backends/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+def set_backend(backend="custom"):
+    """Sets backend used to implement gates.
+
+    Args:
+        backend (str): possible options are 'custom' for the gates that use
+            custom tensorflow operator and 'defaulteinsum' or 'matmuleinsum'
+            for the gates that use tensorflow primitives (``tf.einsum`` or
+            ``tf.matmul`` respectively).
+    """
+    if not config.ALLOW_SWITCHERS and backend != K.gates:
```
scarrazza, line 36: What is the motivation for the `ALLOW_SWITCHERS`?

src/qibo/backends/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+    K = TensorflowBackend()
+else: # pragma: no cover
+    # CI uses tensorflow as default backend
+    K = NumpyBackend()
+
+def set_backend(backend="custom"):
+    """Sets backend used to implement gates.
+
+    Args:
+        backend (str): possible options are 'custom' for the gates that use
```
scarrazza, line 31: Here I would be more explicit with the current supported options for each backend (numpy, tensorflow).

src/qibo/backends/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+                      category=RuntimeWarning)
+
+    gate_backend = backend.split("_")
+    if len(gate_backend) == 1:
+        calc_backend, gate_backend = BACKEND_NAME, gate_backend[0]
+    elif len(gate_backend) == 2:
+        calc_backend, gate_backend = gate_backend
+    else:
+        raise_error(ValueError, "Unknown backend {}.".format(backend))
+    bk = construct_backend(calc_backend)
```
scarrazza, line 47: As you did in `set_precision`, maybe we should keep both backends in sync, for example configurations:
- tensorflow: custom, numpy: defaulteinsum
- tensorflow & numpy: defaulteinsum
- tensorflow & numpy: matmuleinsum

What do you think?

src/qibo/backends/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+            CPU or GPU.
+    """
+    if not config.ALLOW_SWITCHERS and name != K.default_device: # pragma: no cover
+        # no testing is implemented for warnings
+        warnings.warn("Device should not be changed after allocating gates.",
+                      category=RuntimeWarning)
+    K.set_device(name)
+    for bk in CONSTRUCTED_BACKENDS.values():
+        if bk.default_device is not None:
+            bk.set_device(name)
```
scarrazza, line 107: In principle here you have to specialize the `set_device` in NumpyBackend to reject GPUs, but if you do that we break the symmetry. In that case, shall we include at least some warning message about falling back to CPU?

examples/reuploading_classifier/qlassifier.py
```diff
@@ -2,7 +2,6 @@
 from qibo import gates
 import numpy as np
 from datasets import create_dataset, create_target, fig_template, world_map_template
-from qibo.config import matrices
```
scarrazza, line 5: General remark, I think we should reconsider all examples and use K instead of manual numpy imports.

doc/source/qibo.rst
```diff
@@ -376,4 +385,17 @@ The user can switch backends using
...
 
 before creating any circuits or gates. The default backend is ``"custom"`` and
-uses the custom Tensorflow operators.
+uses the custom Tensorflow operators. One can switch to a numpy backend using
+the same approach:
+
+.. code-block::  python
+
+    import qibo
+    qibo.set_backend("numpy_defaulteinsum")
```
scarrazza, line 394: As discussed in `__init__.py`, I think a clear separation between backend and computational engine should be cleaner, and possibly avoid misleading information.

For example, if I follow this instruction, I get this output (which may be very confusing):
```python
import qibo
qibo.set_backend("numpy_defaulteinsum")
print(qibo.get_backend()) 
# prints: defaulteinsum

from qibo.config import BACKEND_NAME
print(BACKEND_NAME)
# prints: tensorflow instead of numpy

from qibo import K
print(K)
# prints: qibo.backends.tensorflow.TensorflowBackend instead of Numpy
```

doc/source/general.rst
```diff
@@ -3,9 +3,12 @@ Code Overview
 
 The Qibo framework in this repository implements a common system to deal with classical hardware and future quantum hardware.
 
-The code abstraction is located in ``qibo/base``. The default simulation engine is implemented using tensorflow and is located in ``qibo/tensorflow``.
+The code abstraction is located in ``qibo/base`` and the core simulation engine is located in ``qibo/core``.
```
scarrazza, line 6: Just to simplification, shall we rename `base` in `abstractions` or `interfaces`?

src/qibo/base/callbacks.py
```diff
@@ -11,7 +11,8 @@ class Callback(ABC):
 
     Results of a callback can be accessed by indexing the corresponding object.
     """
-    from qibo.config import K
+    # FIXME: Remove this because base should be independent of backends
+    from qibo import K
```
scarrazza, line 15: Does this still necessary?

src/qibo/backends/tensorflow.py
```diff
@@ -0,0 +1,150 @@
...
+
+        self.tensor_types = (self.np.ndarray, tf.Tensor, tf.Variable)
+        self.Tensor = tf.Tensor
+        self.random = tf.random
+        self.newaxis = tf.newaxis
+        from tensorflow.python.framework import errors_impl # pylint: disable=E0611
+        self.oom_error = errors_impl.ResourceExhaustedError
+        self.optimization = Optimization()
+
+        from qibo.tensorflow import custom_operators as op
```
scarrazza, line 46: If I understand correctly, `custom/defaulteinsum/matmuleinsum` require custom operators, right? I see that the initial_state operator is always used.

src/qibo/config.py
```diff
@@ -1,22 +1,31 @@
...
 
+# characters used in einsum strings
+EINSUM_CHARS = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+
+# Entanglement entropy eigenvalue cut-off
+# Eigenvalues smaller than this cut-off are ignored in entropy calculation
+EIGVAL_CUTOFF = 1e-14
+
+# Flag for raising warning in ``set_precision`` and ``set_backend``
+ALLOW_SWITCHERS = True
```
scarrazza, line 27: Following my comment above, what is the motivation for this flag?

src/qibo/core/cgates.py
```diff
@@ -113,78 +112,74 @@ def __init__(self, q):
         MatrixGate.__init__(self)
         gates.H.__init__(self, q)
 
-    def construct_unitary(self) -> np.ndarray:
-        return (np.array([[1, 1], [1, -1]], dtype=DTYPES.get('NPTYPECPX'))
-                / np.sqrt(2))
+    def construct_unitary(self):
+        return qnp.cast([[1, 1], [1, -1]]) / qnp.sqrt(2)
```
scarrazza, line 116: Should we move this to the matrices.py?

src/qibo/core/cgates.py
```diff
@@ -336,10 +328,10 @@ def __init__(self, q, theta, trainable=True):
         MatrixGate.__init__(self)
         gates.RY.__init__(self, q, theta, trainable)
 
-    def construct_unitary(self) -> np.ndarray:
+    def construct_unitary(self):
         theta = self.parameters
-        cos, sin = np.cos(theta / 2.0), np.sin(theta / 2.0)
-        return np.array([[cos, -sin], [sin, cos]], dtype=DTYPES.get('NPTYPECPX'))
+        cos, sin = qnp.cos(theta / 2.0), qnp.sin(theta / 2.0)
```
scarrazza, line 333: Ok, that's a good point. So ideally the code is trying to do:
- use math for constants and expected python numerical types
- use numpy (`from qibo import numpy`) for objects which must be or are expected to be numpy objects
- use K in places where objects rely on the typo of the state.

The idea is good, however in places like here, it supposes the parameters will be as numpy arrays or compatible formats (tf which casts to numpy) or python lists. I am fine with these choices, but would be great to have some place where we summarize how to use `K`, `qibo.numpy` and `math`.

src/qibo/hamiltonians.py
```diff
@@ -69,9 +36,9 @@ def XXZ(nqubits, delta=0.5, numpy=False, trotter=False):
             h = XXZ(3) # initialized XXZ model with 3 qubits
     """
     if trotter:
-        hx = np.kron(matrices.X, matrices.X)
-        hy = np.kron(matrices.Y, matrices.Y)
-        hz = np.kron(matrices.Z, matrices.Z)
+        hx = qnp.kron(matrices.X, matrices.X)
+        hy = qnp.kron(matrices.Y, matrices.Y)
+        hz = qnp.kron(matrices.Z, matrices.Z)
```
scarrazza, line 41: For example here, if we implement jax we may prefer to use K.kron in order to use the GPU allocation.

scarrazza: This is just a comment, from a practical perspective we don't need this level of flexibility for these objects.

scarrazza (approved):
@stavros11 thank you very much for this masterpiece!
The implementation looks very good to me, and matches the layout needed by our projects.

You can find some minor comments below, but in general looks pretty good, my only concern is about the usage documentation of `math`, `K`, `qnp`.

This PR with #293 and eventually a new state interface #300 should be sufficient for the time being.

stavros11:
Thank you for the detailed review. I fixed some of the above comments. I also removed `measurements.py` and `fusion.py` from `base`. These objects are not used by any other abstract objects so it is not required to have them in `base` and I believe the code is simplified if we keep them only in `core`. If in the future we add a completely different backend (eg. hardware) that requires to abstract some of this (for example measurements) we can do it accordingly.

I will continue working on the remaining comments and will try to also improve documentation.

src/qibo/core/cgates.py
stavros11: I fixed this by using `matrices` here.

src/qibo/backends/__init__.py
stavros11: I renamed these to `_CONSTRUCTED_BACKENDS` and `_construct_backend`.

src/qibo/backends/__init__.py
stavros11: > I am not sure to understand why there is a distinction between say `defaulteinsum` and `numpy_defaulteinsum`, if I understand correctly, it falls back to ['defaulteinsum', 'matmuleinsum', 'custom'], right?

The difference is that `custom`, `defaulteinsum` and `matmuleinsum` all use Tensorflow, while `numpy_defaulteinsum` and `numpy_matmuleinsum` use only numpy and can be used in an environment without Tensorflow installed.

> Another question, concerning `custom` with the `numpy` backend, I think the `set_backend` should prevent this operation.

In the most recent push, `set_backend` raises an error if one attempts to use the custom backend without tensorflow. However, in principle (I have not tested), it should be possible to use the tensorflow custom operators with the numpy backend. In this case all ops excluding the custom operators will be done in numpy. However there may be performance deterioration (particularly for small qubit numbers) because the custom operators will automatically cast the numpy arrays to `tf.Tensor` when a custom operator is called, so I am not sure if we would like to have this option.

src/qibo/backends/__init__.py
stavros11: It is a flag that "remembers" if a gate has been allocated so that if the user uses `set_backend` or `set_precision` after allocating gates, a warning is raised. It is just a warning so it doesn't have any particular effect on the code or results. 
Note that this exists in our current master and I think its behavior is not changed here. I am not sure if there is a better way to handle this warning.

src/qibo/backends/__init__.py
stavros11: To my understanding there are two approaches on how the backend/model can be selected. 

1. What I implemented here, where `set_backend` handles both the backend/engine (numpy or tensorflow) and the model used for gates (custom, defaulteinsum, matmuleinsum). 
2. Have two different switchers, for example `set_backend` for selecting the engine (numpy or tensorflow) and `set_computational_model` (or `set_gate_model`) for selecting the gate model (custom, defaulteinsum or matmuleinsum). 

Personally I am not sure which approach is better. This may be clearer when we implement a third backend which may not support all gate models. An advantage of the first approach was that it was easier to integrate with our current tests which use `set_backend`, however this can be updated as I am revising all tests in #293 anyway.

Let me know which approach you prefer.

doc/source/qibo.rst
stavros11: I fixed these issues temporarily by completely removing the `BACKEND_NAME` variable as it is not necessarily needed now and defining the `__str__` method for `K`, so that `print(K)` is consistent with the current backend and `get_backend()`. The final implementation of this depends on which approach we will use for the backend switcher.

scarrazza: Thanks.

src/qibo/backends/__init__.py
scarrazza: Thanks for the clarification, we can keep your original implementation (point 1) and then see how it handles different scenarios.

src/qibo/backends/__init__.py
stavros11: I added all currently supported backends in this string.

src/qibo/backends/__init__.py
stavros11: I added a warning when `set_device` is used to set a GPU with the numpy backend that numpy does not support GPUs and that the device change is aborted.

doc/source/general.rst
stavros11: I renamed `base` to `abstractions`, `BaseCircuit` to `AbstractCircuit` and `backends.base.BaseBackend` to `backends.abstract.AbstractBackend`.

src/qibo/base/callbacks.py
stavros11: I removed this and moved some functions that were using it to `core/callbacks.py`.

src/qibo/backends/tensorflow.py
stavros11: Yes, that's right, all Tensorflow backends use the same custom operator for the initial state. This is the case in our current master too. In principle we could write the state initialization using Tensorflow primitives for the defaulteinsum/matmul backend. This way it would be possible to use these backends without compiling the custom operators. I don't think there will be much change in performance since these backends are much slower than custom anyway. If you agree we can implement this, it is relatively simple to do.

The numpy backends already have their own initial state:
```Python
state = np.zeros(2 ** nqubits)
state[0] = 1
```
so that they are completely custom operator free, in order to be usable in an environment without Tensorflow.

stavros11:
I believe I fixed most of the above comments except the following two:

First, regarding the initial state custom operator in the defaulteinsum/matmuleinsum, as I wrote above we could probably do the state initialization using Tensorflow primmitives for these backends so that we keep them completely custom operator free.

Second, regarding the `math`/`qnp`/`K` usage. 

`math` is used only in very simple cases where we really need just a scalar mathematical operation (like log or square root) for which we don't need to import a full linear algebra library. This is mostly done to keep `abstractions` free from any backend (numpy, Tensorflow, etc.). I don't think this would be useful for external users, it is up to them if they prefer to use numpy even for these operations as it is usually more convenient.

Regarding `qnp`, the reason it exists is that in some cases (eg. gate matrix creation or Trotter Hamiltonians) we force using `numpy` for some specific operations even when the user selected backend `K` is Tensorflow. This is mainly because numpy is faster for small matrix operations. An alternative and possibly simpler design that I currently thought is to completely remove `qnp` and just have a "small operation" backend within `K`, which we could call as `K.np`. Then we would replace for example `qnp.kron` with `K.np.kron`. 

The difference between `K.matmul` and `K.np.matmul` would be that the first would be either numpy or tf, while the second will be forced to numpy for all `K`. If we add a different backend (eg. Jax) that is fast for both small and big operations, we could define its `K.np` to be Jax and in this case both `K.matmul` and `K.np.matmul` would be Jax.

src/qibo/backends/__init__.py
scarrazza: Good, thanks.

src/qibo/backends/tensorflow.py
scarrazza: Ok, we can keep for the time being the custom init for all tf backends.

scarrazza (commented):
@stavros11 thanks for the updates.

> First, regarding the initial state custom operator in the defaulteinsum/matmuleinsum, as I wrote above we could probably do the state initialization using Tensorflow primmitives for these backends so that we keep them completely custom operator free.

Lets keep the custom init for all tf backends.

> The difference between K.matmul and K.np.matmul would be that the first would be either numpy or tf, while the second will be forced to numpy for all K. If we add a different backend (eg. Jax) that is fast for both small and big operations, we could define its K.np to be Jax and in this case both K.matmul and K.np.matmul would be Jax.

I tend to prefer this idea, so we reduce the number of objects which may refer to numpy but maybe not, say with jax, or tf.numpy.

Some other comments when testing the code:
- would be great to have a more helpful error message for `set_backend(backend)` if `backend` is not supported. Ideally we could simply print the available backend names.
- if `qibo.K.name == 'numpy'` and I try to `set_device('/device:CPU:0')` (same for GPU), I get "Device /device:CPU:0" does not exist". This error message may be misleading, so we could simply raise a warning message, saying that numpy does not support CPU:0 e GPU:0, or keep the default not implemented error.

stavros11:
> I tend to prefer this idea, so we reduce the number of objects which may refer to numpy but maybe not, say with jax, or tf.numpy.

I replaced `qnp` with `K.qnp` and I removed it from `__init__` so that `from qibo import numpy` is no longer possible to avoid confusing users. Note that `K.qnp` is still the Qibo numpy backend and not the full numpy that we get with `import numpy as np`. The reason I am using the Qibo backend is that it has the proper dtypes coded in it.

> * would be great to have a more helpful error message for `set_backend(backend)` if `backend` is not supported. Ideally we could simply print the available backend names.

I updated this error message and now it prints all the available backend names.

> * if `qibo.K.name == 'numpy'` and I try to `set_device('/device:CPU:0')` (same for GPU), I get "Device /device:CPU:0" does not exist". This error message may be misleading, so we could simply raise a warning message, saying that numpy does not support CPU:0 e GPU:0, or keep the default not implemented error.

I disabled `set_device` completely for the numpy backend. Now if this is used when `K` is the numpy backend it will only show a warning that device placement is not supported by numpy and will not make any changes.

scarrazza:
@stavros11 thank you, everything looks good. Let's wait for the test and then merge.

PR #305, opened 2020-12-27
custom initial_state

scarrazza:
Following the discussion in #304 here I replace the tf.zeros with a full custom operator which builds our desired input state.
Interesting to note that we are faster than tf for small number of qubits while tf (eigen) wins for larger numbers.

Even if this PR solves the issue identified by @stavros11 for nqubits > 15, I have observed issues with the multiprocessing start method and openmp/tf-threadpool if we try to run code in parallel using more than 1 thread. For example, when setting `set_threads(2)` the multiprocessing run gets stuck silently. I have the suspicious that this is related to how the `fork` mode works, limiting the thread creation. 

I prefer to not merge this PR until I understand if the code can run in `spawn` mode. My only concern is about performance, the fork mode is fast on single-thread simulation, thanks to the light initialization method, while spawn will presumably take longer to initialize and thus the benefit of parallel simulation with small number of qubits will disappear.

@stavros11 please let me know your opinion.

### DIFF
--- a/src/qibo/tensorflow/circuit.py
+++ b/src/qibo/tensorflow/circuit.py
@@ -4,7 +4,7 @@ import numpy as np
 import tensorflow as tf
 from tensorflow.python.framework import errors_impl # pylint: disable=no-name-in-module
 from qibo.base import circuit
-from qibo.config import DTYPES, DEVICES, BACKEND, raise_error
+from qibo.config import DTYPES, DEVICES, BACKEND, raise_error, get_threads
 from qibo.tensorflow import measurements
 from qibo.tensorflow import custom_operators as op
 from typing import List, Optional, Tuple, Union
@@ -211,9 +211,7 @@ class TensorflowCircuit(circuit.BaseCircuit):
 
     def _default_initial_state(self) -> tf.Tensor:
         """Creates the |000...0> state for default initialization."""
-        zeros = tf.zeros(self.shapes.get('TF_FLAT'), dtype=DTYPES.get('DTYPECPX'))
-        state = op.initial_state(zeros)
-        return state
+        return op.initial_state(self.shapes.get('TF_FLAT'), DTYPES.get('DTYPECPX'), get_threads())
 
     def get_initial_state(self, state: Optional[InitStateType] = None
                            ) -> tf.Tensor:
--- a/src/qibo/tensorflow/custom_operators/cc/kernels/initial_state.h
+++ b/src/qibo/tensorflow/custom_operators/cc/kernels/initial_state.h
@@ -1,13 +1,15 @@
 #ifndef KERNEL_INITIAL_STATE_H_
 #define KERNEL_INITIAL_STATE_H_
 
+#include "tensorflow/core/framework/op_kernel.h"
+
 namespace tensorflow {
 
 namespace functor {
 
 template <typename Device, typename T>
 struct InitialStateFunctor {
-  void operator()(const Device &d, T *in);
+  void operator()(const Device &d, T *in, int64 shape, int nthreads);
 };
 
 }  // namespace functor
--- a/src/qibo/tensorflow/custom_operators/cc/kernels/initial_state.h
+++ b/src/qibo/tensorflow/custom_operators/cc/kernels/initial_state.h
@@ -9,7 +9,7 @@ namespace functor {
 
 template <typename Device, typename T>
 struct InitialStateFunctor {
-  void operator()(const Device &d, T *in, int64 shape);
+  void operator()(const Device &d, T *in, int64 size);
 };
 
 }  // namespace functor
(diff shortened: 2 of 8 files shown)


stavros11:
> Even if this PR solves the issue identified by @stavros11 for nqubits > 15, I have observed issues with the multiprocessing start method and openmp/tf-threadpool if we try to run code in parallel using more than 1 thread. For example, when setting `set_threads(2)` the multiprocessing run gets stuck silently. I have the suspicious that this is related to how the `fork` mode works, limiting the thread creation.

Thanks for implementing this. I can confirm that the nqubits > 15 is solved by the custom operator and multithreading works with the default initial state. I also tried a few different configurations of number of threads / processes and everything seemed to work. I did not keep track of execution times to see if they make sense but I did not observe any crashes. Is it easy to reproduce the issue with `set_threads(2)`?

Other than that, I believe there is an issue with the CUDA operator, as I get a segfault when I try to run the tests or any other (non multiprocessing script) on GPU.

stavros11:
> > Even if this PR solves the issue identified by @stavros11 for nqubits > 15, I have observed issues with the multiprocessing start method and openmp/tf-threadpool if we try to run code in parallel using more than 1 thread. For example, when setting `set_threads(2)` the multiprocessing run gets stuck silently. I have the suspicious that this is related to how the `fork` mode works, limiting the thread creation.
> 
> Thanks for implementing this. I can confirm that the nqubits > 15 is solved by the custom operator and multithreading works with the default initial state. I also tried a few different configurations of number of threads / processes and everything seemed to work. I did not keep track of execution times to see if they make sense but I did not observe any crashes. Is it easy to reproduce the issue with `set_threads(2)`?
> 
> Other than that, I believe there is an issue with the CUDA operator, as I get a segfault when I try to run the tests or any other (non multiprocessing) script on GPU.

scarrazza:
@stavros11 thanks. Yes, if I take the example in test_parallel and `set_threads(2)` the code gets stuck.

I will have a look at the GPU, the crash there is due to the TensorShape which is accessing the content of the array on GPU. 
We should probably cast the shape as a numpy array on CPU instead of a tf on GPU to avoid this issue.

scarrazza:
Just a quick feedback, I have tried the spawn approach and as I was expecting this fixes the crash when the number of threads is > 1, however the overall performance, even for 1 thread is deteriorated a lot for our problem size, making the parallel implementation useless when compared to sequential code.

There are other mechanisms to run parallel code, based on external libraries, however I don't expect performance improvements if we keep the python as the job manager. 

That being said, one option is to keep the parallel execution to single thread.

scarrazza:
@stavros11 here the code I am using (which hangs) on all my machines:
```python
import numpy as np
import qibo
from qibo.models import QFT, Circuit
from qibo import gates
from qibo.parallel import parallel_execution, parallel_reuploading_execution


def main():
    qibo.set_threads(2)
    nqubits = 16
    nlayers  = 20
    c = Circuit(nqubits)
    for l in range(nlayers):
        c.add((gates.RY(q, theta=0) for q in range(nqubits)))
        c.add((gates.CZ(q, q+1) for q in range(0, nqubits-1, 2)))
        c.add((gates.RY(q, theta=0) for q in range(nqubits)))
        c.add((gates.CZ(q, q+1) for q in range(1, nqubits-2, 2)))
        c.add(gates.CZ(0, nqubits-1))
    c.add((gates.RY(q, theta=0) for q in range(nqubits)))
    size = len(c.get_parameters())
    np.random.seed(0)
    parameters = [ np.random.uniform(0, 2*np.pi, size) for i in range(20) ]
    state = None

    print('First evaluation:')
    r3 = []
    t0 = time.time()
    for params in parameters:
        c.set_parameters(params)
        r3.append(c(state))
    print('Time', time.time()-t0)

    t0 = time.time()
    r4 = parallel_reuploading_execution(c, parameters=parameters, initial_state=state, processes=4)
    print('Time', time.time()-t0)

    np.testing.assert_allclose(r3, r4)


if __name__ == "__main__":
    main()
```

scarrazza:
Btw, the issue comes probably forking the parent process after allocating the openmp threads, i.e. the childs loose control over parent's threads. For example, if you erase the sequential evaluation from my code above and keep the parallel call the system will just run fine.

stavros11:
> Btw, the issue comes probably from forking the parent process after allocating the openmp threads, i.e. the child loose control over parent's threads. For example, if you erase the sequential evaluation from my code above and keep the parallel call the system will just run fine.

Thanks for posting your code. Indeed I also tried having both the sequential and the parallel evaluation in the same script and it hangs for nthreads > 1. In my script I was only doing the parallel one and it worked without problem. If I move the parallel evaluation before the sequential it is also fine.

scarrazza:
Thanks for checking. It looks like an issue affecting libomp and gcc, so if you agree, I think we should at least write down this limitation in the docs, given that I don't believe there is something we can do here.

stavros11:
> Thanks for checking. It looks like an issue affecting libomp and gcc, so if you agree, I think we should at least write down this limitation in the docs, given that I don't believe there is something we can do here.

Yes, writing in the docs should be fine. I don't think it is very important to do something about this in terms of code since the parallel execution methods are mostly useful for small circuits (due to memory limitations) where in most cases using `set_threads(1)` is the optimal choice in terms of performance.

PR #306, opened 2020-12-28
custom initial_state GPU

scarrazza:
I am opening this PR to complement the implementation in #305 for GPU.
Now the code compiles and passes several tests on GPU, however there is something wrong with some of distributed circuit tests.

@stavros11 could you please have a look? Looks like the local state vectors are not recovered from the devices after applying gates, this may be due to the replacement of `.assign` with `=`.

### DIFF
--- a/src/qibo/tensorflow/distutils.py
+++ b/src/qibo/tensorflow/distutils.py
@@ -404,9 +404,10 @@ class DistributedState(DistributedBase):
       """Creates the |000...0> state for default initialization."""
       state = cls(circuit)
       with tf.device(state.device):
-          state.pieces[0] = op.initial_state(nqubits=state.nlocal,
-                                             dtype=DTYPES.get('DTYPECPX'),
-                                             is_matrix=False, omp_num_threads=get_threads())
+          piece = op.initial_state(nqubits=state.nlocal,
+                                   dtype=DTYPES.get('DTYPECPX'),
+                                   is_matrix=False, omp_num_threads=get_threads())
+          state.pieces[0] = tf.Variable(piece, dtype=piece.dtype)
       return state
 
     @classmethod
@@ -441,7 +442,7 @@ class DistributedState(DistributedBase):
                                            self.qubits.transpose_order,
                                            get_threads())
             for i in range(self.ndevices):
-                self.pieces[i] = new_state[i]
+                self.pieces[i].assign(new_state[i])
 
     @property
     def vector(self) -> tf.Tensor:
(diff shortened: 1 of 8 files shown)


src/qibo/tensorflow/distutils.py
```diff
@@ -404,7 +404,9 @@ def default(cls, circuit: "DistributedCircuit"):
       """Creates the |000...0> state for default initialization."""
       state = cls(circuit)
       with tf.device(state.device):
-          state.pieces[0] = op.initial_state(state.pieces[0].shape, state.pieces[0].dtype, get_threads())
+          state.pieces[0] = op.initial_state(nqubits=state.nlocal,
+                                             dtype=DTYPES.get('DTYPECPX'),
+                                             is_matrix=False, omp_num_threads=get_threads())
```
stavros11, line 409: We can leave all the `.assign` calls as they are and change this to:
```suggestion
          piece = op.initial_state(nqubits=state.nlocal,
                                   dtype=DTYPES.get('DTYPECPX'),
                                   is_matrix=False, omp_num_threads=get_threads())
         state.pieces[0] = tf.Variable(piece, dtype=piece.dtype)
```
This way all tests pass and performance is the same as master.

stavros11 (approved):
> @stavros11 could you please have a look? Looks like the local state vectors are not recovered from the devices after applying gates, this may be due to the replacement of `.assign` with `=`.

Thanks for the fix, now it works on GPU. Regarding the `.assign`, is there any important reason it should be changed to `=`? I am not sure why tests fail with `=` but the reason we were using `tf.Variables` and `.assign` was that it was easier to keep track of the device that each part of the state is stored. State is stored as pieces represented by `tf.Variables` and the pieces are transferred to each GPU for the calculation. I guess by changing `state.pieces[i].assign(piece)` to `state.pieces[i] = piece` after each calculation we are not transfering the updated piece back to the CPU but we refer `state.pieces[i]` to the GPU tensor which is not what we want.

I think the simplest solution is to leave all `.assign` calls as they are and just pass the new initial state op inside a `tf.Variable` when initializing `state.pieces[0]`. I checked this and all tests pass while performance is equivalent to master for both single and multi GPU.

src/qibo/tensorflow/distutils.py
scarrazza: Thanks for spotting this issue. 

However, I had to change all `assign` with `=` because, at least on CPU, the code crashes with:
```     
>               state.pieces[i].assign(piece)
E               AttributeError: 'tensorflow.python.framework.ops.EagerTensor' object has no attribute 'assign'
```

Does this happens to you too?

stavros11: If I do the above change `.assign` works both on CPU and GPU. I think changing the initialization as I wrote above using `tf.Variable(op.initial_state(...))` will solve this crash because all `.assign` calls will happen on variables not tensors.

scarrazza: Thanks, let me try to revert all assigns after applying your initialization.

PR #307, opened 2021-01-07
small fixes

scarrazza:
Few fixes to preserve the new code conventions.

### DIFF
--- a/src/qibo/hep/qpdf.py
+++ b/src/qibo/hep/qpdf.py
@@ -66,7 +66,7 @@ class qPDF:
         if len(parameters) != self.nparams: # pragma: no cover
             raise_error(
                 RuntimeError, 'Mismatch between number of parameters and model size.')
-        pdf = K.np.zeros(shape=(len(x), len(self.hamiltonian)))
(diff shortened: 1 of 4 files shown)


src/qibo/hep/qpdf.py
```diff
@@ -68,7 +66,7 @@ def predict(self, parameters, x):
         if len(parameters) != self.nparams: # pragma: no cover
             raise_error(
                 RuntimeError, 'Mismatch between number of parameters and model size.')
-        pdf = np.zeros(shape=(len(x), len(self.hamiltonian)))
+        pdf = K.np.zeros(shape=(len(x), len(self.hamiltonian)))
```
stavros11, line 69: Just a note that if you like to explicitly specify the type of this tensor based on the Qibo precision you can use `K.qnp.zeros(shape, dtype='DTYPE')`, otherwise the default `np.float64` will always be used. I think it is is simpler to keep it as it is with the numpy default type.

stavros11 (approved):
Thanks for the fixes, looks good.

src/qibo/hep/qpdf.py
scarrazza: Thanks for this comment, I think `K.qnp` makes sense here.

PR #308, opened 2021-01-11
Circuit viewer

### LINKED ISSUE #145: Qibo visualization
We mentioned it at the last meeting, but we should consider if we want to implement a Qibo function for different kinds of visualization.

On one hand, there is circuit visualization that could be implemented using qasm, or different latex libraries such as 'qcircuit', which has been already discussed.

Then, there is the issue of visualizing circuit results. The most common way is to plot a histogram with 'matplotlib' from a frequencies dictionary (this dictionary is already available in qibo).

We may want to provide users with a histogram module for their circuit results. If it is easy to use, with few customizable options like titles, and axis customization it will be one of the first things user do after completing a circuit run.

Furthermore, if the default design has something recognisable, such as some specific color or design choice, it can be a way for people to recognize that qibo was being used from just looking at a histogram they might see on some presentation or website.

On the other hand, we can decide that everyone should plot their data in their preferred way, and not provide that for ourselves, which is also an acceptable option.

This may not be something at the top of our priority list, but I think is worth considering.

scarrazza:
Provides a text based circuit viewer, closes #145.

**Example:**
```python
from qibo.models import QFT
c = QFT(5)
print(c.draw())
```
image

@stavros11 @igres26 before polishing the code could you please let me know your opinion?

### DIFF
--- a/src/qibo/abstractions/circuit.py
+++ b/src/qibo/abstractions/circuit.py
@@ -913,7 +913,7 @@ class AbstractCircuit(ABC):
 
         return len(qubits), gate_list
 
-    def draw(self, line_wrap=None) -> str: # pragma: no cover
+    def draw(self, line_wrap=None) -> str:
         """Draw text circuit using unicode symbols.
 
         Args:
@@ -938,7 +938,7 @@ class AbstractCircuit(ABC):
 
         for gate in self.queue:
             if gate.name not in labels:
-                raise_error(NotImplementedError, f"{gamte.name} gate is not supported by `circuit.draw`")
+                raise_error(NotImplementedError, f"{gate.name} gate is not supported by `circuit.draw`")
             gate_name = labels.get(gate.name)
             targets = list(gate.target_qubits)
             controls = list(gate.control_qubits)
--- a/src/qibo/abstractions/circuit.py
+++ b/src/qibo/abstractions/circuit.py
@@ -926,7 +926,7 @@ class AbstractCircuit(ABC):
         labels = {"h": "H", "x": "X", "y": "Y", "z": "Z",
                   "rx": "RX", "ry": "RY", "rz": "RZ",
                   "u1": "U1", "u2": "U2", "u3": "U3",
-                  "cx": "NOT", "swap": "x", "cz": "Z",
+                  "cx": "X", "swap": "x", "cz": "Z",
                   "crx": "RX", "cry": "RY", "crz": "RZ",
                   "cu1": "U1", "cu3": "U3", "ccx": "X",
                   "id": "I", "collapse": "M", "fsim": "f",
@@ -937,6 +937,8 @@ class AbstractCircuit(ABC):
         idx = [0] * self.nqubits
 
         for gate in self.queue:
+            if gate.name not in labels:
+                raise_error(NotImplementedError, f"{gamte.name} gate is not supported by `circuit.draw`")
             gate_name = labels.get(gate.name)
             targets = list(gate.target_qubits)
             controls = list(gate.control_qubits)
@@ -952,7 +954,7 @@ class AbstractCircuit(ABC):
 
             # extend matrix
             for iq in range(self.nqubits):
-                matrix[iq] += [ '' for _ in range(1 + col - len(matrix[iq]))]
+                matrix[iq].extend((1 + col - len(matrix[iq]))* [''])
 
             # fill
             for iq in range(min_qubits_id, max_qubits_id + 1):
@@ -990,19 +992,20 @@ class AbstractCircuit(ABC):
             output = output.splitlines()
             def chunkstring(string, length):
                 return (string[i:length+i] for i in range(0, len(string), length))
-
             for row in range(self.nqubits):
                 chunks = chunkstring(output[row], line_wrap)
                 for i, c in enumerate(chunks):
                     output += ['' for _ in range(self.nqubits)]
+                    suffix = ' ...\n'
                     if i == 0:
                         prefix = ''
                     elif row == 0:
                         prefix = '\n... '
                     else:
                         prefix = '... '
-
-                    output[row + i * self.nqubits ] = prefix + c + '...\n'
+                    if i == len(range(0, len(output[row]), line_wrap))-1:
+                        suffix = '\n'
+                    output[row + i * self.nqubits] = prefix + c + suffix
             output = ''.join(output)
 
         return output.rstrip('\n')
--- a/src/qibo/abstractions/circuit.py
+++ b/src/qibo/abstractions/circuit.py
@@ -913,8 +913,15 @@ class AbstractCircuit(ABC):
 
         return len(qubits), gate_list
 
-    def draw(self) -> str: # pragma: no cover
+    def draw(self, line_wrap=None) -> str: # pragma: no cover
         """Draw text circuit using unicode symbols.
+
+        Args:
+            line_wrap (int): maximum number of characters per line. This option
+                split the circuit text diagram in chunks of line_wrap characters.
+
+        Return:
+            String containing text circuit diagram.
         """
         labels = {"h": "H", "x": "X", "y": "Y", "z": "Z",
                   "rx": "RX", "ry": "RY", "rz": "RZ",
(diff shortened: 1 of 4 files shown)


stavros11 (commented):
Thanks for implementing this, I believe it is simple and provides a good visualization. Some comments regarding appearance:

* Multi-control gates (eg. `gates.X(0).controlled_by(1, 2, 3, 4)`) do not appear properly. Specifically only the first control qubit is considered while the rest are ignored.
* We could consider using capital letters for the gate representation (eg. `X`, `Y`, `Z`, `H`, `I`, `U1`, etc.), simlar to Cirq.
* The controlled gate dot/bullet symbol appears a bit weird in my case:
Screenshot from 2021-01-12 14-03-05
so we may want to check compatibility with different terminal versions.
* Although it may complicate the code slightly, I believe it will be better intuitively if we remove the `c` from the controlled gate names for the visualization. Since we explicitly show the control qubits using the dot symbols, we don't need the `c` in the target qubit symbol.
* Currently the `|` symbol that connects the dot with the gate for controlled gates is not used when the control and target qubits are neighbors (for example see second `cx` in my figure). Again this would complicate the code, but it could look better if we used empty lines between qubits and use the `|` symbol for neighboring qubits too. In this case the `matrix` would have `2 * nqubits - 1` rows where every other row is empty or has `|` symbols.

src/qibo/abstractions/gates.py
```diff
@@ -12,7 +12,7 @@
               "cx": "CNOT", "swap": "SWAP", "cz": "CZ",
               "crx": "CRX", "cry": "CRY", "crz": "CRZ",
               "cu1": "CU1", "cu3": "CU3",
-              "ccx": "TOFFOLI"}
+              "ccx": "TOFFOLI", "id": "identity"}
```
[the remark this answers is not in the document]
scarrazza: I did a quick check and id seems to work, at least using the ibm q editor.

scarrazza:
Thanks for the comments, I agree with all of them, except for the "|" for neighbours, I think the plot is cleaner/more compact if we avoid an extra empty line with these connections, what do you think?

stavros11:
> Thanks for the comments, I agree with all of them, except for the "|" for neighbours, I think the plot is cleaner/more compact if we avoid an extra empty line with these connections, what do you think?

Personally I am fine with either approach, I just proposed the other one because it is what Cirq uses. They also use `@` instead of the bullet which may be universal in terms of different terminal versions. Perhaps @igres26 has some more concrete preferences regarding these choices.

igres26:
Looks good to me. 

Having `|` between neighboring qubits might not be necessary since the main use of this feature is to display the circuit to someone who will be familiar, and not go to a paper. And it will be more compact.

I am not sure if using @ instead of dots might make it more universal, but we can look into that if that would fix the issue on Stavros' device. 

I also agree with Stavros that only `x` instead of `cx` is enough when the controls are visible.

Does it work with all gates from Qibo or it only supports QASM gates?

Overall I think this is useful.

scarrazza:
@stavros11 @igres26 I have implemented the features we have discussed last week, could you please have a look/test the code?

@stavros11 for simplicity, I decided to avoid the moments construction, I believe the current version is much more readable and easy to follow, let me know.

src/qibo/abstractions/circuit.py
```diff
@@ -912,3 +912,97 @@ def read_args(args):
...
+                  "cu1": "U1", "cu3": "U3", "ccx": "X",
+                  "id": "I", "collapse": "M", "fsim": "f",
+                  "generalizedfsim": "gf"}
+
+        # build string representation of gates
+        matrix = [[] for _ in range(self.nqubits)]
+        idx = [0] * self.nqubits
+
+        for gate in self.queue:
+            gate_name = labels.get(gate.name)
```
stavros11, line 940: There are some gates such as the `CallbackGate`, `Flatten` and the channels which are not included in `labels`. In this case using `labels.get` here will return a `None` which will raise some errors later which may be hard for the user to understand. I would suggest to include a check like:
```Python
if gate.name not in labels:
    raise_error(NotImplementedError, f"{gate.name} gate is not supported by `circuit.draw`")
```

src/qibo/abstractions/circuit.py
```diff
@@ -912,3 +912,97 @@ def read_args(args):
...
+            line_wrap (int): maximum number of characters per line. This option
+                split the circuit text diagram in chunks of line_wrap characters.
+
+        Return:
+            String containing text circuit diagram.
+        """
+        labels = {"h": "H", "x": "X", "y": "Y", "z": "Z",
+                  "rx": "RX", "ry": "RY", "rz": "RZ",
+                  "u1": "U1", "u2": "U2", "u3": "U3",
+                  "cx": "NOT", "swap": "x", "cz": "Z",
```
stavros11, line 929:
```suggestion
                  "cx": "X", "swap": "x", "cz": "Z",
```
I believe `X` is more commonly used than `NOT` when plotting quantum circuits, even for the `CNOT` gate (despite calling it `CNOT`). If we use `X` we are also more consistent with how we plot `TOFFOLI` and multi-control `X`.

src/qibo/abstractions/circuit.py
```diff
@@ -912,3 +912,97 @@ def read_args(args):
...
+            qubits.sort()
+            min_qubits_id = qubits[0]
+            max_qubits_id = qubits[-1]
+
+            # identify column
+            col = idx[targets[0]] if not controls and len(targets) == 1 else max(idx)
+
+            # extend matrix
+            for iq in range(self.nqubits):
+                matrix[iq] += [ '' for _ in range(1 + col - len(matrix[iq]))]
```
stavros11, line 955: I believe the following is slightly more efficient because it doesn't recreate the list:
```suggestion
                matrix[iq].extend((1 + col - len(matrix[iq]))* [''])
```

stavros11 (commented):
Thanks for the updates, it looks good to me. No problem about not using the moments, I agree that the code is easy to follow as it is. The only two points I noticed are the following:
* Perhaps we could expand the testing of this a bit, for example the case where `line_wrap` is not `None` is not tested. That being said, I played with various gate cases and combinations as well as the `line_wrap` and everything seems to work as expected.
* When using the `line_wrap`, every chunk gets appended with `...` in the beginning and the end. This also happens for the last chunk, however I believe we do not need `...` at the end for the last chunk.

Some additional comments which are probably less important:
* Although as I said it is easy to follow, it may be easier to test (if we extend tests) if we split this to a few auxiliary functions, instead of a big function as it is now. This point may require additional work and won't change anything in practice so I don't believe it is mandatory for now.
* Depending on the `line_wrap` one selects, some gates that involve more than one character (eg. `NOT`) may get split in different chunks (eg. `N` and `OT`) which looks weird. Personally, I don't think we should do anything about this, just leave it up to the user to play and select a good `line_wrap` for each case. Most gates are single character anyway so this is not a big issue.

scarrazza:
@stavros11, thanks for the comments, please let me know if you agree with the current changes. Concerning the line_wrap, I also think that the user can check what is better for his circuit setup, now, concerning the split in multiple function, I think we can avoid that with the current tests, but let me know your opinion.

stavros11:
Thanks for the update. The tests are okay and I don't think splitting the function is really needed, I just observed that there is still a problem with `...` when using the `line_wrap` when the resulting chunks are more than two. For example
```Python
circuit = QFT(5)
print(circuit.draw(line_wrap=30))
```
gives
```
q0: ─H─U1─U1─U1─U1──────────── ...
q1: ───o──|──|──|──H─U1─U1─U1─ ...
q2: ──────o──|──|────o──|──|── ...
q3: ─────────o──|───────o──|── ...
q4: ────────────o──────────o── ...

... ───────────────x───
... ───────────────|─x─
... H─U1─U1────────|─|─
... ──o──|──H─U1───|─x─
... ─────o────o──H─x───
```
which is okay, but with `line_wrap=20` it gives
```
q0: ─H─U1─U1─U1─U1── ...
q1: ───o──|──|──|──H ...
q2: ──────o──|──|─── ...
q3: ─────────o──|─── ...
q4: ────────────o─── ...

... ────────────────────
... ─U1─U1─U1───────────
... ─o──|──|──H─U1─U1───
... ────o──|────o──|──H─
... ───────o───────o────

... ─────x─── ...
... ─────|─x─ ...
... ─────|─|─ ...
... U1───|─x─ ...
... o──H─x─── ...
```
which is wrong because the second chunk does not have `...` while the last one has. The same happens in any case where the chunks are more than two, I think the `...` are always removed from the second chunk.

scarrazza:
@stavros11 many thanks, this should be fixed now!

stavros11 (approved):
> @stavros11 many thanks, this should be fixed now!

Yes, everything seems to work as expected now.

PR #310, opened 2021-01-18
Tests that run without Tensorflow

stavros11:
Because #293 aims to test the full codebase and will be complicated, I am opening this PR which implements a mini version in order to discuss the mechanism that tests the numpy and Tensorflow backends. 
Specifically this adds the following groups of tests:

1. Abstractions test that require only numpy to execute and have no backend (`K`) depedency:
* `test_abstract_callbacks.py`
* `test_abstract_circuit.py`
* `test_abstract_circuit_qasm.py`
* `test_abstract_gates.py`

2. Core tests that have backend `K` dependency:
* `test_core_gates.py`
* `test_core_measurements.py`

The approach that I use currently is to include the following in all core tests:
```Python
try:
    import tensorflow as tf
    BACKENDS = ["custom", "defaulteinsum", "matmuleinsum", "numpy_defaulteinsum", "numpy_matmuleinsum"]
except ModuleNotFoundError:
    BACKENDS = ["numpy_defaulteinsum", "numpy_matmuleinsum"]
```
Therefore if these tests are executed in environment with Tensorflow installed they will test all backends, while if they are executed in an environment without Tensorflow they will not fail but will test the numpy backends only. @scarrazza, because there are various other ways we can implement this, let me know what you think.

3. Backend tests which require both numpy and tensorflow:
* `test_backends_init.py`: Tests the backend, precision and device switchers and backend instantation.
* `test_backends_matrices.py`: Tests the Qibo hardcoded matrices.
* `test_backends_agreement.py`: Tests that each method of the Tensorflow backend agrees with the corresponding method in the numpy backend.

Once we decide on the mechanism that selects the test backend we can proceed by refactoring all our current tests to follow the above structure (this is essentially what I am doing in #293).


stavros11:
In the last push I made the following updates:
* Moved all new test files to a new directory `new_tests`. The goal is to eventually migrate all tests to this directory and replace our current `tests`. I will probably do this for the basic functionality (circuit, gates) in this PR and leave the more advanced features (Hamiltonians, minimization, etc.) for #293.
* Added tests for the core circuit in addition to the files mentioned in the first post.
* Replaced the mechanism that parametrizes the `backend` fixture in tests to something closer to what we discussed today. Now the parametrization is handled by the [`pytest_generate_tests` function defined in `src/qibo/new_tests/conftest.py`](https://github.com/Quantum-TII/qibo/blob/06c25393f2305104c14e2c9be5bcb479f4206e23/src/qibo/new_tests/conftest.py#L45). 

I believe this parametrization approach is clean since we don't have to do `@pytest.mark.parametrize("backends", [...])` in every test. It also allows to control which backends are tested using flags. For example `pytest src/qibo/new_tests --backends custom,numpy_defaulteinsum` will test only these two backends. By default all available backends are tested. @scarrazza please have a look at `new_tests` and specifically `new_tests/conftest.py` and let me know if you agree with this approach.

scarrazza:
@stavros11 thanks, the `conftest.py` looks good. Are you planning to include more changes in this PR?

stavros11:
> @stavros11 thanks, the `conftest.py` looks good. Are you planning to include more changes in this PR?

Thanks for checking. I would like to complete all the tests that are currently under the `new_tests/` directory, by completely migriting the old `test_gates.py` and `test_circuits.py`. This should provide the possibility to test the basic functionality (circuit simulation) with or without Tensorflow. I would leave the migration of the rest of tests for #293 at a later time. Meanwhile I would also try to look at the state abstraction (#300) as a new feature so that we don't spend all the time in the tests.

I will make these changes and let you know when I think this is good for a second review.

stavros11:
I believe this is in a good state for a final review. The following tests were updated/moved to the `tests_new` directory:

* `test_backends.py` --> `test_backends_agreement.py`
* `test_backpropagation.py` --> `test_core_circuit_backpropagation.py`
* `test_base.py` --> `test_abstract_circuit.py`, `test_backends_init.py` (for testing the backend/precision/device switchers)
* `test_circuits.py` --> `test_abstract_circuit.py`, `test_core_circuit.py`, `test_core_circuit_features.py`
* `test_cirq.py` --> `test_cirq.py`
* `test_custom_operators.py` --> `test_tensorflow_custom_operators.py`
* `test_decompositions.py` --> `test_abstract_circuit.py`, `test_core_circuit_features.py`
* `test_gates.py` --> `test_abstract_gates.py`, `test_core_gates.py`, `test_core_gates_controlled.py`, `test_core_gates_features.py`
* `test_gradients.py` --> `test_tensorflow_custom_operators.py`
* `test_qasm.py` --> `test_abstract_circuit_qasm.py`, `test_core_circuit_qasm.py`

The main new feature is that all tests under `tests_new` can execute without Tensorflow. This can be done either using `pytest src/qibo/tests_new --engines numpy` or executing in environment where Tensorflow is not installed. 

Currently `tests_new` check the basic gate and circuit simulation functionality. The more advanced features such as Hamiltonians, variational models, parallel execution, time evolution, density matrices and noise are still tested using the old tests (under `tests`). Once we migrate all tests, I will rename `tests_new` to just `tests` as we have now, but I guess that it is fine to keep both directories temproarily until the update is completed.

src/qibo/tests_new/test_core_gates.py
```diff
@@ -0,0 +1,519 @@
...
+    np.testing.assert_allclose(final_state, target_state)
+    qibo.set_backend(original_backend)
+
+
+@pytest.mark.parametrize("nqubits,targets,results",
+                         [(2, [1], [0]),
+                          (3, [1], 0),
+                          (4, [1, 3], [0, 1]),
+                          (5, [0, 3, 4], [1, 1, 0]),
+                          (6, [1, 3], np.ones(2, dtype=np.int)),
```
scarrazza, line 112: I think `np.int` is deprecated, so you can use `int` directly.

src/qibo/tests_new/test_core_circuit_qasm.py
```diff
@@ -0,0 +1,173 @@
...
+def test_simple_cirq():
+    c1 = Circuit(2)
+    c1.add(gates.H(0))
+    c1.add(gates.H(1))
+    final_state_c1 = c1()
+
+    c2 = circuit_from_qasm(c1.to_qasm())
+    c2depth = len(cirq.Circuit(c2.all_operations()))
+    assert c1.depth == c2depth
+    final_state_c2 = cirq.Simulator().simulate(c2).final_state
```
scarrazza, line 46: I think we should replace `final_state` with `final_state_vector` in all places.

src/qibo/tests_new/test_cirq.py
```diff
@@ -238,16 +234,13 @@ def test_unitary_matrix_gate_controlled_by(backend, nqubits, ntargets, ndevices)
...
-def test_qft(backend, nqubits, accelerators):
+def test_qft(backend, accelerators, nqubits):
     original_backend = qibo.get_backend()
     qibo.set_backend(backend)
-    initial_state = utils.random_numpy_state(nqubits)
     c = models.QFT(nqubits, accelerators=accelerators)
-    final_state = c(np.copy(initial_state)).numpy()
+    initial_state = random_state(nqubits)
+    final_state = c(np.copy(initial_state))
     cirq_gates = [(cirq.QFT, list(range(nqubits)))]
```
scarrazza, line 244: I think we should replace `cirq.QFT` with `cirq.qft`.

src/qibo/abstractions/abstract_gates.py
```diff
@@ -142,6 +149,24 @@ def nqubits(self, n: int):
...
 
+    @property
+    def density_matrix(self) -> bool:
+        """Controls if the gate acts on state vectors or density matrices."""
+        return self._density_matrix
+
+    @density_matrix.setter
+    def density_matrix(self, x: bool):
+        """Density matrix flag switcher."""
+        if self._nqubits is not None:
```
scarrazza, line 160: Why did you not use `is_prepared` here?

src/qibo/abstractions/callbacks.py
```diff
@@ -97,7 +77,7 @@ def nqubits(self, n: int):
         if self.partition is None: # pragma: no cover
             self.partition = list(range(n // 2 + n % 2))
 
-        if len(self.partition) < n // 2:
+        if len(self.partition) <= n // 2:
```
scarrazza, line 80: Does this affects some of our results?

src/qibo/abstractions/circuit.py
```diff
@@ -553,6 +545,8 @@ def set_parameters(self, parameters: Union[Dict, List]):
         """
         if isinstance(parameters, (list, tuple)):
             self._set_parameters_list(parameters, len(parameters))
+        elif isinstance(parameters, self.param_tensor_types):
```
scarrazza, line 548: So this opens the possibility to use tensors as parameters, shall we update the docs?

src/qibo/abstractions/gates.py
```diff
@@ -46,6 +46,7 @@ def __init__(self, q):
 
     def controlled_by(self, *q):
         """Fall back to CNOT and Toffoli if there is one or two controls."""
+        self._controlled_by_error()
```
scarrazza, line 49: Optiona: shall we convert this to a decorator?

src/qibo/backends/__init__.py
```diff
@@ -28,6 +31,9 @@ def _construct_backend(name):
     else: # pragma: no cover
         # CI uses tensorflow as default backend
         K = NumpyBackend()
+        AVAILABLE_BACKENDS = [b for b in AVAILABLE_BACKENDS
+                              if "tensorflow" not in b]
```
scarrazza, line 35: Does we really need the `if tensorflow` condition?

src/qibo/backends/__init__.py
```diff
@@ -28,6 +31,9 @@ def _construct_backend(name):
...
+        AVAILABLE_BACKENDS.remove("custom")
```
scarrazza, line 36: Also here, if I understand correctly, you are removing tf if the code is initialized with a non tensorflow backend, but this does limit the possibility to switch back to tensorflow in a later stage, right?

src/qibo/backends/__init__.py
```diff
@@ -36,18 +42,14 @@ def _construct_backend(name):
     except ModuleNotFoundError: # pragma: no cover
         # case not tested because CI has tf installed
         log.warning("Tensorflow is not installed. Falling back to numpy.")
-        log.warning("Numpy does not support custom operators and GPU.")
         K = NumpyBackend()
+        AVAILABLE_BACKENDS = [b for b in AVAILABLE_BACKENDS
+                              if "tensorflow" not in b]
+        AVAILABLE_BACKENDS.remove("custom")
+
```
scarrazza, line 49: Here instead is ok because you know that tf is not available.

src/qibo/backends/__init__.py
```diff
@@ -96,6 +95,13 @@ def get_backend():
         return "_".join([K.name, K.gates])
 
 
+if _BACKEND_NAME != "tensorflow": # pragma: no cover
+    # CI uses tensorflow as default backend
+    log.warning("Numpy does not support Qibo custom operators and GPU. "
```
scarrazza, line 100: I think we can replace Numpy with the backend name. Ideally we should store the "allowed" support for each backend, so then this message may be more useful.

src/qibo/tests_new/conftest.py
```diff
@@ -0,0 +1,85 @@
...
+    config.addinivalue_line(
+        "markers", "linux: mark test to run only on linux"
+    )
+
+
+def pytest_addoption(parser):
+    parser.addoption("--engines", type=str, default=_ENGINES)
+    parser.addoption("--backends", type=str, default=_BACKENDS)
+    parser.addoption("--accelerators", type=str, default=_ACCELERATORS)
+    parser.addoption("--target-backend", type=str, default="numpy")
```
scarrazza, line 42: I think this is the only flag which is obscure, could you please add some comment about it?

src/qibo/tests_new/test_abstract_gates.py
```diff
@@ -0,0 +1,398 @@
...
+    pairs = [(0, 1)]
+    with pytest.raises(ValueError):
+        gate = gates.VariationalLayer(targets, pairs, gates.RY, gates.CZ, params)
+
+
+def test_variational_layer_errors():
+    import numpy as np
+    pairs = [(i, i + 1) for i in range(0, 5, 2)]
+    with pytest.raises(ValueError):
+        gate = gates.VariationalLayer(range(6), pairs,                                     gates.RY, gates.CZ,
```
scarrazza, line 222: Something funny with indentation here.

src/qibo/tests_new/test_abstract_gates.py
```diff
@@ -0,0 +1,398 @@
...
+                                      np.zeros(6), np.zeros(7))
+    #with pytest.raises(ValueError):
```
scarrazza, line 224: Shall we remove this comment?

src/qibo/tests_new/conftest.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def pytest_generate_tests(metafunc):
+
```
scarrazza, line 46: Please add few lines explaining the logics (here). I understand it, but probably is better to write a couple of lines.

src/qibo/tests_new/test_backends_agreement.py
```diff
@@ -0,0 +1,104 @@
...
+    ("sqrt", [rand(5)]),
+    ("log", [rand(5)]),
+    ("abs", [rand(5)]),
+    ("expm", [rand((4, 4))]),
+    ("trace", [rand((6, 6))]),
+    ("sum", [rand((4, 4))]),
+    ("sum", {"x": rand((4, 4, 3)), "axis": 1}),
+    ("matmul", [rand((4, 6)), rand((6, 5))]),
+    ("outer", [rand((4,)), rand((3,))]),
+    # ("kron", [rand((4, 4)), rand((5, 5))]),
```
scarrazza, line 41: Would be great to keep the `kron` and `inv` always on.
Can we add some `pytest.raise` for `tensorflow`?

scarrazza (commented):
Thank you very much for this masterpiece of work. I believe it is in a very good shape and also clear to follow and update. I have tested this PR on CPU and GPU, and with environments with and without tf, everything seems to work fine. I have placed some comments above.

Concerning our discussion during the last meeting, unfortunately there is no simple way to add a setup.py `extra_dependency` which blacklist tensorflow. So here we have to install with --no-deps, or eventually create another requirements/setup file for a tf-free version.

stavros11:
> Thank you very much for this masterpiece of work. I believe it is in a very good shape and also clear to follow and update. I have tested this PR on CPU and GPU, and with environments with and without tf, everything seems to work fine. I have placed some comments above.

Thank you very much for the review and the comments. I believe I fixed everything noted above.

> Concerning our discussion during the last meeting, unfortunately there is no simple way to add a setup.py `extra_dependency` which blacklist tensorflow. So here we have to install with --no-deps, or eventually create another requirements/setup file for a tf-free version.

I believe that for now we should stick to our default installation with Tensorflow as a requirement, at least for the `pip` releases. Although circuit simulation is tested and should work with numpy alone, we have not done extensive benchmarks for this case, while for Tensorflow we have the custom operators and have spent several months optimizing everything. Numpy is currently based on the einsum/matmul implementations of tf but we are not sure if this is the optimal way to apply gates when using numpy. I would first do some more benchmarks and perhaps try some alternative approaches such as slicing and once we have more insight on the performance we could think about releasing tf-free versions. If someone would like to try the numpy-only version now they can just use `--no-deps`. Perhaps we can mention this in the documentation too.

src/qibo/abstractions/callbacks.py
stavros11: This should not affect results when the entanglement entropy is concerned. This is because of the entropy's property that it doesn't matter which bi-partition we trace out in order to calculate it. The idea of this `if` is to always trace out the larger bi-partition so that we then calculate `log(rho)` (which requires diagonalization) for a smaller matrix.

For example if we have 10 qubits and the bipartition is [0, 1] and [2, 3, ..., 9], we can either trace out [0, 1] and diagonalize a 2^8 x 2^8 matrix or trace out [2, 3, ..., 9] and diagonalize a 4x4 matrix. Clearly the latter is more efficient. The reason I put the equal here is so that we flip to the smaller bi-partition when the number of qubits is odd. For example for 5 qubits [0, 1, 2] should be flipped to [3, 4] so that the diagonalization happens for one qubit less (it is a minor change).

src/qibo/abstractions/abstract_gates.py
stavros11: I updated this to use `is_prepared`.

src/qibo/abstractions/gates.py
stavros11: I changed this to a decorator. This was actually a good example to play with decorators, thanks!

src/qibo/abstractions/circuit.py
stavros11: This is not a new feature, I think it exists in our current master, here I just changed slightly how the type of `parameters` is checked. Currently in the [How to use parametrized gates?](https://qibo.readthedocs.io/en/stable/advancedexamples.html#how-to-use-parametrized-gates) example in the docs we mention that a `np.array` or a `tf.Tensor` can be used in place of the flat list in `circuit.set_parameters` and we have an example with the variational layer. Do you think we should update something else regarding this?

src/qibo/backends/__init__.py
stavros11: This is a good point. I removed these lines so that if one uses the `QIBO_BACKEND` environment variable to set the backend to numpy, then this does not exclude potential future change to tensorflow if that is installed. Note however that in this case `defaulteinsum` and `matmuleinsum` will refer to the corresponding numpy backends, because numpy will be logged as the default backend. In order to switch to tensorflow one should then do `qibo.set_backend("tensorflow_defaulteinsum")` or just `qibo.set_backend("custom")` since custom is not available for numpy anyway.

src/qibo/tests_new/conftest.py
stavros11: I added some explanations both in this function and in the parser arguments above.

src/qibo/abstractions/circuit.py
scarrazza: Right. I think we can update the info for `parameters` in the docstring of this function.

scarrazza (approved):

PR #311, opened 2021-01-18
MaxCut Hamiltonian

scarrazza:
Implements the maxcut hamiltonian with optional random connections.


scarrazza:
In these last commits we convert the numpy implementation to the symbolic approach, however:
- I am using `hasattr` to check if the map contains a number. Ideally I would prefer to use `isinstance(ivalues, K.numeric_values)` however this file never imports K, so prefer to don't break this assumption.
- when `trotter=True`, I get a `Matrix of invalid type <class 'numpy.complex128'> given during Hamiltonian initialization` not sure if this is related to the scalar symbol or something else.

@stavros11 could you please have a look and let me know if you agree with the changes?

stavros11:
Thank you for implementing this. I believe the update to allow scalar symbols is properly implemented. The issue that happens with `TrotterHamiltonian` seems to be related to an existing "bug"/missing feature (I will try to explain below).

> * I am using `hasattr` to check if the map contains a number. Ideally I would prefer to use `isinstance(ivalues, K.numeric_values)` however this file never imports K, so prefer to don't break this assumption.

Actually I believe it would make sense to completely remove the `SymbolicHamiltonian` from `abstractions.hamiltonians` and move the corresponding functions to `core.hamiltonians.SymbolicHamiltonian` where we import K. There are three reasons I am proposing this:
1. Currently `SymbolicHamiltonian` imports `matrices` which internally import a backend so strictly speaking the file is not completely backend free (as the files in `abstractions` should be).
2. `SymbolicHamiltonian` is not really meant to be used by the user, it is just used internally when the `from_symbolic` method is used to construct Hamiltonians. In the end the user will get either a `Hamiltonian` or a `TrotterHamiltonian` object but never a `SymbolicHamiltonian` one. So it is not really required to have this in abstractions.
3. We would save some lines of code as we could define all functions directly on core instead of defining `@abstractmethod` in abstractions and then overriding on core.

If you agree, we can move everything under `SymbolicHamiltonian` to core and use `K.numeric_values` there.

> * when `trotter=True`, I get a `Matrix of invalid type <class 'numpy.complex128'> given during Hamiltonian initialization` not sure if this is related to the scalar symbol or something else.

This is a bit more complicated because it requires changes in two places. The initial error is indeed related to the fact that it attempts to construct a Hamiltonian term from the scalar symbol. A potential fix is to go to the [trotter_terms](https://github.com/Quantum-TII/qibo/blob/e3628a2fdf1284908ce54a0f7f73d51b14bcdd84/src/qibo/abstractions/hamiltonians.py#L301) and do:
```Python
terms = {t: m for t, m in self.partial_matrices()}
if tuple() in terms:
    constant = terms.pop(tuple()) + self.constant
else:
    constant = self.constant
if set(len(t) for t in terms.keys()) == {1, 2}:
    terms = self.merge_one_qubit(terms)
return terms, constant
```
This will remove the scalar term from the `TrotterHamiltonian` terms and will pass it as part of the scalar constant instead.

If you try this a different error will appear which is more like a "bug" in the `TrotterHamiltonian` construction. The fix is to go to [line 255 of `core.hamiltonians`](https://github.com/Quantum-TII/qibo/blob/master/src/qibo/core/hamiltonians.py#L255) and change to:
```Python
matrices = self.terms[pair]
pair = (pair[1], pair[0])
matrix = 0
for i in range(0, len(matrices), 3):
    matrix += matrices[i] * self.multikron(matrices[i + 2: i:-1])
```
This construction is a bit obscure so it may be worth updating in a different PR to make it easier to follow.

For the MaxCut case, after the two fixes proposed above, the TrotterHamiltonian agrees with the corresponding Hamiltonian in terms of the dense matrix. Just please make sure that the matrix you get also agrees with the one you expect (eg. if you do the `kron` products with just `numpy` without Qibo).

scarrazza:
@stavros11 thanks for the comments:

> If you agree, we can move everything under SymbolicHamiltonian to core and use K.numeric_values there.

Yes, sounds reasonable.
I will try the fix for trotter you propose.

> For the MaxCut case, after the two fixes proposed above, the TrotterHamiltonian agrees with the corresponding Hamiltonian in terms of the dense matrix. Just please make sure that the matrix you get also agrees with the one you expect (eg. if you do the kron products with just numpy without Qibo).

Yes, I will include a test for based on this approach.

scarrazza:
@stavros11 this is now ready.

src/qibo/tests/test_hamiltonians.py
```diff
@@ -263,13 +266,13 @@ def test_tfim_model_hamiltonian(model, kwargs, filename, numpy):
...
 @pytest.mark.parametrize("nqubits", [3, 4])
-@pytest.mark.parametrize("model", [TFIM, XXZ, Y])
+@pytest.mark.parametrize("model", [TFIM, XXZ, Y, MaxCut])
 def test_trotter_hamiltonian_to_dense(nqubits, model):
     """Test that Trotter Hamiltonian dense form agrees with normal Hamiltonian."""
     local_ham = model(nqubits, trotter=True)
     target_ham = model(nqubits, numpy=True)
     final_ham = local_ham.dense
-    np.testing.assert_allclose(final_ham.matrix, target_ham.matrix)
+    np.testing.assert_allclose(final_ham.matrix, target_ham.matrix, atol=1e-15)
```
stavros11, line 275: Just a note that I was doing some checks yesterday and this tolerance may still fail if you increase `nqubits` for example to 7 or 10.

stavros11 (approved):
It should be good now, thanks.

PR #312, opened 2021-01-27
Symbolic Hamiltonian fixes

scarrazza:
Following the discussion in #311, this PR:
- moves the `SymbolicHamiltonian` to core
- uses `K.numeric_types` to intercept scalar symbols
- fixes the Trotter from symbolic expression.


scarrazza:
@stavros11 I think the fix for the trotter may not be working well (see test failure), could you please have a look?

src/qibo/core/hamiltonians.py
```diff
@@ -252,14 +399,36 @@ def merge_one_qubit(self, terms):
...
+            constant (float): The overall constant term of the Hamiltonian.
+        """
+        terms = {t: m for t, m in self.partial_matrices()}
+        if tuple() in terms:
+            constant = terms.pop(tuple()) + self.constant
+        else:
+            constant = self.constant
+        if set(len(t) for t in terms.keys()) == {1, 2}:
+            terms = self.merge_one_qubit(terms)
+        return terms, self.constant
```
stavros11, line 430: Thanks for the update, the issue is here:
```suggestion
        return terms, constant
```
Note that even after this fix there are some numerical issues with the Trotter of the MaxCut and we should set for example `atol=1e-12` to make this test pass. This is probably because the construction of the terms within the Trotter introduces a small numerical error and some values that should be 0 become `1e-16`. I don't think it is due to a bug.

scarrazza: Thank you, this works.

PR #314, opened 2021-02-03
State abstraction

stavros11:
Solves #300 by creating a `qibo.abstractions.AbstractState` object and the corresponding core objects: 
* `VectorState`: returned by standard circuits that use state vectors,
* `MatrixState`: returned by circuits that used density matrices,
* `DistributedState`: returned by the `DistributedCircuit`.

`VectorState` and `MatrixState` contain the state vector / density matrix tensor by default. `DistributedState` contains the state pieces and creates the full tensor only if the `state.tensor` is explicitly called for memory efficiency (this functionality already existed).

The main change from the user perspective is that now all circuits return state objects (which inherit the `AbstractState`) regradless of whether measurements are used or no. If measurements are used then the resulting state provides the `.samples` and `.frequencies` so the experience should be the same with what we have now. If no measurements are used and one tries to access `.samples` or `.frequencies` a `RuntimeError` will be raised. The state vector/density matrix is still accessible via `state.tensor`, `state.numpy()` or `np.array(state)` and indexing is also supported (eg. `state[0]` or `state[1:5]`) in all cases (with or without measurements).

Regarding today's discussion on whether measurements should be part of the circuit or no, in principle the functionality implemented in this PR simplifies things because states have a `.measure` method allowing the following manipulation:
```Python
# execute a circuit without measurements
c = Circuit(4)
c.add(...)
final_state = c()

print(final_state.numpy()) # will print the final state vector as np.array
print(final_state.samples()) # will raise RuntimeError because samples are not available

# measure the state independently of the circuit
mgate = gates.M(0, 2) # measurement gate on qubits 0, 2)
final_state.measure(mgate, nshots=100)

print(final_state.numpy()) # will print the same state vector as above (no collapse)
print(final_state.samples()) # will print the samples as a tensor of shape (100, 2)
print(final_state.frequencies()) # will print the corresponding bitstring frequencies
```

One can also pass registers in the `.measure` method:
```Python
mgate = gates.M(0, 1, 2, 3)
registers = {"a": (0, 1), "b": (2, 3)}
final_state.measure(mgate, nshots=100, registers=registers)
```


scarrazza:
@stavros11 thanks for the implementation. I have some practical usage comments:
- in `VectorState` I think looks better to have `.wavefunction()` or `.state(numpy=True)` instead of `.numpy`.
- we could have a `__call__` which prints the `.numpy` by default, or even better, a dict with the 3 output values ['numpy', 'samples', 'frequencies'] so we reduce the number of instructions required by the user to see the result.
- if I do c.add(gates.M(0)), this seems to breaks `final_state.samples()`, I would prefer to maintain both approaches.

stavros11:
> * in `VectorState` I think looks better to have `.wavefunction()` or `.state(numpy=True)` instead of `.numpy`.

I implemented a `.state()` method which returns the state vector/density matrix and updated the docs to the following proposed usage:
```Python
c = Circuit(2)
c.add([gates.H(0), gates.H(1)])
result = c()

print(result.state()) # tf.Tensor([1, 1, 1, 1])
print(result.state(numpy=True)) # np.array([1, 1, 1, 1])
```

> * we could have a `__call__` which prints the `.numpy` by default, or even better, a dict with the 3 output values ['numpy', 'samples', 'frequencies'] so we reduce the number of instructions required by the user to see the result.

I implemented the `__repr__` method to match the tensor representation. From the user perspective it now looks exactly the same with the current master:
```Python
result = c()
print(result) # prints tf.Tensor([...])
```
Similary if one executes just `result` in a notebook cell it will print the `tf.Tensor` as it does now.

I am not sure if the `__call__` method would be useful because it would mean that one has to do `result()` which is not valid with our current API anyway.

> * if I do `c.add(gates.M(0))`, this seems to breaks `final_state.samples()`, I would prefer to maintain both approaches.

The following works for me:
```Python

c = models.Circuit(2)
c.add((gates.H(0), gates.H(1)))
c.add(gates.M(0))

result = c(nshots=100)
print(result) # prints tf.Tensor([1, 1, 1, 1])
print(result.state()) # prints tf.Tensor([1, 1, 1, 1])
print(result.samples()) # prints tensor with samples
```
Note that in order to get measurements (samples and frequencies) you have to pass `nshots` when executing the circuit. If you do `result = c()` it will ignore the measurement gate and the resulting state will not have samples (so `result.samples()` will raise error). 

This is also the behavior of our current master: `c()` without shots returns just a `tf.Tensor` with the state vector and there are no samples.

src/qibo/abstractions/states.py
```diff
@@ -0,0 +1,267 @@
...
+            numpy (bool): If ``True`` the returned tensor will be a numpy array,
+                otherwise it will follow the backend tensor type.
+                Default is ``False``.
+        """
+        if numpy:
+            return self.numpy()
+        return self.tensor
+
+    def __repr__(self):
+        return str(self.tensor)
```
scarrazza, line 94: Ok, after playing with the code I think that we should move that to `.summary()` method, and keep the default `__repr__`, because the user may get confused about the return type of `Circuit.__call__`.

src/qibo/abstractions/states.py
```diff
@@ -0,0 +1,267 @@
...
+                If ``None`` it is calculated automatically from the tensor
+                representation shape.
+        """
+        obj = cls(nqubits)
+        obj.tensor = x
+        return obj
+
+    @classmethod
+    @abstractmethod
+    def zstate(cls, nqubits): # pragma: no cover
```
scarrazza, line 117: Please replace `zstate` with `zero_state` or `vacuum_state` so it is easier to interpret.

scarrazza (commented):
Thank you for this implementation and changes. 
Looks good to me, I have just some minor comments above.

doc/source/advancedexamples.rst
```diff
@@ -160,7 +158,7 @@ however the user may create the full state as follows:
     final_state = c()
 
     # Access the full state (will double memory usage)
-    full_final_state = final_state.vector
+    full_final_state = final_state.state()
```
scarrazza, line 161: Please rename final_state with something different.

scarrazza:
Thanks, look good. I am still not fully convinced about the `.summary` method, let me propose to remove it.

scarrazza (approved):

PR #316, opened 2021-02-06
adding Windows support

scarrazza:
This is a minimalist approach to support windows, in view of lab requirements.
- removed dependencies which are not well supported on windows
- disabled custom operators compilation. From a technical point of view, the compilation of custom operators is possible however may require complex configurations. This PR supports only the numpy backends. Would be great to support other tensorflow backends, however this may require changes at the level of the abstraction layer, in particular for those objects such as initial state allocated through the custom operator library.
- tests for windows are limited to `tests_new`. Would be nice to have tests for a `custom` free approach.
- missing docs for windows users.

@stavros11 please have a look and let me know your opinion.


src/qibo/tests_new/test_tensorflow_custom_operators.py
```diff
@@ -4,7 +4,7 @@
 import pytest
 import numpy as np
 from qibo import K, get_threads
-from qibo.tensorflow import custom_operators as op
```
stavros11, line 7: Another possibility to avoid importing this in every test is to just use `K.op`. `K.op` should provide access to the custom operators if `K` is the Tensorflow backend and raises an error if `K` is numpy, but this file is skipped for numpy anyway.

stavros11 (commented):
Overally, I believe this is good as it is. Regarding these points:

> * removed dependencies which are not well supported on windows

If I am correct only `blessings` is removed which was just used for appearance and was not very useful otherwise, right?

> * disabled custom operators compilation. From a technical point of view, the compilation of custom operators is possible however may require complex configurations. This PR supports only the numpy backends. Would be great to support other tensorflow backends, however this may require changes at the level of the abstraction layer, in particular for those objects such as initial state allocated through the custom operator library.

If I am correct only the initial state custom operator is used by the einsum tensorflow backends. I can think of two, simple to implement, solutions:
1. Replace the `initial_state` backend method for the einsum backends with one that does the initialization using tf native methods. I think the state initialization is much faster than circuit execution so there won't be much of a difference in terms of performance.
2. Do the above replacement but only for Windows and keep as it is for Linux and mac. We just need a check like the `os.name != 'nt'` you already have somewhere in the code, so it shouldn't be very complex.

> * tests for windows are limited to `tests_new`. Would be nice to have tests for a `custom` free approach.

I think that if we fix the previous point then `tests_new` (and eventually all tests once I finish migrating them) will become `custom` free assuming that the custom backend is not used. Only the custom operator tests have direct custom operator dependence but these we can easily skip from `conftest.py` if `K` is not `custom`.

src/qibo/tests_new/test_tensorflow_custom_operators.py
scarrazza: Thanks, fixed.

scarrazza:
@stavros11 the current code here is almost ready, however I would like to remove the special `pytest` call for windows, and use tf with `defaulteinsum`. 

We still have some tests in `tests` which explicitly set the backend to `custom` (e.g. in test_callbacks), I think we can skip those tests using conftest, similarly to the tests_new setup.

PR #317, opened 2021-02-07
Detach custom operators

scarrazza:
This PR:
- improves the mechanism to detect if custom operators are available
- if custom operators are not available but tf is available the defaulteinsum and matmuleinsum will be available.
- fixed tests_new to support `--engines numpy`

@stavros11 please have a look. I am still not able to run the `tests_new` with `--engine tensorflow` (when no custom operators are available) because circuits from some tests are still calling cgates (even if the backend is matmuleinsum and defaulteinsum).

### DIFF
--- a/src/qibo/tests_new/test_abstract_circuit_qasm.py
+++ b/src/qibo/tests_new/test_abstract_circuit_qasm.py
@@ -1,5 +1,6 @@
 """Tests creating abstract Qibo circuits from OpenQASM code."""
 import pytest
+import qibo
 from qibo import __version__
 from qibo.abstractions import gates
 from qibo.tests_new.test_abstract_circuit import Circuit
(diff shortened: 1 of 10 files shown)


scarrazza:
@stavros11 you can check [here](https://github.com/Quantum-TII/qibo/pull/317/checks?check_run_id=1849225752) the issue for the tf defaulteinsum backend, some tests are still importing cgates...

src/qibo/tests_new/test_abstract_circuit_qasm.py
```diff
@@ -346,7 +347,9 @@ def test_from_qasm_invalid_script():
         c = Circuit.from_qasm(target)
 
 
-def test_from_qasm_measurements():
+def test_from_qasm_measurements(backend):
```
stavros11, line 350: I do not mind with testing the different backends here, just noting that my initial idea of `test_abstract_*` was to be completely backend free and just test the basic functionality.

scarrazza: Right, but here (and in the other places) I am forced to pass backend, otherwise when you set `--engines numpy` and you have tf installed, it will use `defaulteinsum` from tensorflow instead of numpy.

PR #318, opened 2021-02-08
Fix custom operator imports in tests

stavros11:
This fixes the issue of importing custom operators when the non-custom tensorflow operators are used on a machine that cannot compile custom operators. Particularly I identified two issues:

1. Custom operators were indeed imported by some gates that  are only defined in `cgates.py` and not in `gates.py` (eg. the measurement gate). This was easy to fix because these gates do not actually use the custom operators.
2. The non-custom `initial_state` was not properly implemented when `is_matrix=True`.

### DIFF
--- a/src/qibo/backends/numpy.py
+++ b/src/qibo/backends/numpy.py
@@ -2,25 +2,6 @@ from qibo.backends import abstract
 from qibo.config import raise_error, log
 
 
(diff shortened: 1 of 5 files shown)


scarrazza (approved):
Perfect, thanks!

PR #319, opened 2021-02-09
Fix old tests for Windows

stavros11:
Fixes the issue with the old `tests` mentioned in #316. This is mostly a temporary fix until we migrate all tests to `tests_new`.

Regarding the new tests, there may be a possibility to move the
```Python
qibo.set_backend(backend)
...
```
mechanism to `conftest.py` and avoid all the code repetition, however I am not yet sure if this could work.


scarrazza:
Thanks, looks good. I would suggest to disable the parallel tests for windows or raise an exception when importing the multiprocessing module in parallel.py.

PR #323, opened 2021-02-11
Make DistributedState custom operator free

### LINKED ISSUE #320: make DistributedState custom operator free
Following the report [here](https://github.com/Quantum-TII/qibo/runs/1873154622?check_suite_focus=true#step:7:178), DistributedState should not rely on `K.op` when the backend is not `custom`.

stavros11:
Fixes #320 by:
* Removing the `K.op` in the `DistributedState` by defining the `transpose_state` as a seperate backend method implemented using tf primitives when `backend != "custom"`.
* Moves the `DistributedState` tests in a different files which is skipped when the custom backend is not available as this state is not useful without the `DistributedCircuit` and this is only available for the custom backend.

Note that this does not implement the `DistributedCircuit` for the einsum backends, but this should be not hard to do if we need it.

Everything seems to work locally but let's see if the CI also passes.


scarrazza:
Perfect thanks!

scarrazza (approved):

PR #325, opened 2021-02-13
Measurement fixes

### LINKED ISSUE #321: measurements for numpy backend
I think the numpy backend does not support measurement shots, we should provide this feature.

### LINKED ISSUE #322: Errors when using a large number of shots
When using a large number of shots for measuring a state, different errors pop out. Running a code similar to this one:

```
nshots=10**10
circuit = models.Circuit(9)
circuit.add(gates.M(0,1,2,3,4,5,6,7,8))
final_state = circuit2(np.copy(initial_state), nshots=nshots).frequencies(binary=True)
```

For nshots below 10^10 I believe it works fine. From nshots=10^10 to 10^11 I face this error:

_ResourceExhaustedError: OOM when allocating tensor with shape[1,1215752192] and type int64 on /job:localhost/replica:0/task:0/device:CPU:0 by allocator cpu [Op:Multinomial]_

For nshots=10^12 or more:

_InvalidArgumentError: num_samples should be nonnegative, got -727379968 [Op:Multinomial]_

stavros11:
Implements the following:

1. Mirgrates `tests/test_measurements.py` to `tests_new/test_measurement_gate.py`, `test_measurement_gate_probabilistic.py` and `test_measurement_gate_registers.py`. Now these tests are also done using the numpy backends (fixes #321?).

I have not made any changes in the code in order to support numpy, however the following script works both when only numpy is installed (no Tensorflow) and when both numpy and Tensorflow are installed:
```Python
import qibo
qibo.set_backend("numpy_defaulteinsum")

circuit = models.Circuit(2)
circuit.add([gates.H(0), gates.H(1)])
circuit.add(gates.M(0, 1))

result = circuit(nshots=1000)
print(result.state()) # prints [0.5, 0.5, 0.5, 0.5]
print(type(result.state())) # prints <class 'numpy.ndarray'>
print(result.samples()) # prints a (1000, 2) binary array with shots
print(type(result.samples())) # prints <class 'numpy.ndarray'>
print(result.frequencies()) # prints Counter({'11': 264, '01': 263, '00': 251, '10': 222})
```
Note that your results may be difference due to randomness. @scarrazza, if you have spotted a different issue with measurements in numpy that still exists, please let me know.

2. Disables the shot evaluation when the measurement gate is called or the circuit is executed, which fixes #322. Now frequencies and/or shots are only calculated when the user asks for it. Now the example:
```Python
import qibo

circuit = models.Circuit(2)
circuit.add([gates.H(0), gates.H(1)])
circuit.add(gates.M(0, 1))

result = circuit(nshots=10 ** 10)
print(result.frequencies())
```
will work and return `Counter({'00': 2500000000, '01': 2500000000, '10': 2500000000, '11': 2500000000})`, while `result.samples()` will fail with OOM as before.

TODO:
- [x] The frequencies calculated without shots are currently exact (no statistical noise), as I mentioned in #322. This can be seen in the above example. It may be good to find an efficient way to add proper noise without calculating samples.
- [x] I believe the `GateResult`, `CircuitResult` in `core.measurements` are a bit obscure in terms of code readability and I would rename them to something else, such as `MeasurementResult` and `MeasurementRegisterResult`.
- [x] Complete `test_core_measurements.py`


stavros11:
> Thanks for this, looks good. Seems to fix the numpy issues I have observed with master. However, for my example (carlos code) on CPU, I observe that the numpy_defaulteinsum backend seems 2x faster than tf custom. Did you observe something similar?

Thanks for checking. I believe there is an issue with performance due to the way the tensor of frequencies [is transformed to a dictionary](https://github.com/Quantum-TII/qibo/blob/bccf4feb8d40bb1de27a303fff490df5bf76b4e7/src/qibo/core/measurements.py#L123), especially with Tensorflow where this line contains exponentially many castings to Python int. The current frequency calculation is more like a placeholder though, I will try to implement something close to what Carlos suggested in #322 which should also include the noise, so I would check again the final performance once we have this version.

scarrazza:
Great, thanks for the checks.

stavros11:
I updated the frequency calculation to include noise. The approach that is followed currently uses the old shot generator but is called using a predefined batch of shots (2^18) and the shots are not cached in memory to avoid OOMs. Here are some benchmarks for 10 qubits on my machine:
```Python
circuit = models.Circuit(10)
circuit.add(gates.M(*range(10)))
start_time = time.time()
freqs = result.frequencies()
```

nshots | tensorflow (sec) | numpy (sec)
-- | -- | --
1e3 | 0.00062 | 0.00052
1e4 | 0.00086 | 0.00059
1e5 | 0.00331 | 0.00308
1e6 | 0.02705 | 0.02877
1e7 | 0.23152 | 0.29086
2e7 | 0.46103 | 0.56954
4e7 | 0.90392 | 1.17014
6e7 | 1.36139 | 1.69035
8e7 | 1.83547 | 2.29594
1e8 | 2.34076 | 2.91795

This approach also gets quite slow for more shots (>1e9), especially for more qubits, however I believe it avoids the issues found in #322 when only frequencies are required. It may be worth trying the cummulative probability vector approach described by Carlos, however I am not yet sure if there is a way to implement this efficiently using numpy/tensorflow primitives, without going to C++.

scarrazza:
Ok, this solution seems acceptable to me.

stavros11:
As we discussed, I implemented the `set_batch_size` functionality and used the batch method also for the sample calculation. This will still be limited by memory as it will try to construct a tensor of shape `(nshots,)`, but at least it should not be limited from the `int32` so it should be possible to go to higher `nshots` if sufficient memory is available. The frequency calculation is not limited by memory.

I will try to do some benchmarks on the DGX perhaps later because now the machine is almost fully used.

bpcarlos:
Ok, thanks @stavros11. @scarrazza should we wait for the benchmarks or do we go ahead with the 10^12 shots and see what happens?

scarrazza:
@stavros, GPU:2 is free now please go ahead with benchmarks first.

stavros11:
> @stavros11, GPU:2 is free now please go ahead with benchmarks first.

Yes, I already noticed this and started doing some benchmarks:

nshots | GPU (sec) | CPU (sec)
-- | -- | -- 
1e8 | 5.594280958175659 | 2.548774480819702
2e8 | 9.247143030166626 | 5.103588104248047
4e8 | 19.837314128875732 | 10.239906311035156
6e8 | 27.93835163116455 | 15.513289213180542
8e8 | 37.154372692108154 | 20.775641202926636
1e9 | 46.273950815200806 | 25.593887329101562
2e9 | 92.27190232276917 | 51.146870136260986
4e9 | 187.6790692806244 | 104.10990810394287

This is for the 10-qubit example. The CPU benchmarks are done in my laptop which seems to be faster than the GPU. I suspect this has to do with the `tf.random.categorical` implementation on GPU but I will confim it by doing some benchmarks of this method alone.

Regarding the time for more shots, by extrapolating the above without any modification 1e10 shots should take around 5 minutes and 1e12 shots 8-9 hours.

scarrazza:
Right, so the circuit size is probably too small or maybe the batch size should be increased? 
There are several CPU processes in the dgx right now, so this may be blocking the performance.

stavros11:
I think the performance issue on the GPU has also to do with the fact that there is a numpy casting during the frequency calculation. Particularly the calculation code for each batch is:
```Python
samples = self.sample_shots(probs, nsamples)
res, cnts = self.unique(samples, return_counts=True)
for k, v in zip(res, cnts):
    frequencies[k] += v
```
Here `self.sample_shots` uses `tf.random.categorical` which is executed on GPU but `self.unique` uses `np.unique` which runs on CPU. So perhaps the bottleneck is CPU-GPU communication. I will check if there is a tf primitive equivalent to `np.unique` that runs on GPU as it will probably make the process faster.

Regardless of that, I tried the following tf-only script that mimicks our sampling code
```Python
nqubits = 10
batch = int(1e6)

probs = tf.random.uniform((2 ** nqubits,))
probs = probs / tf.reduce_sum(probs)
logits = tf.math.log(probs)[tf.newaxis]

start_time = time.time()
for _ in range(nreps):
    samples = tf.random.categorical(logits, batch, dtype=tf.int64)
```

nreps | GPU (sec) | CPU (sec)
-- | -- | --
1 | 0.0003330707550048828 | 0.06560993194580078
10 | 0.0015027523040771484 | 0.6254217624664307
100 | 0.013147115707397461 | 5.948432683944702
200 | 0.058065176010131836 | 12.373946905136108
300 | 2.810572385787964 | 18.484578371047974
400 | 5.730774164199829 | 24.764498233795166
500 | 8.660166263580322 | 31.14205837249756

So here GPU is indeed faster but it still has a strange scaling behavior with the number of reps.

> Right, so the circuit size is probably too small or maybe the batch size should be increased?

I am not sure if the circuit size matters here because I am only measuring the sampling time, not execution. This is a toy example where the circuit has no other gates so execution time is minimal anyway. In the actual application we may consider running the circuit on CPU since it is small and doing only the sampling on GPU. 
The batch size on all of the above is 1e6. I played with various numbers, when the batch is small (< 10^4) performance indeed deteriorates but after 1e5 it stays almost constant. I am not sure if we can do more than 1e6 because it starts throwing OOMs.

stavros11:
In the latest push, I implemented the frequency calculation fully on Tensorflow however I do not observe any performance improvement compared to the table I posted above. At this point I am not sure what is making the GPU slower because the part of the calculation that takes almost all time is the frequency calculation (which supposedly happens on GPU) and not any castings.

scarrazza:
@stavros11 could you please send me the code example for the table numbers?

stavros11:
> @stavros11 could you please send me the code example for the table numbers?

I use the following for `nqubits=10` and `batch=int(1e6)`.

```Python
import argparse
import time
import qibo


def main(nqubits, nshots, batch):
    if batch is not None:
        qibo.set_batch_size(batch)

    data = [{
        "backend": qibo.get_backend(),
        "batch": qibo.get_batch_size(),
        "nqubits": nqubits, "nshots": nshots,
        "samples": samples
    }]

    start_time = time.time()
    result = circuit(nshots=nshots)
    data[-1]["execution_time"] = time.time() - start_time
    print("backend:", data[-1]["backend"])
    print("nqubits:", data[-1]["nqubits"])
    print("nshots:", data[-1]["nshots"])
    print("batch:", data[-1]["batch"])
    print("samples:", data[-1]["samples"])
    print("execution_time:", data[-1]["execution_time"])

    start_time = time.time()
    freqs = result.frequencies()
    data[-1]["measurement_time"] = time.time() - start_time
    print("measurement_time:", data[-1]["measurement_time"])
    print("\n")
```
The table shows the measurement time.

src/qibo/backends/numpy.py
```diff
@@ -193,9 +193,26 @@ def transpose_state(self, pieces, state, nqubits, order): # pragma: no cover
...
+        samples = self.sample_shots(probs, nsamples)
+        res, counts = self.unique(samples, return_counts=True)
+        frequencies[res] += counts
+        return frequencies
+
+    def sample_frequencies(self, probs, nshots):
+        from qibo.config import SHOT_BATCH_SIZE
+        n = int(probs.shape[0])
+        frequencies = self.zeros(n, dtype=self.dtypes('DTYPEINT'))
+        for _ in range(nshots // SHOT_BATCH_SIZE):
```
scarrazza, line 209: Are you sure this for loop is needed? Looks like you are running this nrep times and the calling the categorical nrep times. This is the big difference with your pure tf code, where there is just a single for loop.

scarrazza: Being a little be more specific, the `sample_frequencies` function is called several times in your circuit example, this looks like the main difference when comparison to your pure tf code.

stavros11: I think the loop is needed otherwise the frequencies will not correspond to `nshots` measurements. Particularly the sum of all frequencies should be `nshots` which is the case now but would not be if this loop is removed. Also I think `tf.random.categorical` is called once for every loop repetition because `self.sample_shots` is called with `nshots=SHOTS_BATCH_SIZE` so the inner loop corresponds to just `range(1)` (so there is no really nested loop happening).

On the other hand I agree that there is a discrepancy somewhere in the code that I am missing right now because according to the tf-only example the GPU should be faster than CPU.

scarrazza: Indeed, other observations, the `self.unique` seems to take 20% of the overall time, but the remaining comes from the `sample_shots`. 

I think your pure tf code does not forces tensorflow to copy data back to CPU. Please try to append the results to a `sample` list and then print e.g. `sample[0]` before the timing computation (remember this behaviour when we have been playing with tf.function).

We have to find a mechanism to reduce data transfers.

scarrazza: (maybe making the transfer async)

stavros11: I modified `sample_frequencies` to not use `sample_shots` so that we don't have any issue with the nested for loop. From a tf perspective now the code is equivalent to:

```Python
import time
import tensorflow as tf

def sample_frequencies(probs, nshots, batch=1000000):
    """Samples measurement frequencies given a probability distribution."""
    logits = tf.math.log(probs)[tf.newaxis]
    def update_frequencies(logits, nsamples, frequencies):
        samples = tf.random.categorical(logits, nsamples, dtype=tf.int64)[0]
        res, idx, counts = tf.unique_with_counts(samples, out_idx=tf.int64)
        return tf.tensor_scatter_nd_add(frequencies, res[:, tf.newaxis], counts)

    frequencies = tf.zeros(int(probs.shape[0]), dtype=tf.int64)
    start_time = time.time()
    for _ in range(nshots // batch):
        frequencies = update_frequencies(batch, frequencies)
    frequencies = update_frequencies(nshots % batch, frequencies)
    return frequencies
```

@scarrazza do you have any example that this Qibo-free code runs faster on GPU? For me for probs of size (1024,) (10 qubits) and nshots between 1e8-5e8 it is faster in my notebook CPU than the DGX. However I suspect that the reason for this is that the DGX is full of processes because the times I get there show large variations.

Regarding data transfer, do you think there is any transfer happening in this? I thought that since the whole function uses tf-primitives everything should run on GPU without any transfer.

scarrazza: @stavros11 thanks. All examples, involving the tf.random.categorical are faster on CPU, on both setups Qibo and pure tf, including your latest commit.

I the bottleneck could be at the level of communication, given that for a single batch evaluation the performance on CPU and GPU are similar.

stavros11: > I the bottleneck could be at the level of communication, given that for a single batch evaluation the performance on CPU and GPU are similar.

Ideally for the frequency calculation there should be no communication in the sense that we don't want to transfer the big tensor of samples on the CPU. When the next batch is calculated, the previous batch should be discarded completely (that's why the same variable `samples` is used). I am not sure if Tensorflow does exactly that though.

> @stavros11 thanks. All examples, involving the tf.random.categorical are faster on CPU, on both setups Qibo and pure tf, including your latest commit.

An alternative to categorical would be to follow the method Carlos proposed. A way to do this with tf primitives is the following:
```Python
def update_frequencies(probs, nsamples, frequencies):
    cumprobs = tf.math.cumsum(probs)
    cumprobs = tf.repeat(cumprobs[tf.newaxis], nsamples, axis=0)
    samples = tf.random.uniform((nsamples,), dtype=tf.float64)
    samples = tf.math.greater(samples[:, tf.newaxis], cumprobs)
    samples = tf.reduce_sum(tf.cast(samples, dtype=tf.int64), axis=1)
    res, idx, counts = tf.unique_with_counts(samples, out_idx=tf.int64)
    return tf.tensor_scatter_nd_add(frequencies, res[:, tf.newaxis], counts)
```
The main disadvantage is that `cumprobs` will become a tensor of shape (batch size, 2^nqubits) so this will require to reduce the batch size. This is much slower than categorical on CPU but I am not sure if it could help on GPU.

Alternatively we could do a custom operator that combines the use of sampling + unique for this example.

scarrazza: I am trying the current code with your test code on GPU (12GB) with 20 qubits but I get an OOM with the default batch size. I am just wondering if the performance we see on GPU is due to a "small" circuit. If so, then we should just benchmark the alternative and see what is the most efficient.

stavros11: > I am trying the current code with your test code on GPU (12GB) with 20 qubits but I get an OOM with the default batch size. I am just wondering if the performance we see on GPU is due to a "small" circuit. 

I am not sure if the circuit size matters so much because I am only measuring sampling time (not execution) but indeed the GPU may be faster than CPU for 20 qubits. Is that the case for smaller batch size? Unfortunately I cannot test this on DGX because if there are other processes running the timings do not make much sense.
Regarding the batch size, perhaps we should find a better way to set the default value because the optimal value that does not go OOM depends both on the GPU memory and the circuit size.

> If so, then we should just benchmark the alternative and see what is the most efficient.

The alternative I wrote above is much slower and gets OOM with smaller batch sizes, at least in my notebook. The only alternative that I can think of right now and could help would be doing the following in C++:
```C++
for (int i = 0; i < nsamples; i++) {
    for (long j = 0; j < nstates; j++) {
        if (random() < cumprobs[j]) {
            frequencies[j]++;
            break;
    }
}
```
where `random()` is just uniform in [0, 1]. This does the frequency calculation directly without generating any samples and calling unique, so we also don't need to specify any batch size and won't have OOM issues.

scarrazza:
Perfect, thanks.

PR #328, opened 2021-02-16
Fix einsum in density matrix traceout

### LINKED ISSUE #326: Tensorflow einsum error when measuring density matrices
**Describe the bug**
Measuring density matrices with more than 6 qubits using any Tensorflow backend leads to an einsum error. Thanks @AdrianPerezSalinas for finding this.

**To Reproduce**
```Python

circuit = models.Circuit(7, density_matrix=True)
circuit.add(gates.M(*range(7)))
result = circuit(nshots=100)
```
gives 

```
tensorflow.python.framework.errors_impl.UnimplementedError: Unsupported rank: 7 while handling repeated indices. Up to rank 6 is supported. [Op:Einsum]
```

**Additional context**
Einsum is used here to calculate the probability distribution from the density matrix, [by tracing out unmeasured qubits](https://github.com/Quantum-TII/qibo/blob/0422b84a189ff86dd6a900b76bd76cac90735d72/src/qibo/core/states.py#L149). Numpy backends do not have this issue.

stavros11:
Fixes #326 by modifying the einsum call used to trace-out density matrices to calculate the corresponding probability distribution for measurements.

### DIFF
--- a/src/qibo/core/states.py
+++ b/src/qibo/core/states.py
@@ -54,42 +54,31 @@ class VectorState(AbstractState):
         matrix = K.outer(self.tensor, K.conj(self.tensor))
         return MatrixState.from_tensor(matrix, nqubits=self.nqubits)
 
-    def _traceout(self, qubits=None, measurement_gate=None):
-        """Helper method for :meth:`qibo.core.states.VectorState.probabilities`.
-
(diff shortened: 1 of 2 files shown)


scarrazza (approved):
Looks good, thanks!

PR #329, opened 2021-02-20
Fix gap callback for degenerate Hamiltonians

### LINKED ISSUE #324: Gap callback does not take into account degeneration of the ground state
I have realized that the `Gap` callback does not take into account the degeneration of the ground state, i.e. it returns a gap equal to zero when the ground state is degenerated (take for example the `TFIM` Hamiltonian with `h=0` and `nqubits=4`).

stavros11:
Fixes #324 by going to the next eigenenergy if the gap is zero (`< EIGVAL_CUTOFF` for precision reasons). If this happens a warning is raised. I also moved `test_callbacks.py` to the new tests and enabled testing with the numpy backends.

### DIFF
--- a/src/qibo/abstractions/callbacks.py
+++ b/src/qibo/abstractions/callbacks.py
@@ -235,6 +235,11 @@ class Gap(Callback):
             first excited state energy (gap) is calculated.
             If ``mode`` is an integer, then the energy of the corresponding
             eigenstate is calculated.
+        check_degenerate (bool): If ``True`` the excited state number is
+            increased until a non-zero gap is found. This is used to find the
+            proper gap in the case of degenerate Hamiltonians.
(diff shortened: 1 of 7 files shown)


scarrazza (approved):
Thanks, looks perfect. I believe the coverage got stuck with the wrong commit.

Haowen-Zou:
> Fixes #324 by going to the next eigenenergy if the gap is zero (`< EIGVAL_CUTOFF` for precision reasons). If this happens a warning is raised. I also moved `test_callbacks.py` to the new tests and enabled testing with the numpy backends.

Can you show me how to plot out other Ground states when the system has two degenerations of ground states in linear.py? I'm struggling to find the other ground states when running linear.py ( it only plots out one ground state). 
(Sorry, I'm a beginner in Python.)

PR #330, opened 2021-02-22
Custom operator for measurement frequencies

stavros11:
This implements a CPU custom operator to calculate measurement frequencies without creating the samples tensor following the approach mentioned in #325. I have not integrated with measurements yet because performance is actually worse from what is implemented in #325:

nshots | #325  | this PR | this PR (no OMP) | this PR (`rand_r`)
-- | -- | -- | -- | --
1e7 | 0.7590808868408203 | 2.5838165283203125 | 2.062943696975708 | 0.4355289936065674
2e7 | 1.4452285766601562 | 5.975822925567627 | 4.169142961502075 | 0.8755536079406738
5e7 | 3.944162368774414 | 14.452663660049438 | 10.261326789855957 | 2.622830867767334
1e8 | 7.314714670181274 | 28.989457607269287 | 21.530482530593872 | 4.610459089279175

The performance issue may have to do with the way I parallelized using OMP because the custom operator is slightly faster without parallelization.


src/qibo/tensorflow/custom_operators/cc/kernels/measurements.cc
```diff
@@ -0,0 +1,99 @@
...
+    int64 nstates = 1 << nqubits;
+    #pragma omp parallel shared(cumprobs)
+    {
+        std::map<int64, int64> frequencies_private;
+        #pragma omp for
+        for (auto i = 0; i < nshots; i++) {
+          Tfloat random_number = ((Tfloat) rand() / RAND_MAX);
+          for (auto j = 0; j < nstates; j++) {
+            if (random_number < cumprobs[j]) {
+                if (frequencies_private.find(j) == frequencies_private.end()) {
```
scarrazza, line 30: I think, this is the killer check for the parallel evaluation.

scarrazza:
@stavros11 could you please send me a test code to reproduce your numbers using the new custom op?

stavros11:
> @stavros11 could you please send me a test code to reproduce your numbers using the new custom op?

This is the code I used after compiling the custom operators from this branch:

```Python
import time
import qibo
import tensorflow as tf
from qibo.tensorflow import custom_operators as op


def main(nqubits, nshots):
    probs = tf.random.uniform((2 ** nqubits,))
    probs = probs / tf.reduce_sum(probs)
    cumprobs = tf.cumsum(probs)

    frequencies = tf.zeros(int(probs.shape[0]), dtype=tf.int64)
    start_time = time.time()
    frequencies = op.measure_frequencies(frequencies, cumprobs, nqubits, nshots, qibo.get_threads())
    print("Execution time:", time.time() - start_time)
```

To test the version without OMP I just modified the C++ code to use a simple single-thread loop.

src/qibo/tensorflow/custom_operators/cc/kernels/measurements.cc
stavros11: In principle (and if I am correct) the `map`/`unordered_map` should implement a hash table so lookups should be O(1). Alternatively one can just use a simple vector for `frequencies_private` and then add it to the total `frequencies` vector, I just thought the map approach would be more memory friendly for more qubits (because `frequencies` and `frequencies_private` would have size 2^nqubits).

What is more interesting though is that even the non-OMP version which uses a simple loop:
```C++
for (auto i = 0; i < nshots; i++) {
    Tfloat random_number = ((Tfloat) rand() / RAND_MAX);
    for (auto j = 0; j < nstates; j++) {
        if (random_number < cumprobs[j]) {
            frequencies_private[j]++;
            break;
        }
    }
}
```
is slower than #325, which is also single thread but much more memory intensive.

stavros11:
I tried an alternative approach based on Metropolis algorithm which I believe (but not tested) gives the correct measurement statistics:
```C++
int64 shot = ((int64) rand() % nstates);
int64 i = 0;
while (i < nshots) {
      int64 new_shot = ((int64) rand() + shot) % nstates;
      Tfloat ratio = probs[new_shot] / probs[shot];
      if (ratio >= rand() / RAND_MAX) {
        shot = new_shot;
        frequencies[shot]++;
        i++;
      }
}
```
The single-thread performance is the best so far:

nshots | time (sec)
-- | --
1e7 | 0.32917308807373047
2e7 | 0.656909704208374
5e7 | 1.6262402534484863
1e8 | 3.3006410598754883

I also tried several ways to parallelize it but I could not get it faster than these single thread times. This would still need minutes for 10^10 shots and hours for 10^12 but it is 2x better than #325.

scarrazza:
@stavros11 thanks for the metropolis, but before using it, could you please try to replace `rand` with `rand_r`? I `rand` is blocking the openmp threads.

For example, something like this should be scale with the number of threads:
```c++
    unsigned seed;
    #pragma omp parallel
    {
        std::map<int64, int64> frequencies_private;
        seed = 12345 + 17*omp_get_thread_num();
        #pragma omp for
        for (auto i = 0; i < nshots; i++) {
          Tfloat random_number = ((Tfloat) rand_r(&seed) / RAND_MAX);
          ...
```

stavros11:
> @stavros11 thanks for the metropolis, but before using it, could you please try to replace `rand` with `rand_r`? I think `rand` is blocking the openmp threads.

Thanks for noting this, indeed performance improves significantly from this change. I posted the new timings in an additional column in the first post. Now multi-threading is a bit less than 2x faster compared to the tf-primitive approach of #325. 

An issue that remains is that sometimes the sum of frequencies is not exactly `nshots`. For example when `nshots=5e7` the sum of frequencies may be something like 49999998 and the same happens for 1e8. I thought that this is a precision issue so I changed `nshots` from attribute to input in the custom operator definition however this did not fix the issue. Changing nshots to input increased performance by a few seconds though (not really sure why).

scarrazza:
Very good, thanks for the update. Concerning the nshots, what happens if you print it from the c++ code? Did you get the original input? If so, then we may have an issue with the frequency type.

stavros11:
> Very good, thanks for the update. Concerning the nshots, what happens if you print it from the c++ code? Did you get the original input? If so, then we may have an issue with the frequency type.

I have been looking at this now and I think the issue is that it misses the cases `random_numer == 1` which should go to the last entry of frequencies. Initially I thought that the fix would be to change 
```random_number < cumprobs[j]``` 
to 
```random_number <= cumprobs[j]```
however this did not work. Strangely if `random_number` and `cumprobs[j]` are both 1 this returns `false` even with the `<=`. If I make the check
```random_number <= cumprobs[j] || j == nstates - 1```
it works as expected and the sum of frequencies is always `nshots` however there is a strange performance deterioration, eg. for 1e8 shots it goes from 4.6sec to 5.9sec.

**Comment by scarrazza** [comment]
Does this happens always, or only when multi-threading is enabled?

**Comment by stavros11** [comment]
> Does this happens always, or only when multi-threading is enabled?

Both the issue and the performance deterioration also happen in single thread, even if I completely remove OMP and just use a single for loop to update the frequencies tensor directly.

**Comment by stavros11** [comment]
By the way I used the seed trick to parallelize the Metropolis version and performance seems to be pretty good:

nshots | time (sec)
-- | --
1e7 | 0.05114936828613281
2e7 | 0.10785055160522461
5e7 | 0.2578613758087158
1e8 | 0.5184485912322998

I still need to confirm that the resulting frequencies make sense though but if it is perhaps this is an acceptable solution that may also get some speed-up from GPU.

**Comment by scarrazza** [comment]
Great, indeed, this on GPU will fly.

**Comment by stavros11** [comment]
@scarrazza I just pushed the latest version after the fix we discussed. The CPU times I get from my notebook now are the following:
nshots | time (sec)
-- | --
1e7 | 0.5260567665100098
2e7 | 1.0730626583099365
5e7 | 3.3192508220672607
1e8 | 5.539575815200806

The difference compared to the table in the first post is either explained by the notebook being not very stable (other processes running, etc.) or just the fact that we switched the `cumprobs` type to `float64` because so far it was using `float32` that is default for tf.

I still need to integrate the custom operator to the actual qibo measurements, fix the pylint issue of the CI (not sure why it appears here) and update the measurement tests. We may need to actually implement the mechanism to set the C++ random seed for the tests. However the C++ part of the code is already there and seems to work well so you can have a look at the CUDA whenever you can.

**Comment by scarrazza** [comment]
@stavros11 great, thanks I will have a look at the cuda approach, while you take care of the other points, including metropolis.

**Comment by stavros11** [comment]
Regarding the tests, I believe there is an issue with the seed in `rand_r`. Right now we are using a hard-coded seed so running the same code in a specific machine produces the same random results always. Eg:
```Python
c = models.Circuit(nqubits)
c.add((gates.H(i) for i in range(nqubits)))
result = c(nshots=nshots)
print(result.frequencies()
```
produces 
```
Counter({'1110': 63055, '0111': 62767, '1100': 62756, '1001': 62716, '0001': 62646, '1000': 62617, '0000': 62543, '0100': 62487, '0110': 62482, '0010': 62322, '1101': 62321, '1011': 62318, '1111': 62296, '0011': 62282, '0101': 62215, '1010': 62177})
```
every time I run it, however the result is different if I use a different machine. I tried on the DGX and I get a different result there.

I think we need to replace `rand_r` with a generator that has the same behavior in all machines once the seed is specified, otherwise it will not be possible to create stable probabilistic measurement tests.

scarrazza:
> I think we need to replace rand_r with a generator that has the same behavior in all machines once the seed is specified, otherwise it will not be possible to create stable probabilistic measurement tests.

Not really, suppose the following algorithm:
- request a seed from the user
- before performing the rejection sampling / metropolis:
  - use rand or rand_r to generate an unique sequence of seeds for the length of omp_num_thread, store these seeds in an array.
  - enter the omp loop, pick the corresponding seed for the given thread
  - continue as usual with rand_r

stavros11:
> > I think we need to replace rand_r with a generator that has the same behavior in all machines once the seed is specified, otherwise it will not be possible to create stable probabilistic measurement tests.
> 
> Not really, suppose the following algorithm:
> 
> * request a seed from the user
> * before performing the rejection sampling / metropolis:
>   
>   * use rand or rand_r to generate an unique sequence of seeds for the length of omp_num_thread, store these seeds in an array.
>   * enter the omp loop, pick the corresponding seed for the given thread
>   * continue as usual with rand_r

I believe that even with this algorithm the results will be different for machines that have different number of threads. Even for a single machine, one can change the results by limiting the number of OMP threads using `qibo.set_threads`. I can confirm though that the results depend only on the number of threads. For example if I limit the DGX to 6 threads (same as my notebook) then results agree with my notebook. If I limit both machines to 4 threads results also agree, etc.

scarrazza:
@stavros11, but then I am not sure I understand the problem. If the problem is reproducibility, then this algorithm works, because we can emulate the behaviour by setting the seed and the maximum number of threads. I don't think we need more than that at the moment, given that when moving to GPU, the random generator and the seed sequence will change.

stavros11:
> If the problem is reproducibility, then this algorithm works, because we can emulate the behaviour by setting the seed and the maximum number of threads. 

Right, I implemented the functionality for the user to pass a seed to the new custom operator and updated the tests by setting both the seed and threads to 1. Tests pass both on my machine and the DGX CPU, if the CI works the CPU part of this should be completed.

> when moving to GPU, the random generator and the seed sequence will change.

That's correct, in `tests_new/test_measurement_gate_probabilistic.py` we already have different hardcoded results when the GPU is used and these are different from the CPU results. Once we have the CUDA implementation for the frequency op we can update these numbers accordingly.

PR #335, opened 2021-03-02
PartialTrace gate for density matrices

stavros11:
Following a discussion we had with @AdrianPerezSalinas and @igres26 this implements the following:

* A `PartialTrace` gate that acts on density matrices only (no state vectors) and implements the following transformation: rho -> (|0><0|)_A x Tr_A (rho), where A is a selected subsystem of qubits.
* Removes the `PartialTrace` callback to avoid code repetition and confusion since its action is substituted by the new gate.
* Migrates some gate tests from `tests/test_density_matrix.py` to `tests_new/test_core_gates_density_matrix.py`.


scarrazza (approved):
Thanks, looks good to me.

PR #340, opened 2021-03-05
Move state reshape in gates

### LINKED ISSUE #337: Cannot allocate circuits with large nqubits
As discussed today, there is a crash due to the flat shape allocation in the Circuit constructor.

stavros11:
Moves the state reshape required for non-custom backends inside the gates instead of the circuit. This fixes some compatibility issues between gates of different backends and makes testing easier. 

It also fixes #337 since the `tensor_shape` and `flat_shape` attributes are no longer required during circuit creation. For this point I also added a test that initializes a QFT for n=100 qubits and checks that the depth is 2n and the number of gates n^2/2 + n.


scarrazza (approved):
@stavros11 thanks for this implementation and fix, looks good!

PR #343, opened 2021-03-05
Fix controlled by Y gates for custom backend

stavros11:
This PR fixes a bug related to using controlled by Y gates with density matrices with the custom backend. To reproduce:
```Python

c = models.Circuit(2, density_matrix=True)
c.add(gates.Y(1).controlled_by(0))
state = c().state()
print(np.trace(state))
```
prints -1 while the trace of the density matrix should always be 1. This appears only in the custom backend, all other backends work fine.


scarrazza (approved):
Thanks, looks good!

PR #344, opened 2021-03-06
Hamiltonian expectation for density matrices

### LINKED ISSUE #339: Hamiltonian expectation for density matrices
**Describe the bug**
Hamiltonian expectation value calculation is wrong for density matrices, as they are treated as state vectors.

**To Reproduce**
```Python
import numpy as np
from qibo import hamiltonians

def random_density_matrix(nqubits):
    shape = 2 * (2 ** nqubits,)
    rho = np.random.random(shape) + 1j * np.random.random(shape)
    rho = rho + rho.T.conj()
    # Normalize
    ids = np.arange(2 ** nqubits)
    rho[ids, ids] = rho[ids, ids] / np.trace(rho)
    return rho


nqubits = 5
ham = hamiltonians.XXZ(nqubits)
state = random_density_matrix(nqubits)
print(ham.expectation(state))

matrix = np.array(ham.matrix)
# EV calculation for density matrices
print(np.trace(matrix @ state))
# EV calculation for state vectors
print(np.sum(np.conj(state) * (matrix @ state)))
```

**Expected behavior**
The calculated expectation value (first printed number) should agree with the `np.trace` calculation (second printed number) but instead it agrees with the third printed value, which is the formula for state vectors.

stavros11:
Fixes #339.

### DIFF
--- a/src/qibo/abstractions/states.py
+++ b/src/qibo/abstractions/states.py
@@ -262,3 +262,7 @@ class AbstractState(ABC):
                 ``p0`` will be used for both bitflips.
         """
         raise_error(NotImplementedError)
+
+    @abstractmethod
+    def expectation(self, hamiltonian, normalize=False): # pragma: no cover
+        raise_error(NotImplementedError)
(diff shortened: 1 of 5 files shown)


scarrazza (approved):
Thanks, looks good.

PR #345, opened 2021-03-07
Fix issue with tf slicing

### LINKED ISSUE #342: Controlled by gates with density matrices
**Describe the bug**
Controlled gates that are created using the `controlled_by` method do not work with density matrices of 5 or more qubits if the `defaulteinsum` or `matmuleinsum` Tensorflow backends are used.

**To Reproduce**
```Python
import qibo
qibo.set_backend("defaulteinsum")

c = models.Circuit(5, density_matrix=True)
c.add(gates.H(1).controlled_by(0))
result = c()
```
leads to the following Tensorflow error:

```tensorflow.python.framework.errors_impl.UnimplementedError: Unhandled input dimensions 10 [Op:StridedSlice] name: strided_slice/```

The issue appears only for 5 or more qubits and only with the native tensorflow backends only. The numpy backends and the custom backend work without problem.

stavros11:
Fixes #342. In principle this is a tf issue as the existing code in master still works with numpy, I just changed the slicing to `K.gather` so that it works with Tensorflow too.

I think the problem of #342 should now be fixed but I would like to do a few more tests with more qubits because based on some other similar tf issues we fixed recently I suspect the tf einsum may fail in some cases. I am not 100% certain though.

### DIFF
--- a/src/qibo/backends/abstract.py
+++ b/src/qibo/backends/abstract.py
@@ -332,6 +332,11 @@ class AbstractBackend(ABC):
         raise_error(NotImplementedError)
 
     @abstractmethod
+    def squeeze(self, x, axis=None): # pragma: no cover
+        """Removes axis of unit length."""
+        raise_error(NotImplementedError)
+
+    @abstractmethod
     def gather(self, x, indices=None, condition=None, axis=0): # pragma: no cover
         """Indexing of one-dimensional tensor."""
         raise_error(NotImplementedError)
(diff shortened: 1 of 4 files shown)


stavros11:
I did a few tests with various controlled by configurations for up to 14 qubits and I did not observe any crashes after this fix, so I believe the einsum issue does not appear here. If you agree we can probably merge this and release, as there aren't any other bug related issues for now.

scarrazza (approved):
@stavros11, thanks for this fix and checks, looks good.

PR #349, opened 2021-03-10
Fix `.on_qubits()` with `.controlled_by()` gates

### LINKED ISSUE #348: the .on_qubits() feature fails when the quantum circuit has .controlled_by() gates.
**Describe the bug**
When trying to add a `Circuit` with `.controlled_by()` gates into a bigger circuit with the `.on_qubits()` method, an error occurs.

**To Reproduce**
```Python
circ1 = Circuit(1)
circ1.add(gates.H(0))

circ2 = Circuit(2)
circ2.add(gates.CNOT(0,1))
circ2.add(gates.H(1).controlled_by(0))

circ3 = Circuit(2)
circ3.add(circ1.on_qubits(*[0]))
circ3.add(circ2.on_qubits(*range(2)))
```

Seems to only happens when `controlled_by` gates are applied, as the CNOT gate seems to go though fine when doing:

```Python
for gate in circ2.on_qubits(*range(2)):
      print(gate)
```

stavros11:
Fixes #348. I added some tests that cover the problematic case but it would be good to also extensively test the example of @igres26 before merging this.

### DIFF
--- a/src/qibo/abstractions/abstract_gates.py
+++ b/src/qibo/abstractions/abstract_gates.py
@@ -184,15 +184,20 @@ class Gate:
         b = not (t1 & set(gate.qubits) or t2 & set(self.qubits))
         return a or b
 
-    def on_qubits(self, *q, controls=None) -> "Gate":
+    def on_qubits(self, *q) -> "Gate":
         """Creates the same gate targeting different qubits.
 
         Args:
             q (int): Qubit index (or indeces) that the new gate should act on.
         """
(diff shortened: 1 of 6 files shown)


scarrazza (approved):
Looks good, let's wait for @igres26 checks before merging.

igres26 (approved):
Thanks for implementing this so quickly. Everything is working as intended in the programs that raised the issue now.

PR #350, opened 2021-03-10
Use random seed for frequency custom operator

stavros11:
As @igres26 pointed out, currently the seed used in the frequency custom operator is generated using the system's time during the backend initialization. Because the backend is normally initialized once when a script is executed, all calls of the `K.sample_frequencies` method return the same frequencies if executed in the same runtime. For example in:
```Python
probs = np.random.random(8)
probs = probs / np.sum(probs)
frequencies1 = K.sample_frequencies(probs, nshots=int(1e6))
frequencies2 = K.sample_frequencies(probs, nshots=int(1e6))
```
`frequencies1` will be exactly the same as `frequencies2`, while usually the user would expect them to be different. This is very relevant in notebooks where the same cell may be executed several times in the same runtime with the user expecting different results.

This PR fixes this issue by generating a new random seed each time `K.sample_frequencies` is called.


igres26 (approved):
Everything works as expected now in notebooks. Thank you very much for implementing this.

Issue #313, opened 2021-02-03
Conditional circuit design after measurements


scarrazza:
We should consider the possibility to provide conditional circuit operations from measurement input.

PR #352, opened 2021-03-11
Conditioning gates on measurement outcomes

### LINKED ISSUE #327: Measurements during circuit execution
As we aim to tackle more complex quantum algorithms we will need to upgrade our system of performing measurements during the execution of a circuit.

Until now, measurements do not alter the wave function of the quantum state, and in order to fix that, the Collapse gate was introduced. This can be used to simulate measurement but requires to start a new Circuit object every time a measurement is taken (see Shor example). 

Collapse also appears to have issues when trying to collapse a state into a non-existent output, try:
```
c = Circuit(2)
c.add(gates.H(0))
c.add(gates.X(1))
c.add(gates.Collapse(1, result=0))
result = c()
```
where `nan` values appear in the wave function.

Ideally one would want to add measurements during a circuit, and then add gates or make alterations to the circuit conditioned to the result of the measurement result. This approach would make sense in a setting where every circuit execution yields one result, as keeping track of all options at the same time would require exponential memory.

This would not only be useful for the semi-classical Shor example, but on future simulations of quantum error correcting codes.

stavros11:
Implements the following in order to fix #327 and #313:

1. Removes `gates.Collapse` because it has the division with zero / `nan` issues explained in #327.
2. Integrates the functionality of collapse in the measurement gate (`gates.M`) using a `collapse` flag passed during the gate initialization: `gates.M(0, 1, collapse=True)`. Default is `False` so that it agrees with current usage.
3. `M` gates with `collapse=False` can be added only in the end of the circuit and return shot results (samples/frequencies) without affecting the state vector.
4. `M` gates with `collapse=True` can be added anywhere in the circuit and they return the state vector collapsed according to the measurement result. Only single-shot measurements are supported here.
5. The measurement outcome of `collapse=True` gates can be accessed during their addition to the circuit:
```Python
output = c.add(gates.M(0, collapse=True))
```
and supports all methods of measurement results (eg. `.samples()`, `.frequencies()`, etc.)
6. The measurement outcome of `collapse=True` can be used as parameter in other gates:
```Python
output = c.add(gates.M(0, collapse=True))
c.add(gates.RX(1, theta = np.pi * output / 4))
```

Below I list three questions regarding the next steps:

* **Completing the Shor example**: Currently the semiclassical Shor example is broken because the Collapse gate is removed (*I disabled the related test temporarily*). We can fix this following Sergi's proposal in #327. To do this we should implement one of the following two:
    1. A `Reset` gate which resets qubits to |00...0> regardless of their current state.
    2. The possibility to condition gates on measurement outcomes.

Based on our discussion I am inclined to go with (i) and leave (ii) for a future PR because it can be partially substitute by controlled gates, eg:
```Python
output = c.add(gates.M(0, collapse=True)
if output:
    c.add(gates.X(1))

# is equivalent to

output = c.add(gates.M(0, collapse=True)
c.add(gates.CNOT(0, 1))
```

My question regarding the `Reset` gate is how superposition is handled. For example if I have the state |111> and I reset the second qubit I get |101>, but what do I get if I have (|000> + |111>)? Do I lose the whole |111> component and go to |000> or do I just factor out a |0> and go to (|00> + |11>)|0>? I am not sure how this gate is actually implemented experimentally.

* **Performing multiple shots**: The `collapse=True` measurement cannot be used with multiple shots because the collapse operation is not well defined in such case. Currently if the user attempts to do `c(nshots=1000)` in a circuit that contains `collapse=True` measurements only then **`nshots` will be ignored** and the circuit will perform a single shot only. The circuit has to be re-executed if more shots are required and this can only be done with a manual loop:
```Python
c = Circuit(2)
c.add(gates.H(0))
output = c.add(gates.M(0, collapse=True))

samples = []
for _ in range(1000):
    c() # execute the circuit
    samples.append(output.samples())
```
The disadvantage of this is that the user has to take care of collecting the samples as shown above. Alternatively we could implement `c(nshots=1000)` and automatically fall-back to execution with repetition when `collapse=True`. The loop method is more transparent, while the `c(nshots=1000)` easier to use.

* **Parametrizing using outcome of more than one qubit**: Regarding point 6, only single-qubit measurement outputs can be used as gate parameters for consistency. It is still possible to measure multiple qubits and use the outcomes as long as seperate gates are used. For example the following is currently possible:
```Python
output0 = c.add(gates.M(0, collapse=True))
output1 = c.add(gates.M(1, collapse=True))
c.add(gates.RX(2, theta = np.pi * (output0 + output1) / 4))
```

Alternatively we could implement something like following:
```Python
output0, output1 = c.add(gates.M(0, 1, collapse=True))
c.add(gates.RX(2, theta = np.pi * (output0 + output1) / 4))
```
Do you think such functionality is useful or shall we leave it as it is and perhaps consider it in the future? The advantage of this is that the collapse kernel will be called once instead of twice so it may be slightly faster.


scarrazza:
@stavros11 thanks for this proposal.

> Completing the Shor example: 

I agree with implementing first (i) and then (ii) later. Concerning the Reset, I think this gate could get different initializations to support the multiple reset possibilities, however probably @igres26 has a clear picture of the most common reset approach, which can be our default.

> Performing multiple shots:

My preference is to delegate the for loop to the circuit execution method, it is less transparent but preserves the symmetry with a non conditional circuit.

> Parametrizing using outcome of more than one qubit: 

Both are clear, however the second is faster to code. 
We could do what matplotlib does with subfigures, by just returning an array of outputs (i.e. the `fig, axs`).

stavros11:
#355 fixes point 2 and 3 by implementing the functionalities described above, while #356 fixes the Shor example without the use of `Reset`. Once we merge these two PRs to this one and after adding some documentation, this should be good to go.

stavros11:
I updated the docs and added an example explaining the new functionalities. I believe coverage will be fixed once we merge #355.

doc/source/advancedexamples.rst
```diff
@@ -348,6 +348,117 @@ gates during optimization.
...
 
+.. _collapse-examples:
+
+How to collapse state during measurements?
+------------------------------------------
+
+As mentioned in the :ref:`How to perform measurements? <measurement-examples>`
+measurement can by default be used only in the end of the circuit and they do
+not have any effect on the state. In this section we describe how to collapse
+the state during measurements and re-use measured qubits in the circuit.
```
scarrazza, line 359: Looks good, however I would add a sentence explaining what is a collapse.

doc/source/advancedexamples.rst
```diff
@@ -348,6 +348,117 @@ gates during optimization.
...
+Conditioning gates on measurement outcomes
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The output of ``collapse=True`` measurements can be used as a parameter in
+any parametrized gate as follows:
+
+.. code-block:: python
+
+    from qibo.models import Circuit
+    from qibo import gates
```
scarrazza, line 431:
```suggestion
    from qibo import gates
    import numpy as np
```

doc/source/advancedexamples.rst
```diff
@@ -348,6 +348,117 @@ gates during optimization.
...
+when the measurement is performed. As explained above, if a ``nshots > 1`` is
+given during circuit execution the execution is repeated using a loop.
+
+If more than one qubits are used in a ``collapse=True`` measurement gate the
+``output`` can be indexed accordingly:
+
+.. code-block:: python
+
+    from qibo.models import Circuit
+    from qibo import gates
```
scarrazza, line 452:
```suggestion
    from qibo import gates
    import numpy as np
```

scarrazza (approved):
@stavros11 thank you for this PR. Looks pretty good to me, I have tested the code and everything seems consistent. @igres26 could you please let us know if this works for your applications?

scarrazza:
@stavros11 btw, this branch is failing on GPU with the following:

<details open>
====================================== short test summary info =======================================
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_random[custom-None]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_random[defaulteinsum-None]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_random[matmuleinsum-None]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_random[custom-accelerators5]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_random[custom-accelerators6]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_random[custom-accelerators7]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_repeated_execution[custom-None-True]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_repeated_execution[custom-None-False]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_repeated_execution[defaulteinsum-None-True]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_repeated_execution[defaulteinsum-None-False]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_repeated_execution[matmuleinsum-None-True]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_repeated_execution[matmuleinsum-None-False]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_repeated_execution[custom-accelerators5-True]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_repeated_execution[custom-accelerators5-False]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_repeated_execution[custom-accelerators6-True]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_repeated_execution[custom-accelerators6-False]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_repeated_execution[custom-accelerators7-True]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_repeated_execution[custom-accelerators7-False]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_repeated_execution_final_measurements[custom-None]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_repeated_execution_final_measurements[defaulteinsum-None]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_repeated_execution_final_measurements[matmuleinsum-None]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_repeated_execution_final_measurements[custom-accelerators5]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_repeated_execution_final_measurements[custom-accelerators6]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_repeated_execution_final_measurements[custom-accelerators7]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_multiple_qubits[custom]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_multiple_qubits[defaulteinsum]
FAILED src/qibo/tests_new/test_measurement_gate_collapse.py::test_measurement_result_parameters_multiple_qubits[matmuleinsum]
</details>

stavros11:
> @stavros11 thank you for this PR. Looks pretty good to me, I have tested the code and everything seems consistent.

Thank you for checking. I fixed the suggested changes for the docs.

> @stavros11 btw, this branch is failing on GPU with the following:

Thanks for finding this, I actually forgot to run the tests on GPU. This should be fixed now. Another thing I noticed, not related to this PR, is that tests may fail for GPUs with small memory. For example when I run the tests from master on my notebook GPU (4GB) I get `core dumped` during `test_backends_agreement.py`, while all tests pass successfully on DGX (32GB). We may want to investigate this issue on a different PR.

scarrazza:
@stavros11 thanks. Test are fine now. When using a 4gb card, I get a crash also in `test_density_matrix.py`.

scarrazza:
Refers #361, if @igres26 approves we should merge this PR.

igres26 (approved):
Thanks for implementing this. If the tests are passing, and since the Shor example works now I think we can merge. I am still in talks with the experimental team in order to properly establish what happens to the quantum system when the Reset gate is applied. But we can keep going with the manual reset (measure and apply an X gate accordingly) for now.

stavros11:
> Refers #361, if @igres26 approves we should merge this PR.

I would suggest to check performance before merging. I did the following comparison:
```Python
circuit.add((gates.H(i) for i in range(nqubits)))
output = circuit.add(gates.M(0, collapse=True))
circuit.add(gates.RX(1, theta=math.pi * output / 3))
circuit.add(gates.RX(2, theta=math.pi * (1 - output) / 4))
circuit.add(gates.RX(3, theta=math.pi * output / 5))
result = circuit()
```
using this branch, vs
```Python
circuit.add((gates.H(i) for i in range(nqubits)))
circuit.add(gates.M(0))
result = circuit(nshots=1)
circuit.add(gates.Collapse(0, result=r))
x = int(r[0])
circuit.add(gates.RX(1, theta=math.pi * x / 3))
circuit.add(gates.RX(2, theta=math.pi * (1 - x) / 4))
circuit.add(gates.RX(3, theta=math.pi * x / 5))
new_result = circuit(initial_state=result.state())
```
using master.

Here are the results using 100 repetitions of the above on GPU:

nqubits | total_time_mcollapse | total_time_master
-- | -- | --
6 | 0.677322 | 0.425897
8 | 0.653290 | 0.500260
10 | 0.766721 | 0.535297
12 | 0.826574 | 0.705908
14 | 0.850507 | 0.604870
16 | 1.004402 | 0.756466
18 | 1.256482 | 1.128693
20 | 2.117982 | 2.169466
22 | 6.161532 | 6.650354
24 | 20.264377 | 21.664483

and on CPU

nqubits | total_time_mcollapse | total_time_master
-- | -- | --
6 | 0.478488 | 0.245857
8 | 0.487205 | 0.273919
10 | 0.557703 | 0.285446
12 | 0.558613 | 0.308191
14 | 0.853567 | 0.359790
16 | 0.692799 | 0.458283
18 | 1.270047 | 1.343697
20 | 4.821236 | 4.383812
22 | 24.102943 | 25.035523
24 | 104.277009 | 103.903807

There is no big performance difference, however master is slightly faster for small circuits, not really sure why yet.

stavros11:
I get the following times after disabling the rotations / symbol evaluation:

nqubits | mcollapse GPU | master GPU | mcollapse CPU | mcollapse CPU
-- | -- | -- | -- | -- 
6 | 0.175398 | 0.158728 | 0.141566 | 0.115476
8 | 0.171014 | 0.158669 | 0.117421 | 0.104842
10 | 0.182583 | 0.152684 | 0.114812 | 0.134738
12 | 0.183026 | 0.156103 | 0.119232 | 0.108770
14 | 0.220729 | 0.150761 | 0.128082 | 0.118374
16 | 0.251556 | 0.249089 | 0.152948 | 0.142831
18 | 0.202999 | 0.229926 | 0.382487 | 0.337713
20 | 0.332628 | 0.384404 | 1.015209 | 1.070440
22 | 1.095007 | 0.999076 | 4.004397 | 3.801437
24 | 3.302335 | 3.254082 | 14.292217 | 14.589004

scarrazza:
@stavros11 ok, thanks. So presumably the overhead is due to the symbol evaluation and I think we can't do much for this, except if we replace sympy with an lighter custom symbol class.

PR #354, opened 2021-03-16
Updating multiline circuit draw

### LINKED ISSUE #353: Improvements on drawing circuits.
Following some discussions with @joseignaciolatorre I would like to make some proposals on changes do the `.draw()` feature for ease of reading.

The `line_wrap` flag should be initialized at some value instead of being set to unlimited by default. We should decide a sensible value in order to not affect small and simple circuits, but at the same time if a large circuit needs to be drawn, have it split into chunks so that it can be properly interpreted.

Additionally, when splitting the circuits into different lines, it would be very useful to have the qubit labels repeated also on each new line in order to easily follow gates acting on a qubit on the following lines as well. The same argument can be made for having labels on the right side of the circuit as well.

If it does not crowd the screen too much, having extra qubit labels helps a lot when analyzing large circuits.

scarrazza:
Closes #353. @igres26 could you please give a try and let me know if you like it?


igres26 (approved):
Thanks for implementing this. The extra qubit labels work as I asked, and make the circuit more easily interpreted. I am still unsure if 70 is the correct `line_wrap` for most users but it looks good on my end. Could probably push it to 75 or 80, but that will depend on the user, so I think its fine.

PR #355, opened 2021-03-16
Multiple collapse measurements with single circuit call

stavros11:
Implements the second question/point analyzed in #352 to allow performing multiple `collapse=True` shots using a single circuit call. The usage is the following:

```Python
c = Circuit(1)
c.add(gates.H(0))
output = c.add(gates.M(0, collapse=True))
result = c(nshots=100)
```
will perform 100 measurements by repeating the circuit execution. The samples are accessible via the `output` object, that is `output.samples()` will return a `(100, 1)` tensor. Note that execution with repetition is equivalent to:
```Python
samples = []
for _ in range(100):
    c()
    samples.append(output.samples())
```

It is also possible to use `output` to parametrize other gates:
```Python
c = Circuit(2)
c.add(gates.H(0))
output = c.add(gates.M(0, collapse=True))
c.add(gates.RX(1, theta=np.pi * output / 4))
result = c(nshots=100)
```
In this case the latest measurement outcome will be used to control subsequent gates in each repetition.

Normal measurement gates can also be used at the end of the circuit:
```Python
c = Circuit(2)
c.add(gates.H(0))
output = c.add(gates.M(0, collapse=True))
c.add(gates.RX(1, theta=np.pi * output / 4))
c.add(gates.M(0, 1))
result = c(nshots=100)
```
In this case `output.samples()` will have shape `(100, 1)` containing the results of the first (collapse) measurement, while `result.samples()` will have shape `(100, 2)` with the results of the second (non-collapse) measurement.

### DIFF
--- a/examples/shor/functions.py
+++ b/examples/shor/functions.py
@@ -363,47 +363,38 @@ def quantum_order_finding_semiclassical(N, a):
     x = [n+1+i for i in range(n)]
     ancilla = 2*n + 1
     q_reg = 2*n + 2
-    circuit = Circuit(2*n+3)
     print(f'  - Total number of qubits used: {2*n+3}.\n')
-    r = []
+    results = []
     exponents = []
     exp = a%N
     for i in range(2*n):
         exponents.append(exp)
         exp = (exp**2)%N
-    
+
+    circuit = Circuit(2 * n + 3)
     # Building the quantum circuit
     circuit.add(gates.H(q_reg))
-    circuit.add(gates.X(x[len(x)-1]))
+    circuit.add(gates.X(x[len(x) - 1]))
     #a_i = (a**(2**(2*n - 1)))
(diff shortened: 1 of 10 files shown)


stavros11:
The latest push implements point 3 from #352. Particularly the user can now measure multiple qubits with a single `collapse=True` measurement gate and use the corresponding outputs to control subsequent gates as follows:

```Python
output = c.add(gates.M(0, 1, collapse=True))
c.add(gates.RX(1, theta = np.pi * output[0] / 4)) # uses result of qubit-0 measurement
c.add(gates.RX(2, theta = np.pi * output[1] / 4)) # uses result of qubit-1 measurement
```

Doing
```Python
output0, output1 = c.add(gates.M(0, 1, collapse=True))
```
is also possible.

scarrazza (approved):

PR #356, opened 2021-03-17
Update semiclassical Shor example to use measurements with collapse

stavros11:
Updates the semiclassical version of the Shor example to use the collapse and gate conditioned on measurement outcome functionality implemented in #352. The whole procedure can now implemented using a single circuit, without requiring to recreate a new one after each measurement.

Note that measured qubits are reset to |0> using an RX gate conditioned on the measurement outcome, like the following example:
```Python
results.append(c.add(gates.M(0, collapse=True)))
c.add(gates.RX(0, theta=np.pi * results[-1]))
```

We may substitute this by the `Reset` gate in the future, once we know how this is implemented experimentally.

### DIFF
--- a/examples/shor/functions.py
+++ b/examples/shor/functions.py
@@ -363,47 +363,39 @@ def quantum_order_finding_semiclassical(N, a):
     x = [n+1+i for i in range(n)]
     ancilla = 2*n + 1
     q_reg = 2*n + 2
-    circuit = Circuit(2*n+3)
     print(f'  - Total number of qubits used: {2*n+3}.\n')
-    r = []
+    results = []
     exponents = []
     exp = a%N
     for i in range(2*n):
         exponents.append(exp)
         exp = (exp**2)%N
-    
+
+    circuit = Circuit(2 * n + 3)
     # Building the quantum circuit
     circuit.add(gates.H(q_reg))
-    circuit.add(gates.X(x[len(x)-1]))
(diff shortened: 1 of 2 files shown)


igres26 (approved):
I have run some instances of the factorization algorithm with the semiclassical update and everything seems to be working as it was before. 

Maybe we could do some benchmarks between this method and the old one just to know is there is a performance increase or loss with the changes. Even if the performance is a bit worse I think we should still keep this approach nonetheless.

stavros11:
Thanks for the check. Yes, I was planning to do some benchmarks comparing with the current version. Here are some numbers on the DGX GPU

N | nqubits | Current version (sec) | New version (sec)
-- | -- | -- | --
15 | 11 | 1.905610 | 2.097330
35 | 15 | 6.979931 | 7.279466
77 | 17 | 11.735047 | 12.348415
95 | 17 | 11.404145 | 12.327344
1007 | 23 | 64.107888 | 64.746143
2021 | 25 | 176.283828 | 178.652594
4087 | 27 | 694.024918 | 696.610630

I would actually expect the current version to be faster because it uses a single circuit so it should reallocate less objects, however it seems to be a few seconds slower. I am not sure if this is because of the `sympy.Symbol` evaluation, I guess we would a simpler circuit example to investigate this. As far as Shor is concerned though the difference is small.

PR #358, opened 2021-03-21
Standarizing optimizers input and output

### LINKED ISSUE #357: Result from optimizers
Hi!
I was working with optimizers implemented in Qibo and I think it would be better if the optimizers returns more than only the value of the function and optimal parameters for example why the minimisation process has stoped, number of function evaluation etc...
Methods like "Powell" have also the possibility in SciPy to set bounds (not in dictionary "options") but this is not possible in Qibo.

Let me know what do you think about it. Thank you all!

scarrazza:
Closes #357. This PR propagates the extra flags from scipy to our optimizer and vqe/qaoa minimizer methods, and returns an extra object which holds the original result output from the underlying optimizer library.

### DIFF
--- a/src/qibo/models.py
+++ b/src/qibo/models.py
@@ -172,7 +172,8 @@ class VQE(object):
         self.circuit = circuit
         self.hamiltonian = hamiltonian
 
-    def minimize(self, initial_state, method='Powell', options=None, compile=False, processes=None):
+    def minimize(self, initial_state, method='Powell', options=None, compile=False, processes=None,
+                 jac=None, hess=None, hessp=None, bounds=None, constraints=(), tol=None, callback=None):
         """Search for parameters which minimizes the hamiltonian expectation.
 
         Args:
@@ -184,6 +185,14 @@ class VQE(object):
             options (dict): a dictionary with options for the different optimizers.
             compile (bool): whether the TensorFlow graph should be compiled.
             processes (int): number of processes when using the paralle BFGS method.
+            jac (dict): Method for computing the gradient vector for scipy optimizers.
+            hess (dict): Method for computing the hessian matrix for scipy optimizers.
+            hessp (callable): Hessian of objective function times an arbitrary
+                vector for scipy optimizers.
+            bounds (sequence or Bounds): Bounds on variables for scipy optimizers.
+            constraints (dict): Constraints definition for scipy optimizers.
+            tol (float): Tolerance of termination for scipy optimizers.
+            callback (callable): Called after each iteration for scipy optimizers.
 
         Return:
             The final expectation value.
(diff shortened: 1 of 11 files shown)


src/qibo/optimizers.py
```diff
@@ -1,5 +1,6 @@
 def optimize(loss, initial_parameters, args=(), method='Powell',
-             options=None, compile=False, processes=None):
+             options=None, compile=False, processes=None, jac=None, hess=None,
+             hessp=None, bounds=None, constraints=(), tol=None, callback=None):
```
stavros11, line 3: Do you think it would be useful to add the `args` argument here to allow user to pass additional arguments to the loss? I think it is not useful for existing Qibo models but could be if the user defines a custom variational model.

Also, shall we make the order of arguments consistent with scipy by moving the `options` flag in the end?

src/qibo/models.py
```diff
@@ -212,12 +225,15 @@ def _loss(params, circuit, hamiltonian):
...
+        result, parameters, extra = self.optimizers.optimize(loss, initial_state,
+                                                             args=(self.circuit, self.hamiltonian),
+                                                             method=method, options=options,
+                                                             compile=compile, processes=processes,
+                                                             jac=jac, hess=hess, hessp=hessp,
+                                                             bounds=bounds, constraints=constraints,
+                                                             tol=tol, callback=callback)
         self.circuit.set_parameters(parameters)
-        return result, parameters
+        return result, parameters, extra
```
stavros11, line 236: We could make the optimizers and the `minimize` methods of models return a single result object instead of three objects. For scipy and the cma optimizer this can be the `extra` object since it already contains `result` and `parameters`, while for SGD we could create a small custom object that follows the scipy structure, eg. the result and parameters are accessible via `result.fun` and `result.x`. This would change our existing API slightly but I think it is easier to maintain. Otherwise the current structure is also fine with me. Let me know what you think.

stavros11 (approved):
Thank you for this fix. It looks good to me, just some small comments to perhaps consider below:

src/qibo/optimizers.py
scarrazza: > Do you think it would be useful to add the args argument here...

We have already this flag here, right (3rd argument)?

> Also, shall we make the order of arguments consistent with scipy by moving the options flag in the end?

yes, I agree.

src/qibo/models.py
scarrazza: Concerning this point, are you suggesting to define a new return object for all optimizer?
If we return the extra object for scipy and cma they will have different attributes (.fun, .x vs .result.fbest, .result.xbest).

src/qibo/optimizers.py
stavros11: > > Do you think it would be useful to add the args argument here...
> 
> We have already this flag here, right (3rd argument)?

Indeed, sorry I was looking at the `minimize` method arguments in `models.py` when writing this comment where we don't have the `args` argument but there it is probably not needed anyway. No need to change anything here.

src/qibo/models.py
stavros11: Yes, the suggestion was to return a single object for all optimizers which could follow the scipy conventions. For cma, indeed the attributes differ from scipy so we could either return our own custom object that follows scipy or return the cma object but break consistency with either optimizers. I am not sure if returning a single object is simpler than what we have now from the user perspective so we can keep whatever you prefer (or ask people who use the optimizers for their preference)

scarrazza: Right, I think we can keep the layout as it is for now and then see if there are more specific request in the future.

PR #363, opened 2021-03-24
Improving performance for circuits with many gates

### LINKED ISSUE #360: Memory usage when allocating large circuits
When allocating large circuits there seems to be a larger memory usage than expected. A lot of memory is used in the CPU regardless if the circuit is going to be executed on CPU or GPU.

That also seems to affect the running time of the algorithm, up to a point where circuit allocation might take longer than circuit execution.

**To Reproduce**

```Python
from qibo import gates

c = Circuit(21)

for i in range(2**9):
     for j in range(2**9):
          for k in range(2**1):
               for l in range(2**1):
                    c.add(gates.RY(20, 0.2).controlled_by(*range(20))
```

This might not be an issue normally as circuits are **encouraged** to not scale exponentially with the amount of qubits. Nonetheless, I think the memory issue is something we should look into.

stavros11:
I made some small modifications in some castings called by `gate.prepare()` in an attempt to fix #360. With this the memory usage for the following script
```Python
circuit = models.Circuit(10)
circuit.add((gates.H(0) for i in range(2 ** 20)))
```
drops from about 3GB to 1.2GB and execution time from 106.74sec to 21.08sec. The catch is that by removing these castings we lose in performance when executing a circuit multiple times. For example look at the following benchmark:
```Python
circuit.add((gates.H(i) for i in range(nqubits)))

for _ in range(nreps)
    result = circuit()
```

for 10 qubits on my notebook CPU:

nreps | master (sec) | this branch (sec)
-- | -- | --
1 | 0.000706 | 0.001770
10 | 0.002385 | 0.008641
50 | 0.011132 | 0.039351
100 | 0.021371 | 0.075836
200 | 0.033434 | 0.153833
500 | 0.087286 | 0.382295
1000 | 0.158653 | 0.811011
5000 | 0.816121 | 3.885552
10000 | 1.680684 | 7.897133

and on my notebook GPU:

nreps | master (sec) | this branch (sec)
-- | -- | --
1 | 0.002156 | 0.003461
10 | 0.003980 | 0.021935
50 | 0.012403 | 0.093961
100 | 0.022166 | 0.200843
200 | 0.043246 | 0.317211
500 | 0.108625 | 0.793593
1000 | 0.210205 | 1.976617
5000 | 1.043224 | 9.315798
10000 | 2.074535 | 19.654179

If I remember correctly repeated execution was the reason we added these castings in the first place. Also, removing `tf.device` from `prepare` will most likely break parallelization for multi-GPU set-ups. On the other hand, single repetition for larger circuits is not affected much by whether we cast or not:

* CPU:

nqubits | master (sec) | this branch (sec)
-- | -- | --
10 | 0.000706 | 0.001770
12 | 0.000900 | 0.001993
14 | 0.000957 | 0.002309
16 | 0.002036 | 0.003730
18 | 0.003834 | 0.005417
20 | 0.048215 | 0.034231
22 | 0.178869 | 0.181978
24 | 0.804867 | 0.816635

* GPU:

nqubits | master (sec) | this branch (sec)
-- | -- | --
10 | 0.002156 | 0.003461
12 | 0.002279 | 0.004582
14 | 0.002708 | 0.003884
16 | 0.003073 | 0.005941
18 | 0.004284 | 0.005147
20 | 0.010675 | 0.012997
22 | 0.034972 | 0.043789
24 | 0.242973 | 0.171259

With these in mind, a potential solution is to remove the castings from prepare and add a `.compile` like method that allows the user to do them manually. This would remove the casting overhead for deep circuits and for cases where the circuit is only run a few times. The multi-GPU case would need to be handled seperately. @scarrazza let me know what you think and once we decide on a solution I will try to implement it for all backends.


scarrazza:
@stavros11 thanks for this suggestion, given that the single execution seems not affected by this a posteriori casting, could we have an automatic compile/cast when the user calls the circuit.execute method for the first time?

stavros11:
> @stavros11 thanks for this suggestion, given that the single execution seems not affected by this a posteriori casting, could we have an automatic compile/cast when the user calls the circuit.execute method for the first time?

In the latest push I implemented something close to this idea by making `qubits_tensor` a property that it is calculated and cached the first time the gate is executed. With this implementation I get the following times for repeated execution:

* CPU

nreps | master (sec) | this branch (sec)
-- | -- | --
1 | 0.002803 | 0.001804
10 | 0.003977 | 0.003703
50 | 0.012083 | 0.009690
100 | 0.019403 | 0.019936
200 | 0.046936 | 0.035625
500 | 0.109860 | 0.091680
1000 | 0.201468 | 0.226800
5000 | 0.949477 | 0.985850
10000 | 1.893997 | 1.813349

* GPU

nreps | master (sec) | this branch (sec)
-- | -- | --
1 | 0.003900 | 0.002573
10 | 0.005564 | 0.004463
50 | 0.013515 | 0.012250
100 | 0.024330 | 0.022517
200 | 0.042546 | 0.042032
500 | 0.104766 | 0.102012
1000 | 0.206608 | 0.208275
5000 | 1.024028 | 1.024395
10000 | 2.162794 | 2.064779

I believe this is the best solution both in terms of performance and code simplicity, as we could perhaps remove the `prepare` methods and simplify the code a lot. Regarding the large circuits mentioned in #360, I attempted the following benchmark:
```Python
circuit = models.Circuit(10)
circuit.add((gates.H(0) for i in range(ngates)))
result = circuit()
```
for 10 qubits and large number of gates and I get the following numbers on CPU:

ngates | master (sec) | this branch (sec)
-- | -- | --
2^15 | 4.350579 | 2.253695
2^16 | 8.272054 | 4.860859
2^17 | 16.680923 | 8.546744
2^18 | 32.049377 | 17.529772
2^19 | 61.688870 | 33.915008
2^20 | 125.347635 | 70.605023

The times are total (creation + execution). The main difference is that in master most of the time is spent during circuit creation, while in this branch all time is spent in execution (because no preperation/castings happen during creation). The performance gain is probably due to using the pre-created Tensorflow matrix for H instead of re-casting it and creating copies every time.

scarrazza:
@stavros11 great thanks for this implementation. Maybe we should ask @igres26 to check if this branch fixes his large circuit.
I presume that the performance and memory (knowing that we start from ~900MB) used here is comparable to the numbers from cirq, right?

stavros11:
> @stavros11 great thanks for this implementation. Maybe we should ask @igres26 to check if this branch fixes his large circuit.
> I presume that the performance and memory (knowing that we start from ~900MB) used here is comparable to the numbers from cirq, right?

Running and executing the circuit in the above example for 2^20 gates takes about 1.3GB when the H gate is used and about 2GB when a rotation is used (because rotation requires creation of new matrices). Cirq takes about 700MB for H and 1GB for rotation.

Using this branch could work for Sergi's example, at least based on the numbers above the performance should be better than master. However there are many fixes I would like to do before merging, particularly propagate the fix in all backends (currently it is only implemented for custom) and fix the multi-GPU case where all tensors should be casted seperately in each device to avoid breaking the `joblib` parallelization.

Regarding the backends, I am trying to refactor the code a bit so that we have a single `gates.py` file (instead of `cgates.py` and `gates.py`). This may take some effort but I believe it will make future maintenance much easier.

src/qibo/core/cgates.py
```diff
@@ -50,16 +48,18 @@ def reprepare(self):
     def prepare(self):
         """Prepares the gate for application to state vectors."""
         self.is_prepared = True
```
scarrazza, line 50: Do we really need prepare at this point?

scarrazza (commented):
Thanks for the confirmation, very good, please go ahead with the refactor.

src/qibo/core/cgates.py
stavros11: I think no. I would like to remove `prepare` completely, I am just planning to do it after "merging" `gates` and `cgates` so that it works for all backends.

scarrazza: Ok, great.

stavros11:
I checked the multi-GPU performance of this branch and it is the same as master so there are no issues as I was expecting. This should be good to merge. #372 makes the refactor I wrote above and removes the `prepare` method from gates as we discussed.

stavros11:
@scarrazza, I fixed the conflicts on this branch, it should be good for a second review.

PR #364, opened 2021-03-25
Grover model

igres26:
Added a model in order to easily perform Grover's algorithm for any Oracle and Superposition.

Added a `grover-playground` inside examples in order to perform early tests, but `pytests` have to be built still in order to make sure everything is working as intended.

Any changes or improvements in order to better include it into Qibo are welcome.


src/qibo/models.py
```diff
@@ -1,10 +1,12 @@
 import math
 from qibo import get_backend
-from qibo.config import raise_error
+from qibo.config import raise_error, log
 from qibo.core.circuit import Circuit as StateCircuit
 from qibo.core.circuit import DensityMatrixCircuit
 from qibo.evolution import StateEvolution, AdiabaticEvolution
 from typing import Dict, Optional
+import numpy as np
+from qibo import gates
```
scarrazza, line 9: I would suggest to perform this imports inside the Grover class, to avoid importing overhead for other models.

src/qibo/models.py
```diff
@@ -429,3 +431,161 @@ def _loss(params, qaoa, hamiltonian):
...
+    """Model that performs Grover's algorithm.
+
+    # TODO: Add a few more details on Grover's algorithm and/or reference.
+
+    Args:
+        oracle (:class:`qibo.core.circuit.Circuit`): quantum circuit that flips
+            the sign using a Grover ancilla initialized with -X-H-
+            and expected to have the total size of the circuit.
+        initial_state (:class:`qibo.core.circuit.Circuit`): quantum circuit
+            that initializes the state. Leave empty if |000..00>
```
scarrazza, line 446: Do you mean "Default is |000..00>", right?

src/qibo/models.py
```diff
@@ -429,3 +431,161 @@ def _loss(params, qaoa, hamiltonian):
...
+        superposition (:class:`qibo.core.circuit.Circuit`): quantum circuit that
```
scarrazza, line 447: Maybe we should rename it `superposition_circuit` in order to make it transparent.

src/qibo/models.py
```diff
@@ -429,3 +431,161 @@ def _loss(params, qaoa, hamiltonian):
...
+            takes an initial state to a superposition. Expected to use the first
+            set of qubits to store the relevant superposition.
+        sup_qubits (int): number of qubits that store the relevant superposition.
+            Leave empty if superposition does not use ancillas.
+        sup_size (int): how many states are in a superposition.
+            Leave empty if its an equal superposition of quantum states.
+        num_sol (int): number of expected solutions. Needed for normal Grover.
```
scarrazza, line 454: I would suggest to name these variables with the full prefix name, so it is easier for interpretation.

src/qibo/models.py
```diff
@@ -429,3 +431,161 @@ def _loss(params, qaoa, hamiltonian):
...
+            Leave empty for iterative version.
+        check (function): function that returns True if the solution has been
+            found. Required of iterative approach.
+            First argument should be the bitstring to check.
+        check_args (tuple): arguments needed for the check function.
+            The found bitstring not included.
+    """
+    def __init__(self, oracle, superposition=None, initial_state=None, sup_qubits=None, sup_size=None, num_sol=None, check=None, check_args=()):
```
scarrazza, line 462: Break this into 2 or 3 lines.

src/qibo/models.py
```diff
@@ -429,3 +431,161 @@ def _loss(params, qaoa, hamiltonian):
...
+        sup_qubits (int): number of qubits that store the relevant superposition.
+            Leave empty if superposition does not use ancillas.
+        sup_size (int): how many states are in a superposition.
+            Leave empty if its an equal superposition of quantum states.
+        num_sol (int): number of expected solutions. Needed for normal Grover.
+            Leave empty for iterative version.
+        check (function): function that returns True if the solution has been
+            found. Required of iterative approach.
+            First argument should be the bitstring to check.
+        check_args (tuple): arguments needed for the check function.
```
scarrazza, line 459: These last 2 arguments may be obscure for the user, e.g. what are the function signatures?

src/qibo/models.py
```diff
@@ -429,3 +431,161 @@ def _loss(params, qaoa, hamiltonian):
...
+
+        if superposition:
+            self.superposition = superposition
+        else:
+            if not sup_qubits:
+                raise_error(ValueError)
+                # TODO: Consider a better way for the user to pass superposition qubits
+                # and the number of ancillas
+            self.superposition = Circuit(sup_qubits)
+            self.superposition.add((gates.H(i) for i in range(sup_qubits)))
```
scarrazza, line 474: Do you really need the `()` or you can just use a list?

src/qibo/models.py
```diff
@@ -429,3 +431,161 @@ def _loss(params, qaoa, hamiltonian):
...
+            Leave empty for iterative version.
+        check (function): function that returns True if the solution has been
+            found. Required of iterative approach.
+            First argument should be the bitstring to check.
+        check_args (tuple): arguments needed for the check function.
+            The found bitstring not included.
+    """
+    def __init__(self, oracle, superposition=None, initial_state=None, sup_qubits=None, sup_size=None, num_sol=None, check=None, check_args=()):
+        self.oracle = oracle
+        self.initial_state = initial_state
```
scarrazza, line 464: I think it is probably worth add a dimension check taking the nqubits from the oracle and then checking the length of the initial state.

src/qibo/models.py
```diff
@@ -429,3 +431,161 @@ def _loss(params, qaoa, hamiltonian):
...
+                    if self.check(i[0], *self.check_args):
+                        log.info('Solution checked and successful.')
+                    else:
+                        log.info('Not a solution of the problem. Something went wrong.')
+        else:
+            if not self.check:
+                raise_error(ValueError, "Check function needed for iterative approach.")
+            measured, total_iterations = self.iterative()
+            log.info('Solution found in an iterative process.')
+            log.info(f'Solution: {measured}\n')
```
scarrazza, line 590:
```suggestion
            log.info(f'Solution: {measured}')
```

scarrazza (commented):
@igres26 thank you very much for this great model.
I have some minor comments above, but I assume you are planning to complete the points marked with #TODO. 

Concerning examples and tests, let me suggest to have:
- tests for all the different combination of parameters.
- the jupyter notebook as example (similar to the qPDFs), or if you prefer, a script.

src/qibo/models.py
stavros11: @igres26, following our discussion, if you move these imports inside the class, you have to use `self.np` and `self.gates` in order to access them in the class methods. I had a quick look at the code and there aren't many calls to gates and np so this seems an acceptable solution to me. 

Something additional to keep in mind. If you do:
```Python
class Grover:
    import numpy as np
    from qibo import gates
    ...
```
I don't think there is a big difference with doing global imports in terms of overhead because I believe that Python will still import these libraries if the user does `from qibo import models`. On the other hand if you do:
```Python
class Grover:
    
    def __init__(self, ...):
        import numpy as np
        self.np = np
```
then numpy will be imported only when a Grover model is initialized. In both cases you have to do `self.np` to use numpy later in the model. This note is true for existing models too, not only Grover.

igres26: For now I have followed @stavros11 advice on importing inside the class.

src/qibo/models.py
igres26: I hope the wording is clearer now.

src/qibo/models.py
igres26: These I am not sure how to present them more clearly. Its supposed to be a function that checks the result, and the argument are the one needed for the check function to work.

src/qibo/models.py
igres26: Have reworded `initial_state` into `initial_state_circuit` as I was looking for a quantum circuit that prepares the state.

src/qibo/models/grover.py
```diff
@@ -0,0 +1,194 @@
...
+
+        if superposition_qubits:
+            self.sup_qubits = superposition_qubits
+        else:
+            self.sup_qubits = self.superposition.nqubits
+
+        if superposition_size:
+            self.sup_size = superposition_size
+        else:
+            self.sup_size = int(2 ** self.superposition.nqubits)
```
stavros11, line 61: Is this `if` check required? Are there cases where the user has to pass a superposition size that is different than `2 ** nqubits`?

src/qibo/models/grover.py
```diff
@@ -0,0 +1,194 @@
...
+from qibo import gates
+from qibo.config import log, raise_error
+from qibo.models.circuit import Circuit
+
+
+class Grover(object):
+    """Model that performs Grover's algorithm.
+
+    # TODO: Add a few more details on Grover's algorithm and/or reference
+    for example `arXiv:quant-ph/9605043 <https://arxiv.org/abs/quant-ph/9605043>`_
```
stavros11, line 11: I added an example reference here but can you please add any other paper you think it is relevant, for example for the iterative method. If there is a different paper for the iterative method it can also go to the docstring of the `iterative` method.

src/qibo/models/grover.py
```diff
@@ -0,0 +1,194 @@
...
+                        log.info('Solution checked and successful.')
+                    else:
+                        log.info('Not a solution of the problem. Something went wrong.')
+        else:
+            if not self.check:
+                raise_error(ValueError, "Check function needed for iterative approach.")
+            measured, total_iterations = self.iterative_grover()
+            log.info('Solution found in an iterative process.')
+            log.info(f'Solution: {measured}')
+            log.info(f'Total Grover iterations taken: {total_iterations}')
```
stavros11, line 186: I would probably remove all printing from the model and make sure that all the relevant information related to the execution is returned instead of printed. This is more flexible as it allows the user to print or use this information in any way. Alternatively we can add a flag in the `execute` arguments that chooses whether messages are printed or no.

stavros11 (commented):
I added some tests based on the `grover_playground` notebook. I would suggest to remove this notebook and keep this simple model only for testing. We can use the existing examples (3SAT and/or hash) to demonstrate how the Grover model can be used.

@igres26, if you like to play with the tests you can add some more tests so that all different functionality and parameter configurations are tested. Otherwise let me know and I can also have a look. The goal is to get coverage to 100% so that every line of code is executed at least once by the tests.

src/qibo/models/grover.py
igres26: Yes, although not commonly, one can search over the space of quantum states with, for example, k qubits out of n in state |1>. Although this could be worded differently, as its role is to get the number of Grover iterations to be performed, which technically its 1/amplitude of the solution state. This amplitude si most of the time 1/sqrt(superposition_size), and commonly 1/sqrt(2**superposition_qubits).

src/qibo/models/grover.py
igres26: I think adding a flag that is False by default might be the best compromise.

AdrianPerezSalinas:
Hi all, 

I have been checking the grover model and I have written an example for it. In the corresponding folder `examples/grover` you can see two files. `example_1.py` is copied from what Sergi wrote, and `example_2.py` is mine. In this second file I have written a Grover code for finding all the entries with `num_1` 1's. For instance, if we want no 1's, the only option is `000...000`, and so on. I have checked both the iterative and standard way, and I have seen that the standard way is not working as expected.

@igres26 could you please take a look at that? I'll do it myself and will let you know in case I see something

AdrianPerezSalinas:
Ok, forget what I just said, I had understood the oracle in a bad way with regard to reseting ancillas to 0. It's working fine

igres26:
Things seem to be working on the examples. I would suggest upping the number of samples when the number of solutions is large in the exact Grover, as some non-solutions might appear with small probability.

AdrianPerezSalinas:
I have deleted examples to work on it on another branch, so we can merge this one and leave that for the near future

AdrianPerezSalinas (approved):

AdrianPerezSalinas (changes requested):
I have seen a new feature that can make this model much more applicable. Right now, the superposition circuit may have ancilla qubits, but the oracle is supposed to include all ancillas and qubits and take into account the superposition. I suggest having independent circuits to be fused inside the model. 

There are still discussions to happen for this suggestion between @igres26  and me

scarrazza:
@igres26, concerning pylint, I have fixed that in https://github.com/Quantum-TII/qibo/commit/d1b55d6d74338f3689054023f654bfdc8f9ac8bc (master). Merging master should fix the issue in this and other PRs.

scarrazza (approved):
@igres26 @AdrianPerezSalinas looks very good to me.
I think we can merge asap.

AdrianPerezSalinas:
Nice! There is another branch (grovermodel_ex) with some examples. Should we merge this one too?

scarrazza:
@AdrianPerezSalinas could you please open another PR pointing to this?

AdrianPerezSalinas:
Do we merge that one to main or grovermodel?

scarrazza:
Just open the PR pointing to the `grovermodel` branch.

AdrianPerezSalinas (approved):

AdrianPerezSalinas:
Hi @scarrazza , I am not able to relate `grovermodel` and `grovermodel_ex`, and do not know why. It looks like the `grovermodel` branch is gone, but we have this PR. I do not know what is going on. Can't we open a PR on a branch which is already a PR?

scarrazza:
@AdrianPerezSalinas if you have the code, can you create a new branch and then open a PR?

AdrianPerezSalinas:
Do you mean a new branch with both code and examples?

scarrazza:
> Do you mean a new branch with both code and examples?

not really, take the branch in this PR as baseline, create a new branch and then add the examples.

AdrianPerezSalinas:
Okey now there is a branch called  `examples_grover` pointing to main for PR. I do not know why I cannot merge that one with `grovermodel`. Maybe is this branch deleted?

scarrazza:
@AdrianPerezSalinas thanks, I have restored the branch.

PR #365, opened 2021-03-28
Proposal for new model code structure

stavros11:
Since the number of Qibo models is growing, it may be useful from a code readability perspective to have a folder `models/` instead of `models.py`. The folder can include all the model definitions in separate files and we can make them accessible from `qibo.models` by importing them in `models/__init__.py`. This PR proposes a structure that follows this idea. For example the new Grover model (#364) can be added as a new file in `models/grover.py`.

@scarrazza let me know if you agree with this, otherwise we can leave it as it is. There is absolutely no difference in terms of performance or from the user perspective.


scarrazza (approved):
@stavros11 looks very good and clear, thanks!

igres26:
I agree with @stavros11 that this might be better in an effort to separate the models. Also will be clearer once more models are added to qibo. Should I port the Grover model (#364) into something like this too?

PR #370, opened 2021-04-01
Simplify backend switch

stavros11:
Simplifies the backend switching mechanism by creating a seperate backend class for each of the einsum backends. For example currently we have a single `NumpyBackend` class that handles both `numpy_defaulteinsum` and `numpy_matmuleinsum`. In this PR I seperate the backends by create different classes `NumpyDefaultEinsumBackend` and  `NumpyMatmulEinsumBackend`. I believe this allows a cleaner switching function and may be easier to maintain if we add more backends either for simulation or hardware.

There is no change from the user perspective, that is the user can still switch the backends using `qibo.set_backend("custom")`, `qibo.set_backend("defaulteinsum")`, `qibo.set_backend("numpy_defaulteinsum")`, etc..

### DIFF
--- a/src/qibo/backends/__init__.py
+++ b/src/qibo/backends/__init__.py
@@ -21,8 +21,8 @@ class Backend:
         }
 
         self.constructed_backends = {}
-        self.qnp = self.construct_backend("numpy_defaulteinsum")
         self._active_backend = None
+        self.qnp = self.construct_backend("numpy_defaulteinsum")
         # Create the default active backend
         if "QIBO_BACKEND" in os.environ: # pragma: no cover
             self.initialize_from_env_variable(os.environ.get("QIBO_BACKEND"))
@@ -37,7 +37,6 @@ class Backend:
 
     @active_backend.setter
     def active_backend(self, name):
-        # TODO: Move `construct_backend` here
         self._active_backend = self.construct_backend(name)
 
     def construct_backend(self, name):
@@ -59,7 +58,12 @@ class Backend:
                 raise_error(ValueError, "Unknown backend {}. Please select one of "
                                         "the available backends: {}."
                                         "".format(name, available))
-            self.constructed_backends[name] = self.available_backends.get(name)()
+            new_backend = self.available_backends.get(name)()
+            if self.active_backend is not None:
+                new_backend.set_precision(self.active_backend.precision)
+                if self.active_backend.device is not None:
+                    new_backend.set_device(self.active_backend.default_device)
+            self.constructed_backends[name] = new_backend
         return self.constructed_backends.get(name)
 
     def initialize_from_env_variable(self, name): # pragma: no cover
--- a/src/qibo/backends/__init__.py
+++ b/src/qibo/backends/__init__.py
@@ -1,59 +1,65 @@
 import os
 from qibo import config
 from qibo.config import raise_error, log, warnings
-from qibo.backends.numpy import NumpyBackend
-from qibo.backends.tensorflow import TensorflowBackend
+from qibo.backends.numpy import NumpyDefaultEinsumBackend, NumpyMatmulEinsumBackend
+from qibo.backends.tensorflow import TensorflowCustomBackend, TensorflowDefaultEinsumBackend, TensorflowMatmulEinsumBackend
 
+
+AVAILABLE_BACKENDS = {
+    "custom": TensorflowCustomBackend,
+    "tensorflow": TensorflowCustomBackend,
+    "defaulteinsum": TensorflowDefaultEinsumBackend,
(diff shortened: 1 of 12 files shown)


scarrazza (approved):
Thanks looks good!

src/qibo/backends/__init__.py
```diff
@@ -1,59 +1,71 @@
...
+    "custom": TensorflowCustomBackend,
+    "tensorflow": TensorflowCustomBackend,
+    "defaulteinsum": TensorflowDefaultEinsumBackend,
+    "matmuleinsum": TensorflowMatmulEinsumBackend,
+    "tensorflow_defaulteinsum": TensorflowDefaultEinsumBackend,
+    "tensorflow_matmuleinsum": TensorflowMatmulEinsumBackend,
+    "numpy": NumpyDefaultEinsumBackend,
+    "numpy_defaulteinsum": NumpyDefaultEinsumBackend,
+    "numpy_matmuleinsum": NumpyMatmulEinsumBackend
+}
```
scarrazza, line 18: I am checking this PR, and I think we could be a little bit more specific for this dict.
What do you think if we include some docs so when you raise an error we can show some message like:
```
ValueError: Unknown backend *. Please select one of the available backends: 
- custom: uses precompiled primitives, fastest simulation engine...
- tensorflow: same as custom...
- defaulteinsum: uses tf primitives, and einsum, faster on GPU...
- matmuleinsum: ...
```

scarrazza: Another motivation to do that is the follow: suppose I set `qibo.set_backend("numpy")`, then when calling `qibo.get_backend()` it prints `numpy_defaulteinsum`, so this may confuse the user, and in the worst case break some programmatic check.

src/qibo/backends/tensorflow.py
```diff
@@ -1,6 +1,6 @@
 import os
 from qibo.backends import abstract, numpy
-from qibo.config import raise_error, LOG_LEVEL
+from qibo.config import raise_error, log, LOG_LEVEL
```
scarrazza, line 3: I think we can remove the `log`.

scarrazza (changes requested):
@stavros11 just few comments concerning the layout, otherwise looks good to merge.

src/qibo/backends/__init__.py
```diff
@@ -71,26 +83,12 @@ def set_backend(backend="custom"):
...
-
-    gate_backend = backend.split("_")
-    if len(gate_backend) == 1:
-        calc_backend, gate_backend = _BACKEND_NAME, gate_backend[0]
-    elif len(gate_backend) == 2:
-        calc_backend, gate_backend = gate_backend
-    if gate_backend == "custom":
-        calc_backend = "tensorflow"
-    bk = _construct_backend(calc_backend)
     K.assign(bk)
```
scarrazza, line 90: If we replace this line with `K = bk` the bug in #374 seems to be fixed.

This sounds like a problem with the __init__ global allocation and the `_CONSTRUCTED_BACKENDS`. 
One possibility to avoid this issue consists in always reuse the backends from `_CONSTRUCTED_BACKENDS`.

stavros11: I just pushed some updates in the backend switching mechanism that fixes the device issue and avoids the `.assign` function. I think the idea is close to one of the solutions we were discussing earlier today. I still need to polish several things, fix the tests and your above comments but let me know if you have any comments on this implementation.

scarrazza: Thanks, looks pretty good, and indeed it is quite close to my draft in #374. I think we can keep your approach (without the new method), given that you already expose all required features!

src/qibo/backends/__init__.py
stavros11: I updated the error when an invalid backend is given so that a list of the available backends is printed. You can check the error message by passing a wrong value on purpose (for example do `qibo.set_backend("test")`).

Regarding the second point, I agree that it may be a bit confusing that `numpy` and `numpy_defaulteinsum` are essentially the same backend (same for `custom` and `tensorflow`). The reason I did it this way is because I wanted to enable the user to do `qibo.set_backend("numpy")` as a shortcut for `qibo.set_backend("numpy_defaulteinsum")`. If you think it is confusing we can remove `numpy` and keep only the `numpy_defaulteinsum` option.

src/qibo/backends/tensorflow.py
stavros11: Done.

stavros11:
Thanks for checking. I made some additional updates regarding your previous comments above. If CI passes I think this PR is okay now.

PR #372, opened 2021-04-01
Use single gates.py file for all backends

stavros11:
This simplifies our code base by removing the `cgates.py` file and using a single file to implement gates for all different backends. To do so, I moved the functions that apply gates (`state_vector_call` and `density_matrix_call`) to the backends. As discussed in #363 I also removed the `prepare` and `reprepare` methods from gates as they are not needed anymore.

I tested the multi-GPU performance of this branch and it is the same as master so there is no fix required for that regard. The memory and performance for very deep circuits is also the same with what reported in #363.


src/qibo/abstractions/abstract_gates.py
```diff
@@ -290,6 +290,61 @@ def on_qubits(self, *q):
                     "Cannot use special gates on subroutines.")
 
 
+class Channel(Gate):
+    """Abstract class for channels."""
+
+    def __init__(self):
+        super().__init__()
+        self.gates = tuple()
+        # create inversion gates to rest to the original state vector
```
scarrazza, line 299: rest -> restore/reset?

src/qibo/abstractions/abstract_gates.py
```diff
@@ -290,6 +290,61 @@ def on_qubits(self, *q):
...
+        # because of the in-place updates used in custom operators
+        self._inverse_gates = None
+
+    @property
+    def inverse_gates(self):
+        if self._inverse_gates is None:
+            self._inverse_gates = self.calculate_inverse_gates()
+            for gate in self._inverse_gates:
+                if gate is not None:
```
scarrazza, line 308: Could you please remind me, why we may get None here?

scarrazza (commented):
@stavros11 thanks for this cleanup, the code is much simpler to follow.

PR #376, opened 2021-04-05
[WIP] Benchmarks based on pytest-benchmarks

stavros11:
Adds some basic functionality to perform benchmarks using the [`pytest-benchmark`](https://pytest-benchmark.readthedocs.io/en/latest/index.html) library. Currently only QFT and measurement frequency benchmarks are included, however we can easily extend this with other circuits.

The benchmark configurations should be set according to the used machine in `conftest.py`, by modifying the `NQUBITS` and `NQUBITS_DISTRIBUTED` lists. The benchmarks can be executed using:
```console
pytest examples/benchmarks --benchmark-save=benchmark_name
```

### DIFF
--- a/.github/workflows/rules.yml
+++ b/.github/workflows/rules.yml
@@ -32,11 +32,11 @@ jobs:
         pylint src -E -d E1123,E1120
     - name: Test with pytest core
       run: |
-        pytest --cov=qibo --cov-report=xml --pyargs qibo
+        pytest --cov=qibo --cov-report=xml --pyargs qibo --ignore examples/benchmarks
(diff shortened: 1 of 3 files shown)


scarrazza:
That is quite interesting indeed. If I understand correctly, it retries several times similarly to `%%timeit`.
We could integrate this tool in our CI, however given that we do not have control over the worker setup, this may be not reliable to keep track of the code performance evolution. However, if we store the runs for specific hardware, e.g. dgx or aws instances, we could store the logs in our repo and  before this tool periodically.

PR #377, opened 2021-04-06
Fix CPU custom operators for more than 32 qubits

stavros11:
Based on the answer of [https://stackoverflow.com/questions/31465893/using-unsigned-long-long-as-iteration-range-in-for-loop-using-openmp](https://stackoverflow.com/questions/31465893/using-unsigned-long-long-as-iteration-range-in-for-loop-using-openmp), I think OpenMP doesn't like the `auto` option in parallelized for loops when this refers to an int64 integer.

Here I replace `auto` with `long` and this seems to fix the issue of not applying gates on CPU for more than 32 qubits.

### DIFF
--- a/src/qibo/tensorflow/custom_operators/cc/kernels/apply_gate_kernels.cc
+++ b/src/qibo/tensorflow/custom_operators/cc/kernels/apply_gate_kernels.cc
@@ -37,15 +37,15 @@ struct BaseOneQubitGateFunctor<CPUDevice, T> {
     // Apply gate
     if (ncontrols == 0) {
       #pragma omp parallel for
-      for (long g = 0; g < nstates; g += 1) {
-        long i = ((int64)((int64)g >> m) << (m + 1)) + (g & (tk - 1));
+      for (int64 g = 0; g < nstates; g += 1) {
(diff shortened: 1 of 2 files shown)


stavros11:
By the way, it would be useful to have some extreme case tests that catch issues like these which appear in large systems, however I am not sure if this can be integrated to the CI given the limited hardware resources of these machines. Perhaps we could run the "heavy" tests locally and skip them in CI?

scarrazza (commented):
@stavros11 thanks for this, I confirm that it fixes the issue.

However, given that we are using int64 everywhere, I think we should use int64 instead of long, for consistency. (in principle int64 is an alias of long and equivalent to size_t).

scarrazza:
> By the way, it would be useful to have some extreme case tests that catch issues like these which appear in large systems, however I am not sure if this can be integrated to the CI given the limited hardware resources of these machines. Perhaps we could run the "heavy" tests locally and skip them in CI?

Yes, sounds a good idea. In future we can setup a special runner on AWS.

stavros11:
Thanks for the additional fixes. I replaced all `long` with `int64` as you suggested. I think all OMP loops are now fixed. Initial state uses `size_t` which I believe is also okay.

scarrazza (approved):

PR #378, opened 2021-04-08
Falqon

AdrianPerezSalinas:
FALQON model, trotter multiplication missing


src/qibo/models/variational.py
```diff
@@ -292,3 +292,199 @@ def _loss(params, qaoa, hamiltonian):
...
+        energy = [inf]
+
+        if callback is not None:
+            callback_result = []
+
+        for it in range(1, max_layers + 1):
+            beta = _loss(parameters, self, self.evol_hamiltonian)
+
+            if tol is not None:
+                energy.append(_loss(parameters, self, self.hamiltonian).numpy())
```
stavros11, line 476: Prefer using `np.array` instead of `.numpy()`, because the latter will break if a numpy backend is used. I am aware that this problem exists in existing models (QAOA and VQE) so I will fix it in a different PR (thanks for helping in finding this!).

src/qibo/models/variational.py
```diff
@@ -292,3 +292,199 @@ def _loss(params, qaoa, hamiltonian):
...
+
+    Example:
+        ::
+
+            import numpy as np
+            from qibo import models, hamiltonians
+            # create XXZ Hamiltonian for four qubits
+            hamiltonian = hamiltonians.XXZ(4)
+            # create FALQON model for this Hamiltonian
+            falqon = models.falqon(hamiltonian)
```
stavros11, line 327:
```suggestion
            falqon = models.FALQON(hamiltonian)
```

stavros11 (commented):
Thanks for implementing this. I had a quick look at the code and if I am not missing something, all functions except `minimize` and `__init__` are the same with the `QAOA`. With that in mind, you can avoid code repetition by inheriting the `QAOA` class:

```Python
class FALQON(QAOA):
    
    def minimize(self, ...):
        ...
```
This will automatically put all `QAOA` methods under `FALQON` without the need to re-write them. You can substitute a method by re-defining it with the same name, like I did with `minimize` above.

Regarding `__init__`, if I am correct, it is the same with `QAOA.__init__` with the addition of the `self.evol_hamiltonian` line. The way to do this with inheritance is the following:
```Python
class FALQON(QAOA):
    
    def __init__(self, hamiltonian, mixer=None, solver="exp", callbacks=[],
                 accelerators=None, memory_device="/CPU:0"):
        # call parent class `__init__`
        super().__init__(hamiltonian, mixer, solver, callbacks, accelerators, memory_device)
        # additional code not included in parent class
        self.evol_hamiltonian = 1j * (self.hamiltonian @ self.mixer - self.mixer @ self.hamiltonian)
```

Some additional comments are in code.

AdrianPerezSalinas:
Hi @stavros11 , thanks for your review. I have seen that. However, I wanted to keep things separated just to be clean and do not mix methods. However, if you think that is more convenient, I have no problem with that. I will change it

I have doubts however about how to substitute the `minimize` method in such a way that it works similarly to the one implemented in QAOA (which is mostly scipy-inherited)

stavros11:
> Hi @stavros11 , thanks for your review. I have seen that. However, I wanted to keep things separated just to be clean and do not mix methods. However, if you think that is more convenient, I have no problem with that. I will change it

I think inheritance is fine because it avoids code repetition and easier to maintain (if we find an error we only have to fix one place instead of two). I don't think there will be any mix issues at least for the existing code, we have more complicated inheritance in other places in Qibo already.

> I have doubts however about how to substitute the `minimize` method in such a way that it works similarly to the one implemented in QAOA (which is mostly scipy-inherited)

If you just do `class FALQON(QAOA)`, remove all methods from `FALQON` and leave only the `__init__` I wrote above and the `minimize` you already have it should work as you expect. When you do `def minimize(...)`  you define FALQON's minimize from scratch, as if the QAOA minimize doesn't exist.

AdrianPerezSalinas:
I think this change accomplishes more or less what you said

AdrianPerezSalinas:
Hi @stavros11 
what is the status of this PR? Is there anything I can do to help you to have it ready for tomorrow at the meeting?

stavros11:
> Hi @stavros11
> what is the status of this PR? Is there anything I can do to help you to have it ready for tomorrow at the meeting?

Thanks for checking @AdrianPerezSalinas. I pushed some tests for the FALQON model that fix the coverage complaints. 

My only question to you is regarding the `minimize` function. I see that unlike other models this does not have a `method` argument which is probably because this model always uses the specific optimization you coded following the paper, so the user cannot chose a different method (eg. BFGS, cma, etc.). If that is the case, could you please revise the arguments of `minimize`? For example the `compile` and `processes` arguments are not used in the function body. If that's correct, I would suggest to completely remove them (I added a `TODO` in the code about this).

src/qibo/models/variational.py
```diff
@@ -294,3 +294,96 @@ def _loss(params, qaoa, hamiltonian):
...
+
+        Args:
+            delta_t (float): initial guess for the time step. A too large delta_t will make the algorithm fail.
+            max_layers (int): maximum number of layers allowed for the FALQON.
+            initial_state (np.ndarray): initial state vector of the FALQON.
+            tol (float): Tolerance of energy change. If not specified, no check is done.
+            callback (callable): Called after each iteration for scipy optimizers.
+            options (dict): a dictionary with options for the different optimizers.
+            compile (bool): whether the TensorFlow graph should be compiled.
+            processes (int): number of processes when using the parallel BFGS method.
```
stavros11, line 357: Please update the docstring for the new arguments too.

AdrianPerezSalinas:
You're right, done. 

I wanted to ask something about these arguments. I understand that they are useful in the optimization scenario, but could there be any advantage in compiling this kind of circuits?

scarrazza:
@AdrianPerezSalinas, if I understand this PR is ready, except that we need to the example tests and render the docs in sphinx.

AdrianPerezSalinas:
Yes, exactly. 

I don't know how to check the sphinx documentation and so on, do I just have to read it in the code?

AdrianPerezSalinas:
Hi @scarrazza , do you want to talk for five minutes so we can finish this PR?

scarrazza:
@AdrianPerezSalinas let me have a final review.

scarrazza:
@AdrianPerezSalinas when you parametrize you should provide at least 2 or more parameters.

examples/falqon/main.py
```diff
@@ -0,0 +1,23 @@
...
+    print('The optimal energy found is', best_energy)
+
+    return best_energy, final_parameters
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--nqubits", default=5, type=int)
+    parser.add_argument("--delta_t", default=.1, type=float)
+    parser.add_argument("--max_layers", default=100, type=int)
```
scarrazza, line 21: Please include the `help=` flag with a minimal explanation of the parameters.

examples/falqon/main.py
```diff
@@ -0,0 +1,23 @@
+import argparse
+from qibo import models, hamiltonians
+
```
scarrazza, line 3: Please include 2 empty lines between includes and function definition.

examples/falqon/main.py
```diff
@@ -0,0 +1,23 @@
...
+    return best_energy, final_parameters
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--nqubits", default=5, type=int)
+    parser.add_argument("--delta_t", default=.1, type=float)
+    parser.add_argument("--max_layers", default=100, type=int)
+    args = vars(parser.parse_args())
+    main(**args)
```
scarrazza, line 23: Please add a new line at the end of the file.

scarrazza: @AdrianPerezSalinas if you use vscode, you can set this automatically when saving the file, and for the PEP8 specs (like 2 empty lines) you can do `crtl+shift i` and this will align the code.

examples/test_examples.py
```diff
@@ -246,6 +246,19 @@ def test_shor(N, times, A, semiclassical, enhance):
     run_script(args)
 
 
+@pytest.mark.parametrize("nqubits", 5)
+@pytest.mark.parametrize("delta_t", 0.1)
+@pytest.mark.parametrize("max_layers", 100)
```
scarrazza, line 251: You need more than 1 parameter per parametrize.

src/qibo/models/variational.py
```diff
@@ -294,3 +294,94 @@ def _loss(params, qaoa, hamiltonian):
                                                              compile=compile, processes=processes)
         self.set_parameters(parameters)
         return result, parameters, extra
+
```
scarrazza, line 297: remove extra empty line.

src/qibo/models/variational.py
```diff
@@ -294,3 +294,94 @@ def _loss(params, qaoa, hamiltonian):
...
+            import numpy as np
+            from qibo import models, hamiltonians
+            # create XXZ Hamiltonian for four qubits
+            hamiltonian = hamiltonians.XXZ(4)
+            # create FALQON model for this Hamiltonian
+            falqon = models.FALQON(hamiltonian)
+            # optimize using random initial variational parameters
+            # and default options and initial state
+            delta_t = 0.01
+            best_energy, final_parameters, extra = falqon.minimize(delta_t)
```
scarrazza, line 334: If I run this code, I get:
```
TypeError: minimize() missing 1 required positional argument: 'max_layers'
```

src/qibo/models/variational.py
```diff
@@ -294,3 +294,94 @@ def _loss(params, qaoa, hamiltonian):
...
+    """
+    from qibo.core import states
+
+    def __init__(self, hamiltonian, mixer=None, solver="exp", callbacks=[],
+                 accelerators=None, memory_device="/CPU:0"):
+        # call parent class `__init__`
```
scarrazza, line 340: Please remove unnecessary comments.

src/qibo/models/variational.py
```diff
@@ -294,3 +294,94 @@ def _loss(params, qaoa, hamiltonian):
...
+        super().__init__(hamiltonian, mixer, solver, callbacks, accelerators, memory_device)
+        # additional code not included in parent class
+        self.evol_hamiltonian = 1j * (self.hamiltonian @ self.mixer - self.mixer @ self.hamiltonian)
+
```
scarrazza, line 344: Remove empty line (inside classes the methods are separated by 1 line).

src/qibo/models/variational.py
```diff
@@ -294,3 +294,94 @@ def _loss(params, qaoa, hamiltonian):
...
+            delta_t (float): initial guess for the time step. A too large delta_t will make the algorithm fail.
+            max_layers (int): maximum number of layers allowed for the FALQON.
+            initial_state (np.ndarray): initial state vector of the FALQON.
+            tol (float): Tolerance of energy change. If not specified, no check is done.
+            callback (callable): Called after each iteration for scipy optimizers.
+            options (dict): a dictionary with options for the different optimizers.
+
+        Return:
+            The final energy (expectation value of the ``hamiltonian``).
+            The corresponding best parameters.
```
scarrazza, line 359: You should probably explain what `extra` contains.

scarrazza (commented):
Thanks for this implementation, please find above some minor comments.

AdrianPerezSalinas:
@scarrazza   I have implemented everything. As soon as I know how to generate documentation, I will make the `push`

scarrazza (approved):
@AdrianPerezSalinas thanks!

PR #379, opened 2021-04-08
Fix variational models for numpy backends

stavros11:
Variational model (VQE and QAOA) optimizations will fail if a numpy backend is used because there are some `.numpy()` calls in the loss function which are only supported by Tensorflow. This PR fixes this issue and extends the variational model optimization tests to numpy backends.


scarrazza (commented):
The changes look ok, but from the tests failure, looks like there is something wrong when changing `K.op`.

scarrazza (approved):
Thanks for the fixes, looks good.

PR #383, opened 2021-04-13
renaming cma function

### LINKED ISSUE #381: Optimizers documentation
There is a section in the documentation that I think that might lead to a confusion (as it happened to me). In particular, from [components/optimizers](https://qibo.readthedocs.io/en/stable/qibo.html#qibo.optimizers.cma), one could think that to access the options of the `cma` optimizer it suffices to type 
```python
from qibo.optimizers import cma
cma.CMAOptions()
```
which is not the case. Maybe the code to access the options,
```python
import cma
cma.CMAOptions()
```
could be stated more explicitly so as to avoid misunderstandings.

Also (although I found it clear enough) perhaps a sentence specifying that the user is not supposed to use the `qibo.optimizers.cma`, `qibo.optimizers.newtonian` and `qibo.optimizers.sgd` directly is worth including.

scarrazza:
Closes #381. @DiegoGM91 could you please confirm this solves the ambiguity?

### DIFF
--- a/src/qibo/optimizers.py
+++ b/src/qibo/optimizers.py
@@ -1,3 +1,6 @@
+from qibo.parallel import ParallelResources, _executor
+
+
 def optimize(loss, initial_parameters, args=(), method='Powell',
              jac=None, hess=None, hessp=None, bounds=None, constraints=(),
              tol=None, callback=None, options=None, compile=False, processes=None):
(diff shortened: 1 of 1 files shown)


DiegoGM91 (commented):
@scarrazza this removes the ambiguity, yes. 

But I still think that a sentence in the documentation may be in place, e.g. saying that one first needs to `import cma` to be able to see the `CMAOptions`.

DiegoGM91 (approved):

PR #387, opened 2021-04-16
removing undesired prints

### LINKED ISSUE #386: set_parameters returns prints
**Describe the bug**
I believe this is a bug of the new Qibo version. When you use the `circuit.set_parameters(params)` you receive as well some prints, so if you use it inside a loop you are basically receiving prints from circuit.set_parameters constantly.

`<qibo.backends.numpy.NumpyDefaultEinsumBackend object at 0x7f5042d8d240>`

scarrazza:
Closes #386. I don't think there is some simple way to catch prints from pytest but let me know.

### DIFF
--- a/src/qibo/core/gates.py
+++ b/src/qibo/core/gates.py
@@ -327,16 +327,12 @@ class RZ(MatrixGate, abstract_gates.RZ):
         abstract_gates.RZ.__init__(self, q, theta, trainable)
 
     def construct_unitary(self):
-        print(self.parameters)
         if isinstance(self.parameters, K.native_types): # pragma: no cover
             p = K
             theta = K.cast(self.parameters)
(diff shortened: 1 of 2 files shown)


bpcarlos (approved):
Fixed the undesired prints

stavros11 (approved):
I had a quick look too and did not find any other prints.

> Closes #386. I don't think there is some simple way to catch prints from pytest but let me know.

I am also not aware of any method to do this directly with pytest. A naive solution would be to write a test that loads each `.py` file from the project as string and checks if it contains the word `print`, excluding the cases where `print` appears in docstrings. If you agree I can try to implement this test in a different PR.

scarrazza:
@stavros11 ok, let's do that.

PR #391, opened 2021-04-20
Test that code does not contain `print` statements

stavros11:
Following what was discussed in #387, this adds a test that checks that our code does not contain any undesired `print` statements.


scarrazza (approved):
@stavros11 thanks, it works for me. 
I have played with the code and indeed it seems to detect prints correctly.

PR #394, opened 2021-04-28
backends auto-discovery

scarrazza:
A quick module parsing for the backend activations.
@stavros11 please let me know if you agree.

### DIFF
--- a/src/qibo/backends/__init__.py
+++ b/src/qibo/backends/__init__.py
@@ -23,7 +23,7 @@ class Backend:
             self.available_backends["numpy_defaulteinsum"] = NumpyDefaultEinsumBackend
             self.available_backends["numpy_matmuleinsum"] = NumpyMatmulEinsumBackend
         else:  # pragma: no cover
-            raise_error("Numpy is not installed.")
+            raise_error(ModuleNotFoundError, "Numpy is not installed.")
 
         # check if tensorflow is installed and use it as default backend.
         if _check_availability("tensorflow"):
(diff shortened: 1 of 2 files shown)


src/qibo/backends/__init__.py
```diff
@@ -1,35 +1,63 @@
 import os
+from pkgutil import iter_modules
 from qibo import config
 from qibo.config import raise_error, log, warnings
 from qibo.backends.numpy import NumpyDefaultEinsumBackend, NumpyMatmulEinsumBackend
 from qibo.backends.tensorflow import TensorflowCustomBackend, TensorflowDefaultEinsumBackend, TensorflowMatmulEinsumBackend
 
 
+def _check_availability(module_name):
```
stavros11, line 9: This can also be a `@staticmethod` of the `Backend` class since we do not use it anywhere outside this class.

src/qibo/backends/__init__.py
```diff
@@ -1,35 +1,63 @@
 import os
+from pkgutil import iter_modules
 from qibo import config
 from qibo.config import raise_error, log, warnings
 from qibo.backends.numpy import NumpyDefaultEinsumBackend, NumpyMatmulEinsumBackend
 from qibo.backends.tensorflow import TensorflowCustomBackend, TensorflowDefaultEinsumBackend, TensorflowMatmulEinsumBackend
```
stavros11, line 6: We can could also move these imports under the `if _check_availability("tensorflow")` part below (line 29) to keep the rest of the file clean of backends that are not available.

stavros11 (approved):
Thanks, looks good and works as expected when Tensorflow is not installed. Some optional suggestions.

scarrazza:
@stavros11 thanks, I have refactor the code.

PR #395, opened 2021-04-28
moving distributed circuit into core

scarrazza:
Moving the distributed circuit into core.


stavros11 (approved):
Thanks for moving this and fixing the docs. I pushed an additional fix for the docs as some `qibo.tensorflow.distcircuit` were remaining in `advancedexamples.rst`. Other than that I believe it is good to merge.

scarrazza:
@stavros11 thanks.

PR #398, opened 2021-04-29
Load custom operators from external library

stavros11:
Removes all custom operator code and imports this functionality from the external library `qibo_sim_tensorflow`. CI is not expected to test the custom backend until we make this library available (via pip or somehow else). Once this happens we need to modify the installation instructions accordingly (eg. do `pip install qibo_sim_tensorflow` explicitly).


scarrazza:
@stavros11, qibotf is now in pypi https://pypi.org/project/qibotf.

scarrazza:
@stavros11 in #400 I have implemented some fixes for setup.py so if you prefer we can merge that here asap.

stavros11:
> @stavros11 in #400 I have implemented some fixes for setup.py so if you prefer we can merge that here asap.

Yes, I believe we can merge. I was looking at the matrices pylint issue you mentioned and I will open another PR that simplifies this point, particularly makes matrices backend agnostic, which hopefully will solve the issue.

scarrazza:
In order to complete this PR we need:
- [ ] to cleanup the backends, as discussed the only available backends should be `[qibotf, numpy, tensorflow]` using defaulteinsum.
- [ ] adjust tests accordingly, and if possible bring all tests to the new format.
- [ ] document how to install qibo and qibotf from pip and git.
- [ ] update the documentation explaining the new backends, adding some plots for visual reference (before doing so, I would suggest to review and merge #399.

stavros11:
Just noting that a last piece remaining open for this PR other than the documentation in #411 is the version compatibility check between qibo and qibotf.

scarrazza:
@stavros11 indeed, thanks, I have opened #412 with a simple approach, let me know.

scarrazza (approved):

scarrazza:
@stavros11 shall we merge this?

stavros11:
> @stavros11 shall we merge this?

Yes, I believe it has everything we need for now.

PR #399, opened 2021-04-29
Docs improvements

scarrazza:
Here we include some extra subsections to simplify the navigation from the user point of view.


scarrazza:
@stavros11 I would like to merge this before completing the docs in #398, could you please let me know if you agree?

doc/source/qibo.rst
```diff
@@ -156,9 +181,16 @@ all the gates in the group.
 Time evolution
 ^^^^^^^^^^^^^^
 
+State evolution
+"""""""""""""""
```
stavros11, line 185: I would move the "Time evolution" subsection above the "Circuit addition" and "Circuit fusion" sections because I think it would look better in the side contents list.

Also, since now we list each model as a seperate section in the contents, do you think we should still keep the general / application specific circuit model distinction or shall we put all circuits in the same section and just have one main section for circuit-based computation and one for time evolution?

doc/source/qibo.rst
```diff
@@ -241,23 +501,62 @@ Trotter decomposition. The Hamiltonians represented by this object are sums of
 commuting terms, following the description of Sec. 4.1 of
 `arXiv:1901.05824 <https://arxiv.org/abs/1901.05824>`_.
 
+Trotter hamiltonian
```
stavros11, line 504:
```suggestion
Trotter Hamiltonian
```
I would also move the text right above (including the reference) under the "Trotter Hamiltonian" section as it refers to Trotter Hamiltonians.

doc/source/qibo.rst
```diff
@@ -241,23 +501,62 @@ Trotter decomposition. The Hamiltonians represented by this object are sums of
...
 In addition to these abstract models, Qibo provides the following pre-coded
 Hamiltonians:
 
-.. automodule:: qibo.hamiltonians
-   :members:
-   :member-order: bysource
+.. note::
+    Note that all pre-coded Hamiltonians can be created as either
+    :class:`qibo.abstractions.hamiltonians.Hamiltonian` or
+    :class:`qibo.abstractions.hamiltonians.TrotterHamiltonian` using the ``trotter`` flag.
```
stavros11, line 517: I would move this note above the `TrotterHamiltonian` function documentation or at least above the line "... Qibo provides the following pre-coded Hamiltonians", as this refers to the next section.

stavros11 (approved):
Thanks for writing this, I agree with having a seperate section for each model / gate / Hamiltonian, as this makes it easier for the user to navigate the docs through the side contents. Some small comments below.

Also remember to change the base branch to `simtf` before merging this.

PR #400, opened 2021-04-29
refactoring for qibotf

scarrazza:
Small adjustments to interface with qibotf.


scarrazza:
@stavros11 we have some pylint import errors for windows. I think we can disable this check for the specific files given that we are not importing these libraries, except for the matrices.py import.

stavros11 (approved):
Thanks for the update, this looks good to merge.

> @stavros11 we have some pylint import errors for windows. I think we can disable this check for the specific files given that we are not importing these libraries, except for the matrices.py import.

I think that #402 fixes the matrices.py import issue. For the rest tensorflow imports I believe it should be fine to just do `# pylint: disable`. I am not sure if there is a better solution.

scarrazza:
@stavros11 thanks, I will merge #402 as soon as test finish, and then fix this PR so we can continue in your #398.

scarrazza:
@stavros11 #402 helps but there are several tests relying on `.numpy`, we should probably have a `to_numpy` method in the abstract backend to avoid casting with `K.qnp.dtypes` everywhere.

PR #401, opened 2021-04-30
Reuploading fix

AdrianPerezSalinas:
I have solve the issue with the `saved_parameters.pl` file from the reuploading example. 

What I did is essentially add some code such that:

- It tries to read the file and stores the contain as `data`
- If it is not, then an empty dict is created as `data = {}`.

Then the code looks into `data` for parameters. If it is found, the figures are painted. If it is not, and we do not mind whether it was because the file did not exist or because this calculation was never done, then the optimization is performed. Afterwards it dumps data into the file in any case. 

Let me know if this solves your problem


AdrianPerezSalinas:
@jubaglio let me talk to you in this PR. IN response to https://github.com/Quantum-TII/qibo/issues/397#issuecomment-829922359

I am sorry, if does not work because of my fault, I did not check it properly. 

I have added the following:
```
        try:
            data[dataset][layers] = parameters
        except:
            data[dataset] = {}
            data[dataset][layers] = parameters
```

The problem is that it is not possible to create a dict inside something that does not exist. For this reason we must add a previous step. You were completely right. However, I did it in such a way that it is not necessary to create all the list of dataset, and then we do not have to change it if we add more examples.

examples/reuploading_classifier/main.py
```diff
@@ -31,7 +36,11 @@ def main(dataset, layers):
         print('Problem never solved, finding optimal parameters...')
         result, parameters = ql.minimize(
             method='l-bfgs-b', options={'disp': True})
-        data[dataset][layers] = parameters
+        try:
+            data[dataset][layers] = parameters
+        except:
+            data[dataset] = {}
```
scarrazza, line 42: I think you can avoid this by making sure e.g. in line 30 that `data[dataset]` is not empty / not defined.

AdrianPerezSalinas: Indeed

examples/reuploading_classifier/main.py
```diff
@@ -25,8 +25,8 @@ def main(dataset, layers):
             # Load previous results. Have we ever run these problem?
             data = pickle.load(f)
     except:
-        print('new data')
         data = {}
+        data[dataset] = {}
```
scarrazza, line 29:
```suggestion
        data = {dataset: {}}
```

scarrazza: @AdrianPerezSalinas here a even shorter version, so you can remove the first `data = {}`.

scarrazza (approved):

PR #402, opened 2021-04-30
Make matrices backend agnostic

stavros11:
Simplifies the mechanism that creates `qibo.matrices` so that we don't need a separate matrix class for each backend. Hopefully fixes the issue described in #400.

@scarrazza I opened this PR with master as base because this simplification is independent of qibo_sim but if you prefer I can rebase it for `simtfrename` in order to merge there and check if the pylint issue is fixed.


scarrazza (approved):
@stavros11 thanks looks good and works for me.

PR #403, opened 2021-05-01
fixing tests for numpy/qibotf

scarrazza:
Fixes test issues reported in #400.


src/qibo/backends/abstract.py
```diff
@@ -14,7 +14,7 @@ def __init__(self):
 
         self.cpu_devices = []
         self.gpu_devices = []
-        self.default_device = None
+        self.default_device = []
```
stavros11, line 17: Is this change required? Because for me if `self.default_device` is not set to `None` here a test in `test_backends_init.py` fails sometimes (not always). If I make it `None` everything seems to work.

scarrazza: This change is not strictly necessary, however some tests are using `if 'GPU' in get_device()` so None will break this approach, in particular when installing qibo without qibotf. 

What exactly is failing for you? The CI seems happy with these changes but let me know.

stavros11: I get the following failure:
```Python
test_abstract_circuit_qasm.py:369: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../backends/__init__.py:136: in set_backend
    K.active_backend = backend
../backends/__init__.py:62: in active_backend
    self._active_backend = self.construct_backend(name)
../backends/__init__.py:89: in construct_backend
    new_backend.set_device(self.active_backend.default_device)
../backends/tensorflow.py:50: in set_device
    abstract.AbstractBackend.set_device(self, name)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <qibo.backends.tensorflow.TensorflowCustomBackend object at 0x7f6a80b3e0d0>, name = []

    def set_device(self, name):
>       parts = name[1:].split(":")
E       AttributeError: 'list' object has no attribute 'split'
```
for `test_from_qasm_measurements` in `tests_new/test_abstract_circuit_qasm.py` but strangely it doesn't happen always, only sometimes. I guess it the order the backends are tested changes every time I run?

I think that if we decide to go with the list, then we should change [line 88 of `backends/__init__.py`](https://github.com/Quantum-TII/qibo/blob/463ceebc39597ccc61e9e737b3c435a6f2d78eb6/src/qibo/backends/__init__.py#L89) from
```Python
if self.active_backend.default_device is not None:
```
to
```Python
if self.active_backend.default_device:
```
and this should fix the issue.

scarrazza: Ok, please go ahead and push the option works for you.

scarrazza:
@stavros11 thanks. Let me know if you agree with these changes, so we can start merging towards your original PR.

requirements.txt
```diff
@@ -0,0 +1,9 @@
+# Runtime requirements for qibo
```
stavros11, line 1: I thought the plan was to remove this file and have the dependencies in `setup.py` directly. Do you prefer the current approach that has requirements.txt? I am fine with either way.

stavros11 (approved):
> @stavros11 thanks. Let me know if you agree with these changes, so we can start merging towards your original PR.

Yes, the changes look good. I am currently working on removing the matmuleinsum backend on a branch based on this. I would suggest to start merging to avoid confusion between all these branches. I will open the matmuleinsum PR later today or tomorrow.

requirements.txt
scarrazza: Yes, but then I realized that the test for wheels was not working as expected. So to avoid code duplication I have restore the previous mechanism.

PR #404, opened 2021-05-03
Remove matmuleinsum backend

stavros11:
As discussed in #398 this removes the matmuleinsum backend and renames the existing backends as follows:

* qibotf: The old custom backend which now loads the custom operators from the external qibotf library.
* numpy: The numpy defaulteinsum backend.
* tensorflow: The old defaulteinsum backend which uses the numpy implementation replacing `np.einsum` with `tf.einsum`.


src/qibo/models/variational.py
```diff
@@ -82,14 +83,12 @@ def _loss(params, circuit, hamiltonian):
 
         if method == 'sgd':
             # check if gates are using the MatmulEinsum backend
-            from qibo import K
-            if K.name == "custom":
+            if K.op is not None:
```
scarrazza, line 86: If I understand, K.op is None for tensorflow and numpy, right? Does this work for numpy?

src/qibo/backends/__init__.py
```diff
@@ -11,43 +11,34 @@ def __init__(self):
...
                 log.warning("Einsum will be used to apply gates with Tensorflow. "
                             "Removing custom operators from available backends.")
-                self.available_backends["tensorflow"] = TensorflowDefaultEinsumBackend
-            active_backend = "tensorflow"
         else:  # pragma: no cover
             # case not tested because CI has tf installed
             log.warning("Tensorflow is not installed. Falling back to numpy. "
                         "Numpy does not support Qibo custom operators and GPU. "
                         "Einsum will be used to apply gates on CPU.")
-            # use numpy for defaulteinsum and matmuleinsum backends
```
scarrazza, line 44: Looks good. Let me suggest to update the warning messages with actions that should be taken from the user in order to enable the backend, e.g. where we could say "in order to activate the tensorflow backend please do `pip install tensorflow`, same for `qibotf` and `numpy`.

scarrazza (commented):
@stavros11 thank you, looks pretty good.
Please above minor comments.

src/qibo/backends/__init__.py
stavros11: Thanks for the review. I updated the warnings, please have a second look and let me know if you agree.

src/qibo/models/variational.py
stavros11: Indeed `K.op` is `None` for all backends without custom operators (currently tensorflow and numpy). On the other hand SGD should only work with the `tensorflow` backend. I decided to remove this check completely because if you see in the `sgd` method defined in `optimizers.py` we already check if an SGD-compatible backend is used. I believe it is cleaner to have the check inside the sgd method instead of every model we call sgd.

src/qibo/models/variational.py
```diff
@@ -76,20 +75,15 @@ def _loss(params, circuit, hamiltonian):
...
-            if K.op is not None:
-                raise_error(RuntimeError, 'SGD VQE requires native Tensorflow '
-                                          'gates because gradients are not '
-                                          'supported in the custom kernels.')
-            loss = _loss
         else:
+            loss = _loss
+
+        if method != "sgd":
             loss = lambda p, c, h: K.qnp.dtypes("DTYPE")(_loss(p, c, h))
```
scarrazza, line 85: Ah, I think this can be updated to K.to_numpy...

stavros11:
Thanks for the fixes, I checked using the `test_prints.py` script and I think there aren't any leftover "defaulteinsum" and "matmuleinsum" references in the source code or the examples.

I will continue working on the points you wrote in #398 and particularly bring all the tests in the `tests_new` format.

PR #406, opened 2021-05-05
Adjusts all tests to new format

stavros11:
Fixes the second point from the TODO list in #398 by moving all the remaining old tests from the old `tests` folder to `tests_new`. The `tests_new` is now renamed to `tests` and is the only test folder.


scarrazza (approved):
Thanks, looks good and seems to work fine for me.

PR #407, opened 2021-05-05
Set test backend in conftest

stavros11:
Currently we use the following three lines
```Python
qibo.set_backend(backend)
# ... test code here ...
```
in every test that runs with all different backends. This PR creates a pytest fixture that does this manipulation through the `conftest.py` so that we don't have to include the above lines in every test and avoid code repetition.

### DIFF
--- a/src/qibo/tests/conftest.py
+++ b/src/qibo/tests/conftest.py
@@ -5,9 +5,9 @@ Pytest fixtures.
 """
 import sys
 import pytest
-from qibo import K
+import qibo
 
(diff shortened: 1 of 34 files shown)


scarrazza (approved):
@stavros11 thanks a lot, I was thinking about that when looking at the other pull request.

I have just a comment, I believe there must be a way to drop the `backend` argument of each test and use a decorator instead. However, this might be slightly complicated, so I believe we can merge this PR as it is.

PR #409, opened 2021-05-06
removing zpow and czpow gates

### LINKED ISSUE #396: cleanup gates
I think we can deprecate the ZPow and CZPow gates.

scarrazza:
Closes #396 . Removes ZPow and CZPow in favour of U1 and CU1.


src/qibo/tests_new/test_core_gates.py
```diff
@@ -252,11 +241,11 @@ def test_cun(backend, name, params):
     qibo.set_backend(original_backend)
 
 
-def test_czpow(backend):
+def test_cu1(backend):
```
stavros11, line 244: I think this test is now redudant because `cu1` is tested in `test_cun` which is right above this one anyway.

src/qibo/tests_new/test_core_gates_density_matrix.py
```diff
@@ -68,7 +68,7 @@ def test_rygate_density_matrix(backend):
                          [("H", {}), ("X", {}), ("Y", {}), ("Z", {}), ("I", {}),
                           ("RX", {"theta": 0.123}), ("RY", {"theta": 0.123}),
                           ("RZ", {"theta": 0.123}), ("U1", {"theta": 0.123}),
-                          ("ZPow", {"theta": 0.123}),
+                          ("U1", {"theta": 0.123}),
```
stavros11, line 71: Similarly here we already test for U1 (see right above line) so no need to do this twice.

src/qibo/tests_new/test_core_gates_density_matrix.py
```diff
@@ -108,7 +108,7 @@ def test_controlled_by_one_qubit_gates(backend, gatename):
                          [("CNOT", {}), ("CZ", {}), ("SWAP", {}),
                           ("CRX", {"theta": 0.123}), ("CRY", {"theta": 0.123}),
                           ("CRZ", {"theta": 0.123}), ("CU1", {"theta": 0.123}),
-                          ("CZPow", {"theta": 0.123}),
+                          ("CU1", {"theta": 0.123}),
```
stavros11, line 111: Same here.

PR #411, opened 2021-05-13
documentation upgrade

scarrazza:
Improves the documentation following the discussion in #398.


scarrazza:
@stavros11 as discussed briefly yesterday. I think the current approach to `pip install qibo[backend_name]` may not be sufficient because we can introduce some version mismatch which are not controlled by qibo. 

So I think we should include in the backend `__init__.py` file some checks for versions, and then modify setup.py to read and set this versions in the `extra_requires`.

stavros11:
> So I think we should include in the backend `__init__.py` file some checks for versions, and then modify setup.py to read and set this versions in the `extra_requires`.

If I understand correctly, the only issue that may arise from version mismatch, at least regarding the simulation backends, is if at some point we decide to change some custom operator function signature (eg. we remove / rearrange its arguments). In that case, checking for example the `qibotf` version when we first import it in `qibo` should be sufficient. Perhaps we can have a `QIBOTF_VERSION` variable in `config.py` which is imported in `backends/__init__.py` and the `setup.py`. 

But from the user perspective the `pip install qibo[qibotf]` should still work and install the proper version, right?

scarrazza:
> But from the user perspective the pip install qibo[qibotf] should still work and install the proper version, right?

indeed, following this approach the version will always be correct, but this syntax is probably less known for several users, thus adding tests which prevents mismatches when running `pip install qibotf` should be a good feature.

doc/source/installation.rst
```diff
@@ -1,24 +1,117 @@
...
+specializes the code for quantum simulation on different classical hardware
+configurations, such as multi-threading CPU, single GPU and multi-GPU, and
+similarly for quantum hardware control, from superconducting to ion trap
+technologies including FPGA and AWG devices.
+
+_______________________
+
+.. _backend-drivers:
+
+Backends drivers
```
stavros11, line 37:
```suggestion
Backend drivers
```

doc/source/installation.rst
```diff
@@ -1,24 +1,117 @@
...
+* :ref:`installing-qibo` is the base package for coding and using the API. This package contains all primitives and algorithms for start coding quantum circuits, adiabatic evolution and more (see :ref:`Components`). This package comes with a lightweight quantum simulator which works on multiple CPU architectures such as x86 and arm64.
+
+.. _simulation-backends:
+
+Simulation backends
+^^^^^^^^^^^^^^^^^^^
+
+We provide multiple simulation backends for Qibo, which are automatically loaded if the corresponding packages are installed, following the hierarchy below:
+
+* :ref:`installing-qibotf` (*recommended*): an efficient simulation backend for CPU, GPU and multi-GPU based on TensorFlow primitives. Install this package if you need to simulate quantum circuits with large number of qubits or complex quantum algorithms which may benefit from computing parallelism.
```
stavros11, line 70: I am not sure if "based on TensorFlow primitives" is correct here. I think it would be good to mention that `qibotf` uses custom operators for the multiplication of gates to states and some other operations to distinguish it from the pure `tensorflow` backend which uses only tensorflow primitives.

doc/source/installation.rst
```diff
@@ -1,24 +1,117 @@
...
+
+In the table below we summarize the status of *pre-compiled binaries
+distributed with pypi* for the packages listed above.
+
++------------------+------+------------------+------------+
+| Operating System | qibo | qibotf (cpu/gpu) | tensorflow |
++==================+======+==================+============+
+| Linux x86        | Yes  | Yes/Yes          | Yes        |
++------------------+------+------------------+------------+
+| MacOS >= 10.15   | Yes  | Yes/No           | Yes        |
```
stavros11, line 87: Is qibotf GPU not supported on mac? I know that we probably never tested but is there a serious compatibility constraint with this?

doc/source/installation.rst
```diff
@@ -1,24 +1,117 @@
...
++------------------+------+------------------+------------+
+| Windows          | Yes  | No/No            | No         |
```
stavros11, line 89: Similarly, if it is possible to install tensorflow on windows (which is, right?) it should be possible to use the `tensorflow` bakend, right?

doc/source/installation.rst
```diff
@@ -1,24 +1,117 @@
...
++------------------+------+------------------+------------+
+
+.. note::
+      All packages are supported for Python 3.6, 3.7 and 3.8.
```
stavros11, line 93: Just a reminder to include 3.9 here once we make the relevant qibotf upgrade and include 3.9 in qibo CI.

doc/source/installation.rst
```diff
@@ -105,17 +244,50 @@ or if you prefer to manually execute all installation steps:
...
+      qibo.set_backend("tensorflow")
+
+In order to install the package, we recommend the installation using:
+
+.. code-block:: bash
+
+      pip install qibo[tensorflow]
+
+.. note::
+      The TensorFlow can be installed following its `documentation
```
stavros11, line 273:
```suggestion
      TensorFlow can be installed following its `documentation
```

stavros11 (approved):
Thank you for writing this, looks good and I believe the installation instructions are easy to follow from the user perspective. Some small comments below:

doc/source/installation.rst
scarrazza: I have never tested this explicitly, but in principle should work if you compile from source.
In this table we are quoting the status of the pip package so I believe the information is correct.

doc/source/installation.rst
scarrazza: Indeed, this is a typo, thanks for spotting!

doc/source/installation.rst
scarrazza: Thanks, you are right.

PR #412, opened 2021-05-17
adding version checks for backends

scarrazza:
Implements a simple version check for backends, including the mirroring with setup.py.

### DIFF
--- a/src/qibo/backends/__init__.py
+++ b/src/qibo/backends/__init__.py
@@ -2,6 +2,9 @@ import os
 from qibo import config
 from qibo.config import raise_error, log, warnings
 
+# versions requirements
+TF_MIN_VERSION = '2.2.0'
+
 
 class Backend:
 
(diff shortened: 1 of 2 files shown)


src/qibo/backends/__init__.py
```diff
@@ -2,6 +2,9 @@
 from qibo import config
 from qibo.config import raise_error, log, warnings
 
+# versions requirements
+TF_MIN_VERSION = '2.2.0'
```
stavros11, line 6: We could also move this to config.py.

stavros11 (commented):
Thanks, the check looks good but only checks the tensorflow version. If I understood correctly your comment [here](https://github.com/Quantum-TII/qibo/pull/411#issuecomment-841206778) we would also like to check that the qibotf version is compatible with the qibo version.

scarrazza:
I believe the only strong requirement we have now is the tf version, given that for qibotf we don't have any specific API that requires locking their version.

src/qibo/backends/__init__.py
scarrazza: Any special reason to move that to `config.py`?

stavros11: No, other than that we define all other "constants" used throughout the qibo code there. I don't have any strong preference on what is the right position for `TF_MIN_VERSION` though.

scarrazza:
@stavros11, yes, thanks, I agree.

PR #420, opened 2021-05-24
fixing CI

### LINKED ISSUE #418: parallel tests hangs with tensorflow backend v2.5.0
The `test_parallel.py` is failing when the backend is tensorflow (only v2.5.0).
The numpy and qibotf backends are passing the tests.

@stavros11 could you please confirm this happens for you? This is probably why our [CI](https://github.com/qiboteam/qibo/actions) is failing for linux.

scarrazza:
Fixes #418.

### DIFF
--- a/src/qibo/tests/test_parallel.py
+++ b/src/qibo/tests/test_parallel.py
@@ -12,7 +12,7 @@ from qibo.parallel import parallel_parametrized_execution, parallel_execution
 
 def test_parallel_circuit_evaluation(backend):
     """Evaluate circuit for multiple input states."""
-    if 'GPU' in qibo.get_device() or os.name == 'nt': # pragma: no cover
+    if 'GPU' in qibo.get_device() or os.name == 'nt' or qibo.get_backend() == "tensorflow": # pragma: no cover
         pytest.skip("unsupported configuration")
     original_threads = qibo.get_threads()
     qibo.set_threads(1)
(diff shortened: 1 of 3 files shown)


src/qibo/tests/test_parallel.py
```diff
@@ -12,7 +12,7 @@
 
 def test_parallel_circuit_evaluation(backend):
     """Evaluate circuit for multiple input states."""
-    if 'GPU' in qibo.get_device() or os.name == 'nt': # pragma: no cover
+    if 'GPU' in qibo.get_device() or os.name == 'nt' or qibo.get_backend() == "tensorflow": # pragma: no cover
```
stavros11, line 15: Optionally we could also cover the error cases here using:
```Python
with pytest.raises(RuntimeError):
    ...
```
instead of `pytest.skip`. This way we could remove `# pragma: no cover` from the errors above and include them in the coverage check.

stavros11 (approved):
Looks good to me and CI seems to pass.

src/qibo/tests/test_parallel.py
scarrazza: I was thinking about it, however then we need to wrap the parallel function call in a if/else condition, right? (and duplicate the test basically)

stavros11: Yes, I am not sure if it is the best way but this is what I do in a few other tests where I check for similar errors. For these tests I think it is sufficient to do something like:
```Python
if 'GPU' in qibo.get_device() or os.name == 'nt' or qibo.get_backend() == "tensorflow":
    with pytest.raises(RuntimeError):
        r2 = parallel_execution(c, states=states, processes=2)
else:
    r2 = parallel_execution(c, states=states, processes=2)
    np.testing.assert_allclose(r1, r2)
```
so the code duplication is not that large. Not very necessary to do this though.

scarrazza: Yeah, let's keep as it is, thanks.

PR #421, opened 2021-05-24
Small Grover change

igres26:
Changed the Grover model where the iterative model always outputs something even if it runs for too long of a time.


stavros11 (approved):
Everything looks good, thanks.

PR #422, opened 2021-05-31
icarusq

scarrazza:
Porting icarusq interface to main qibo package.


scarrazza:
@stavros11 could you please have a look at the few coverage issues?

stavros11:
> @stavros11 could you please have a look at the few coverage issues?

Thank you for porting this here. I fixed the coverage by adding some exclusions and some tests, let's see if CI passes now. We should keep in mind that the hardware backend is not tested right now and that we should add some basic tests of it once the qiboicarusq library is available for downloading in the CI. If we merge this PR as it is, perhaps we should open an issue about the missing tests?

scarrazza:
@stavros11 thank you very much, I will merge and open the issue.

Issue #413, opened 2021-05-20
Computing hamiltonians expectation values with the custom backend


AdrianPerezSalinas:
hi all, 

I am trying to compute the expectation value of some Fermi Hubbard hamiltonian where the terms are of this kind
```
1.0 [] +
-0.5 [X0 Z1 X2] +
-0.5 [X0 Z1 Z2 Z3 X4] +
-0.5 [Y0 Z1 Y2] +
-0.5 [Y0 Z1 Z2 Z3 Y4] +
-0.25 [Z0] + ...
```

If I do it with custom backend, I obtain the error message
```
[Qibo|ERROR|2021-05-20 16:26:50]: Unitary gate supports one or two-qubit gates when using custom operators, but 3 target qubits were given. Please switch to a different backend to execute this operation.

NotImplementedError: Unitary gate supports one or two-qubit gates when using custom operators, but 3 target qubits were given. Please switch to a different backend to execute this operation.
```

Does it behave as expected? Should we consider implement it for custom backend?

stavros11:
> Does it behave as expected?

I think yes, because custom operators contain kernels for up to two qubit gates. The trotterization procedure that is coded in qibo just splits the Hamiltonian terms but does not reduce the target qubits of each term, so for example the `[Y0 Z1 Y2]` will be mapped to a three qubit `Unitary` gate which cannot be applied using the qibotf backend. It should work with numpy and tensorflow though but ofcourse will be slower.

> Should we consider implement it for custom backend?

Regarding implementing this, there are probably two ways:
1. Reduce the >2 qubit terms to two qubit gates. This can be done exactly since two qubit gate sets are universal but we do not have any compilation algorithm in qibo to do this automatically yet. Also there may be a performance tradeoff between doing it exactly but slow, or approximately but faster that we may want to consider if we try to implement this approach.
2. Write kernels for >2 qubit gates. The main issue with this solution is that it doesn't scale, for example we can write up to four qubit kernels to accomodate the terms you wrote, but then we won't be able to accomodate a five qubit term, etc.. The reason the custom kernels were up to two qubit is that these are the gates that are more relevant for hardware.

scarrazza:
@AdrianPerezSalinas could you please measure the execution time for your code using the latest master (0.1.6-dev) with `set_backend("numpy")` and `set_backend("tensorflow")`?

scarrazza:
@AdrianPerezSalinas thanks for this numbers.

[...]

PR #424, opened 2021-06-01
Hamiltonian expectation using terms

stavros11:
Following the discussion in #413 this calculates the `hamiltonian.expectation` term-by-term instead of constructing the full matrix. The custom backend can also be used for this calculation even for Hamiltonians with terms acting on more than two qubits because the calculation involves only single qubit gates. 

Here are some benchmarks using a Fermi-Hubbard Hamiltonian provided by @igres26:

nsites | nqubits | qibotf (this branch) | tensorflow (master) | numpy (master)
-- | -- | -- | -- | --
3 | 9 | 0.043816 | 0.053682 | 0.008743
4 | 12 | 0.061650 | 0.094803 | 0.030591
5 | 15 | 0.117778 | 0.200112 | 0.208575
6 | 18 | 0.280394 | 0.840739 | 2.054210
7 | 21 | 3.864714 | 9.843416 | 21.948270
8 | 24 | 29.415665 | 90.591800 | 232.787657

The way this is implemented currently is the following:

* When a `Hamiltonian` is constructed using `Hamiltonian.from_symbolic`, instead of constructing the full matrix directly a dictionary with terms is created. The matrix is calculated only when needed.
* The Hamiltonian-state multiplication is implemented using single qubit `gates.Unitary` gates for each term in the Hamiltonian.

Before finalizing this, I would propose to change the API for defining Hamiltonians and Trotter Hamiltonians with symbols. For example a possible API would be the following:
```Python
from qibo.symbols import X, Y, Z
ham = X(0) * X(1) + Y(0) * Y(1) + Z(0) * Z(1) + ...
```
The advantage of this is:
1. simplified Hamiltonian construction as the user does not to define symbol maps with matrices,
2. Pauli symbols which are the most common for Hamiltonian construction can be mapped to the corresponding gates for which we already have very fast specialized kernels in qibotf. These kernels can be used for expectation value calculation.

### DIFF
--- a/doc/source/advancedexamples.rst
+++ b/doc/source/advancedexamples.rst
@@ -676,7 +676,7 @@ to the state vector. For more information on how solvers work we refer to the
 :ref:`How to simulate time evolution? <timeevol-example>` section.
 When a :class:`qibo.abstractions.hamiltonians.Hamiltonian` is used then solvers will
 exponentiate it using its full matrix. Alternatively, if a
-:class:`qibo.abstractions.hamiltonians.TrotterHamiltonian` is used then solvers
+:class:`qibo.core.hamiltonians.TrotterHamiltonian` is used then solvers
 will fall back to traditional Qibo circuits that perform Trotter steps. For
 more information on how the Trotter decomposition is implemented in Qibo we
 refer to the :ref:`Using Trotter decomposition <trotterdecomp-example>` example.
@@ -1141,7 +1141,7 @@ functionality to perform this transformation automatically, if the underlying
 Hamiltonian object is defined as a sum of commuting parts that consist of terms
 that can be exponentiated efficiently.
 Such Hamiltonian can be implemented in Qibo using
-:class:`qibo.abstractions.hamiltonians.TrotterHamiltonian`.
+:class:`qibo.core.hamiltonians.TrotterHamiltonian`.
 The implementation of Trotter decmoposition is based in Sec.
 4.1 of `arXiv:1901.05824 <https://arxiv.org/abs/1901.05824>`_.
 Below is an example of how to use this object in practice:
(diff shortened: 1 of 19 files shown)


igres26:
Thanks for implementing this. The numbers look great, and the implementation of dedicated symbols for the common X, Y and Z terms is a really good idea.

stavros11:
I added the functionality to create Qibo symbols in this PR in order to simplify the construction of symbolic Hamiltonians. The user can now construct symbolic Hamiltonians as follows:
```Python
from qibo.symbols import X, Y, Z
from qibo.hamiltonians import SymbolicHamiltonian

# Construct Heisenberg model using symbols:
nqubits = 5
symbolic_ham = sum(X(i) * X(i + 1) for i in range(nqubits - 1))
symbolic_ham += sum(Y(i) * Y(i + 1) for i in range(nqubits - 1))
symbolic_ham += sum(Z(i) * Z(i + 1) for i in range(nqubits - 1))
# this is a `sympy` expression

# construct the qibo Hamiltonian object
ham = SymbolicHamiltonian(symbolic_ham) # no symbol map is needed
```
Unlike the standard qibo `Hamiltonian`, the `SymbolicHamiltonian` can be multiplied to states and used to calculate expectation values:
```Python
# multiply to states
new_state = ham @ state
# calculate expectation value
ev = ham.expectation(state)
```
 without creating the full matrix. This works with all backends including qibotf. 

If the user still wants to construct the full matrix this can be done using `ham.dense` (returns the equivalent `Hamiltonian`) or `ham.matrix` (returns the matrix directly).

Note that using `Hamiltonian.from_symbolic` now will construct a `SymbolicHamiltonian` as above. The old `symbol_map` functionality is still supported for compatibility reasons but falls back to the mechanism described above.

@scarrazza, I believe this is okay in terms of code, I would just like to repeat the benchmarks from the first post using the latest version to make sure the numbers still hold before merging.

The `TrotterHamiltonian` is left untouched for now, but I plan to update it in a different PR to allow its construction using Qibo symbols. I think eventually the `TrotterHamiltonian` object will be substituted by the `SymbolicHamiltonian` implemented here by properly implementing the `SymbolicHamiltonian.exp` method to calculate the exponential using the Trotter decomposition.

stavros11:
Here are the benchmarks of the first post repeated to use the latest version of this and Qibo symbols to construct the Fermi-Hubbard Hamiltonian:

nsites | nqubits | qibotf (this branch) | tensorflow (master) | numpy (master)
-- | -- | -- | -- | --
3 | 9 | 0.021513 | 0.048610 | 0.008733
4 | 12 | 0.050942 | 0.087957 | 0.030417
5 | 15 | 0.082954 | 0.213923 | 0.210788
6 | 18 | 0.214952 | 0.846481 | 2.159981
7 | 21 | 2.787341 | 9.452677 | 22.286445
8 | 24 | 23.434635 | 88.888486 | 235.670543

I believe this is good for review.

scarrazza (approved):
@stavros11 thanks looks good to me.

PR #425, opened 2021-06-02
Tensorflow not installed warning

stavros11:
The latest master after merging the hardware backend switcher gives a warning that Tensorflow is not installed even if it is installed. For example the following script
```Python
import qibo
print(qibo.get_backend)
```
prints
```
[Qibo|WARNING|2021-06-02 13:22:24]: Tensorflow is not installed, falling back to numpy. Numpy backend uses `np.einsum` and supports CPU only. To enable GPU acceleration please install Tensorflow with `pip install tensorflow`. To install the optimized Qibo custom operators please use `pip install qibotf` after installing Tensorflow.
qibotf
```
so the warning appears even though the default backend is properly set to `qibotf`. @scarrazza could you please confirm that you also get this warning?

This PR should fix this issue.

### DIFF
--- a/src/qibo/backends/__init__.py
+++ b/src/qibo/backends/__init__.py
@@ -42,13 +42,6 @@ class Backend:
                             "high performance custom operators please use "
                             "`pip install qibotf`.")
 
-        # check if IcarusQ is installed
-        if self.check_availability("qiboicarusq"): # pragma: no cover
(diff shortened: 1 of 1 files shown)


scarrazza (approved):
I confirm, and your changes fix the issue.

PR #426, opened 2021-06-08
Remove TrotterHamiltonian

### LINKED ISSUE #413: Computing hamiltonians expectation values with the custom backend

stavros11:
Removes the `TrotterHamiltonian` object and implements the same functionality using the new `SymbolicHamiltonian`. I believe the new approach is cleaner in terms of code, the basic Hamiltonian operations (addition / subtraction) are more flexible when using Trotter Hamiltonians and also expectation values do not raise errors when qibotf is used for more than two qubit terms (fixes #413 for Trotter Hamiltonians - note that this is already fixed when using the `SymbolicHamiltonian` object or `Hamiltonian.from_symbolic`).

Performance of Trotterization is equivalent to the current for simple state evolution, for example here are the results using the XXZ Hamiltonian:

<details>
<summary> State evolution of the XXZ Hamiltonian </summary>

nqubits | master | this branch
-- | -- | --
8 | 0.073656 | 0.107406
9 | 0.116725 | 0.083405
10 | 0.116073 | 0.099602
11 | 0.106919 | 0.101791
12 | 0.150131 | 0.168620
13 | 0.131640 | 0.139686
14 | 0.178208 | 0.204391
16 | 0.374865 | 0.353153
18 | 1.138154 | 0.911783
20 | 3.306129 | 3.442376
22 | 15.685223 | 15.767390
24 | 72.801820 | 72.644368

</details>

but it is currently slower for adiabatic evolution:

<details>
<summary> Adiabatic evolution using XXZ as problem and X as initial Hamiltonian </summary>

nqubits | master | this branch
-- | -- | --
8 | 0.573128 | 1.734088
9 | 0.517800 | 1.993964
10 | 0.514065 | 2.085160
11 | 0.645079 | 2.437488
12 | 0.662311 | 2.595809
13 | 0.737495 | 2.919230
14 | 0.826153 | 2.818743
16 | 1.100191 | 3.976513
18 | 1.817827 | 5.310719
20 | 4.554980 | 10.354011
22 | 17.522003 | 31.458710
24 | 70.051189 | 125.452080

</details>

This difference is more likely due to the way Hamiltonian addition is handled when constructing the adiabatic evolution Hamiltonian (1 - s(t)) * H0 + s(t) * H1. Although slightly complicated to explain here, master does some "fusion" of terms (eg. the X(0) term is merged with X(0) * X(1) + ... terms) before exponentiating which is not yet implemented here.

TODO:

- [x] Fix term fusion to match master performance in adiabatic evolution but without compromising the flexibility in `SymbolicHamiltonian` addition.
- [x] Update docs.
- [x] Update examples.
- [x] Remove `NumpyHamiltonian` and the `numpy` argument from all Hamiltonian constructors. This was used by the old `TrotterHamiltonian` so it is no longer needed as I don't think it's useful for the user. There is always the option to switch the backend to numpy anyway.
- [x] Optional: Keep a `TrotterHamiltonian` (including `TrotterHamiltonian.from_symbolic`) object that falls back to `SymbolicHamiltonian` to avoid breaking existing code.


stavros11:
I believe this is now good for review. Note that it implements a few API changes that may break existing code:

* The `TrotterHamiltonian` object is completely removed. Qibo now has two types of Hamiltonians:
    - The standard `Hamiltonian` object which uses the full matrix representation (straightforward but memory inefficient).
    - The `SymbolicHamiltonian` which can be created using the new `qibo.symbols` implemented in #424. This contains a more efficient term-by-term reprenentation based on the new object `qibo.core.terms.HamiltonianTerm` (this object is not exposed to user). When this Hamiltonian is used to calculate expectation values, they will be calculated using term-by-term sum (not constructing the full matrix) and when this Hamiltonian is exponentiated for time evolution then the Trotter decomposition will be used as we had before.

* The `trotter=False` flag in pre-coded Hamiltonian models (TFIM, XXZ, etc.) is replaced by the `dense=True` flag. If `dense=True` a `Hamiltonian` will be created, otherwise a `SymbolicHamiltonian`. The reason I removed `trotter` from these names is because the `SymbolicHamiltonian` is useful not only for time evolution/Trotter but also for calculating expectation values without the full matrix, so I believe `dense` is less confusing than `trotter` for the user.

* Removed the `numpy=False` flag from pre-coded Hamiltonians and the `NumpyHamiltonian` all over. This is no longer useful as the terms of the `SymbolicHamiltonian` rare handled by the new classes in `qibo.core.terms` and not by creating smaller `NumpyHamiltonian` objects.

In terms of performance, I repeated the benchmarks of the first post after the latest pushes with the "term fusion". Overally it looks okay, although for adiabatic evolution there is some overhead for small number of qubits.

<details>
<summary>State evolution of the XXZ Hamiltonian</summary>

nqubits | master | this branch
-- | -- | --
3 | 0.02075 | 0.02121
4 | 0.02662 | 0.02623
5 | 0.03282 | 0.03284
6 | 0.04037 | 0.04032
7 | 0.04619 | 0.04532
8 | 0.05212 | 0.05306
9 | 0.05717 | 0.05637
10 | 0.06272 | 0.06496
11 | 0.06998 | 0.06994
12 | 0.07703 | 0.07870
13 | 0.08666 | 0.08898
14 | 0.09521 | 0.09953
16 | 0.15770 | 0.16420
18 | 0.30743 | 0.28132
20 | 0.81168 | 0.80415
22 | 2.90282 | 2.34681
24 | 16.75837 | 12.71948
26 | 94.35569 | 94.69983
28 | 409.50615 | 410.97905

</details>

<details>
<summary>Adiabatic evolution using XXZ as problem and X as initial Hamiltonian</summary>

nqubits | master | this branch
-- | -- | --
3 | 0.10067 | 0.23183
4 | 0.12943 | 0.30897
5 | 0.15415 | 0.38251
6 | 0.18480 | 0.45640
7 | 0.21234 | 0.53682
8 | 0.24353 | 0.60803
9 | 0.26677 | 0.68971
10 | 0.29086 | 0.78475
11 | 0.32021 | 1.18974
12 | 0.35343 | 0.92855
13 | 0.37937 | 1.01895
14 | 0.41986 | 1.08080
16 | 0.52311 | 1.31599
16 | 0.52311 | 1.28847
16 | 0.52647 | 1.31599
16 | 0.52647 | 1.28847
18 | 0.78431 | 1.62515
20 | 1.51273 | 2.85305
22 | 4.59003 | 5.90768
24 | 19.20760 | 19.47585
26 | 95.64385 | 96.34195
28 | 408.32744 | 412.96284

</details>

doc/source/advancedexamples.rst
```diff
@@ -1359,12 +1344,12 @@ How to define custom Hamiltonians using symbols?
...
 
 In order to use the VQE, QAOA and time evolution models in Qibo the user has to
-define Hamiltonians based on :class:`qibo.abstractions.hamiltonians.AbstractHamiltonian`, or
-:class:`qibo.core.hamiltonians.TrotterHamiltonian` when the Trotter
-decomposition is to be used. Qibo provides pre-coded Hamiltonians for some
-common physics models, such as the transverse field Ising model (TFIM) and the
-Heisenberg model (see :ref:`Hamiltonians <Hamiltonians>` for a complete list
-of the pre-coded models).
+define Hamiltonians based on :class:`qibo.core.hamiltonians.Hamiltonian` which
+uses the full matrix represntation of the corresponding operator or
```
scarrazza, line 1348:
```suggestion
uses the full matrix representation of the corresponding operator or
```

doc/source/advancedexamples.rst
```diff
@@ -1141,7 +1141,7 @@ functionality to perform this transformation automatically, if the underlying
 Hamiltonian object is defined as a sum of commuting parts that consist of terms
 that can be exponentiated efficiently.
 Such Hamiltonian can be implemented in Qibo using
-:class:`qibo.core.hamiltonians.TrotterHamiltonian`.
+:class:`qibo.core.hamiltonians.SymbolicHamiltonian`.
 The implementation of Trotter decmoposition is based in Sec.
```
scarrazza, line 1145:
```suggestion
The implementation of Trotter decomposition is based in Sec.
```

examples/benchmarks/qaoa.py
```diff
@@ -10,20 +10,20 @@
...
-def main(nqubits, nangles, trotter=False, solver="exp",
+def main(nqubits, nangles, dense=True, solver="exp",
          method="Powell", maxiter=None):
     """Performs a QAOA minimization test."""
 
     print("Number of qubits:", nqubits)
     print("Number of angles:", nangles)
 
-    hamiltonian = hamiltonians.XXZ(nqubits=nqubits, trotter=trotter)
+    hamiltonian = hamiltonians.XXZ(nqubits=nqubits, dense=dense)
```
scarrazza, line 26: I am trying to read the docs from the user perspective and the usage of the dense flag and trotterization, e.g. what happens if we sent dense to True or False and trotterization, etc. We should include a comment about that in the advanced examples "Using Trotter decomposition", explaining the advantage of dense=False and the interoperability with dense=True.

src/qibo/core/adiabatic.py
```diff
@@ -0,0 +1,107 @@
...
+        self.groups = terms.TermGroup.from_terms(all_terms)
+
+    def circuit(self, dt, accelerators=None, memory_device="/CPU:0", t=0):
+        if self.trotter_circuit is None:
+            self.trotter_circuit = TrotterCircuit(self.groups, dt, self.nqubits,
+                                                  accelerators, memory_device)
+        st = self.schedule(t / self.total_time) if t != 0 else 0 # pylint: disable=E1102
+        coefficients = {self.h0: 1 - st, self.h1: st}
+        self.trotter_circuit.set(dt, coefficients)
+        return self.trotter_circuit.circuit
```
scarrazza, line 107: Could you please include some docstrings for the methods of the 4 classes above?

src/qibo/core/terms.py
```diff
@@ -0,0 +1,273 @@
...
+    """Single term in a :class:`qibo.core.hamiltonians.SymbolicHamiltonian`.
+
+    This Hamiltonian is represented by a sum of such terms which are held in
+    its ``.terms`` attribute which is a list.
+
+    Args:
+        matrix (np.ndarray): Full matrix corresponding to the term representation
+            in the computational basis. Has size (2^n, 2^n) where n is the
+            number of target qubits of this term.
+        q (int): Target qubit ids.
```
scarrazza, line 16: This is a list of ints, right?

src/qibo/core/terms.py
```diff
@@ -0,0 +1,273 @@
...
+    """
+
+    def __init__(self, matrix, *q):
+        self.target_qubits = tuple(q)
```
scarrazza, line 20: Maybe include some test for qubits positivenes and matching with matrix shape.

src/qibo/core/terms.py
```diff
@@ -0,0 +1,273 @@
...
+    @property
+    def gate(self):
+        """Qibo gate that implements the action of the term on states."""
+        if self._gate is None:
+            self._gate = gates.Unitary(self.matrix, *self.target_qubits)
+        return self._gate
+
+    def exp(self, dt):
+        """Matrix exponentiation of the term."""
+        return K.qnp.expm(-1j * dt * self.matrix)
```
scarrazza, line 42: Is this correct? Or should we drop the `qnp`?

src/qibo/core/terms.py
```diff
@@ -0,0 +1,273 @@
...
+        return self._matrix
+
+    @property
+    def gate(self):
+        """Qibo gate that implements the action of the term on states."""
+        if self._gate is None:
+            self._gate = gates.Unitary(self.matrix, *self.target_qubits)
+        return self._gate
+
+    def exp(self, dt):
```
scarrazza, line 40: Maybe add some check for dt > 0.

src/qibo/core/terms.py
```diff
@@ -0,0 +1,273 @@
...
+        """Matrix exponentiation of the term."""
+        return K.qnp.expm(-1j * dt * self.matrix)
+
+    def expgate(self, dt):
```
scarrazza, line 44: Maybe add some check for dt > 0.

src/qibo/core/terms.py
```diff
@@ -0,0 +1,273 @@
...
+        """Unitary gate implementing the matrix exponentiation of the term."""
+        return gates.Unitary(self.exp(dt), *self.target_qubits)
+
+    def merge(self, term):
```
scarrazza, line 48: Maybe add some check for tems size and content.

src/qibo/core/terms.py
```diff
@@ -0,0 +1,273 @@
...
+        """Creates a new term by merging the given term to the current one.
+
+        The target qubits of the given term should be a subset of the target
+        qubits of the current term.
+        """
+        matrix = K.np.kron(term.matrix, K.eye(2 ** (len(self) - len(term))))
+        matrix = K.np.reshape(matrix, 2 * len(self) * (2,))
```
scarrazza, line 55: I must admit that that K.np, K.qnp or simply K can be misleading. Do you still need this triple distinction?
Ideally with cupy/numpy we could use directly K, I believe.

src/qibo/core/terms.py
```diff
@@ -0,0 +1,273 @@
...
+        return self.gate(state) # pylint: disable=E1102
+
+
+class SymbolicTerm(HamiltonianTerm):
+    """:class:`qibo.core.terms.HamiltonianTerm` constructed from ``sympy`` symbols.
+
+    Example:
+        ::
+
+            from qibo.symbols import X, Y, SymbolicTerm
```
scarrazza, line 92: I think the `SymbolicTerm` is not exposed by `qibo.symbols`.

src/qibo/core/terms.py
scarrazza: Btw, we are doing `self.qnp = self.construct_backend("numpy")` so most likely we could replace that with `K.np`.

scarrazza (approved):
@stavros11 thank you very much for this implementation and cleanup.
Please find above some comments.

src/qibo/core/terms.py
stavros11: > I must admit that that K.np, K.qnp or simply K can be misleading. Do you still need this triple distinction?

I think the distinction between K and K.np is still useful because in some cases we prefer to use numpy instead of the backend. This was first introduced for the tensorflow backend because we noticed that numpy is faster than tensorflow in performing small matrix operations. In such cases we use K.np to force use of numpy regardless of the activated backend. These operations are for example gate multiplications of small (2x2 or 4x4 matrices) in gate fusion, or the exponentiation of small matrices for Trotterized evolution which is used in the current PR. I have not tested but it is likely that numpy is still faster than numba/cupy for such small operations so performance-wise having K.np should be helpful.

Regarding K.np vs K.qnp, the difference is that the first is the actual numpy while the second is qibo's `NumpyBackend`. There are small API differences between the two, but I am not sure if these are enough for this distinction. I can have a look over the whole qibo code and perhaps we could remove one of the two. For these specific lines you note both K.qnp and K.np would work the same.

src/qibo/core/terms.py
stavros11: That is correct, this is a leftover from a previous version. The `SymbolicTerm` is now only accessed via `qibo.core.terms` and is not exposed to the user.

src/qibo/core/terms.py
stavros11: Check my comment below regarding the K vs K.qnp distinction. I think leaving K.qnp here would generally be faster because we are exponentiating small matrices (2x2 or 4x4).

Also this is an example case where K.qnp has different behavior from K.np. K.np would not work here because numpy does not provide an `expm` method and we have to use the one from scipy (which is available via `K.qnp`).

src/qibo/core/terms.py
stavros11: We could add that but in principle it is okay if one calls this method with dt < 0 as all it does is calculating exp(-i dt H) which is a valid unitary for all dt. Perhaps the name dt is not the best here, perhaps we could switch it to an arbitrary x?

src/qibo/core/adiabatic.py
stavros11: Done.

src/qibo/core/terms.py
stavros11: From the code side `q` is interpreted as a list of ints, but in terms of API the ints are just passed as seperate arguments because `*q` is used in definition, eg.:
```Python
HamiltonianTerm(matrix, 0, 1, 2, 3)
```
So I am not sure what is the proper way of implementing this. It is the same API with gates with multiple targets (eg. the `gates.Unitary` or `gates.I`).

scarrazza: I think you can safely say list of ints.

src/qibo/core/terms.py
scarrazza: Ok, fair enough.

src/qibo/core/terms.py
scarrazza: Ok, I agree. Replacing with `x` is fine.

examples/benchmarks/qaoa.py
stavros11: We already had a note regarding the `dense` flag in the Hamiltonians section, so I extended this and I added a reference to this in the "Using Trotter decomposition example". Please have a look and let me know if you agree.

stavros11:
> @stavros11 thank you very much for this implementation and cleanup.
> Please find above some comments.

Thank you for the review. I updated according to your comments, including adding the checs, changing `dt` to `x` and explaining the `dense` a bit more in the docs. I left the `K.np`, `K.qnp` as they are, if we would like to change this at some point I think it is better to do in a different PR as it will affect other parts of code too, not only Hamiltonians.

doc/source/advancedexamples.rst
```diff
@@ -1164,8 +1164,12 @@ exponentials of the Trotter decomposition and can be executed on any state.
 Note that in the transverse field Ising model (TFIM) that was used in this
 example is among the pre-coded Hamiltonians in Qibo and could be created as
 a :class:`qibo.core.hamiltonians.SymbolicHamiltonian` simply using the
-``dense=False`` flag. Defining custom Hamiltonians from terms can be more
-complicated, however Qibo simplifies this process by providing the option
+``dense=False`` flag. For more information on the difference between dense
+and non-dense Hamiltonians we refer to the :ref:`Hamiltonians <Hamiltonians>` 
+section.
```
scarrazza, line 1169: Looks good, however I still believe someone will ask "what happens if dense if True and I do trotterization"? Maybe include a sentence about this here.

stavros11: Oh I see. The answer to this is that if `dense=True` then no trotterization will be used because the full Hamiltonian matrix is available so Qibo will directly exponentiate this in order to perform evolution. I added an additional sentence to this paragraph explaining this.

Do you think we should allow trotterization using dense (full matrix) Hamiltonians? Generally the idea is that one can construct the SymbolicHamiltonian which has a more compact term-by-term representation and use this for memory efficient expectation values and Trotter evolution and in case the full matrix is needed it is still possible to do `SymbolicHamiltonian.dense` to get the dense for or `SymbolicHamiltonian.matrix` to get the 2^n x 2^n matrix directly.

scarrazza: Indeed, I think the current logics is fine.

stavros11:
As we discussed, I added a `TrotterHamiltonian` object in Hamiltonians which raises an error when constructed.

Furthermore, @igres26 pointed out to me some problems in the documentation of the solvers and how/when trotterization is used. So I decided to remove the "Solvers" section from the documentation, since Qibo's solver objects are not directly exposed to user and expanded the docstring of the `solver` argument in evolution objects. @igres26, @scarrazza let me know if you agree with the new documentation.

scarrazza (approved):
Looks good to me, thanks.

stavros11:
I fixed the conflicts with master after the qibojit merge. Tests should be passing on GPU with both cupy and tensorflow too. @scarrazza you can have a second look if you like.

scarrazza:
@stavros11 thank you. On GPU, I am getting 1 failure:

<details>

```
=============================================================== FAILURES ================================================================
________________________________________ test_adiabatic3sat[BFGS-1-params0-False-rk4-4-1-10-0.1] ________________________________________

nqubits = 4, instance = 1, T = 10, dt = 0.1, solver = 'rk4', dense = False, params = [0.5, 0.5, 10, 10, 10, 10], method = 'BFGS'
maxiter = 1, plot = False

    @pytest.mark.parametrize("nqubits,instance,T,dt", [(4, 1, 10, 1e-1)])
    @pytest.mark.parametrize("solver", ["exp", "rk4"])
    @pytest.mark.parametrize("dense", [True, False])
    @pytest.mark.parametrize("params", [[0.5, 0.5]])
    @pytest.mark.parametrize("method,maxiter", [("BFGS", 1)])
    def test_adiabatic3sat(nqubits, instance, T, dt, solver, dense, params,
                           method, maxiter, plot=False):
        if "functions" in sys.modules:
            del sys.modules["functions"]
        args = locals()
        path = os.path.join(base_dir, "adiabatic3sat")
        sys.path[-1] = path
        os.chdir(path)
>       run_script(args)

/home/carrazza/repo/qiboteam/qibo/examples/test_examples.py:222: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/home/carrazza/repo/qiboteam/qibo/examples/test_examples.py:54: in run_script
    exec(code, {"args": args})
<string>:113: in <module>
    ???
<string>:88: in main
    ???
/home/carrazza/repo/qiboteam/qibo/src/qibo/models/evolution.py:303: in minimize
    result, parameters, extra = optimizers.optimize(
/home/carrazza/repo/qiboteam/qibo/src/qibo/optimizers.py:73: in optimize
    return newtonian(loss, initial_parameters, args, method,
/home/carrazza/repo/qiboteam/qibo/src/qibo/optimizers.py:141: in newtonian
    m = minimize(loss, initial_parameters, args=args, method=method,
/home/carrazza/anaconda3/lib/python3.8/site-packages/scipy/optimize/_minimize.py:614: in minimize
    return _minimize_bfgs(fun, x0, args, jac, callback, **options)
/home/carrazza/anaconda3/lib/python3.8/site-packages/scipy/optimize/optimize.py:1135: in _minimize_bfgs
    sf = _prepare_scalar_function(fun, x0, jac, args=args, epsilon=eps,
/home/carrazza/anaconda3/lib/python3.8/site-packages/scipy/optimize/optimize.py:261: in _prepare_scalar_function
    sf = ScalarFunction(fun, x0, args, grad, hess,
/home/carrazza/anaconda3/lib/python3.8/site-packages/scipy/optimize/_differentiable_functions.py:136: in __init__
    self._update_fun()
/home/carrazza/anaconda3/lib/python3.8/site-packages/scipy/optimize/_differentiable_functions.py:226: in _update_fun
    self._update_fun_impl()
/home/carrazza/anaconda3/lib/python3.8/site-packages/scipy/optimize/_differentiable_functions.py:133: in update_fun
    self.f = fun_wrapped(self.x)
/home/carrazza/anaconda3/lib/python3.8/site-packages/scipy/optimize/_differentiable_functions.py:130: in fun_wrapped
    return fun(x, *args)
/home/carrazza/repo/qiboteam/qibo/src/qibo/models/evolution.py:300: in <lambda>
    loss = lambda p, ae, h1, msg, hist: K.to_numpy(self._loss(p, ae, h1, msg, hist))
/home/carrazza/repo/qiboteam/qibo/src/qibo/models/evolution.py:271: in _loss
    final_state = super(AdiabaticEvolution, adiabatic_evolution).execute(params[-1])
/home/carrazza/repo/qiboteam/qibo/src/qibo/models/evolution.py:126: in execute
    state = self.solver(state)
/home/carrazza/repo/qiboteam/qibo/src/qibo/solvers.py:96: in __call__
    k1 = ham1 @ state
/home/carrazza/repo/qiboteam/qibo/src/qibo/core/hamiltonians.py:391: in __matmul__
    return self.apply_gates(o)
/home/carrazza/repo/qiboteam/qibo/src/qibo/core/hamiltonians.py:369: in apply_gates
    total += term(K.copy(state), density_matrix)
/home/carrazza/repo/qiboteam/qibo/src/qibo/core/terms.py:207: in __call__
    state = factor.gate(state)
/home/carrazza/repo/qiboteam/qibo/src/qibo/abstractions/abstract_gates.py:549: in __call__
    return getattr(self, self._active_call)(state)
/home/carrazza/repo/qiboteam/qibo/src/qibo/core/gates.py:69: in state_vector_call
    return K.state_vector_call(self, state)
/home/carrazza/repo/qiboteam/qibo/src/qibo/backends/numpy.py:552: in state_vector_call
    return gate.gate_op(state, gate.nqubits, *gate.target_qubits,
/home/carrazza/repo/qiboteam/qibojit/src/qibojit/custom_operators/__init__.py:59: in apply_x
    return backend.one_qubit_base(state, nqubits, target, "apply_x", qubits)
/home/carrazza/repo/qiboteam/qibojit/src/qibojit/custom_operators/backends.py:196: in one_qubit_base
    ktype = self.get_kernel_type(state)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <qibojit.custom_operators.backends.CupyBackend object at 0x7f5717a1b9d0>
state = array([0.25, 0.25, 0.25, 0.25, 0.25, 0.25, 0.25, 0.25, 0.25, 0.25, 0.25,
       0.25, 0.25, 0.25, 0.25, 0.25])

    def get_kernel_type(self, state):
        if state.dtype == self.cp.complex128:
            return self.kernel_double_suffix
        elif state.dtype == self.cp.complex64:
            return self.kernel_float_suffix
>       raise TypeError("State of invalid type {}.".format(state.dtype))
E       TypeError: State of invalid type float64.

/home/carrazza/repo/qiboteam/qibojit/src/qibojit/custom_operators/backends.py:182: TypeError
----------------------------------------------------------- Captured log call -----------------------------------------------------------
INFO     qibo.config:evolution.py:87 Normalizing state during RK solution.

```
</details>

Seems like from some ground_state used in adiabatic3sat is not complex.

scarrazza:
@stavros11 thanks, this fixes the issue.

PR #427, opened 2021-06-10
Add qibojit backend

stavros11:
Adds new backend that loads custom operators from the qibojit library. The new backend is not tested in CI until the qibojit library is released.


src/qibo/backends/numpy.py
```diff
@@ -370,3 +370,140 @@ def density_matrix_collapse(self, gate, state, result):
         state = substate / norm
         state = self._append_zeros(state, sorted_qubits, density_matrix_result)
         return self.reshape(state, gate.cache.flat_shape)
+
+
+class NumpyCustomBackend(NumpyBackend): # pragma: no cover
```
scarrazza, line 375: As discussed today, I would suggest to rename this class to `JITCustomBackend`.

src/qibo/backends/numpy.py
```diff
@@ -370,3 +370,140 @@ def density_matrix_collapse(self, gate, state, result):
...
+    def __init__(self):
+        from qibo.backends import Backend
+        if not Backend.check_availability("qibojit"): # pragma: no cover
+            # CI can compile custom operators so this case is not tested
+            raise_error(RuntimeError, "Cannot initialize qibojit "
+                                      "if the qibojit is not installed.")
+        from qibojit import custom_operators as op # pylint: disable=E0401
+        super().__init__()
+        self.name = "qibojit"
+        self.op = op
```
scarrazza, line 386: Associate the `self.backend` to the appropriate library numpy/cupy.

src/qibo/backends/tensorflow.py
```diff
@@ -290,11 +290,18 @@ def density_matrix_half_matrix_call(self, gate, state):
                             2 * gate.nqubits, *gate.target_qubits,
                             self.get_threads())
 
+    def _result_tensor(self, result):
+        n = len(result)
+        result = sum(2 ** (n - i - 1) * r for i, r in enumerate(result))
+        return self.cast(result, dtype="DTYPEINT")
```
scarrazza, line 296: Does this operation is strictly necessary?

src/qibo/core/circuit.py
```diff
@@ -128,7 +128,7 @@ def compile(self):
             raise_error(RuntimeError, "Circuit is already compiled.")
         if not self.queue:
             raise_error(RuntimeError, "Cannot compile circuit without gates.")
-        if K.op is not None:
+        if K.name == "qibotf":
```
scarrazza, line 131: What do you think if we keep the `return func` in the qibotf backend class and remove this if condition?

src/qibo/backends/tensorflow.py
stavros11: I think yes, because the custom operators for collapse expect the result to be integer (in decimal form) while the other backends use the array form which represents the same number in binary. Note that this is a new addition here, we already do exactly the same conversion inside `core.gates.M` (see [here](https://github.com/qiboteam/qibo/blob/b706ed3c014f139ff43794bef03db9e464f28774/src/qibo/core/gates.py#L250) in master). I just moved the conversion at the backend level in this PR because it is slightly different for qibojit and qibotf so it is easier to implement this way.

There may be a way to avoid this conversion by using decimal as the default representation for `result` and convert to binary when using the numpy/tensorflow backends.

stavros11:
Thanks for this first review, I fixed the three points you wrote above. I'm currently testing the cupy part of the new backend and it seems to require a few changes. I will probably open a different PR based on this for this update.

scarrazza:
@stavros11 I believe this is almost ready to go, except for:
- moving qibojit to default instead of tensorflow
- adding the appropriate docs concerning the package.

Could you please take care of the first?

stavros11:
> Could you please take care of the first?

Sure, but do we want `qibojit` to be the default if available? If I understand correctly the backend order will be the following:
1. If everything is available, default is `qibojit`.
2. If `qibojit` is not available default falls back to `qibotf`.
3. If `qibojit` and `qibotf` not available it falls back to `tensorflow`.
4. If `tensorflow` is not available it falls back to `numpy`.

Do we need any warnings to appear for missing libraries? Currently a warning appears prompting to install `qibotf` if this is not available.

scarrazza:
@stavros11 yes, perfect.

> Do we need any warnings to appear for missing libraries? Currently a warning appears prompting to install qibotf if this is not available.

More I think, I believe we should have a minimal sentence, like in keras (Using ... backend) to make clear what is going on. What do you think? The only issue is that ideally this printing must be controlled by the logger, and eventually turned off (similarly to what we already do with tensorflow).

scarrazza (approved):

PR #428, opened 2021-06-10
Set threshold for using Metropolis custom op

stavros11:
Following a discussion with @igres26, this implements a getter/setter for the threshold value beyond which measuring shot frequencies will fall to the Metropolis custom operator.

The reason we need this is that the current Metropolis implementation does not seem to work very well for specific (usually sparse) distributions. Of course the proper solution would be to improve the custom operator itself, however the threshold setter is a good temporary solution that allows the user to disable the custom operator if its results are not good, by setting a very high threshold (eg. 10^12). The default tensorflow/numpy implementation is quite fast for up to 10^7 shots anyway, which is sufficient for many applications.

### DIFF
--- a/src/qibo/config.py
+++ b/src/qibo/config.py
@@ -78,6 +78,7 @@ def get_batch_size():
     return SHOT_BATCH_SIZE
 
 def set_batch_size(batch_size):
+    """Sets batch size used for sampling measurement shots."""
     if not isinstance(batch_size, int):
         raise_error(TypeError, "Shot batch size must be integer.")
     elif batch_size < 1:
@@ -93,6 +94,7 @@ def get_metropolis_threshold():
     return SHOT_METROPOLIS_THRESHOLD
 
 def set_metropolis_threshold(threshold):
+    """Sets threshold for using Metropolis algorithm for sampling measurement shots."""
     if not isinstance(threshold, int):
         raise_error(TypeError, "Shot threshold must be integer.")
     elif threshold < 1:
--- a/src/qibo/config.py
+++ b/src/qibo/config.py
@@ -26,7 +26,7 @@ EIGVAL_CUTOFF = 1e-14
 SHOT_BATCH_SIZE = 2 ** 18
 
 # Threshold size for sampling shots in measurements frequencies with custom operator
-SHOT_CUSTOM_OP_THREASHOLD = 100000
+SHOT_METROPOLIS_THRESHOLD = 100000
 
 # Flag for raising warning in ``set_precision`` and ``set_backend``
 ALLOW_SWITCHERS = True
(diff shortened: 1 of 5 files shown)


src/qibo/config.py
```diff
@@ -88,6 +89,20 @@ def set_batch_size(batch_size):
...
+    return SHOT_METROPOLIS_THRESHOLD
+
+def set_metropolis_threshold(threshold):
+    """Sets threshold for using Metropolis algorithm for sampling measurement shots."""
+    if not isinstance(threshold, int):
+        raise_error(TypeError, "Shot threshold must be integer.")
+    elif threshold < 1:
+        raise_error(ValueError, "Shot threshold be a positive integer.")
+    global SHOT_METROPOLIS_THRESHOLD
+    SHOT_METROPOLIS_THRESHOLD = threshold
```
scarrazza, line 103: If I understand correctly, these last 2 functions are required in order to update the `SHOT_METROPOLIS_THRESHOLD` dynamically during the code execution, correct? So far, however the code doesn't use this feature anywhere, so this is a feature for the user, right?

stavros11: No, this is used in the `sample_frequencies` function of the qibotf backend [here](https://github.com/qiboteam/qibo/blob/48928bf1873cd7a067dcc2fc7b8ff2a7eb4ec599/src/qibo/backends/tensorflow.py#L227). If the number of shots requested by the user is smaller than this number then sampling falls back to tf primitives (`tf.random.categorical`) while if it is larger the Metropolis custom operator is used.

scarrazza: Sorry, I was not clear, I mean set/get_metropolis_threshold() functions. These functions are not used anywhere in the code, I believe.

stavros11: Oh yes, that's correct, it is just for the user to use if they want to disable the Metropolis (by setting very high threshold) for some reason (for example if it does not produce good results for their distribution). I think this is the case with most set/get functions we already have, like `set_precision`, `set_batch_size`, etc., they are not used anywhere in the code.

scarrazza: Ok, thanks for confirmations. We might include some instructions about that in the "Advanced examples" section.

scarrazza (approved):
@stavros11 thanks for this. Could you please include a couple of lines about the missing set/get_* methods we have discussed today?

stavros11:
> @stavros11 thanks for this. Could you please include a couple of lines about the missing set/get_* methods we have discussed today?

I added some lines in the existing "How to perform measurements?" example which I think is the more relevant place for these functions. Please have a look and let me know if you agree.

scarrazza (approved):
Thank you. Looks good to me!

PR #432, opened 2021-06-16
Fixes for cupy backend

stavros11:
Updates tests and other parts to work well with the cupy part of the qibojit backend. The main issue to handle is the conversion from cupy to numpy array (copy from GPU device to host memory) which can only be done explicitly by calling the `.get()` method of cupy arrays.

The main issue that still remains is that several tests that involve the collapse operator hang. We observed the same problem in qibojit tests and we solved it by deleting the reference to the state and adding a `.free_all_blocks()` command. A similar solution cannot be trivially applied here because deleting the references is slightly more complicated for some tests.

### DIFF
--- a/src/qibo/backends/numpy.py
+++ b/src/qibo/backends/numpy.py
@@ -487,6 +487,14 @@ class JITCustomBackend(NumpyBackend): # pragma: no cover
             return self.backend.asarray(result)
         return super().gather(x, indices, condition, axis)
 
+    def device(self, device_name):
+        # assume tf naming convention '/GPU:0'
+        if self.op.get_backend() == "numba":
+            return super().device(device_name)
+        else: # pragma: no cover
+            device_id = int(device_name.split(":")[-1])
+            return self.backend.cuda.Device(device_id)
+
     def initial_state(self, nqubits, is_matrix=False):
         return self.op.initial_state(nqubits, self.dtypes('DTYPECPX'),
                                     is_matrix=is_matrix)
--- a/src/qibo/backends/numpy.py
+++ b/src/qibo/backends/numpy.py
@@ -371,6 +371,9 @@ class NumpyBackend(abstract.AbstractBackend):
         state = self._append_zeros(state, sorted_qubits, density_matrix_result)
         return self.reshape(state, gate.cache.flat_shape)
 
+    def assert_allclose(self, value, target):
+        self.np.testing.assert_allclose(value, target)
+
 
 class NumpyCustomBackend(NumpyBackend): # pragma: no cover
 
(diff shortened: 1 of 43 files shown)


src/qibo/backends/numpy.py
```diff
@@ -429,28 +432,64 @@ def set_device(self, name):
...
+    def reshape(self, x, shape):
+        return super().reshape(x, self.check_shape(shape))
+
+    def eye(self, shape, dtype='DTYPECPX'):
+        return super().eye(self.check_shape(shape), dtype=dtype)
+
+    def zeros(self, shape, dtype='DTYPECPX'):
+        return super().zeros(self.check_shape(shape), dtype=dtype)
+
+    def ones(self, shape, dtype='DTYPECPX'):
```
scarrazza, line 458: Just a question about this functions, if I understand correctly, the shapes should not be `self.Tensor`, but rather numpy/list/numbers right?

scarrazza:
@stavros11 thanks for checking. I have tested this PR in a environment without tf and the code still hangs on `test_core_gates`, but the memory usage is much lower (taking just 523MB after executing 38% of tests), so I assume this issue is not related to memory, right?

stavros11:
With the latest pushes here and the open PR in the qibojit package all tests should pass on both CPU and GPU with no hangs. By the way the hanging was happening because of a type mismatch between Python and the CUDA kernel. The kernel was expecting a `long` tensor while I was passing `int32`. It just seems that cupy doesn't raise proper errors in these situations (unlike numba), so it was a bit harder to debug. No deletion of vectors and freeing the blocks is required now.

I will proceed by doing some cupy vs qibotf benchmarks to check if there is something we need to fix in terms of performance.

scarrazza:
Great, thanks for investigating this issue. I confirm that `pytest` is working fine on my GPUs.

stavros11:
> Great, thanks for investigating this issue. I confirm that `pytest` is working fine on my GPUs.

Thank you for checking. I think there is an issue with cupy when using large number of qubits. While
```Python
x = cp.zeros(10)
x[0] = 1
```
works ok, for larger size for example
```Python
x = cp.zeros(2 ** 30)
x[0] = 1
```
I get a compilation error. I only noticed that on GPUs with large memory because it appears on 28 or more qubits. Have you noticed something similar?

scarrazza:
@stavros11 yes, this happens also to me, looks like indexing issues. I have opened [this](https://github.com/cupy/cupy/issues/5367). Meanwhile we could use a custom kernel for this operation.

scarrazza:
Following the discussion in https://github.com/cupy/cupy/issues/5367, there is already a fix in the master.

PR #434, opened 2021-07-01
Set number of threads for qibojit

stavros11:
As we discussed, this moves the `set_threads` and `get_threads` methods at the backend level so that we can set the number of threads for numba properly. I checked and it seems to work properly for qibojit, for example
```Python
import qibo
from qibo.models import QFT

qibo.set_backend("qibojit")
qibo.set_threads(10)
c = QFT(28)
state = c()
```
should use 10 threads. The default number of threads is read from `psutil` (unless the `OMP_NUM_THREADS` variable is provided) and is equal to the physical number of cores. This is usually half of total threads.

Note that the above mechanism works for qibotf, but not for pure tensorflow operators. So when using the pure `tensorflow` backend, `set_threads` is completely ignored and when using `qibotf`, custom operators will use the proper number of threads while all other tf operations will use all threads (which is tensorflow's default). I am not sure if there is a way around this because tensorflow does not allow use of `tf.config.threading.set_inter_op_parallelism_threads` after import. This is in contrast to `numba.set_num_threads()` which can be called anytime during execution.

### DIFF
--- a/src/qibo/backends/abstract.py
+++ b/src/qibo/backends/abstract.py
@@ -17,14 +17,9 @@ class AbstractBackend(ABC):
         self.gpu_devices = []
         self.default_device = []
         self.nthreads = None
-        if "OMP_NUM_THREADS" in os.environ: # pragma: no cover
-            self.nthreads = int(os.environ.get("OMP_NUM_THREADS"))
-        elif "NUMBA_NUM_THREADS" in os.environ: # pragma: no cover
-            self.nthreads = int(os.environ.get("NUMBA_NUM_THREADS"))
-        else:
-            import psutil
-            # using physical cores by default
-            self.nthreads = psutil.cpu_count(logical=False)
+        import psutil
+        # using physical cores by default
+        self.nthreads = psutil.cpu_count(logical=False)
 
         self.op = None
         self._matrices = None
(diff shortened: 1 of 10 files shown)


src/qibo/backends/abstract.py
```diff
@@ -15,6 +16,13 @@ def __init__(self):
         self.cpu_devices = []
         self.gpu_devices = []
         self.default_device = []
+        self.nthreads = None
+        if "OMP_NUM_THREADS" not in os.environ:
```
scarrazza, line 20: I think would be great to include the `NUMBA_NUM_THREADS` environment flag, so we do not break the layout when e.g. running numba with tbb.

scarrazza (approved):
Looks very good.

src/qibo/backends/abstract.py
stavros11: Good point. I moved the `OMP_NUM_THREADS` in the qibotf backend as it is relevant only for this and added the check for `NUMBA_NUM_THREADS` in the numba backend. If None of these flags are given it will fall back to the default value from psutil.

scarrazza: Perfect, thanks.

PR #435, opened 2021-07-03
documentation update

scarrazza:
This PR updates the documentation for the following points:
- applications are now sorted by topic and algorithms.
- installation instructions include qibojit and qiboicarusq.


doc/source/installation.rst
```diff
@@ -65,11 +65,22 @@ Base package
 Simulation backends
 ^^^^^^^^^^^^^^^^^^^
 
-We provide multiple simulation backends for Qibo, which are automatically loaded if the corresponding packages are installed, following the hierarchy below:
+We provide multiple simulation backends for Qibo, which are automatically loaded
+if the corresponding packages are installed, following the hierarchy below:
 
-* :ref:`installing-qibotf` (*recommended*): an efficient simulation backend for CPU, GPU and multi-GPU based on TensorFlow custom operators. Install this package if you need to simulate quantum circuits with large number of qubits or complex quantum algorithms which may benefit from computing parallelism.
+* :ref:`installing-qibojit` (*experimental*): an efficient simulation backend for CPU, GPU and multi-GPU based on just-in-time (JIT) compiled custom operators. Install this package if you need to simulate quantum circuits with large number of qubits or complex quantum algorithms which may benefit from computing parallelism.
```
stavros11, line 71: qibojit does not support multi-GPU currently so we should probably remove the multi-GPU from this line and put it back once we enable this feature.

stavros11 (approved):
Thanks for the update, looks good to me. The only thing is that this documentation assumes that qibotf is the default backend and qibojit as experiment that should be enabled using `qibo.set_backend`. This will not be true after we merge #436 as qibojit will become the default backend. Shall we update the docs for this?

igres26:
It also looks good to me. 

However I would like to make one quick request as we are working on the documentation here already. 

When one searches in the documentation for "Adiabatic evolution" or "Quantum Fourier Transform" the first thing that pops up is the benchmarks, but they are not labelled as benchmarks so it's easy to get confused and enter in the wrong documentation part. I would add "benchmark" to the parts of the documentation that are benchmarks.

doc/source/installation.rst
scarrazza: @stavros11 thanks, I have fixed this issue and resorted the backend priority.

scarrazza:
@igres26 thanks for pointing this out. I have introduced some changes that should help clarifying that is benchmark from what is not. Most likely we should remove the benchmark section soon in the future, given that the code is evolving quickly.

PR #436, opened 2021-07-04
Make qibojit default backend and update warnings

stavros11:
As discussed in #427 this makes qibojit the default backend if all possible backends are available. The backend order now is the following:

1. qibojit
2. If qibojit not available: qibotf
3. If qibojit and qibotf not available: tensorflow
4. If qibojit and tensorflow not available: numpy

If qibojit and qibotf are not available, a warning will appear prompting the user to install them for increased performance. If one of the two is available no warning will appear. Furthermore an info message will be printed in all cases the first time `qibo` is imported showing the activated benchmarks.

The user can control the logging level using the `QIBO_LOG_LEVEL` environment variable. Default value is 1 (everything appears) and additional options are 3 (hide info) and 4 (hide info and warnings). See [here](https://docs.python.org/3/library/logging.html#logging-levels) for more info on logging levels.

As we discussed, my only concern about making qibojit the default is the dry run time. Other than that this PR should be ok.

Note: I noticed that some tests are failing when running on GPU due to some cupy->numpy conversion issues. I will fix this in a different PR, it should not affect what is implemented here.


scarrazza (approved):
Thanks @stavros11 looks perfect.
Let me suggest to include one sentence about the `QIBO_LOG_LEVEL` in the documentation.

stavros11:
> Thanks @stavros11 looks perfect.

Thanks for the review. I fixed the tests and now everything should work on both CPU and GPU. 

> Let me suggest to include one sentence about the `QIBO_LOG_LEVEL` in the documentation.

Sure, shall we do this in the backend sections that are introduced in #435? In this case I can merge the documentation branch here and update this and also the fact that qibojit is now the default backend. Let me know.

Other than that, I believe the qibojit integration is ready. We could try releasing qibojit on pip and adding it to the CI.

scarrazza:
@stavros11 thanks. Please go ahead with the log flag documentation.

stavros11:
> @stavros11 thanks. Please go ahead with the log flag documentation.

I added a paragraph about the log level environment variable under the list of backends in the Packages section. I also made a few updates in the examples and Backends sections of the docs, as they still contained information about the defaulteinsum/matmuleinsum backends which no longer exist. @scarrazza please have a look in the updates and let me know if you agree.

scarrazza:
@stavros11 thanks, I have included an extra note about cupy installation.

scarrazza:
@stavros11 let me put this info here, I believe we should include a warning message about `set_threads` when if the backend is tensorflow or numpy, in fact, I believe this function does not work as expected for these backends.

PR #437, opened 2021-07-05
adding qibojit to CI

scarrazza:


stavros11:
Thanks for adding qibojit in CI. With this enabled we can now remove some `# pylint: disable` statements from the code that we used when qibojit was missing and re-enable coverage for the JIT backend. Shall I do this here?

scarrazza:
Sure, but lets first merge #438 if you agree with these changes.

stavros11:
I just noticed that qibojit is not installed and tested on Windows. I think in the docs we say that qibojit is supported on Windows so shouldn't we enable these tests?

scarrazza:
@stavros11 thanks for pointing this out.

PR #438, opened 2021-07-06
disabling parallel evaluation on darwin

scarrazza:
With qibojit installed the multiprocessing import cannot be changed to "fork" mode on macos. This PR disables the parallel feature for macos. I believe we should schedule a revision of this module in the future, in particular considering better parallel algorithms rather than forcing circuit copies and async evaluation. @stavros11 please let me know your opinion.

### DIFF
--- a/.github/workflows/rules.yml
+++ b/.github/workflows/rules.yml
@@ -43,7 +43,7 @@ jobs:
     - name: Test examples
       if: startsWith(matrix.os, 'ubuntu') && matrix.python-version == '3.9'
       run: |
-        pytest examples/
+        OMP_NUM_THREADS=1 pytest examples/
     - name: Upload coverage to Codecov
       if: startsWith(matrix.os, 'ubuntu')
       uses: codecov/codecov-action@v1
--- a/examples/EF_QAE/main.py
+++ b/examples/EF_QAE/main.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 import numpy as np
 from qibo.models import Circuit
-from qibo import hamiltonians, gates, models, matrices
+from qibo import hamiltonians, gates, models, K
 from qibo.hamiltonians import Hamiltonian
 from scipy.optimize import minimize
 from sklearn.datasets import load_digits
@@ -10,7 +10,7 @@ import argparse
 
 
 def main(layers, autoencoder, example):
-    
+
     def encoder_hamiltonian_simple(nqubits, ncompress):
         """Creates the encoding Hamiltonian.
         Args:
@@ -24,7 +24,7 @@ def main(layers, autoencoder, example):
         m1 = np.eye(2 ** (nqubits - ncompress), dtype=m0.dtype)
         ham = hamiltonians.Hamiltonian(nqubits, np.kron(m1, m0))
         return 0.5 * (ham + ncompress)
-    
+
     def rotate(theta, x):
         new_theta = []
         index = 0
@@ -45,14 +45,14 @@ def main(layers, autoencoder, example):
     compress = 2
     encoder = encoder_hamiltonian_simple(nqubits, compress)
     count = [0]
-    
+
     if example == 0:
         ising_groundstates = []
         lambdas = np.linspace(0.5, 1.0, 20)
         for lamb in lambdas:
             ising_ham = -1 * hamiltonians.TFIM(nqubits, h=lamb)
             ising_groundstates.append(ising_ham.eigenvectors()[0])
-            
+
         if autoencoder == 1:
             circuit = models.Circuit(nqubits)
             for l in range(layers):
@@ -72,36 +72,36 @@ def main(layers, autoencoder, example):
                 circuit.add(gates.CZ(4, 1))
             for q in range(nqubits-compress, nqubits, 1):
                 circuit.add(gates.RY(q, theta=0))
-                
+
             def cost_function_QAE_Ising(params, count):
                 """Evaluates the cost function to be minimized for the QAE and Ising model.
-        
+
                 Args:
                     params (array or list): values of the parameters.
-        
+
                 Returns:
                     Value of the cost function.
-                """                    
+                """
                 cost = 0
                 circuit.set_parameters(params) # this will change all thetas to the appropriate values
                 for i in range(len(ising_groundstates)):
                     final_state = circuit.execute(np.copy(ising_groundstates[i]))
-                    cost += encoder.expectation(final_state).numpy().real
-                    
+                    cost += K.to_numpy(encoder.expectation(final_state)).real
+
                 cost_function_steps.append(cost/len(ising_groundstates)) # save cost function value after each step
-        
+
                 if count[0] % 50 == 0:
                     print(count[0], cost/len(ising_groundstates))
                 count[0] += 1
-        
+
                 return cost/len(ising_groundstates)
-        
+
             nparams = 2 * nqubits * layers + compress
             initial_params = np.random.uniform(0, 2*np.pi, nparams)
-        
+
             result = minimize(cost_function_QAE_Ising, initial_params,
                               args=(count), method='BFGS', options={'maxiter': 5.0e4})
-            
+
         elif autoencoder == 0:
             circuit = models.Circuit(nqubits)
             for l in range(layers):
@@ -121,38 +121,38 @@ def main(layers, autoencoder, example):
                 circuit.add(gates.CZ(4, 1))
             for q in range(nqubits-compress, nqubits, 1):
                 circuit.add(gates.RY(q, theta=0))
-                
+
             def cost_function_EF_QAE_Ising(params, count):
                 """Evaluates the cost function to be minimized for the EF-QAE and Ising model.
-        
+
                 Args:
                     params (array or list): values of the parameters.
-        
+
                 Returns:
                     Value of the cost function.
-                """                                                               
+                """
                 cost = 0
                 for i in range(len(ising_groundstates)):
                     newparams = rotate(params, lambdas[i])
                     circuit.set_parameters(newparams)
                     final_state = circuit.execute(np.copy(ising_groundstates[i]))
-                    cost += encoder.expectation(final_state).numpy().real
-                    
+                    cost += K.to_numpy(encoder.expectation(final_state)).real
+
                 cost_function_steps.append(cost/len(ising_groundstates)) # save cost function value after each step
-        
+
                 if count[0] % 50 == 0:
                     print(count[0], cost/len(ising_groundstates))
                 count[0] += 1
-        
+
                 return cost/len(ising_groundstates)
-        
-            
+
+
             nparams = 4 * nqubits * layers + 2 * compress
             initial_params = np.random.uniform(0, 2*np.pi, nparams)
-        
+
             result = minimize(cost_function_EF_QAE_Ising, initial_params,
                               args=(count), method='BFGS', options={'maxiter': 5.0e4})
-            
+
         else:
             raise ValueError("You have to introduce a value of 0 or 1 in the autoencoder argument.")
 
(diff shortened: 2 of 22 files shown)


stavros11 (approved):
Thanks for the update, I cannot test on mac but code looks okay to me. If I remember correctly though parallel evaluation was not working with qibojit also on linux, that's why the related tests are disabled. What exactly is the new issue? Installing qibojit on mac blocks the parallel evaluation for all backends? For qibotf it used to work on mac, right?

Regarding revising the implementation, perhaps we should merge this and open an issue for now. Is there an alternative approach based on Python?

scarrazza:
> If I remember correctly though parallel evaluation was not working with qibojit also on linux, that's why the related tests are disabled.

ok, I believe this PR already disables if `get_backend() == "qibojit"`, so we can skip the tests or catch the runtime error.

> What exactly is the new issue? 

https://github.com/qiboteam/qibo/runs/2996557250?check_suite_focus=true#step:8:56
Presumably numba (or some other package from qibojit) already loads multiprocessing, so we can't change the start method from `spawn` to `fork` (our previous default only for macos).

> Installing qibojit on mac blocks the parallel evaluation for all backends? For qibotf it used to work on mac, right?

I think so, in particular if we import qibojit, before loading qibotf.

> Regarding revising the implementation, perhaps we should merge this and open an issue for now. Is there an alternative approach based on Python?

I agree, I don't think the parallel implementation we have is robust, we can keep as experimental for the time being, and the propose some better approach.

stavros11:
> ok, I believe this PR already disables if `get_backend() == "qibojit"`, so we can skip the tests or catch the runtime error.

Currently the tests are skipped. We could change to catching the error, however I am not sure if this is very useful.

> I think so, in particular if we import qibojit, before loading qibotf.

Oh, I see now. I am not sure if there is a away to avoid this while keeping qibojit as the default backend. So we either follow this PR and disable parallel completely for mac or we make a different backend the default.

scarrazza:
Yeah, lets keep disable and in future revisit the implementation ref. #440.

scarrazza:
@stavros11 seems like examples are taking much longer to run, so we hit the max time of 6h.
Maybe this is something related to the number of threads? Did you observe this before?

stavros11:
> @stavros11 seems like examples are taking much longer to run, so we hit the max time of 6h.
> Maybe this is something related to the number of threads? Did you observe this before?

Usually example tests take long but do not hit the max time. I am running them locally now and I actually get some failures so I am guessing examples don't work well with the qibojit backend which is the new default.

stavros11:
The failing tests were related to numpy conversion using Tensorflow's `.numpy()`, which I fixed by changing to our backend generic `K.to_numpy`. There is still an issue with some tests that hang which may be the reason the CI timed out. I skip these tests temproarily so let's see if everything works now. The tests that hang are `test_adiabatic3sat` and `test_ef_qae`.

stavros11:
I re-run the example tests using this branch (with the two problematic tests skipped) and they pass within a few minutes both in my laptop and qibo machine. @scarrazza could you please confirm that this works for you too? I am not sure why the CI times out.

scarrazza:
Thanks for checking, I doing that right now.

scarrazza:
@stavros11, yes, I get:
```
94 passed, 12 skipped, 28 warnings in 703.09s (0:11:43)
```
Let me check the memory and reducing the threads.

scarrazza:
Ok, the problem comes from these limitations: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

Try to run the examples with:
```
ulimit -v 7000000
OMP_NUM_THREADS=2 pytest -s examples/
```
After few minutes the shor test fails with:
```
 RuntimeError: State does not fit in /CPU:0 memory.Please switch the execution device to a different one using ``qibo.set_device``.
```

scarrazza:
Given that qibotf was not crashing, maybe there examples importing tensorflow...

scarrazza:
Lets see what happens, in fact it is quite odd that qibotf was taking more than 3h to complete something that qibojit can do in less than 1h. So this sounds like memory limitations...

stavros11:
With the latest version the tests pass on both my laptop and the qibo machine. When using the `ulimit` and `OMP_NUM_THREADS` commands the tests still pass on my laptop but give various errors instantly on qibo machine. So I'm not sure if CI will pass but it is likely that no.

scarrazza:
To me it fails with:
```
------------------------------------------------------------

Using Shor's algorithm to factorize N = 21 with a = 16.

  - Performing algorithm using a fully quantum iQFT.

  - Total number of qubits used: 22.

[Qibo|WARNING|2021-07-07 15:12:54]: Falling back to CPU because the GPU is out-of-memory.

INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/home/carrazza/anaconda3/lib/python3.8/site-packages/_pytest/main.py", line 269, in wrap_session
INTERNALERROR>     session.exitstatus = doit(config, session) or 0
INTERNALERROR>   File "/home/carrazza/anaconda3/lib/python3.8/site-packages/_pytest/main.py", line 323, in _main
INTERNALERROR>     config.hook.pytest_runtestloop(session=session)
...

```
Do you understand why there is this log message about falling back to CPU because the GPU is out-of-memory? Sounds like shor raises a OOM in CPU, maybe during shots, and then this function catches the error.

stavros11:
> Do you understand why there is this log message about falling back to CPU because the GPU is out-of-memory? Sounds like shor raises a OOM in CPU, maybe during shots, and then this function catches the error.

That's strange, I never got something similar. Are you sure the tests are running on CPU and not on GPU?

scarrazza:
> Are you sure the tests are running on CPU and not on GPU?

On CPU for sure, I double checked nvidia-smi. I cannot run on GPU with ulimit and 2 threads, the code crashes silently.
If I run on GPU without ulimit, everything passes except this one:
<details>

```
_______________________________________________ test_entropy_large_circuit[qibojit-None] ________________________________________________

backend = None, accelerators = None

    def test_entropy_large_circuit(backend, accelerators):
        """Check that entropy calculation works for variational like circuit."""
        thetas = np.pi * np.random.random((3, 8))
        target_entropy = callbacks.EntanglementEntropy([0, 2, 4, 5])
        c1 = Circuit(8)
        c1.add((gates.RY(i, thetas[0, i]) for i in range(8)))
        c1.add((gates.CZ(i, i + 1) for i in range(0, 7, 2)))
        state1 = c1()
        e1 = target_entropy(state1)
    
        c2 = Circuit(8)
        c2.add((gates.RY(i, thetas[1, i]) for i in range(8)))
        c2.add((gates.CZ(i, i + 1) for i in range(1, 7, 2)))
        c2.add(gates.CZ(0, 7))
        state2 = (c1 + c2)()
        e2 = target_entropy(state2)
    
        c3 = Circuit(8)
        c3.add((gates.RY(i, thetas[2, i]) for i in range(8)))
        c3.add((gates.CZ(i, i + 1) for i in range(0, 7, 2)))
        state3 = (c1 + c2 + c3)()
        e3 = target_entropy(state3)
    
        entropy = callbacks.EntanglementEntropy([0, 2, 4, 5])
        c = Circuit(8, accelerators)
        c.add(gates.CallbackGate(entropy))
        c.add((gates.RY(i, thetas[0, i]) for i in range(8)))
        c.add((gates.CZ(i, i + 1) for i in range(0, 7, 2)))
        c.add(gates.CallbackGate(entropy))
        c.add((gates.RY(i, thetas[1, i]) for i in range(8)))
        c.add((gates.CZ(i, i + 1) for i in range(1, 7, 2)))
        c.add(gates.CZ(0, 7))
        c.add(gates.CallbackGate(entropy))
        c.add((gates.RY(i, thetas[2, i]) for i in range(8)))
        c.add((gates.CZ(i, i + 1) for i in range(0, 7, 2)))
        c.add(gates.CallbackGate(entropy))
        state = c()
    
        K.assert_allclose(state3, state)
>       K.assert_allclose(entropy[:], K.cast([0, e1, e2, e3]))

src/qibo/tests/test_core_callbacks.py:238: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
src/qibo/backends/numpy.py:460: in cast
    return self.op.cast(x, dtype=dtype)
../qibojit/src/qibojit/custom_operators/__init__.py:47: in cast
    return backend.cast(x, dtype)
../qibojit/src/qibojit/custom_operators/backends.py:170: in cast
    return self.cp.asarray(x, dtype=dtype)
../../../anaconda3/lib/python3.8/site-packages/cupy/_creation/from_data.py:66: in asarray
    return _core.array(a, dtype, False, order)
cupy/_core/core.pyx:2120: in cupy._core.core.array
    ???
cupy/_core/core.pyx:2199: in cupy._core.core.array
    ???
cupy/_core/core.pyx:2267: in cupy._core.core._send_object_to_gpu
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   TypeError: Implicit conversion to a NumPy array is not allowed. Please use `.get()` to construct a NumPy array explicitly.

```
</details>

> cupy_backends.cuda.libs.cusolver.CUSOLVERError: CUSOLVER_STATUS_INTERNAL_ERROR

This sounds like limited GPU memory, try to reboot or use an empty GPU.

Going back to our issue here. I believe that the github action was going to swap memory when using qibotf (that's why 3h), while now with qibojit it crashes directly. If the amount of memory required by shor is really correct, then the only solution is to reduce the number of qubits for the tests.

stavros11:
> On CPU for sure, I double checked nvidia-smi. I cannot run on GPU with ulimit and 2 threads, the code crashes silently.
> If I run on GPU without ulimit, everything passes except this one:

Thanks for the update, actually the qibo tests were passing for me on GPU even before the last push. The example tests still fail on GPU due to some cupy -> numpy conversions though.


> Going back to our issue here. I believe that the github action was going to swap memory when using qibotf (that's why 3h), while now with qibojit it crashes directly. If the amount of memory required by shor is really correct, then the only solution is to reduce the number of qubits for the tests.

If that's the case perhaps the simplest to try would be to remove the `N=21` value from the Shor test and leave `N=15`. I am not sure if it would work but it would definitely make the test lighter.

scarrazza:
> If that's the case perhaps the simplest to try would be to remove the N=21 value from the Shor test and leave N=15. I am not sure if it would work but it would definitely make the test lighter.

Indeed, I am playing with different configurations here: https://github.com/qiboteam/qibo/actions?query=branch%3Atryfixci
I have fixed some extra problems including the adiabatic3sat, however as you can see, the jobs still take a long time, while on my laptop if we skip the EF_QEA these should take max 6 minutes with ulimit and 2 cores.

If these tests hangs, then we should try to isolate which test is breaking the CI.

scarrazza:
With these last changes, I get 5-7 minutes runtime for all tests (no skips) on 2 different machines, using 2 threads and ulimit.

scarrazza:
@stavros11 just to let you know that I am debugging one by one here: https://github.com/qiboteam/qibo/actions?query=branch%3Apruning

scarrazza:
Looks like the scipy minimize is slowing down dramatically all tests on the CI machines, if I set a tiny maxiter test are passing.

stavros11:
What is the current status of this? Can I help in any way? I see that some tests are passing fast in the `prunning` branch.

scarrazza:
Some of then passes, if we reduce the minimization to few iteractions, however it still takes much longer than on my laptop.
I will try to replace the `exec` call to something else (e.g. [this](https://pypi.org/project/pytest-console-scripts/)), I have the suspicious that pytest is accumulating data. 

If that doesn't work then I assume we can disable these tests from CI for the time being.

scarrazza:
@stavros11 believe me or not, but CI seems to work well only with single-thread. Let me polish this PR, then we can merge.

PR #443, opened 2021-07-09
Set threads warnings

### LINKED ISSUE #442: Set threads warning messages
I was going to comment this on the qibojit PR but since this is merged I am writing here. I believe the following point is still open, right?

> @stavros11 let me put this info here, I believe we should include a warning message about `set_threads` when if the backend is tensorflow or numpy, in fact, I believe this function does not work as expected for these backends.

I was adding the warnings now, an issue though is that the [`set_threads` method ](https://github.com/qiboteam/qibo/blob/47ad4adc0674ba15578795e7f91c10b4e3823184/src/qibo/backends/__init__.py#L211) that is exposed to the user changes the threads for all created backends. This means that if numpy/tensorflow backends are created but a different backend (qibotf or qibojit) is active, then the numpy/tensorflow warnings will be raised if the user uses `set_threads`. @scarrazza, do you think we should modify `set_threads` to only change the threads of the active backend?

stavros11:
Fixes #442 by:
* Modifying `qibo.set_threads` to update the number threads only for the currently active backend.
* Raising warnings when using `qibo.set_threads` with the numpy or tensorflow backends which do not support this kind of thread setting.

### DIFF
--- a/src/qibo/backends/numpy.py
+++ b/src/qibo/backends/numpy.py
@@ -33,6 +33,7 @@ class NumpyBackend(abstract.AbstractBackend):
     def set_threads(self, nthreads):
         log.warning("Numpy backend supports only single-thread execution."
                     "Cannot change the number of threads.")
+        abstract.AbstractBackend.set_threads(self, nthreads)
 
     def to_numpy(self, x):
         return x
--- a/src/qibo/backends/numpy.py
+++ b/src/qibo/backends/numpy.py
@@ -25,6 +25,9 @@ class NumpyBackend(abstract.AbstractBackend):
         self.newaxis = np.newaxis
         self.oom_error = MemoryError
         self.optimization = None
+        self.cpu_devices = ["/CPU:0"]
+        self.gpu_devices = []
+        self.default_device = self.cpu_devices[0]
 
     def set_device(self, name):
         log.warning("Numpy does not support device placement. "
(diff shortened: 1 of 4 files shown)


src/qibo/backends/numpy.py
```diff
@@ -30,6 +30,11 @@ def set_device(self, name):
         log.warning("Numpy does not support device placement. "
                     "Aborting device change.")
 
+    def set_threads(self, nthreads):
+        log.warning("Numpy backend supports only single-thread execution."
+                    "Cannot change the number of threads.")
```
scarrazza, line 35:
```suggestion
                    " Cannot change the number of threads.")
```

stavros11: Fixed, thanks.

scarrazza:
Great, thanks. Could you please fix another issues I have just observed, when changing to numpy the device list is [], for symmetry with other backends I think we could simply say something like /CPU:0...

stavros11:
> Great, thanks. Could you please fix another issues I have just observed, when changing to numpy the device list is [], for symmetry with other backends I think we could simply say something like /CPU:0...

I changed the default numpy device to "/CPU:0". Let me know if you agree with the last changes.

scarrazza (approved):
Perfect, thank you.

PR #446, opened 2021-07-24
Fix gate.matrix property

### LINKED ISSUE #444: U1.matrix returns a single value
When trying to get the matrix form of U1, qibo returns a single value of the exponential instead of the full 2x2 matrix.

stavros11:
Fixes #444 by making the `.matrix` property to return the matrix representation of gates in the computational basis. @igres26 you can try now for the U1 and you should get the correct matrix with both `.matrix` and `.unitary`.

The backends now use different matrix properties which should not be used by the user. These are `native_op_matrix` for numpy/tensorflow backends and `custom_op_matrix` for qibojit/qibotf backends.

### DIFF
--- a/src/qibo/abstractions/abstract_gates.py
+++ b/src/qibo/abstractions/abstract_gates.py
@@ -405,7 +405,8 @@ class ParametrizedGate(Gate):
         # pylint: disable=E1101
         if isinstance(self, BaseBackendGate):
             self._matrix = None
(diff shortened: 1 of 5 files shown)


igres26 (approved):
It now works for the U1 case, and for `.controlled_by` gates of up to 2 total qubits.

scarrazza (approved):
@stavros11 thanks, looks good.

PR #447, opened 2021-07-30
replacing unitary with matrix

scarrazza:
Following #446 here we remove `.unitary` in favour of `.matrix`.

### DIFF
--- a/doc/source/qibo.rst
+++ b/doc/source/qibo.rst
@@ -409,36 +409,39 @@ fSim with general rotation
     :member-order: bysource
 
 
+Special gates
+^^^^^^^^^^^^^
+
 Toffoli
-^^^^^^^
+"""""""
 
 .. autoclass:: qibo.abstractions.gates.TOFFOLI
     :members:
     :member-order: bysource
 
 Arbitrary unitary
-^^^^^^^^^^^^^^^^^
+"""""""""""""""""
 
 .. autoclass:: qibo.abstractions.gates.Unitary
     :members:
     :member-order: bysource
 
 Variational layer
-^^^^^^^^^^^^^^^^^
+"""""""""""""""""
 
 .. autoclass:: qibo.abstractions.gates.VariationalLayer
     :members:
     :member-order: bysource
 
 Flatten
-^^^^^^^
+"""""""
 
 .. autoclass:: qibo.abstractions.gates.Flatten
     :members:
     :member-order: bysource
 
 Callback gate
-^^^^^^^^^^^^^
+"""""""""""""
 
 .. autoclass:: qibo.abstractions.gates.CallbackGate
     :members:
--- a/doc/source/qibo.rst
+++ b/doc/source/qibo.rst
@@ -9,7 +9,6 @@ The Qibo package comes with the following modules:
 * Gates_
 * Hamiltonians_
 * Callbacks_
-* Solvers_
 * Optimizers_
 * Parallel_
 * Backends_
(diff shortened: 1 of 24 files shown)


stavros11 (approved):
Thanks for implementing this, looks good. Only comment is regarding the depreciation of features as we were discussing, perhaps we could keep the `unitary` property and raise an error prompting the user to use `.matrix` instead, however I am not sure if this would be very useful in this particular case.

Also regarding documentation, I noticed that the abstract gate models are not included in the docs, so the matrix property is also not documented. Do you think we should include it somewher?

scarrazza:
> Thanks for implementing this, looks good. Only comment is regarding the depreciation of features as we were discussing, perhaps we could keep the unitary property and raise an error prompting the user to use .matrix instead, however I am not sure if this would be very useful in this particular case.

Indeed, I was checking the code and we don't have example relying on this specific feature, so we can probably skip the deprecation notice.

> Also regarding documentation, I noticed that the abstract gate models are not included in the docs, so the matrix property is also not documented. Do you think we should include it somewhere?

Good point. In principle we have to document the BaseBackendGate, given that Gate does not define the matrix attribute, or maybe move this attribute to the Gate class. What do you think?

stavros11:
> Good point. In principle we have to document the BaseBackendGate, given that Gate does not define the matrix attribute, or maybe move this attribute to the Gate class. What do you think?

If I remember correctly, the difference between Gate and BaseBackendGate is that the first would serve as the general abstract gate for all backends (simulation and hardware) while the second is more specialized for simulation. However I am not sure if this distinction makes sense, at least until we have a better idea of how the hardware backends will work and decide on a more robust abstraction scheme. Therefore in principle we could merge these two objects to a single one without changing much of the current code. Alternatively we could just include BaseBackendGate in the docs but this would still skip documenting some methods that could be useful for the user (eg. `target_qubits`, `control_qubits`, etc.) and are defined in `Gate`.

scarrazza:
Yes, in this last commit I have included both classes and improved the indexing.
I believe we can keep the structure as it is while we develop the hardware backends.

stavros11:
> Yes, in this last commit I have included both classes and improved the indexing.
> I believe we can keep the structure as it is while we develop the hardware backends.

Thanks for the update. I agree with keeping the structure for now and make any modifications as we develop the hardware backends. Some comments regarding the updated docs:

I agree with documenting both abstract gate models but I would consider making some changes in the corresponding docstrings. For example, the attributes of `Gate` class are documented in the class docstring and now they appear in a weird way in the docs, eg. check how `init_args`, `init_kwargs`, `is_controlled_by`, etc. appear in rendered docs. Also all of these attributes and a few other methods such as `state_vector_call`, `density_matrix_call`, `density_matrix_half_call`, etc. are intended for internal Qibo usage and are not supposed to be seen by the user so it may be confusing to include them in the docs. Perhaps the easiest solution would be to use empty docstrings to hide methods from docs but we still need some kind of documentation to remain within code.

Also, the new groupping of gate sections looks good, however I would include a fourth section, called for example "Special gates" to include the Toffoli, Unitary, Variational Layer, Flatten and Callback gates so that they are in the same section level (`""""`) as the rest of the gates.

scarrazza:
@stavros11, thanks. I have grouped the special gates in a subsection.

Concerning the attributes, I believe we have 3 options:
1. rename all of then with the _ prefix (in fact I think this happens in the code but not in the docstring)
2. move the docstring to the __init__ method
3. exclude members using the sphinx syntax (directly in the docs source)

Point 1 seems simpler to maintain but may require a non negligible code refactoring.

stavros11:
Thanks for the update. 

I agree that option 1 is probably the proper approach from design perspective. My main concerns are that indeed this would need a major refactoring in many files and also the code could become slightly less elegant as we would have many variables starting with _. The first point is not very important, we would just need one big PR that implements this, but the second may be annoying in the future. Also, the _ prefix works for functions but I'm not sure if it would work for attributes documented in class docstring.

The second option is not a bad solution but it would work only for attributes, not potential functions that we need to hide from docs.

Perhaps we could use the second option for attributes but leave the names as they are (no _) but use the _ for all functions we want to hide. Let me know your preference and how we should proceed, I can do the _ refactoring.

PR #448, opened 2021-07-30
EntanglementEntropy use circuits with fixed nqubits

scarrazza:
This PR raises an error if the user allocates multiple circuits using different number of qubits but sharing the same `EntanglementEntropy` object. @igres26 could you please check if this PR raises a proper error in your example?

### DIFF
--- a/src/qibo/core/callbacks.py
+++ b/src/qibo/core/callbacks.py
@@ -52,7 +52,7 @@ class EntanglementEntropy(BackendCallback, abstract_callbacks.EntanglementEntrop
         from qibo import gates
         if self._nqubits is not None and self._nqubits != n:
             raise_error(RuntimeError,
-                        "Changing nqubits for EntanglementEntropy is not supported.")
+                        f"Changing EntanglementEntropy nqubits from {self._nqubits} to {n}.")
         self._nqubits = n
         if self.partition is None:
             self.partition = list(range(n // 2 + n % 2))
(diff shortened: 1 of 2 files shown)


src/qibo/core/callbacks.py
```diff
@@ -50,6 +50,9 @@ def density_matrix(self, x):
     @abstract_callbacks.Callback.nqubits.setter
     def nqubits(self, n: int):
         from qibo import gates
+        if self._nqubits is not None and self._nqubits != n:
+            raise_error(RuntimeError,
+                        "Changing nqubits for EntanglementEntropy is not supported.")
```
stavros11, line 55: Perhaps we could make this error more explicit by printing the values of `n` and `self._nqubits` to show how they differ.

stavros11 (commented):
Thanks for implementing this. I think it works if one uses the entropy as part of circuit (@igres26 can confirm), but it does not work if the entropy is called directly on a state. For example I tried the following:

```Python
import numpy as np
from qibo import K
from qibo.callbacks import EntanglementEntropy

# 3 qubit state
state1 = np.random.random(8) + 1j * np.random.random(8)
state1 = K.cast(state1 / np.sqrt(np.sum(np.abs(state1) ** 2)))
# 4 qubit state
state2 = np.random.random(16) + 1j * np.random.random(16)
state2 = K.cast(state2 / np.sqrt(np.sum(np.abs(state2) ** 2)))

entropy = EntanglementEntropy()
ent1 = entropy(state1)
ent2 = entropy(state2)
```
and I don't get the `RuntimeError`. I think the reason is that when entropy is called on a state, it calls [`set_nqubits`](https://github.com/qiboteam/qibo/blob/edd2e9c4f3f798d57c368913516f238802d15bd3/src/qibo/core/callbacks.py#L79) which only calls the `nqubits` setter `if self._nqubits is None`. So we should either remove the `if self._nqubits is None` condition from there and always call the setter or move the new check in the `set_nqubits` function. I think the first is a better solution.

Note that in the above example I get a different error:
```
ValueError: cannot reshape array of size 16 into shape 8
```

scarrazza:
Thanks, should work now.

stavros11 (approved):
Thanks for the update, looks good to me now. Both the script from the above post and the following:
```Python
from qibo import callbacks, gates, models

entropy = callbacks.EntanglementEntropy()

c = models.Circuit(3)
c.add(gates.CallbackGate(entropy))

c = models.Circuit(4)
c.add(gates.CallbackGate(entropy))
```
raise the `RuntimeError`.

PR #449, opened 2021-08-03
gates refactor

scarrazza:
Performs minimal refactoring following the discussion in #447.
@stavros11 let me know if you agree.

### DIFF
--- a/src/qibo/abstractions/abstract_gates.py
+++ b/src/qibo/abstractions/abstract_gates.py
@@ -101,7 +101,7 @@ class Gate:
         self._set_control_qubits(qubits)
         self._check_control_target_overlap()
 
-    def set_targets_and_controls(self, target_qubits: Sequence[int],
+    def _set_targets_and_controls(self, target_qubits: Sequence[int],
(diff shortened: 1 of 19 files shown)


stavros11 (approved):
Thank you for the refactor. If you agree, I would additionally use the _ prefactor in the following methods:
* `BaseBackendGate.control_unitary`
* `BaseBackendGate.construct_unitary`
* `BaseBackendGate.set_nqubits`

I think all the rest are okay to be documented.

Alternatively (with a second look), we could also move the `matrix` property to `Gate` and not document `BaseBackendGate` at all but then we would also need to move `control_unitary` and `construct_unitary` so what we have now may be better.

To complete the docs, I would also add a return statement for the `Gate.on_qubits` method:
```Python
def on_qubits(self, *q) -> "Gate":
     """Creates the same gate targeting different qubits.

    Args:
        q (int): Qubit index (or indeces) that the new gate should act on.

    Returns:
        A :class:`qibo.abstractions.gates.Gate` object of the original gate type targeting the given qubits.
    """
    ...
```

PR #453, opened 2021-08-22
[WIP] MultiGPU for qibojit

### LINKED ISSUE #450: MultiGPU for qibojit
As agreed, we should port the qibotf multiGPU to qibojit.

stavros11:
Porting the existing MultiGPU implementation to qibojit to fix #450. Note that the current implementation may not be the most efficient for multi-GPU but the port may be useful for qibotf - qibojit comparisons. @scarrazza here I added an additional backend module accessible via `K.multigpu` to implement the methods required for multi-GPU separately for qibotf and qibojit. However with a second look it may be possible to simplify this and avoid some code repetition by using the `with K.multigpu.on_cpu()` with the custom object defined in `qibo.backends.jit.CupyCpuDevice` to cast operations on CPU. I also created a new file `jit.py` to place the `JITCustomBackend` so that `numpy.py` is not very long.

Tests pass on my local machine. I am not sure why the CI is failing but it may not be related to this PR. For example see [here](https://github.com/qiboteam/qibo/runs/3392377829?check_suite_focus=true) where I created a new branch that clones master and CI failed with the same error.

TODO:
- [x] Simplify code and perhaps remove `K.multigpu` completely.
- [x] The qibotf implementation uses two additional custom operators for CPU only: transpose_state and swap_pieces. Currently I have implemented these within qibo (without numba acceleration), but before merging we should implement them in the numba part of qibojit and merge in the qibojit repository.
- [x] Use the qibojit transpose_state and swap_pieces (we should first release an updated qibojit for CI to pass with this).
- [x] Once the code is finalized we should benchmark and compare performance to qibotf.


scarrazza:
@stavros11 thanks for this cleanup. I am checking the CI, most likely there is numpy version conflict with the latest numba and tensorflow, in particular for python 3.7. I will try to upgrade qibotf and check again.

stavros11:
> On the other hand, we could disable GPU/multi-GPU for macos and keep that only for linux/windows, given that there is no cupy version (or better nvidia/amd drivers) for macos.

Following this, I added an error when one tries to construct a distributed circuit on macos and disabled the related tests. The CI seems to be okay now so I will proceed by doing some benchmarks on the DGX mainly focusing on the following:
- Make sure that this branch is equal or better than master when qibotf is used,
- Compare qibotf and qibojit multi-GPU performance.

I have not made any changes regarding single GPU and macos. I am not sure if there are issues with this as I cannot test and cannot be captured by CI. We could do a different PR if there are issues with this.

scarrazza:
Thanks for the fixes and implementation. I agree with both points.

> I have not made any changes regarding single GPU and macos. I am not sure if there are issues with this as I cannot test and cannot be captured by CI. We could do a different PR if there are issues with this.

That's fine, there is currently no simple procedure to use cuda on macos.

stavros11:
> * Compare qibotf and qibojit multi-GPU performance.

Here are some QFT benchmarks using maximum of two GPUs on DGX:

nqubits | dry run time (sec) | simulation time (sec) | transfer to `np.array` time (sec) 
-- | -- | -- | -- 
30 (qibotf / single GPU) | 8.04281 | 0.64819 | 22.54783
30 (qibojit / single GPU) | 8.50952 | 7.35147 | 4.86743
31 (qibotf / 2x) | 41.11845 | 39.63167 | 13.12209
31 (qibojit / 2x) | 48.66338 | 36.1170 | 13.66040
32 (qibotf / 8x*) | 130.2988 | 133.9675 | 50.0059
32 (qibojit / 8x*) | 117.9901 | 110.8543 | 29.0673

* Note that it should be possible to run 32 qubits using 4x but for some reason I get OOM errors on both qibotf and qibojit when I try to do so.

> * Make sure that this branch is equal or better than master when qibotf is used,

I am not able to run multi-GPU benchmarks using the latest master and benchmark script under examples/benchmarks as I get the following error:
```
tensorflow.python.framework.errors_impl.UnimplementedError: SwapPiecesOp operator not implemented for GPU. [Op:SwapPieces]
```
which seems to be caused by using the swap_pieces op on wrong device (this should not be used on GPU). I am not sure if this was caused by some of the backend refactoring we did or it broke because of a later tf version. It is strange that it is not captured by tests. This issue appears to be fixed in this PR though and the numbers quoted above are better than we we measured in the original multi-GPU benchmark that is quoted [in our docs](https://qibo.readthedocs.io/en/latest/benchmarks/results/QFT.html#double-precision-complex128).

scarrazza:
Thanks for these numbers. The SwapPiecesOp error appears only for qibotf, right? What happens with qibojit?

stavros11:
> Thanks for these numbers. The SwapPiecesOp error appears only for qibotf, right? What happens with qibojit?

It only appears for qibotf on master branch. qibojit multi-GPU is not implemented at all on master so we cannot compare.

scarrazza:
@stavros11 ok, could you please track down this regression, e.g. downgrading tf and/or moving back to rc0 an 0.1.5?

stavros11:
> @stavros11 ok, could you please track down this regression, e.g. downgrading tf and/or moving back to rc0 an 0.1.5?

Ok, I checked and it is just an issue with the benchmark script only, as it passing the GPU as the `memory_device` in the DistributedCircuit unless the user specifies otherwise. I fixed this and now everything works on master too with latest tf and qibotf versions. I get the following times:

nqubits | dry run time (sec) | simulation time (sec) 
-- | -- | --
31 (2x) | 42.1373 | 42.11049
32 (4x) | 109.5348 | 111.60453

It seems that from master I can run 32 qubits using four logical GPUs which is the correct scaling. Another issue I have now is that I cannot run 30 qubits on single DGX GPU (32GB) which is strange because the 30 qubit state should be 16GB.

scarrazza:
> It seems that from master I can run 32 qubits using four logical GPUs which is the correct scaling. Another issue I have now is that I cannot run 30 qubits on single DGX GPU (32GB) which is strange because the 30 qubit state should be 16GB.

OK, good. So presumably there is some copy object in this PR, right?

stavros11:
> OK, good. So presumably there is some copy object in this PR, right?

Indeed an object deletion was not done properly so the GPU memory remained occupied. This is fixed in the last push.

> Another issue I have now is that I cannot run 30 qubits on single DGX GPU (32GB) which is strange because the 30 qubit state should be 16GB.

This was also fixed by deleting the state object after each run, for example after dry run / before the second simulation run. This was only related to the benchmark script, not the Qibo code.

After fixing these two issues the memory usage now agrees with master and is the expected one. Below is some complete comparison between qibotf master - qibotf multigpu (this branch) and qibojit multigpu. Note that the machine was not entirely during these benchmarks so the numbers may not be very accurate. The two GPUs I used were (almost) emtpy, however the CPU was used which may affect results. I believe it is okay for comparing the different backends though.

nqubits (ngpus / backend) | dry run time (sec) | simulation time (sec) 
-- | -- | --
30 (1 / qibotf master) | 8.03795 | 0.02455
30 (1 / qibotf multigpu) | 8.03138 | 0.02423
30 (1 / qibojit multigpu) | 8.37705 | 7.30610
31 (2 / qibotf master) | 41.30851 | 40.19546
31 (2 / qibotf multigpu) | 41.20646 | 40.18874
31 (2 / qibojit multigpu) | 48.48313 | 36.29911
32 (2x2 / qibotf master) | 105.68456 | 106.05286
32 (2x2 / qibotf multigpu) | 107.18271 | 106.10236
32 (2x2 / qibojit multigpu) | 108.48155 | 92.74082
31 (2x1 / qibotf master) | 69.42422 | 71.08769
31 (2x1 / qibotf multigpu) | 68.89664 | 70.01364
31 (2x1 / qibojit multigpu) | 59.84499 | 52.70878

These numbers make sense to me so I would say this PR is good for a review now.

scarrazza:
@stavros11 thank you very much. Indeed, these numbers are reassuring, I will review the implementation.

src/qibo/backends/abstract.py
```diff
@@ -520,6 +512,52 @@ def density_matrix_collapse(self, gate, state, result): # pragma: no cover
         """Collapses density matrix to a given result."""
         raise_error(NotImplementedError)
 
+    @abstractmethod
+    def on_cpu(self): # pragma: no cover
+        """Used as `with K.on_cpu():` to perform following operations on CPU."""
+        raise_error(NotImplementedError)
+
+    @abstractmethod
+    def cpu_tensor(self, x, dtype=None): # pragma: no cover
```
scarrazza, line 521: Could you please include some docstrings?

src/qibo/backends/numpy.py
```diff
@@ -236,16 +232,16 @@ def update_frequencies(nsamples, frequencies):
...
+            pass
 
-            def __exit__(self, *args):
-                pass
+        def __exit__(self, *args):
+            pass
 
-        return DummyModule()
+    def device(self, device_name):
+        return self.DummyModule()
```
scarrazza, line 244: Maybe we could use instead:
```suggestion
        pass
```

scarrazza: Forget, I misread this class.

scarrazza (approved):
@stavros11 thanks for this great implementation.
I have a question concerning `memory_device`, in this PR we consider the CPU as the default device for the full state, correct?

stavros11:
Thank you for the review. I added a docstring in the `cpu_tensor` method, thanks for catching this.

> I have a question concerning `memory_device`, in this PR we consider the CPU as the default device for the full state, correct?

Yes, I removed the `memory_device` argument from distributed circuits and "hard-coded" the CPU as the default device to save state pieces during multi-GPU simulation. This is mostly because the current implementation targets systems with single CPU with plenty of memory (RAM) and multiple GPUs with smaller memories, so it only makes sense to use the CPU as the `memory_device` and allowing user to change is mostly complicating things rather than providing something useful. If you believe it would still be useful to have the `memory_device` let me know.

scarrazza:
Thank you. I fully agree with this choice. 
Let me suggest to merge this PR now and go ahead with the benchmark code.

PR #454, opened 2021-08-24
fixing CI with numba and py3.7

scarrazza:
There is an [issue](https://github.com/numba/numba/issues/7339) in the latest wheels for numba with python 3.7 and linux.
Here we propose an workaround until this issue is solved.


scarrazza:
@igres26 if you can, could you please try to run pytest on this PR using the latest qibotf and qibojit?

PR #457, opened 2021-09-04
adding profile file for backends

scarrazza:
As discussed, this PR implements the backend loading mechanism using a default configuration file located in `qibo/backends/profiles.yml` or by using the `QIBO_PROFILE` environment variable.

At some point in the future, we should consider the possibility to store the backend class definition directly in the external module.

### DIFF
--- a/src/qibo/backends/__init__.py
+++ b/src/qibo/backends/__init__.py
@@ -22,7 +22,7 @@ class Backend:
             with open(profile_path) as f:
                 import yaml
                 profile = yaml.safe_load(f)
-        except FileNotFoundError:
+        except FileNotFoundError:  # pragma: no cover
             raise_error(FileNotFoundError,
                         f"Profile file {profile_path} not found.")
 
@@ -167,15 +167,6 @@ def get_backend():
         A string with the backend name.
     """
     return K.name
-
-
-def get_available_backends():
-    """Get list of available backends.
-
-    Returns:
-        A dictionnary with the names and modules for available backends.
-    """
-    return K.available_backends
 
 
 def set_precision(dtype="double"):
--- a/src/qibo/backends/__init__.py
+++ b/src/qibo/backends/__init__.py
@@ -38,7 +38,7 @@ class Backend:
         for backend in profile.get('backends'):
             name = backend.get('name')
             if self.check_availability(name):
-                if name == 'tensorflow':
+                if name == 'tensorflow' or name == 'qibotf':
                     os.environ["TF_CPP_MIN_LOG_LEVEL"] = str(
                         config.TF_LOG_LEVEL)
                     import tensorflow as tf  # pylint: disable=E0401
@@ -50,7 +50,7 @@ class Backend:
                     backend.get('from')), backend.get('class'))
                 self.available_backends[name] = custom_backend
                 if backend.get('is_hardware', False):  # pragma: no cover
-                    self.hardware_backends['qiboicarusq'] = custom_backend
+                    self.hardware_backends[name] = custom_backend
                 if profile.get('default') == name:
                     active_backend = name
 
--- a/src/qibo/backends/__init__.py
+++ b/src/qibo/backends/__init__.py
@@ -49,7 +49,7 @@ class Backend:
                 custom_backend = getattr(importlib.import_module(
                     backend.get('from')), backend.get('class'))
                 self.available_backends[name] = custom_backend
-                if backend.get('is_hardware', False):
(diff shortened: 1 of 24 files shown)


src/qibo/backends/__init__.py
```diff
@@ -22,34 +35,24 @@ def __init__(self):
...
-        # check if IcarusQ is installed
-        if self.check_availability("qiboicarusq"): # pragma: no cover
-            # hardware backend is not tested until `qiboicarusq` is available
-            from qibo.backends.hardware import IcarusQBackend
-            self.available_backends["icarusq"] = IcarusQBackend
-            self.hardware_backends["icarusq"] = IcarusQBackend
+        for backend in profile.get('backends'):
+            name = backend.get('name')
+            if self.check_availability(name):
+                if name == 'tensorflow':
```
stavros11, line 41: Minor comment: We could change this to
```suggestion
                if name == 'tensorflow' or name == 'qibotf':
```
so that in case someone removes the tensorflow backend from profile.yml or uses a custom profile file without tensorflow but with qibotf, tensorflow warnings are still supressed.

src/qibo/backends/__init__.py
```diff
@@ -22,34 +35,24 @@ def __init__(self):
...
+                    import tensorflow as tf  # pylint: disable=E0401
+                    if tf.__version__ < TF_MIN_VERSION:  # pragma: no cover
+                        raise_error(
+                            RuntimeError, f"TensorFlow version not supported, minimum is {TF_MIN_VERSION}.")
+                import importlib
+                custom_backend = getattr(importlib.import_module(
+                    backend.get('from')), backend.get('class'))
+                self.available_backends[name] = custom_backend
+                if backend.get('is_hardware', False):  # pragma: no cover
+                    self.hardware_backends['qiboicarusq'] = custom_backend
```
stavros11, line 53: I believe this should be
```suggestion
                    self.hardware_backends[name] = custom_backend
```
so that it works for arbitrary hardware backends.

src/qibo/tests/conftest.py
```diff
@@ -11,9 +11,9 @@
 _ACCELERATORS = None
 if ("qibotf" in _available_backends or "qibojit" in _available_backends):
     _ACCELERATORS = "2/GPU:0,1/GPU:0+1/GPU:1,2/GPU:0+1/GPU:1+1/GPU:2"
-if "icarusq" in _available_backends: # pragma: no cover
+if "qiboicarusq" in _available_backends: # pragma: no cover
```
stavros11, line 14: We can make this backend name agnostic by using the `.is_hardware` property (which perhaps we need to add to `AbstractBackend`). Alternatively something that could work and skip the tests for all hardware backends would be to change the `_available_backends` definition to:
```Python
_available_backends = set(qibo.K.available_backends.keys()) - set(qibo.K.hardware_backends.keys())
```

In a similar manner, we could possibly extend this PR to completely remove hard-coded backend names by adding the relevant properties in the abstract backend. For example another case which currently requires hard-coded backend names around the code is the multi-GPU (see the `if "qibotf" in _available_backends:` above). We could make this name agnostic using a `.supports_multigpu` property in the backends. The multi-GPU case may be slightly more complicated than `.is_hardware` so we should probably to this in a different PR, but let me know what you think.

stavros11 (approved):
Thank you for implementing this, I believe it is a great first step to simplify the process of adding new backends, especially from external collaborators. I think now it is possible to add a backend without requiring to touch the qibo internals at all, which is great, however one should still inherit Qibo's abstract backend and define the corresponding methods. 

With that in mind, I believe that as next step we should better document and perhaps somehow simplify this process, for example by making clear which methods are required and optional (eg. the multi-GPU methods could be optional). We should also improve the hardware abstractions, as currently the abstract backend is geared during simulation.

src/qibo/backends/__init__.py
scarrazza: Thanks.

src/qibo/backends/__init__.py
scarrazza: Absolutely.

scarrazza:
Thanks for the review. Indeed, this PR should be the basis for a more general cleanup and simplification.

stavros11:
> Thanks for the review. Indeed, this PR should be the basis for a more general cleanup and simplification.

Thank you for the updates, looks good. Just don't forget to change the base branch to `master` before merging as `multigpu` is already merged.

PR #458, opened 2021-09-06
using abstract backend attributes

scarrazza:
Minor changes to avoid hardcoding backend names.

We should complete the docs about the implementation of a new backend.
In principle we could remove the .compile method, however the mechanism might be useful for hardware optimization so I am keeping this feature.

@stavros11 please let me know if you see some extra generalization that makes sense at this point. Here I decided to keep the profiles.yml as light as possible, given that we are already exposing the backend class inheritance, so we can use that directly instead of reading and setting from the profile.

### DIFF
--- a/src/qibo/core/distcircuit.py
+++ b/src/qibo/core/distcircuit.py
@@ -94,10 +94,9 @@ class DistributedCircuit(circuit.Circuit):
 
         Also checks that there are sufficient qubits to use as global.
         """
-        if K.name not in {"qibotf", "qibojit"}:
-            raise_error(NotImplementedError, "Distributed circuit is implemented "
-                                             "only for the qibotf and qibojit "
-                                             "backends.")
+        if not K.supports_multigpu:
+            raise_error(NotImplementedError, f"Distributed circuit is not supported "
+                                              "by the {K.name} backend.")
         if isinstance(gate, gates.KrausChannel):
             raise_error(NotImplementedError, "Distributed circuits do not "
                                              "support channels.")
--- a/src/qibo/core/distcircuit.py
+++ b/src/qibo/core/distcircuit.py
@@ -46,9 +46,9 @@ class DistributedCircuit(circuit.Circuit):
     """
 
     def __init__(self, nqubits: int, accelerators: Dict[str, int]):
-        if sys.platform == "darwin":  # pragma: no cover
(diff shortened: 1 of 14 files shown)


src/qibo/core/distcircuit.py
```diff
@@ -94,10 +94,9 @@ def _add(self, gate: gates.Gate):
 
         Also checks that there are sufficient qubits to use as global.
         """
-        if K.name not in {"qibotf", "qibojit"}:
-            raise_error(NotImplementedError, "Distributed circuit is implemented "
-                                             "only for the qibotf and qibojit "
-                                             "backends.")
+        if not K.supports_multigpu:
+            raise_error(NotImplementedError, f"Distributed circuit is not supported "
+                                              "by the {K.name} backend.")
```
stavros11, line 99: I think we could move this check in the `DistributedCircuit.__init__` (line 49 same file). Currently we have a `sys.platform == "darwin"` check there which I believe we can change to `if K.supports_multigpu`, since the mac os check is now done within the multi-GPU backends to set the `self.supports_multigpu` property.

stavros11 (approved):
Thank you for these updates, it looks good, exactly what I had in mind in my comment in the other PR. I agree with having the supports multigpu, gradients and is_hardware as backend properties instead of the yml file. 

I do not see any further generalization that we can do at this point. The only places where some hard-coded backend names remain are the following:
* Tests that use random numbers from within each backend, such as in `test_core_measurements.py` and `test_measurement_gate_probabilistic.py`. I do not see an easy way to generalize this given that random number generator behavior is different for each number.
* Setting the accelerators and skipping distributed tests according to whether each backend supports multi-GPU in `conftest.py`. For example the qibojit and qibotf names are still hardcoded in lines 12, 103 and 116 in `conftest.py`. There we cannot always use the `.supports_multigpu` property directly as there is no access to the backend objects. For example, line 12 could be updated to something like:
```Python
for backend in _available_backends:
    if backend.supports_multigpu:
        _ACCELERATORS = "2/GPU:0,1/GPU:0+1/GPU:1,2/GPU:0+1/GPU:1+1/GPU:2"
```
but it is not clear how we could do something similar for the other cases.

scarrazza:
@stavros11 thanks. I think this cleanup round is complete but we have to improve the documentation in another PR. Concerning tests, I think the current setup is acceptable, these are designed to test our own backends.

Issue #459, opened 2021-09-08
"Probabilities do not sum to 1" with NumPy backend


mlazzarin:
**Describe the bug**
The script ``examples/variational_classifier/main.py`` raises a ValueError "probabilities do not sum to 1" when exectued with the NumPy backend. Last two lines of the traceback:
```
  File "../qibo/backends/numpy.py", line 232, in sample_frequencies
    frequencies = update_frequencies(nshots % SHOT_BATCH_SIZE, frequencies)
  File "../qibo/backends/numpy.py", line 224, in update_frequencies
    samples = self.random.choice(range(len(probs)), size=nsamples, p=probs)
  File "mtrand.pyx", line 939, in numpy.random.mtrand.RandomState.choice
ValueError: probabilities do not sum to 1
```
Maybe it is linked to this issue https://github.com/numpy/numpy/issues/6123.
Indeed, adding a normalization step for ``probs`` in the following function seems to solve the issue.
https://github.com/qiboteam/qibo/blob/9af4e773231374a83b3d0e93c47645d47f95cf27/src/qibo/backends/numpy.py#L218-L230

**To Reproduce**
1. Install ``qibo0.1.6rc1``
2. Execute the variational classifier example with the NumPy backend.

If needed, I managed to reproduce it in Google Colab.

scarrazza:
@mlazzarin thank you very much for spotting this. Could you please retry with the branch #460?

mlazzarin:
I confirm that the problem does not occur anymore with branch #460.

scarrazza:
@mlazzarin thanks for checking.

scarrazza:
Fixed in #460.

PR #460, opened 2021-09-08
normalizing probs

scarrazza:
Implements the fix proposed in #459.

@stavros11 please have a look, I dislike this extra operation only for numpy, we could add a threshold or apply the normalization conditionally.

### DIFF
--- a/src/qibo/backends/__init__.py
+++ b/src/qibo/backends/__init__.py
@@ -59,7 +59,7 @@ class Backend:
         self.qnp = self.construct_backend("numpy")
         # Create the default active backend
         if "QIBO_BACKEND" in os.environ:  # pragma: no cover
-            self.active_backend = os.environ.get("QIBO_BACKEND")
+            active_backend = os.environ.get("QIBO_BACKEND")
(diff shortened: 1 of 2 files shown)


stavros11 (approved):
> Implements the fix proposed in #459.

Thanks for the fix, looks good to me.

> @stavros11 please have a look, I dislike this extra operation only for numpy, we could add a threshold or apply the normalization conditionally.

My main issue here was why `probs` was not normalized in the first place, so I tried printing the final state norm for the variational classifier example and I get values like 0.9999999477489163 or 1.0000000645147356 so I assume it is a precision issue. I get such values with any backend, not only numpy, it is just that numpy's frequency generator is "sensitive" to this error.

Regarding the extra step, I am not sure if it is a big issue since the `sample_frequencies` implementation is different for each backend anyway.  Do you have in mind a cleaner solution?

scarrazza:
@stavros11 thanks. Indeed, at the moment, given that numpy does not offer a tolerance flag for this method, I think we can keep this version.

PR #463, opened 2021-09-10
Remove unused flag from AbstractCircuit class

### LINKED ISSUE #462: Flag "is_executed" is defined but not implemented
The "is_executed" flag in the AbstractCircuit class is defined here
https://github.com/qiboteam/qibo/blob/14bb45448209312e68df8176647e796a74109f71/src/qibo/abstractions/circuit.py#L91-L93
but it seems like it is not accessed nor modified anywhere in the source code of Qibo.
Indeed, it always returns False. The "add" method ignores it and simply checks for "_final_state"
https://github.com/qiboteam/qibo/blob/4073ce9d96836cb46c1abf2e2b75b7270c0331d4/src/qibo/abstractions/circuit.py#L390-L392
This is a minor bug but it may cause issues if such flag is used by unaware users.
Should we implement it properly (or simply remove it)?

mlazzarin:
Closes #462. I added some edits that came to my mind while reading `abstractions/circuit.py`, let me know if they are ok or if I should undo them.

### DIFF
--- a/src/qibo/abstractions/circuit.py
+++ b/src/qibo/abstractions/circuit.py
@@ -63,12 +63,13 @@ class AbstractCircuit(ABC):
     All backend-based circuits should inherit ``AbstractCircuit``.
 
     Qibo provides the following circuits:
-    A state vector simulation circuit:
-    :class:`qibo.core.circuit.Circuit`,
-    a density matrix simulation circuit:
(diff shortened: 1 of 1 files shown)


scarrazza (approved):
@mlazzarin thank you. This looks good to me, I have recompiled the docs and the links are working properly.

PR #465, opened 2021-09-12
Fix tests about `print` statements

mlazzarin:
If I understand correctly, the code in ``src/qibo/tests/test_prints.py`` implements a test that checks if the source code contains `print` statement. If this is so, such test is not working properly. In fact, the `print` statement occurs twice:
https://github.com/qiboteam/qibo/blob/28171d29c8654748d2b436338f1fa99de8ae7f92/src/qibo/parallel.py#L56-L57
but no error is raised.

I found two issues in the CodeText class:
- If the code starts with `"""`,  then the first piece will be an empty string, the second will be the content of a docstring, the third will be code, and so on.
  https://github.com/qiboteam/qibo/blob/28171d29c8654748d2b436338f1fa99de8ae7f92/src/qibo/tests/test_prints.py#L28
  On the other hand, if the code doesn't start with `"""`, the first piece will contain proper code, the second will contain a docstring, and so on.
  In both cases, docstrings are in odd position, so
  https://github.com/qiboteam/qibo/blob/28171d29c8654748d2b436338f1fa99de8ae7f92/src/qibo/tests/test_prints.py#L40-L44
  inverts docstrings with code (in files that starts with a docstring).
- If `print` occurs at the beginning of `piece`, than `i=0` but the code checks for `i>0`.
  https://github.com/qiboteam/qibo/blob/28171d29c8654748d2b436338f1fa99de8ae7f92/src/qibo/tests/test_prints.py#L62-L69

I edited the script accordingly and I added some tests to cover the CodeText class.

### DIFF
--- a/doc/source/contributing.rst
+++ b/doc/source/contributing.rst
@@ -46,6 +46,9 @@ The current code standards that are applied to any new changes:
 - **Coverage**: Test coverage should be maintained / be at least at the same level when new features are implemented.
 - **Pylint**: Test code for anomalies, such as bad coding practices, missing documentation, unused variables.
 
+Besides the linter, further custom rules are applied e.g. checks for ``print`` statements that bypass the logging system
+(such check can be excluded line by line with the ``# CodeText:skip`` flag).
+
 Documentation
 -------------
 
--- a/src/qibo/parallel.py
+++ b/src/qibo/parallel.py
@@ -53,8 +53,8 @@ class ParallelResources:  # pragma: no cover
                     # if copy is not implemented just use the original object
                     copy = obj
                 except Exception as e:
-                    # use print otherwise the message will not appear
-                    print('Exception in ParallelResources', str(e))
+                    # use print otherwise the message will not appear # CodeText:skip
+                    print('Exception in ParallelResources', str(e)) # CodeText:skip
                 args.append(copy)
             args = tuple(args)
             self._objects_per_process[pname] = args
--- a/src/qibo/tests/test_prints.py
+++ b/src/qibo/tests/test_prints.py
@@ -44,42 +44,32 @@ class CodeText:
 
         raise StopIteration
 
-    def get_line(self, i):
-        """Calculates line number of the identified `print`.
+    def check_exists(self, target_word):
+        """Checks if a word exists in the code text.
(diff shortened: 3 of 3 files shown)


scarrazza (commented):
@mlazzarin thanks for raising this issue.
The fix and test looks fine by me, I have opned #466 with the fix for parallel.py, we can merge that here and then proceed with the master.

mlazzarin:
Following #466, should we implement the possibility to exclude `print` statements line by line or simply exclude the whole `.py` file?

scarrazza:
Adding a exclusion filter for the parallel.py file should be sufficient for the moment.

mlazzarin:
Done. It is not perfect as files with the same name but different path are not discriminated by the exclusion filter e.g. another hypothetical parallel.py in a different folder may be excluded as well.
I added a comment in the code to highlight that. Let me know if a more robust implementation is required.

scarrazza:
Thanks, this should work, however I believe you could use the `qibo.__file__` to get the absolute location and skip tests only for those files. Could you please take a look at this option before we merge?

mlazzarin:
I'm not sure how I should use `__file__` because it always returns the path of the `test_prints.py` file, but I managed to use the `os.walk` returned value, I hope there are no drawbacks.

scarrazza:
@mlazzarin I believe you can take __file__ the absolute path and go back to the relative path.

mlazzarin:
I managed to do that with the os.walks() returned values but I run into cross-compatibility issues between platform (I'm not used to cross-platform software). Let me propose an alternative line by line filter which may be actually easier (I was already thinking about it), but if you prefer I'll dig into cross-platform path management.

scarrazza:
> Let me propose an alternative line by line filter which may be actually easier 

yes, fine by me

scarrazza:
I like the example with `# CodeText:skip`, I would avoid as much as possible printing docstrings so I don't see a serious plot with the current setup.

scarrazza (approved):
Thanks. Could you please write a section about this flag in the docs, e.g. contributing.rst?
I will then integrate this in #467.

mlazzarin:
Let me know if it is ok.

PR #466, opened 2021-09-12
removing print from parallel.py

scarrazza:
Fix issues spotted in #465.

### DIFF
--- a/src/qibo/parallel.py
+++ b/src/qibo/parallel.py
@@ -53,8 +53,9 @@ class ParallelResources:  # pragma: no cover
                     # if copy is not implemented just use the original object
                     copy = obj
                 except Exception as e:
-                    # use print otherwise the message will not appear
-                    print('Exception in ParallelResources', str(e))
+                    # otherwise the message will not appear
+                    from qibo.config import raise_error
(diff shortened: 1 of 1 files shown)


mlazzarin:
I'm afraid I cannot review this PR because I don't know how the whole parallel code works (yet).
However, I think it might introduce some subtle changes in the behavior of the code, because:
- raise_error now raises an additional exception w.r.t. print
- if I manually raise an error in the `try` body, now the tests `src/qibo/tests/test_parallel.py` get stuck forever, while with print they still work.

A comment hints to a specific reason for the usage of print. However I don't know ``multiprocessing`` well nor what the excepted behavior should be.

scarrazza:
Indeed, the exceptions and logs doesn't work well on a multiprocessing environment, thus the `print` status is the simplest think we can have in order to get some information about the exception.

I think the situation still complex if we decide to log from multiple threads in python.
So we may prefer to include an exclusion filter in #465, and skipping the check for this particular `print` statement.

mlazzarin:
> So we may prefer to include an exclusion filter in #465, and skipping the check for this particular `print` statement.

Ok, I'm on it.

scarrazza:
Ok, I am closing this in favour of #465.

PR #473, opened 2021-09-18
Rename ``Gate.on_qubits()`` to ``Gate._on_qubits()``

### LINKED ISSUE #472: Usage of Gate.on_qubits() method
Hi! I'm trying to understand how to use the ``qibo.abstractions.abstract_gates.Gate.on_qubits()`` method.
By reading the documentation, I understood that such method is used to obtain a gate that acts on some target qubits by cloning a gate that was originally created with different targets.

Indeed, in the following example it works
```python
from qibo import gates

c = Circuit(2)
gate = gates.H(0)
c.add(gate)
c.add(gate.on_qubits(1))
print(c().state())
```
but in this one it doesn't work
```python
from qibo import gates

c = Circuit(2)
gate = gates.H(1)
c.add(gate)
c.add(gate.on_qubits(0))
print(c().state())
```
```
  File "<stdin>", line 1, in <module>
  File "...qibo/src/qibo/abstractions/abstract_gates.py", line 214, in on_qubits
    gate = self.__class__(*qubits, **self.init_kwargs)
  File "...qibo/src/qibo/abstractions/abstract_gates.py", line 213, in <genexpr>
    qubits = (q[i] for i in self.qubits)
IndexError: tuple index out of range

```
Am I using the method properly?

mlazzarin:
Closes #472. I renamed ``Gate.on_qubits()`` to ``Gate._on_qubits()`` in
 - ``qibo.abstractions.abstract_gates.Gate`` (+ some comments)
 -  ``qibo.abstractions.abstract_gates.SpecialGate``
 -  ``qibo.abstractions.abstract_gates.Channel``
 - ``qibo.abstractions.gates.Unitary``
 - ``qibo.abstractions.circuit.AbstractCircuit.on_qubits()``

and I also modified the tests accordingly. I also made some minor fixes in the docstrings.
I should have found all occurences of such method. I'll ask for review as soon as I solve the problem with pylint.

### DIFF
--- a/.github/workflows/rules.yml
+++ b/.github/workflows/rules.yml
@@ -37,7 +37,6 @@ jobs:
       if: startsWith(matrix.os, 'windows')
       run: |
         python -m pip install --upgrade pip
-        pip install pylint
         pip install pytest-cov
         pip install .[qibojit,tests]
     - name: Test with pylint
--- a/src/qibo/abstractions/abstract_gates.py
+++ b/src/qibo/abstractions/abstract_gates.py
@@ -91,7 +91,7 @@ class Gate:
 
     @target_qubits.setter
     def target_qubits(self, qubits: Sequence[int]):
-        """Sets control qubits tuple."""
+        """Sets target qubits tuple."""
         self._set_target_qubits(qubits)
         self._check_control_target_overlap()
 
@@ -196,7 +196,9 @@ class Gate:
         return a or b
 
     def _on_qubits(self, *q) -> "Gate":
-        """Creates the same gate targeting different qubits.
+        """Helper method for :meth:`qibo.abstractions.circuit.AbstractCircuit.on_qubits`.
+
+        Creates the same gate targeting different qubits.
 
         Args:
             q (int): Qubit index (or indeces) that the new gate should act on.
@@ -205,6 +207,10 @@ class Gate:
             A :class:`qibo.abstractions.gates.Gate` object of the original gate
             type targeting the given qubits.
         """
+        # Note that q is interpreted as a map from the original qubits to the new ones, e.g.
+        # gates.CNOT(2, 3).on_qubits(0, 1, 2, 3) is equivalent to gates.CNOT(2, 3)
+        # gates.CNOT(2, 3).on_qubits(1, 2, 3, 0) is equivalent to gates.CNOT(3, 0)
+        # It is required for `len(q)` to be greater than the max qubit id of the original gate
         if self.is_controlled_by:
             targets = (q[i] for i in self.target_qubits)
             controls = (q[i] for i in self.control_qubits)
@@ -349,7 +355,7 @@ class Channel(Gate):
     def _on_qubits(self, *q): # pragma: no cover
         # future TODO
         raise_error(NotImplementedError, "`_on_qubits` method is not available "
-                                         "for the `GeneralChannel` gate.")
+                                         "for the `Channel` gate.")
 
 
 class ParametrizedGate(Gate):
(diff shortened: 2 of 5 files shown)


src/qibo/abstractions/abstract_gates.py
```diff
@@ -205,6 +207,10 @@ def on_qubits(self, *q) -> "Gate":
             A :class:`qibo.abstractions.gates.Gate` object of the original gate
             type targeting the given qubits.
         """
+        # Note that q is interpreted as a map from the original qubits to the new ones, e.g.
+        # gates.CNOT(2, 3)._on_qubits(0, 1, 2, 3) is equivalent to gates.CNOT(2, 3)
+        # gates.CNOT(2, 3)._on_qubits(1, 2, 3, 0) is equivalent to gates.CNOT(3, 0)
+        # It is required for `len(q)` to be greater than the max qubit id of the original gate
```
scarrazza, line 213: I believe this comment can be converted to an actual check.

src/qibo/abstractions/abstract_gates.py
```diff
@@ -205,6 +207,10 @@ def on_qubits(self, *q) -> "Gate":
             A :class:`qibo.abstractions.gates.Gate` object of the original gate
             type targeting the given qubits.
         """
+        # Note that q is interpreted as a map from the original qubits to the new ones, e.g.
+        # gates.CNOT(2, 3)._on_qubits(0, 1, 2, 3) is equivalent to gates.CNOT(2, 3)
+        # gates.CNOT(2, 3)._on_qubits(1, 2, 3, 0) is equivalent to gates.CNOT(3, 0)
```
scarrazza, line 212: Let me suggest to convert these comments to an example in the docstring of this function.

src/qibo/abstractions/abstract_gates.py
```diff
@@ -91,7 +91,7 @@ def _set_control_qubits(self, qubits: Sequence[int]):
 
     @target_qubits.setter
     def target_qubits(self, qubits: Sequence[int]):
-        """Sets control qubits tuple."""
+        """Sets target qubits tuple."""
```
scarrazza, line 94: Good catch!

src/qibo/abstractions/abstract_gates.py
mlazzarin: Ok, I can do it, but it would require something like ``if len(q) <= max(self.qubits): raise_error(...)`` to be executed for every call.
Is the overhead worth it? Because if it is called only by the ``AbstractCircuit.on_qubits()`` method, a more strict check is already performed once before calling ``Gate._on_qubits()`` for each gate in the circuit:
https://github.com/qiboteam/qibo/blob/9e0002072ec43d1d438c2a19733acadf122357b9/src/qibo/abstractions/circuit.py#L176-L181

scarrazza: Ok, that is a good point indeed. If the bad configurations are caught by the `AbstractCircuit.on_qubits` this hidden/private method can stay as it is.

.github/workflows/rules.yml
```diff
@@ -43,6 +42,7 @@ jobs:
         pip install .[qibojit,tests]
     - name: Test with pylint
       run: |
+        pip install pylint
```
mlazzarin, line 45: @scarrazza Could you explain why this edit solved the problem? I'm just curious, as I didn't manage to solve the problem by myself.

scarrazza: There is version conflict between tensorflow and pylint for type_extensions. So this last install overrides the bad version.

mlazzarin: Ok, thanks. Then I guess that line 40 is also redundant, isn't it?

scarrazza: Yes, thanks.

scarrazza (approved):

PR #474, opened 2021-09-22
Fix scalar mutliplication of ``SymbolicTerm``

### LINKED ISSUE #470: trotter evolution and dense evolution does not match
We should check the trotterization evolution code, we are observing non convergence with this approach, while with dense matrices the evolution works well.

stavros11:
Fixes #470, the fix was very simple and is in the first commit only. In the rest I am adding some unit tests that focus on the `core/terms.py` file and improving the related docstrings and comments to help in the future, because as it is now it was quite hard to debug this issue.


stavros11:
I simplified a little bit the `qibo.core.terms.SymbolicTerm` consturctor and added a new test file for the code in `terms.py`. I believe this should be good for review. The notebook that discovered this bug works for me now, but it would be good if someone else could confirm this too before merging.

src/qibo/core/terms.py
```diff
@@ -107,65 +111,72 @@ class SymbolicTerm(HamiltonianTerm):
...
+
+    Args:
+        coefficient (complex): Complex number coefficient of the underlying
+            term in the Hamiltonian.
+        factors (list): List of :class:`qibo.symbols.Symbol` that represent
+            the term factors.
+        matrix_map (dict): Dictionary that maps target qubit ids to a list of
+            matrices that act on each qubit.
+    """
+    """Helper constructor using the ``sympy`` expression directly.
```
scarrazza, line 124: I think there is something wrong with this comment, no?

src/qibo/core/terms.py
```diff
@@ -107,65 +111,72 @@ class SymbolicTerm(HamiltonianTerm):
             from qibo.core.terms import SymbolicTerm
             sham = X(0) * X(1) + 2 * Y(0) * Y(1)
             termsdict = sham.as_coefficients_dict()
-            sterms = [SymbolicTerm(c, f) for f, c in termsdict.items()]
+            sterms = [SymbolicTerm.from_factors(c, f) for f, c in termsdict.items()]
```
scarrazza, line 114: I think `from_factors` was removed, right?

src/qibo/core/terms.py
```diff
@@ -107,65 +111,72 @@ class SymbolicTerm(HamiltonianTerm):
...
+                # if the user is using ``symbol_map`` instead of qibo symbols,
+                # create the corresponding symbols
+                if symbol_map is not None and factor in symbol_map:
+                    from qibo.symbols import Symbol
+                    q, matrix = symbol_map.get(factor)
+                    factor = Symbol(q, matrix, name=factor.name)
+
+                if isinstance(factor, sympy.Symbol):
+                    if isinstance(factor.matrix, K.qnp.tensor_types):
+                        self.factors.extend(pow * [factor])
```
scarrazza, line 166: Given that the comments are quite good, please some sentence explaining how and why the pow term is applied to the matrix_map.

scarrazza (commented):
Thanks for this implementation. Works for me, seems to fix the non convergence observed in the notebooks. Please a look at my minor comments.

src/qibo/core/terms.py
stavros11: Yes, thanks. I was going to update the docstring after removing the `.from_factors` constructor for simplification but I forgot to complete the update. It should be ok now.

stavros11:
> Thanks for this implementation. Works for me, seems to fix the non convergence observed in the notebooks. Please a look at my minor comments.

Thank you for the review. I update the comments for the above points, let me know if it is ok.

scarrazza (approved):
Looks good, thank you.

PR #476, opened 2021-10-02
Remove sympy.expand call when constructing SymbolicHamiltonian

### LINKED ISSUE #469: sympy vs qibo symbols performance
We should verify the impact of the `expand` call in qibo.
It looks like this call deteriorates our performance in comparison to a pure sympy implementation.

stavros11:
Fixes #469. As discussed in the issue, the main bottleneck comes from the `sympy.expand` call that is used on the given sympy expression of the Hamiltonian form when a `SymbolicHamiltonian` is constructed. This call is useful when parsing the symbolic terms to [Qibo terms](https://github.com/qiboteam/qibo/blob/sympyexpand/src/qibo/core/terms.py). Given that the Qibo term representation of Hamiltonian is only useful when trotterizing, here I disable this calculation unless it is required (eg. when the user asks for trotterization). The idea is that Hamiltonians that cause #469 are non-local and therefore trotterization won't be useful for them anyway, so it will be unlikely that the user will get this performance issue.

The second part implemented here is a function that calulates the full Hamiltonian matrix directly from the given expression without the need to expand, because in the current implementation the Qibo terms are used which need the expanded form. In principle this should be straightforward as all we need is the `sympy.Expr.subs` method to substitute symbols with the corresponding matrices (numpy arrays), however this does not seem to work well when using `sympy.Symbol` instead of `sympy.MatrixSymbol`. I have been playing around a bit with sympy and trying to inherit Qibo symbols from `MatrixSymbol` instead of `Symbol` as it is now and here is a short summary of my understanding:

Pros of `MatrixSymbol`:
* Given that quantum operators have the properties of matrices (eg. are non-commuting), strictly speaking, this should be the correct way to go.
* Sympy algebra behaves better when substituting with arrays.

Cons of `MatrixSymbol`:
* New objects are created when used in operations. For example
```Python
import sympy

n = sympy.Symbol("n")
X = sympy.MatrixSymbol("X", n, n)
Y = sympy.MatrixSymbol("Y", n, n)
Z = X * Y
print(id(X), id(Y))
for t in Z.free_terms:
    print(id(t))
```
will print different ids if `MatrixSymbol` is used and same ids if `Symbol` is used. Not creating new objects is important when inheriting in Qibo.
* We would need to use an identity symbol instead of just `1` as `MatrixSymbol` do not like operations with numbers. This would break our current API, for example `2 + Z(0)` will no longer work and we would need to do something like `2 * I + Z(0)`.
* `MatrixSymbol` is non-commutative by default, while in `Symbol` we can select this property during creation.

Taking into account this, I ended up leaving the `Symbol` as it is and just writing [a function that does the substitution](https://github.com/qiboteam/qibo/blob/e5086b90b6dd7818b6113fecb5fc74d3e2b52bc5/src/qibo/core/hamiltonians.py#L317) with numpy arrays myself. This is kind of reinventing the wheel as this should have been possible using just `.subs`, but unfortunately is the cleanest solution I could find.

Following the last comment from #469, I also added the option to make Qibo symbols commutative using `commutative=True` during creation, although this is generally not recommended.

### DIFF
--- a/src/qibo/symbols.py
+++ b/src/qibo/symbols.py
@@ -7,12 +7,31 @@ class Symbol(sympy.Symbol):
     """Qibo specialization for ``sympy`` symbols.
 
     These symbols can be used to create :class:`qibo.core.hamiltonians.SymbolicHamiltonian`.
+    See :ref:`How to define custom Hamiltonians using symbols? <symbolicham-example>`
+    for more details.
+
+    Example:
+        ::
+
+            from qibo import hamiltonians
+            from qibo.symbols import X, Y, Z
+            # consturct an XYZ Hamiltonian on two qubits using Qibo symbols
+            form = X(0) * X(1) + Y(0) * Y(1) + Z(0) * Z(1)
+            ham = hamiltonians.SymbolicHamiltonian(form)
 
     Args:
         q (int): Target qubit id.
         matrix (np.ndarray): 2x2 matrix represented by this symbol.
         name (str): Name of the symbol which defines how it is represented in
             symbolic expressions.
+        commutative (bool): If ``True`` the constructed symbols commute with
+            each other. Default is ``False``.
+            This argument should be used with caution because quantum operators
+            are not commutative objects and therefore switching this to ``True``
+            may lead to wrong results. It is useful for improving performance
+            in symbolic calculations in cases where the user is sure that
+            the operators participating in the Hamiltonian form are commuting
+            (for example when the Hamiltonian consists of Z terms only).
     """
 
     def __new__(cls, q, matrix=None, name="Symbol", commutative=False):
--- a/src/qibo/symbols.py
+++ b/src/qibo/symbols.py
@@ -15,11 +15,11 @@ class Symbol(sympy.Symbol):
             symbolic expressions.
     """
 
-    def __new__(cls, q, matrix=None, name="Symbol"):
+    def __new__(cls, q, matrix=None, name="Symbol", commutative=False):
         name = "{}{}".format(name, q)
-        return super().__new__(cls=cls, name=name, commutative=False)
+        return super().__new__(cls=cls, name=name, commutative=commutative)
 
-    def __init__(self, q, matrix=None, name="Symbol"):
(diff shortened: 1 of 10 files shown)


src/qibo/symbols.py
```diff
@@ -15,11 +15,11 @@ class Symbol(sympy.Symbol):
             symbolic expressions.
     """
 
-    def __new__(cls, q, matrix=None, name="Symbol"):
+    def __new__(cls, q, matrix=None, name="Symbol", commutative=False):
```
scarrazza, line 18: Please document the new commutative argument. Ideally, would be really nice to have some example here or in the docs.

scarrazza (commented):
Thanks for this implementation and tests. I went through the different points you have raised and indeed, I like the current solution. I will start the review of this PR soon.

scarrazza (approved):
@stavros11 thanks for the implementation, looks good.
I assume that the performance improves when using the `commutative=True` flag for expressions which are compatible with commutation, correct?

stavros11:
> @stavros11 thanks for the implementation, looks good.

Thanks for the review. I added some documentation on the new `commutative` argument and also wrote a paragraph about it in the existing "How to define custom Hamiltonians using symbols?" example. Let me know if you agree with these changes.

> I assume that the performance improves when using the `commutative=True` flag for expressions which are compatible with commutation, correct?

Yes, calling `sympy.expand` to an expression consisting of default `sympy.Symbol` has exactly the same performance with creating the same expression using Qibo symbols and `commutative=True`. If `commutative=False` then expand takes much longer but this is true both for sympy and Qibo symbols (which are essentially the same, just that qibo implements a few additional methods for calculating the matrices).

Note however that just creating a symbolic Hamiltonian using `ham = hamiltonians.SymbolicHamiltonian(symbolic_form)` does not use call `sympy.expand` anymore and it takes negligible time regardless of whether the given form contains sympy, Qibo, commutative or non-commutative symbols. Expand is only used when trotterizing, so there using `commutative=True` may lead to some difference but I am not sure how big this will be for local Hamiltonians.

scarrazza:
Perfect, thanks looks good to me.

PR #478, opened 2021-10-02
drawing Unitary gates

### LINKED ISSUE #477: Draw Unitary gate
Currently it is not possible to use the `.draw()` method on a circuit that contains a `gates.Unitary` gate. I think it would be useful to have this feature in Qibo, and I guess it is not difficult to implement.  I encountered this issue while working on a current project. The typical use for me of the `.draw()`  method is to make sure that the circuit I'm building is the desired one.

So here is a small code to reproduce the issue:
```
from qibo import gates
import numpy as np

c = Circuit(3)
u = np.eye(8)
c.add(gates.Unitary(u,0,1,2))
print(c.draw())
```
which gives
`NotImplementedError: Unitary gate is not supported by circuit.draw`

I would propose that we could draw something like:
```
q0: -U-
q1: -U-
q2: -U-
```

scarrazza:
Closes #477.

### DIFF
--- a/src/qibo/abstractions/circuit.py
+++ b/src/qibo/abstractions/circuit.py
@@ -962,7 +962,7 @@ class AbstractCircuit(ABC):
                   "crx": "RX", "cry": "RY", "crz": "RZ",
                   "cu1": "U1", "cu3": "U3", "ccx": "X",
                   "id": "I", "measure": "M", "fsim": "f",
-                  "generalizedfsim": "gf"}
+                  "generalizedfsim": "gf", "Unitary": "U"}
 
         # build string representation of gates
         matrix = [[] for _ in range(self.nqubits)]


scarrazza:
@DiegoGM91 please let me know if this works for you.

stavros11:
Just noting that this seems generally ok to me, for example
```Python
import numpy as np
import qibo
qibo.set_backend("numpy")

c = models.Circuit(6)

c.add(gates.Unitary(np.random.random((8, 8)), 0, 2, 3))
c.add(gates.Unitary(np.random.random((2, 2)), 2))
c.add(gates.Unitary(np.random.random((2, 2)), 4))

print(c.draw())
```
gives
```
q0: ─U───
q1: ─|───
q2: ─U─U─
q3: ─U───
q4: ───U─
q5: ─────
```
which I believe is acceptable, however there is a small discrepancy in the following:
```Python
c = models.Circuit(2)
c.add(gates.Unitary(np.random.random((4, 4)), 0, 1))
c.add(gates.Unitary(np.random.random((2, 2)), 0))
c.add(gates.Unitary(np.random.random((2, 2)), 1))
```
which gives
```
q0: ─U─U─
q1: ─U─U─
```
so when the unitary acts on neighboring qubits, there is no distinction between a two-qubit (4x4) unitary and two one-qubit (2x) unitaries.

scarrazza:
Thanks, good point. However, this spoils that Unitary does not sotre targets and controls as other gates, so it fallsback to the fSim and similar gates. Not a big deal, but can be annoying.

DiegoGM91 (approved):
@scarrazza this works for me. Thanks!

PR #480, opened 2021-10-05
Fix some examples in docstrings

mlazzarin:
In this PR I fixed some broken examples in the docstrings. I haven't made an exhaustive search so some examples may still be broken.

Let me know if it is all ok, in particular with the ``VariationalLayer`` (I had to change a lot to fix it)


scarrazza (approved):
Thanks @mlazzarin looks good to me.

PR #481, opened 2021-10-05
Update benchmark docs

stavros11:
Updates the instructions on how to execute the scripts under `examples/benchmarks` by revising the arguments and adding missing sections (eg. for the time evolution benchmark). I also removed the `SupremacyCircuit` as I do not believe it is correct.

We should have a second pass in this once we have some progress with the new benchmarks and also update the tabulated benchmark results in the docs. Perhaps it would be preferrable to completely remove the benchmarks from the qibo repository and have a seperate benchmark repository which also allows comparing with other libraries. This could also serve as "proof" since we cite such results in our docs.


scarrazza (approved):
Thanks, I went through all points, and looks fine to me.
I agree that as soon as we perform new benchmarks we should move all that to a new docs/repo.

PR #467, opened 2021-09-12
updating documentation

scarrazza:
Improving documentation layout and content. Closes #464.

PR #482, opened 2021-10-05
adding docker images for tags

scarrazza:
This implements a simple docker image with qibo installed for each specific tag version.
Here an example for the 0.1.5 release: https://github.com/qiboteam/qibo/pkgs/container/qibo

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,5 +1,5 @@
 # A single CI script with github workflow
-name: Build wheels
+name: Build wheels and deploy
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


stavros11 (approved):
I have also never used docker before so I cannot comment much on the implementation, however I tried the 0.1.5 image on my machine and it works as expected. My only questions/comments:

* It would be great to add some information about this in the docs mainly because the installation procedure is very easy even for someone without prior docker experience, if we write exactly the few commands one needs to execute.
* I noticed that the docker image for 0.1.5 is quite big in size (2.2GB). Is this because it contains Tensorflow?
* Do we plan to release images containing some custom backend (eg. qibojit)? Because qibo 0.1.6 will come with the numpy backend only.
* Tensorflow installed via the docker image 0.1.5 did not detect my GPU. I am not sure if there is an issue specific to my machine or if it is more complicated to include GPU support via docker.

scarrazza:
>    It would be great to add some information about this in the docs mainly because the installation procedure is very easy even for someone without prior docker experience, if we write exactly the few commands one needs to execute.

I have included some docs in https://github.com/qiboteam/qibo/pull/467/commits/17144ede33869afdcf0b0052cfce14e14bf28539. Please let me know if that is sufficient.

>    I noticed that the docker image for 0.1.5 is quite big in size (2.2GB). Is this because it contains Tensorflow?

Partially, the ubuntu:latest image is already big as soon as you install some packages. We could use another baseline lightweight system, however I believe the ubuntu image is more popular.

>    Do we plan to release images containing some custom backend (eg. qibojit)? Because qibo 0.1.6 will come with the numpy backend only.

We should discuss about it during the meeting today. In principle we could install multiple backends.

>    Tensorflow installed via the docker image 0.1.5 did not detect my GPU. I am not sure if there is an issue specific to my machine or if it is more complicated to include GPU support via docker.

Yes, that's normal, we have to install the cuda drivers on the docker image (+ some extra fine tuning), for the moment I don't think we need something very advanced. In fact, my idea was to build something very simple and check if there is interested from the community.

Issue #483, opened 2021-10-06
implement doctests


scarrazza:
Following our discussion, we should test the code examples in docstrings.
One possibility is to use https://docs.python.org/3/library/doctest.html

andrea-pasquale:
Of course, I will look into it.

stavros11:
> where the line `print(c.draw())` is compared to the drawing of the circuit below when executing the code.

I guess the initial idea was to just check if the code cell works, eg. if it executes or raises any error but not necessarily check the output. This would be a good first step on this. If you can also check the output this would of course be even better but may be much harder to implement as you might find cases in the examples where we don't specify the output and even when we do there is no specific structure to where or how we specify.

> I ask this because in the case of tensors the output depends on the backend used, therefore the example in the docstring must specify correctly the backend and the printed tensor.

Generally in tests if you like to check whether two tensors agree you can use
```Python
from qibo import K
K.assert_allclose(final_tensor, target_tensor)
```
which should work for any backend and regardless of the types. For example `final_tensor` can be `tf.Tensor` while `target_tensor` is `np.ndarray` and this will still pass as long as they have the same shape and values. Following what I wrote above though, I am not sure if this can be incorporated above, for example if you want to compare printed tensors then that's a different story.

mlazzarin:
If I understand correctly, in order to use the **doctest** library it is required to edit all examples in both the source code and the documentation itself.
I guess the rendered docs would also change as well.
It may be tedious, so I would suggest to implement such tests only for a small portion of the code, then maybe open a PR so that we can discuss if it's actually what we want or look for alternatives (maybe this https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html)

mlazzarin:
Shall we close this issue?

scarrazza:
Yes, lets close it now. Thanks.

PR #487, opened 2021-10-08
adding doctest

andrea-pasquale:
This PR aim at closing #483.
In this first commit I modified only docstrings in `src/qibo/abstractions/` as a first attempt. In order to test whether the code examples in the docstrings are working every python file must be executed using the `-m doctest` option. I tested the modified files on my local computer and the test need to be added to qibo.
Let me know if this is ok.
If so, I can start modifying the other docstrings.

### DIFF
--- a/.github/workflows/rules.yml
+++ b/.github/workflows/rules.yml
@@ -47,7 +47,7 @@ jobs:
       run: |
         pytest --cov=qibo --cov-report=xml --pyargs qibo
     - name: Test documentation examples
-      if: startsWith(matrix.os, 'ubuntu') && matrix.python-version == '3.9'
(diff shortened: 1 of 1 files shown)


scarrazza:
@andrea-pasquale thanks, I have two questions:
- does this changes in the docstring works well with sphinx?
- we use pytest for all qibo tests, is it possible to integrate the approach presented with pytest?

scarrazza:
@andrea-pasquale ok, please go ahead.

andrea-pasquale:
I updated also the docstrings in the documentation.
The only thing left is the printed tensors, I temporarily commented all those lines.
I looked for a solution to skip those lines in the test but I couldn't find.
There are also a few errors which I will show.

PR #496, opened 2021-10-14
Reduce import qibo time

stavros11:
As discussed in the benchmarks, qibo's import time is large because tensorflow is imported regardless of the used backend. Here I clean up the backend initialization and avoid importing modules that are not used. I also sorted the backends in `profile.yml` in order of default preference, so if the specified default backend is not installed the first available from this list will be the default. In principle we could remove the default entry completely and use the first backend specified in the profile as default.

Here are some import time measurements from qibomachine:

configuration | import time (master) | import time (this branch)
-- | -- | --
qibo only | 0.27728 | 0.27459
qibo + qibojit | 0.63503 | 0.61032
qibo + qibojit + qibotf | 1.49840 | 0.61790
qibo + qibojit with GPU | 0.56741 | 0.54582
qibo + qibojit + qibotf with GPU | 1.76882 | 0.55866
qiskit | 0.67056 |
cirq | 1.41273 |
tfq | 2.36862 |
qulacs | 0.00495 |

Unfortunately the difference with qiskit is not sufficient to compensate the dry run but the total difference will be a few miliseconds which I believe won't be relevant in most cases.

### DIFF
--- a/src/qibo/backends/__init__.py
+++ b/src/qibo/backends/__init__.py
@@ -17,7 +17,10 @@ class Backend:
         except FileNotFoundError:  # pragma: no cover
             raise_error(FileNotFoundError, f"Profile file {profile_path} not found.")
 
+        # dictionary to cache if backends are available
+        # used by ``self.check_availability``
         self._availability = {}
+        
         # create numpy backend (is always available as numpy is a requirement)
         if self.check_availability("numpy"):
             from qibo.backends.numpy import NumpyBackend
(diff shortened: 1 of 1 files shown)


stavros11:
> Could you please post this observation in the other repo for benchmarks?

Yes, I will post some new plots there where I compare all relevant times (import, creation, dry run and simulation) between various libraries, as well as the total time which is what the user will get in the end when he runs a script. I am using this updated branch for qibo.

PR #498, opened 2021-10-19
adding qgan class

bpcarlos:
to be finished


bpcarlos:
@scarrazza @stavros11 I updated the comments of the code. I should be ready to go, modulo possible code modifications.

scarrazza:
Here some points we have discussed today:
- [ ] remove `set_backend('tensorflow')`, add exception in `__init__`
- [ ] update the docstring example
- [ ] make models optional arguments
- [ ] change interface to match scikit model/train/prediction syntax
- [ ] add docs to the sphinx documentation

stavros11 (commented):
I would add the tests to the above list.

Regarding the hard-coded models for the quantum generator and the classical discriminator, we could have two arguments in the `__init__`, for example a `generator` which accepts a qibo circuit and a `discrimintator` which accepts a keras sequential model and their default value is `None`. Then in `execute` we do:
```Python
if self.discriminator is None:
    discriminator = self.define_discriminator()
```
and the same for the generator circuit. This keeps the default behavior as it is now but adds more flexibility by allowing the user to define custom models. We may need to cleanup/document some of the current arguments if we follow this approach, for example let the user know that `layers` will be neglected if a custom generator is given.

bpcarlos:
@scarrazza @stavros11 I believe I made most of the changes. Please, check them out. We are currently missing the sphinx documentation and tests.

stavros11:
I made some updates in order to fix the tests and coverage, this should be good for another review now. 

@scarrazza, @bpcarlos what do you think about moving the parameters associated with training, such as `initial_params`, `batch_samples`, `n_epochs` and `lr` to the `fit` method instead of `__init__` and not keeping them as class attributes? This would simplify the code a bit and would make it easier to reuse the same model with various training configurations. For example one could try training the same model with many different learning rates, which is typical in machine learning. It is also more natural to pass only parameters that define the model (such as `latent_dim`, `layers` etc.) in `__init__` and define everything else parameters later.

scarrazza:
@stavros11 thank for you including the missing features.

>  ... to the fit method instead of __init__ and not keeping them as class attributes? 

I absolutely agree, we should follow as close as possible the scikit prescription.

bpcarlos:
@stavros11 @scarrazza I updated the model and tests implementing Stavros' suggestion. Now the parameters associated with the training are given to the `fit` method.

scarrazza:
@bpcarlos thank you for the changes. In order to conclude this PR once and for all, we still need to fix the coverage and include an entry in the docs.

stavros11:
> @bpcarlos thank you for the changes. In order to conclude this PR once and for all, we still need to fix the coverage and include an entry in the docs.

I fixed the coverage and added the model to the docs. I’m not sure if we would like to include an example of this, if no this should be good.

scarrazza:
@stavros11 thanks for fixing the missing points, the implementation looks good to me.

> I fixed the coverage and added the model to the docs. I’m not sure if we would like to include an example of this, if no this should be good.

Let me suggest to merge this PR now, and delegate to @bpcarlos the decision to open another PR with a tutorial based on this example.

scarrazza (approved):

PR #501, opened 2021-11-01
Remove some hard-coded ``np.linalg`` calls from tests

mlazzarin:
In this PR and in https://github.com/qiboteam/qibojit/pull/36 I am trying to fix issue https://github.com/qiboteam/qibojit/issues/34 concerning tests with AMD GPUs.

In particular, there are two tests in Qibo with calls to ``np.linalg.eigh`` and ``np.linalg.eigvalsh``.
When these calls are performed with a CuPy object as argument, it seems that somehow there is a automatic call to ``cp.linalg.eigh`` or ``cp.linalg.eigvalsh`` (I think that this behavior is very confusing).
The problem is that this behavior bypasses the fallback mechanism implemented in https://github.com/qiboteam/qibojit/pull/36 and a ``rocblas_not_implemented_status`` error is raised.

The easiest fix is to replace the hard-coded call to NumPy with the backend ``K`` in these tests.
However, I wonder if there was a specific reason to use NumPy in these tests, e.g. to compare the results with ``K``.
@scarrazza @stavros11 what do you think?

### DIFF
--- a/src/qibo/tests/test_core_callbacks.py
+++ b/src/qibo/tests/test_core_callbacks.py
@@ -323,7 +323,7 @@ def test_gap(backend, dense, check_degenerate):
     ham = lambda t: (1 - t) * h0.matrix + t * h1.matrix
     targets = {"ground": [], "excited": [], "gap": []}
     for t in np.linspace(0, 1, 11):
-        eigvals = np.linalg.eigvalsh(ham(t)).real
+        eigvals = K.eigvalsh(ham(t)).real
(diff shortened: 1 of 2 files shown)


stavros11 (approved):
Thanks for this update. I do not see any particular reason for using numpy explicitly here, probably using `K` is a better choice. It may be a "leftover" from before we implemented the backend mechanisms in qibo.

Also, I am not sure if you already looked on this but the same modification should fix #502.

PR #500, opened 2021-10-25
Remove custom backend classes

stavros11:
Following what is discussed in #499, this PR removes the `TensorflowCustomBackend` and `JITCustomBackend` classes from qibo. These classes will be moved to the qibotf and qibojit repositories respectively.

TODO:
- [x] Complete migration to qibojit. Implemented in qiboteam/qibojit#35.
- [x] Complete migration to qibotf. Implemented in qiboteam/qibotf#12
- [x] Move any hard-coded regressions, for example these that are found in tests based on random number instances, to the respective backends. @mlazzarin this should simplify the ROCm testing issues once fixed.
- [ ] (?) Find a way for CI to work before releasing.

The general goal of this is to enable making changes in the backends without having to touch the qibo code, so any other simplification contributes to this should be considered here.

PR #503, opened 2021-11-06
Fix dtype of CZ matrix

mlazzarin:
I may have found an unwanted behavior. The dtype of CZ matrix in ``src/qibo/backends/matrices.py`` is ``int64`` but I think it should be equal to the ``dtype`` property of ``qibo.backends.matrices.Matrices`` class.


stavros11 (commented):
Thanks for finding this. Looking at the [implementation of `self.backend.cast`](https://github.com/qiboteam/qibo/blob/3cbdcf34f979de72b7d04d4fabea459b939b2e2f/src/qibo/backends/numpy.py#L47) for the numpy and other backends, I believe that the final array, here `self._CZ` will have the backend complex type regardless of the type of the original array `m`, right? If this is true the `self.dtype` used in all matrix definitions here is essentially useless.

I am not sure why the macos tests fail though.

stavros11 (approved):

scarrazza (approved):

mlazzarin:
Given that we merged #500, I changed the base branch to ``multiqubit``. Shall we merge this?

scarrazza:
@mlazzarin, yes, sure. I was planning to merge directly to master, but we can instead merge this in `multiqubit`.

PR #475, opened 2021-10-01
first AAVQE added

nicolezattarin:
A basic model for Adiabatically Assisted VQE

PR #507, opened 2021-11-08
Fix AAVQE example with qibotf

### LINKED ISSUE #502: AAVQE example is not working with qibotf
If I run that example with qibotf installed but not with qibojit I got the following error
``AttributeError: 'tensorflow.python.framework.ops.EagerTensor' object has no attribute 'real'``
 in line 62:
https://github.com/qiboteam/qibo/blob/7f9687a935eb11381a620835cc58cb41db95e3ab/examples/aavqe/main.py#L60-L63

mlazzarin:
Closes #502.

### DIFF
--- a/examples/aavqe/main.py
+++ b/examples/aavqe/main.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 import numpy as np
-from qibo import gates, models, hamiltonians
+from qibo import gates, models, hamiltonians, K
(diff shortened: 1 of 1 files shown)


scarrazza (approved):
@mlazzarin thanks, looks good to me.

stavros11 (approved):
Looks good, just to keep in mind that once we merge this, we should merge the new master to #475 and fix the conflict (use the `K.real` version there).

PR #508, opened 2021-11-08
Limiting qibotf actions for py36

scarrazza:


stavros11 (approved):
Looks good to me, thanks.

PR #509, opened 2021-11-10
Remove hardware backend

stavros11:
Following #500, this removes the hardware backend class which will be moved in the respective repository for flexibility. In addition, I implemented to new methods in the `AbstractBackend`: `circuit_class` and `create_gate`. These simplify the selection mechanism between simulation and hardware circuit and gate objects.


andrea-pasquale (approved):
Looks good to me.

src/qibo/backends/__init__.py
```diff
@@ -36,7 +36,8 @@ def __init__(self):
         if not self.check_availability(default_backend):  # pragma: no cover
             for backend in self.profile.get('backends'):
                 name = backend.get('name')
-                if self.check_availability(name):
+                if self.check_availability(name) and not backend.get('is_hardware'):
+                    # hardware backends cannot be default
```
scarrazza, line 40: I am not sure I understand why this cannot be the default. By design it should be possible to have a lab server running exclusively qibolab. Maybe I am missing something?

src/qibo/backends/abstract.py
```diff
@@ -122,6 +118,42 @@ def matrices(self):
...
+                                                 "implemented for density "
+                                                 "matrices.")
+            from qibo.core.circuit import DensityMatrixCircuit
+            return DensityMatrixCircuit
+        elif accelerators is not None:
+            from qibo.core.distcircuit import DistributedCircuit
+            return DistributedCircuit
+        else:
+            from qibo.core.circuit import Circuit
+            return Circuit
```
scarrazza, line 147: Maybe include some comment here talking saying that `qibo.core.circuit` is for simulation.

src/qibo/backends/profiles.yml
```diff
@@ -13,9 +13,9 @@ backends:
     from: qibo.backends.tensorflow
 
   # hardware backends
-  - name: qiboicarusq
-    class: IcarusQBackend
-    from: qibo.backends.hardware
+  - name: qibolab
+    class: QibolabBackend
+    from: qibolab.backend
```
scarrazza, line 18: Just a question about this point that I forgot to discuss. What do you think if we change this with:
```yaml
- name: qibolab
  driver: qibolab.backend.QibolabBackend
```

scarrazza (approved):
Thanks @stavros11 looks good to me, please have a look at my minor comments.

src/qibo/backends/__init__.py
stavros11: The only reason I added this here is that there are some tests in qibo for which the backend is not selected, so they fall back to the default backend, and use features that are not available in the hardware backend, such as density matrices, channels, etc.. So if someone uses an environment with hardware backends and no simulation backends other than numpy and the hardware backend is used as default, qibo tests will fail.

Summarizing this is only for development convenience. Once the hardware backend is developed enough to have all features it can support and proper testing mechanism and qibo tests are flexible enough to skip testing features that are not supported by hardware then I do not see any issue with setting hardware backend as default.

src/qibo/backends/abstract.py
stavros11: I added some comments here and in `create_gates` and also for the setting hardware backend as default point from above. Let me know if you agree.

src/qibo/backends/profiles.yml
stavros11: I am fine with that, particularly given that `from` and `class` are Python words so it may be preferrable to avoid them. If you like I can open a new PR implementing this. If I understand correclty you mean substituting both the `from` and `class` entries with one `driver` entry, right?

src/qibo/backends/__init__.py
scarrazza: Ok, thanks.

src/qibo/backends/profiles.yml
scarrazza: Indeed, sounds reasonable, and makes sense to update this.

andrea-pasquale:
I agree. Looks good to me.

PR #510, opened 2021-11-11
check minimum version for backends

### LINKED ISSUE #506: add minimum version for qibojit/qibotf
We should include a minimum version check for all backends.
We have some simple mechanism for tf but I believe this information could be stored directly in the `profiles.yml`.

scarrazza:
Closes #506.

### DIFF
--- a/src/qibo/backends/__init__.py
+++ b/src/qibo/backends/__init__.py
@@ -156,7 +156,7 @@ class Backend:
         """
         if module_name not in self._availability:
             from pkgutil import iter_modules
-            from importlib_metadata import version
+            from importlib.metadata import version
             is_available = module_name in (name for _, name, _ in iter_modules())
             if check_version:
                 minimum_version = self._backends_min_version.get(module_name)
--- a/src/qibo/backends/__init__.py
+++ b/src/qibo/backends/__init__.py
@@ -158,7 +158,7 @@ class Backend:
             from pkgutil import iter_modules
             from importlib_metadata import version
             is_available = module_name in (name for _, name, _ in iter_modules())
-            if check_version:
+            if is_available and check_version:
                 minimum_version = self._backends_min_version.get(module_name)
                 if version(module_name) < minimum_version:
                     raise_error(ModuleNotFoundError, f"Please upgrade {module_name}. "
--- a/src/qibo/backends/__init__.py
+++ b/src/qibo/backends/__init__.py
@@ -22,7 +22,7 @@ class Backend:
         self._availability = {}
 
         # create numpy backend (is always available as numpy is a requirement)
-        if self.check_availability("numpy"):
+        if self.check_availability("numpy", check_version=False):
             from qibo.backends.numpy import NumpyBackend
             self.qnp = NumpyBackend()
         else:  # pragma: no cover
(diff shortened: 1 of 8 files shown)


src/qibo/backends/__init__.py
```diff
@@ -22,25 +22,37 @@ def __init__(self):
...
             self.qnp = NumpyBackend()
         else:  # pragma: no cover
             raise_error(ModuleNotFoundError, "Numpy is not installed. "
                                              "Please install it using "
                                              "`pip install numpy`.")
 
+        # loading backend names and version
+        self._backends_min_version = {}
+        for backend in self.profile.get("backends"):
+            self._backends_min_version[backend.get("name")] = backend.get("minimum_version")
```
stavros11, line 36: Shorter alternative:
```Python
self._backends_min_version = {backend.get("name"): backend.get("minimum_version") for backend in self.profile.get("backends")}
```

src/qibo/backends/profiles.yml
```diff
@@ -3,19 +3,23 @@ backends:
   - name: qibojit
     class: JITCustomBackend
     from: qibojit.custom_operators
+    minimum_version: 0.0.4.dev0
```
stavros11, line 6: Do we need to use "dev0" here and or would just "0.0.4" be sufficient? Same for qibotf. Also, I think the pre-relased version of qibojit and qibotf is "rc0" no "dev0", right?

src/qibo/backends/__init__.py
```diff
@@ -132,18 +144,25 @@ def __repr__(self):
     def show_config(self):
         log.info(f"Using {self} backend on {self.active_backend.default_device}")
 
-    def check_availability(self, module_name):
+    def check_availability(self, module_name, check_version=True):
         """Check if module is installed.
 
         Args:
             module_name (str): module name.
+            check_version (str): check if minimum version.
```
stavros11, line 152:
```suggestion
            check_version (str): check if module has the required minimum version. A `ModuleNotFoundError` is raised if no.
```

src/qibo/backends/__init__.py
```diff
@@ -132,18 +144,25 @@ def __repr__(self):
...
         Returns:
             ``True`` if the module is installed, ``False`` otherwise.
         """
         if module_name not in self._availability:
             from pkgutil import iter_modules
+            from importlib_metadata import version
             is_available = module_name in (name for _, name, _ in iter_modules())
+            if is_available and check_version:
+                minimum_version = self._backends_min_version.get(module_name)
+                if version(module_name) < minimum_version: # pragma: no cover
```
stavros11, line 163: In principle the `minimum_version` entry in profiles.yml is optional and the user can add a backend there without specifiying this, for example if any version of this backend is supported by qibo. Should we add an exception to the version check in this case? I believe
```suggestion
                if minimum_version is not None and version(module_name) < minimum_version: # pragma: no cover
```
would work.

src/qibo/backends/__init__.py
```diff
@@ -22,25 +22,37 @@ def __init__(self):
...
+
+        # check if default backend is described in the profile file
+        if default_backend != "numpy":
+
+            if default_backend not in (backend.get("name") for backend in self.profile.get("backends")): # pragma: no cover
+                raise_error(ModuleNotFoundError, f"Default backend {default_backend} not set in {profile_path}.")
+
+            # change the default backend if it is not available
+            if not self.check_availability(default_backend):  # pragma: no cover
+                for backend in self.profile.get('backends'):
```
stavros11, line 49: I should have done this in a previous PR but perhaps we could add a comment here explaining that this loop sets as default backend the first available in the order the backends are defined in profiles.yml and if none of these backends is available then we fall back to numpy.

stavros11 (commented):
Thanks for implementing this, please find a few comments below. The only breaking issue I find is that when I do 
```Python
from importlib_metadata import version
version("tensorflow")
``` 
I get the following error:
```
importlib_metadata.PackageNotFoundError: No package metadata was found for tensorflow
```
Have you observed anything like this? Given that it does not appear in CI, it may be an issue with my configuration. `version` works for other libraries, such as `qibo`, `qibojit` and `qibotf`.

src/qibo/backends/__init__.py
scarrazza: Thanks.

src/qibo/backends/profiles.yml
scarrazza: Indeed, I have just fixed the versioning of qibojit and qibotf, given that after 0.0.4rc0 the next natural version is 0.0.4rc1.dev1. I have just included this fix for both repos, and for qibo in this PR.

src/qibo/backends/__init__.py
scarrazza: Thanks.

src/qibo/backends/__init__.py
scarrazza: Thanks. I have also included the versioning parser from packaging because the string comparison doesn't work always.

src/qibo/backends/__init__.py
scarrazza: Thanks, fixed.

stavros11 (approved):
The changes look good to me, we just need to be consistend with the rc/dev notation to avoid errors during development. The `importlib_metadata.PackageNotFoundError` for tensorflow remains but it appears to be independent of qibo as I can reproduce it in an environment without qibo, just by installing tensorflow.

scarrazza:
> The importlib_metadata.PackageNotFoundError for tensorflow remains but it appears to be independent of qibo as I can reproduce it in an environment without qibo, just by installing tensorflow.

ok, thanks for checking. I can't reproduce it, so if you have some conda env setup I can give a try.

PR #511, opened 2021-11-11
updating docs for backends

scarrazza:


stavros11 (approved):
Thanks for catching this. Perhaps it would be helpful to have a longer description of qibolab here but I am not sure exactly what we could write at this stage and I guess that will be easier to write once qibolab is further developed.

PR #513, opened 2021-11-17
disabling py36

scarrazza:
Here we remove python3.6 support and simplify some installation targets.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,12 +64,8 @@ setup(
     extras_require={
         "docs": ["sphinx", "sphinx_rtd_theme", "recommonmark", "sphinxcontrib-bibtex", "sphinx_markdown_tables", "nbsphinx", "IPython"],
         "tests": ["pytest", "cirq", "ply", "sklearn"],
-        # Backends dependencies
-        "qibotf": ["qibotf"],
-        "qibojit": ["qibojit"],
-        "tensorflow": [f"tensorflow>=2.2.0"],
     },
-    python_requires=">=3.6.0",
+    python_requires=">=3.7.0",
     long_description=long_description,
     long_description_content_type='text/markdown',
 )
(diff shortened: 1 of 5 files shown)


mlazzarin (approved):
Looks good to me.
By the way, given that we are fixing the docs here, there are two sentences that need to be fixed in line 56 and 122 of installation.rst (regarding the conda packages)

setup.py
```diff
@@ -64,12 +64,8 @@ def get_version():
     extras_require={
         "docs": ["sphinx", "sphinx_rtd_theme", "recommonmark", "sphinxcontrib-bibtex", "sphinx_markdown_tables", "nbsphinx", "IPython"],
         "tests": ["pytest", "cirq", "ply", "sklearn"],
-        # Backends dependencies
-        "qibotf": ["qibotf"],
-        "qibojit": ["qibojit"],
-        "tensorflow": [f"tensorflow>=2.2.0"],
     },
-    python_requires=">=3.6.0",
+    python_requires=">=3.7.0",
```
stavros11, line 68: I guess with this change it will not be possible to install qibo at all in py3.6. In principle we could relax this given that for the most part, excluding some tensorflow and scipy stuff which are not hard requirements, qibo still works fine on py3.6. I am not sure if it is a good practice to support a version that we don't test and we don't include in the docs.

stavros11 (approved):
Thanks for implementing this, looks good. 

My only comment is that according to the [backend/OS support table](https://qibo.readthedocs.io/en/latest/getting-started/installation.html#operating-systems-support) we provide in the docs `tensorflow` backend should work on Windows however we are currently not testing for it. In the current rules.yml tensorflow is only installed through the qibotf, which is deactivated for Windows, right?

setup.py
scarrazza: I agree, but given that py36 is EOD, if we don't have a strong motivation (= technical need), we can simply skip this requirement and instead, include py3.10 as soon as it is supported by our backend dependencies.

scarrazza:
@mlazzarin @stavros11 thanks, I have implemented the points you have raised.

PR #515, opened 2021-11-19
Minor fixes in the docs

mlazzarin:
The conda instructions for qibo and qibojit were swapped. I also updated the instructions for qibojit, given that now it depends on qibo.


scarrazza (approved):
Thanks!

PR #518, opened 2021-11-22
Add AAVQE example to doctests

mlazzarin:
I added the AAVQE example to the doctests, fixed some typos, and removed some hardcoded calls to NumPy methods that make tests crash with CuPy.


scarrazza (approved):
Looks good to me.

mlazzarin:
@scarrazza what's your opinion about keeping the old main example ``main_old.py``. Is it really necessary?

scarrazza:
I think this file should be removed.

PR #520, opened 2021-11-24
Fix CI running twice

mlazzarin:
In #475 a trigger on pull requests was added to the workflows.
I think we should remove it otherwise the CI runs twice for each push to a branch with an open PR.


scarrazza:
Good point, however this was needed in order to allow fork PRs to run CI, e.g. AAVQE PR #475, but indeed there should be some better approach.

stavros11 (approved):

PR #521, opened 2021-11-24
State in computational basis

### LINKED ISSUE #519: State in computational basis
We should provide a state option to print results in the computational basis.

stavros11:
Closes #519. We provide the following options:
```Python
c = models.Circuit(5)
c.add(gates.H(1))
result = c()

print(result) # (0.70711+0j)|00000> + (0.70711+0j)|01000>
print(result.state(symbolic=True)) # (0.70711+0j)|00000> + (0.70711+0j)|01000>
print(result.symbolic(decimals=1)) # (0.7+0j)|00000> + (0.7+0j)|01000> 
print(result.symbolic(decimals=2)) # (0.71+0j)|00000> + (0.71+0j)|01000>
```
The `decimals` argument controls the number of decimals and the default is 5. Zero amplitudes are ignored.

@scarrazza, @igres26 let me know if you agree. I will implement something similar for density matrices.

### DIFF
--- a/src/qibo/abstractions/states.py
+++ b/src/qibo/abstractions/states.py
@@ -69,6 +69,14 @@ class AbstractState(ABC):
         self._tensor = x
 
     @abstractmethod
+    def symbolic(self, decimals=5):  # pragma: no cover
+        """Dirac notation representation of the state in the computational basis."""
+        raise_error(NotImplementedError)
+
+    def __repr__(self):
+        return self.symbolic()
+
+    @abstractmethod
     def __array__(self): # pragma: no cover
         """State's tensor representation as an array."""
         raise_error(NotImplementedError)
@@ -78,7 +86,7 @@ class AbstractState(ABC):
         """State's tensor representation as a numpy array."""
         raise_error(NotImplementedError)
 
-    def state(self, numpy=False):
+    def state(self, numpy=False, symbolic=False):
         """State's tensor representation as an backend tensor.
 
         Args:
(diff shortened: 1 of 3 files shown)


scarrazza:
@stavros11 thanks for this. Looks good to me, however I would prefer to move the `decimals` option directly in the `state` method, e.g. `result.state(symbolic=True, decimals=2)`. Just to avoid to remember another method, but lets see @igres26 opinion.

igres26:
This looks like a more cleaned up version than the small code snippet that I made. I think implementing this is better.

It might be useful to have a cutoff on the amplitudes that are shown. As in a way to ignore very small but non-zero amplitudes. Or make a cutoff of all amplitudes under a certain absolute value.

stavros11:
Thank you for the comments.

> however I would prefer to move the `decimals` option directly in the `state` method, e.g. `result.state(symbolic=True, decimals=2)`. 

I moved `decimals` to `.state()` and removed the `symbolic=True` argument. Instead if `decimals < 0` the full tensor will be returned (current behavior), while if `decimals >=0` the string representation will be returned. I use `decimals=-1` as default so that the default behavior remains the same we have now and we don't break existing code. Let me know if you agree.

> It might be useful to have a cutoff on the amplitudes that are shown. As in a way to ignore very small but non-zero amplitudes. Or make a cutoff of all amplitudes under a certain absolute value.

I added a cutoff argument in `.state()` so that values smaller than cutoff are ignored in the representation.

When using `.state()` one has full control of all the arguments, eg. `.state(decimals=3, cutoff=1e-5, max_terms=10)`. Using `print(state)` will do the same thing but for the default arguments `decimals=5, cutoff=1e-10, max_terms=20`.

scarrazza (approved):
Thanks, looks good to me.

PR #526, opened 2021-12-22
Add mixed state StateEvolution

DiegoGM91:
Attempt to add support for state evolution of mixed states (i.e. density matrices).

I'll try to do my best and implement this. Once I've finished, I'll let you know so that the experts can take over.

### DIFF
--- a/src/qibo/solvers.py
+++ b/src/qibo/solvers.py
@@ -80,10 +80,13 @@ class Exponential(BaseSolver):
         else:
             return super(Exponential, cls).__new__(cls)
 
-    def __call__(self, state):
+    def __call__(self, state, density_matrix=False):
         propagator = self.current_hamiltonian.exp(self.dt)
         self.t += self.dt
-        return K.matmul(propagator, state[:, K.newaxis])[:, 0]
+        if not density_matrix:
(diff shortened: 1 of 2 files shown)


src/qibo/solvers.py
```diff
@@ -80,10 +80,13 @@ def __new__(cls, dt, hamiltonian):
         else:
             return super(Exponential, cls).__new__(cls)
 
-    def __call__(self, state):
+    def __call__(self, state, density_matrix=False):
```
stavros11, line 83: I had a quick look and it may be more convenient if you pass the `density_matrix` argument in the solver's `__init__` instead of here. For example for the `TrotterizedHamiltonian` solver you need to propagate this through `hamiltonian.circuit` and create the circuit that implements the Trotter operators using `density_matrix=True`. This can be done from `__call__` too but may be slightly easier from `__init__`.

You can have a try if you like and let us know if you need any help.

DiegoGM91: Okay, thanks @stavros11!

I'll have a try, I feel it can be a good exercise for me to get more acquainted with the base code. Hopefully my implementation can be a good starting point.

PR #528, opened 2021-12-23
Fix probabilities do not sum to 1

### LINKED ISSUE #517: Probabilities do not sum to 1
I have a very interesting bug to report. If I run the following circuit with ``NQUBITS>=26`` and ``TARGET>=25``:
```
import qibo
from qibo import gates
qibo.set_backend("numpy")
qibo.set_precision("single")

c = Circuit(NQUBITS)
for i in range(NQUBITS):
    c.add(gates.H(i))
output = c.add(gates.M(TARGET, collapse=True))
for i in range(NQUBITS):
    c.add(gates.H(i))
result = c()
```
I obtain the following error:
```

```Traceback (most recent call last):
  File ".../circuit.py", line 15, in <module>
    result = c()
  File ".../qibo/abstractions/circuit.py", line 712, in __call__
    return self.execute(initial_state=initial_state, nshots=nshots)
  File ".../qibo/core/circuit.py", line 306, in execute
    state = self._device_execute(initial_state)
  File ".../qibo/core/circuit.py", line 235, in _device_execute
    state = self._execute(initial_state=initial_state)
  File ".../qibo/core/circuit.py", line 221, in _execute
    state = self._eager_execute(state)
  File ".../qibo/core/circuit.py", line 188, in _eager_execute
    state = gate(state)
  File ".../qibo/core/gates.py", line 302, in __call__
    return getattr(self, self._active_call)(state)
  File ".../qibo/core/gates.py", line 291, in _state_vector_call
    return K.state_vector_collapse(self, state, self.result.binary[-1])
  File ".../qibo/core/measurements.py", line 111, in binary
    self._binary = self._convert_to_binary()
  File ".../qibo/core/measurements.py", line 167, in _convert_to_binary
    return K.mod(K.right_shift(self.decimal[:, K.newaxis], _range), 2)
  File ".../qibo/core/measurements.py", line 102, in decimal
    self._decimal = self._sample_shots()
  File ".../qibo/core/measurements.py", line 187, in _sample_shots
    result = K.cpu_fallback(K.sample_shots, self.probabilities, self.nshots)
  File ".../qibo/backends/abstract.py", line 106, in cpu_fallback
    return func(*args)
  File ".../qibo/backends/numpy.py", line 232, in sample_shots
    return self.random.choice(range(len(probs)), size=nshots, p=probs)
  File "mtrand.pyx", line 933, in numpy.random.mtrand.RandomState.choice
ValueError: probabilities do not sum to 1
```

Additional details:
 - The problem appears in single precision, with ``TARGET >= 25``. If ``TARGET < 25``, the problem disappears.
 - The probabilities sum to 0.5.

The main issue appears here:
https://github.com/qiboteam/qibo/blob/fe9130003c073c580f026dc7ac4443cd6bf09598/src/qibo/core/states.py#L78-L83
In line 82, state is normalized, but then after the call to ``K.sum``, we obtain a probability distribution which is not normalized.
I will investigate this issue further, in particular I'm curious to see if there is the same problem with non-collapse measurements.

mlazzarin:
Closes #517. In this PR I fixed issue #517 by casting the state vector to double precision before computing the probabilities.
The memory overhead w.r.t. main should be around 33%.

In the meantime, I made a small change in the NumPy cast, to ensure that no arrays are duplicated if not needed. Let me know your opinion.

### DIFF
--- a/src/qibo/core/states.py
+++ b/src/qibo/core/states.py
@@ -92,7 +92,7 @@ class VectorState(AbstractState):
     def probabilities(self, qubits=None, measurement_gate=None):
         unmeasured_qubits = tuple(i for i in range(self.nqubits)
                                   if i not in qubits)
-        state = K.reshape(K.square(K.abs(K.cast(self.tensor))), self.nqubits * (2,))
+        state = K.reshape(K.square(K.abs(K.cast(self.tensor, dtype="complex128"))), self.nqubits * (2,))
         return K.sum(state, axis=unmeasured_qubits)
 
     def measure(self, gate, nshots, registers=None):
--- a/src/qibo/core/states.py
+++ b/src/qibo/core/states.py
@@ -92,8 +92,8 @@ class VectorState(AbstractState):
     def probabilities(self, qubits=None, measurement_gate=None):
         unmeasured_qubits = tuple(i for i in range(self.nqubits)
                                   if i not in qubits)
-        state = K.reshape(K.square(K.abs(K.cast(self.tensor, dtype="complex128"))), self.nqubits * (2,))
-        return K.sum(state, axis=unmeasured_qubits)
+        state = K.reshape(K.square(K.abs(K.cast(self.tensor))), self.nqubits * (2,))
+        return K.sum(K.cast(state, dtype="float64"), axis=unmeasured_qubits)
 
     def measure(self, gate, nshots, registers=None):
         self.measurements = gate(self, nshots)
--- a/src/qibo/backends/numpy.py
+++ b/src/qibo/backends/numpy.py
@@ -64,7 +64,7 @@ class NumpyBackend(abstract.AbstractBackend):
             dtype = self.dtypes(dtype)
         if isinstance(x, self.backend.ndarray):
             return x.astype(dtype, copy=False)
-        return self.backend.array(x, dtype=dtype)
+        return self.backend.array(x, dtype=dtype, copy=False)
 
     def diag(self, x, dtype='DTYPECPX'):
         if isinstance(dtype, str):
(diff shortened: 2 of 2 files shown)


src/qibo/core/states.py
```diff
@@ -92,7 +92,7 @@ def wrapper(self, qubits=None, measurement_gate=None):
     def probabilities(self, qubits=None, measurement_gate=None):
         unmeasured_qubits = tuple(i for i in range(self.nqubits)
                                   if i not in qubits)
-        state = K.reshape(K.square(K.abs(K.cast(self.tensor))), self.nqubits * (2,))
+        state = K.reshape(K.square(K.abs(K.cast(self.tensor, dtype="complex128"))), self.nqubits * (2,))
```
stavros11, line 95: Does it work if we make the cast on the float level, after the abs:
```suggestion
        state = K.reshape(K.cast(K.square(K.abs(self.tensor)), dtype="float64"), self.nqubits * (2,))
```
This way we would save some memory and performance deterioration, right? Not sure if it works though.

mlazzarin: I'm not sure, we may still need the cast because otherwise we need to cast tensor from cupy to numpy in case of fallback, but we may add an additional cast before the sum. I'll give it a try.

mlazzarin: I tried a similar approach and it works, take a look at then latest commit.

stavros11: Yes, that's exactly what I meant. Given that `K.cast` is properly implemented and does not create any copies when the same type is passed, now it should work ok, right? And the sum issue is fixed since we convert to float64 before, while we avoid the complex128 conversion.

stavros11 (approved):

scarrazza (approved):

src/qibo/core/states.py
mlazzarin: Yes, it should work fine.

scarrazza:
Can we merge this?

mlazzarin:
Yes, I believe so.

PR #529, opened 2022-01-06
Fix qibo tests to support CuQuantum Backend

andrea-pasquale:
This PR contains a minor fix on the qibo tests in order to support a new backend for `qibojit` on GPU based on the NVIDIA [CuQuantum](https://github.com/NVIDIA/cuQuantum) library which is currently under-development in PR qiboteam/qibojit#57.

### DIFF
--- a/src/qibo/tests/test_backends_agreement.py
+++ b/src/qibo/tests/test_backends_agreement.py
@@ -50,8 +50,8 @@ def test_backend_methods_list(tested_backend, target_backend, method, args):
     tested_func = getattr(tested_backend, method)
     target_func = getattr(target_backend, method)
     target_result = target_func(*args)
-    if tested_backend.name == "qibojit" and tested_backend.engine.name == "cupy": # pragma: no cover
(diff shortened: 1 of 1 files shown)


mlazzarin (approved):
LGTM, thanks.

stavros11 (approved):
Thank you, it works for me. By the way, I am also NOT getting any issues with regressions of random tests.

In the future, we could also consider disabling `test_backends_agreement.py` from qibo or leaving it only for the backends coded in qibo (numpy and tensorflow) and moving the rest tests to the respective repositories (qibojit, qibotf). I am not sure how this could be done without code repetition between repositories, but it's a bit annoying having to touch qibo (even for small changes) every time we do some backend change.

scarrazza (approved):
Thanks @andrea-pasquale . @stavros11 for the time being, given the GPU support from CI, there isn't much we can do, but maybe in the near future we could have self-hosted runners or custom pipelines that will software the hardware issue.

PR #531, opened 2022-01-13
Fix `test_prints.py` when cuquantum is installed

stavros11:
As mentioned in qiboteam/qibojit#57 `test_prints.py` fails when executed in environment with cuquantum installed and a GPU enabled. For me this happens for both Python 3.8 and 3.9 and only after installing cuqantum, even if I use the master/main branches of qibo and qibojit. This implements a simple fix.

@andrea-pasquale, let me know if it works for you.

### DIFF
--- a/src/qibo/tests/test_prints.py
+++ b/src/qibo/tests/test_prints.py
@@ -17,7 +17,7 @@ class CodeText:
     @classmethod
     def from_file(cls, filedir):
         assert filedir[-3:] == ".py"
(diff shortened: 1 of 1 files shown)


andrea-pasquale (approved):
I can confirm that this fix works for me on python 3.9, which was the only one that was causing problem for me. With python 3.8 `test_print` still passes without problems.

scarrazza (approved):

PR #533, opened 2022-01-17
Upgrade `qibo.set_backend()` to support platform setting

stavros11:
As discussed, this upgrades the `qibo.set_backend()` method with an additional `platform` argument that allows switching the platform for backends that support this, for example:
```Python
import qibo
qibo.set_backend('qibojit', platform='cupy')
qibo.set_backend('qibolab', platform='tiiq')
```
For backends that do not have different platforms (numpy, tensorflow, qibotf) the additional argument is ignored.

I will open a PR in qibojit repo that makes use of this for switching between cupy and cuquantum.


mlazzarin (commented):
Thank you for the implementation of this feature!

On my laptop tests are not passing (when using qibojit in PR https://github.com/qiboteam/qibojit/pull/59)
They are fine if I use the main branch of qibojit, though.

stavros11:
> On my laptop tests are not passing (when using qibojit in PR [qiboteam/qibojit#59](https://github.com/qiboteam/qibojit/pull/59)) They are fine if I use the main branch of qibojit, though.

Thanks for noting this, I guess it's something to do with putting the `yield` inside a for loop in the pytest fixture. It should work now that I removed the platform setting from conftest.py. We can still test the cupy platform if we run in a machine with GPU and switch and test cuquantum using `QIBOJIT_PLATFORM=cuquantum pytest`. I'm not sure if it's very useful testing all platforms via conftest as the GPU platforms won't be tested in the CI anyway. Let me know what you think.

andrea-pasquale (approved):
LGTM.

mlazzarin (approved):
Looks good to me. I guess we have to run the test with cuquantum manually, as we do for the GPU.

stavros11:
Thanks to both for checking!

> Looks good to me. I guess we have to run the test with cuquantum manually, as we do for the GPU.

As it is now, qibo tests will use the default platform for each backend. So if a GPU is available it will test only qibojit-cupy, and if not only qibojit-numba. Cuquantum will never be tested unless you make it default using the environment variable as I wrote above.

There are ways to activate the tests for all available platforms. For example, in a machine with GPU it would test qibojit-cupy, qibojit-numba and qibojit-cuquantum (if installed). I did not do it properly before, that's why it was failing, but it can certainly be done.

For now I leave it as it is. If someone believes the second approach is better, let me know.

mlazzarin:
When a new backend is set, there is a log like this

``[Qibo 0.1.7rc2.dev0|INFO|2022-01-19 15:57:45]: Using qibojit backend on /GPU:0``

Do we want to add something that specifies the platform?

stavros11:
> Do we want to add something that specifies the platform?

I was updating the docs right now and noticed this issue. I was planning to add something like:
```
[Qibo 0.1.7rc2.dev0|INFO|2022-01-19 15:57:45]: Using qibojit (cupy) backend on /GPU:0
```
if platform is available, otherwise leave the message as it is.

stavros11:
I updated the info message to the above when a platform is available and added the 
```Python
qibo.set_backend("qibojit", platform="cuquantum")
```
example in the installation paragraph of docs, as discussed today. @scarrazza let me know if you agree.

scarrazza:
Yes, looks good to me, thanks.

mlazzarin:
After merging my PR, the CI fails. I guess it's because we are using the rc of qibojit and not the main branch.

stavros11:
> After merging my PR, the CI fails. I guess it's because we are using the rc of qibojit and not the main branch.

Indeed, it is due to using 'K.platform' in the examples while the rc still has 'K.engine'. By the way, a quick fix could be to use 'K.get_platform' instead of 'K.platform.name'. I'll give a try.

scarrazza:
@stavros11 thanks, can we merge, right?

stavros11:
> @stavros11 thanks, can we merge, right?

It's complete for me, if everyone agrees we can merge.

scarrazza (approved):

PR #535, opened 2022-01-19
Fix engine not renamed to platform in some tests

mlazzarin:
In some tests the old ``engine`` member of ``JitCustomBackend`` was still used. I renamed it to ``platform``.

### DIFF
--- a/src/qibo/tests/test_backends_agreement.py
+++ b/src/qibo/tests/test_backends_agreement.py
@@ -50,7 +50,7 @@ def test_backend_methods_list(tested_backend, target_backend, method, args):
     tested_func = getattr(tested_backend, method)
     target_func = getattr(target_backend, method)
     target_result = target_func(*args)
-    if tested_backend.name == "qibojit" and tested_backend.engine.name in ["cupy", "cuquantum"]: # pragma: no cover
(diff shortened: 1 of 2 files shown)


src/qibo/tests/test_backends_agreement.py
```diff
@@ -50,7 +50,7 @@ def test_backend_methods_list(tested_backend, target_backend, method, args):
     tested_func = getattr(tested_backend, method)
     target_func = getattr(target_backend, method)
     target_result = target_func(*args)
-    if tested_backend.name == "qibojit" and tested_backend.engine.name in ["cupy", "cuquantum"]: # pragma: no cover
+    if tested_backend.name == "qibojit" and tested_backend.platform.name in ["cupy", "cuquantum"]: # pragma: no cover
```
stavros11, line 53: This change is done in #533, in fact I completely removed this if and updated the casting below so that it works with all backends/platforms and we don't have to modify it every time we update something in qibojit.

stavros11 (commented):
The `test_backends_agreement.py` is already fixed in #533. The `benchmarks/main.py` is not. Could you please merge this fix in #533?

scarrazza (approved):
Looks good to me, please merge in #533.

PR #536, opened 2022-01-19
added fermionic swap, working in numpy

AdrianPerezSalinas:
I have added the fermionic swap to the list of gates, including definition and documentation. It is working in the numpy backend, but I cannot make it work in qibojit. Could you please help?


AdrianPerezSalinas:
For the record, the fermionic swap is 

1 0 0 0
0 0 1 0
0 1 0 0
0 0 0 -1

src/qibo/core/gates.py
```diff
@@ -558,6 +558,15 @@ def __init__(self, q0, q1):
     def _construct_unitary(self):
         return K.matrices.SWAP
 
+class FSWAP(BackendGate, abstract_gates.FSWAP):
```
stavros11, line 561: If you change this to 
```suggestion
class FSWAP(MatrixGate, abstract_gates.FSWAP):
```
it should work with qibojit and qibotf too.

Also please leave two empty lines between this and the previous class to match the rest of the code.

stavros11 (commented):
Thanks for implementing this, looks good. See the comment below for an easy fix to make it work with qibojit. Also make sure to add some tests. You can add a test similar to [`test_swap`](https://github.com/qiboteam/qibo/blob/86fa8b76996762aa25fc995b6a2894facc8e0fcf/src/qibo/tests/test_core_gates.py#L229) in `test_core_gates.py` and also add FSWAP in the `GATES` list in `test_core_gates_features.py` so that the matrix construction is also tested. This should fix the coverage.

By the way, if you follow my `MatrixGate` proposal below, qibojit will apply this gate using the matrix multiplication kernel for two qubit gates. So it will apply the gate by multiplying the state with the 4x4 matrix you defined. 
For this gate is is also possible to use the fSim kernel, which may be slightly faster (difference may be negligible), but this is slightly more complicated. You'd have to define [`custom_op_matrix`](https://github.com/qiboteam/qibo/blob/86fa8b76996762aa25fc995b6a2894facc8e0fcf/src/qibo/core/gates.py#L569) as we do for fSim and update [the dictionary in `AbstractCustomOperators`](https://github.com/qiboteam/qibo/blob/86fa8b76996762aa25fc995b6a2894facc8e0fcf/src/qibo/backends/abstract.py#L624) to map `"fswap"` to `apply_fsim` custom operator. If @scarrazza and @mlazzarin believe it is worth it, I can help you do it this way.

src/qibo/core/gates.py
AdrianPerezSalinas: I can change this, have no problem with that. However, I do not quite understand why, if the code is like the standard SWAP gate, we cannot somehow use the apply_swap kernel, instead of the generic two qubit gates. I think this would increase performance, but the final word is yours.

stavros11: In order to use the `apply_swap` kernel you need to add an `"fswap": self.apply_swap` entry in the [the dictionary in `AbstractCustomOperators`](https://github.com/qiboteam/qibo/blob/86fa8b76996762aa25fc995b6a2894facc8e0fcf/src/qibo/backends/abstract.py#L624). If you do this, you will stop getting the error you get now with qibojit, but it won't apply the gate you want, it will apply the standard SWAP instead (without the -1 phase). This is because the `apply_swap` kernel does the swap directly, without using the matrix (see [here](https://github.com/qiboteam/qibojit/blob/cbe5f27ce54f7460a6029398ef6d0738b3aa73c5/src/qibojit/custom_operators/gates.py#L199) the kernel code where the `gate` argument is ignored).

There's the option to add a new FSWAP kernel or modify the current SWAP kernel to accomodate the phase change but this needs to be implemented in qibojit first. Another solution is the fSim kernel I wrote above, or just the two-qubit matrix kernel. Let's see what others believe is best.

AdrianPerezSalinas:
Would if be possible to add the fSwap as a printable gate?
Something like
f 
|
f

would work

stavros11:
> Would if be possible to add the fSwap as a printable gate? Something like f | f
> 
> would work

Yes, you just need to add it to [`labels` dictionary in `circuit.draw()`](https://github.com/qiboteam/qibo/blob/86fa8b76996762aa25fc995b6a2894facc8e0fcf/src/qibo/abstractions/circuit.py#L950), like I did in #538. If you add an `"fswap": "f"` entry it should do what you say, but from what I see f is already used for fSim so there may be a confusion there. You can change the fSim symbol to something else if you like.

AdrianPerezSalinas:
What do you think about this? 
image

src/qibo/core/gates.py
AdrianPerezSalinas: For the moment I have applied the standard two qubit gates and it works with qibojit. We can change it later

stavros11:
> What do you think about this? 

That's fine for me, I believe everything works now. We are only missing tests and decide whether we will leave this as `MatrixGate` or use a specialized kernel, either a new FSWAP kernel or the existing fSim kernel.

AdrianPerezSalinas:
Sure, I will create the tests! Just give me some time because I have never done it systematically. On the kernel, whatever you decide is fine for me.

doc/source/api-reference/qibo.rst
```diff
@@ -434,6 +434,13 @@ Swap (SWAP)
     :members:
     :member-order: bysource
 
+f-Swap (FSWAP)
+"""""""""""
```
scarrazza, line 438: Please increase the number of ".

PR #537, opened 2022-01-20
using qibojit main code in actions

scarrazza:


stavros11 (approved):
Thanks for the fix, if you merge to `setplatform` I can have a look at the coverage issue (and also update some tests).

PR #539, opened 2022-01-20
Test all available platforms automatically

stavros11:
As explained [here](https://github.com/qiboteam/qibo/pull/533#issuecomment-1014797907) currently tests use the default platform for each backend. Here I try to automate testing for all platforms that are available for the machine that executes the tests. So if only CPU is available (eg. CI) it will test only qibojit-numba but if a GPU is available it will test qibojit-numba AND qibojit-cupy. If cuquantum is installed it will test this too. This is done by passing a `{backend}-{platform}` pair in the `backend` fixture.

It needs some polishing as some cupy and cuquantum tests are failing in my machine, I just want to see if you agree with this approach. I'll let you know once everything is fixed. I'm basing this on `setplatform` but is independent of #533 so it's ok to merge the other one before this is ready.


stavros11:
With the latest commits and qiboteam/qibojit#64 tests are passing for all platforms, however I still get some random failures for cuquantum every now and then. For example a distributed QFT in test_cirq.py fails sometimes. I could reproduce the same issue with a standalone script and I'll have a more detailed look why it happens.

scarrazza:
I have been doing some debugging and it looks like the linux tests involving parallel evaluation are hanging https://github.com/qiboteam/qibo/actions/runs/1760647967. This looks like a current limitation of github actions, rather than a code issue. If you agree, let me suggest to switch off the parallel checks for linux for the actions.

stavros11:
> I have been doing some debugging and it looks like the linux tests involving parallel evaluation are hanging https://github.com/qiboteam/qibo/actions/runs/1760647967. This looks like a current limitation of github actions, rather than a code issue. If you agree, let me suggest to switch off the parallel checks for linux for the actions.

Thank you for the debugging, indeed the CI is passing if we disable the parallel tests. In the last commits I added a `--skip-parallel` option in our pytest which can be used to skip all tests that use functions defined in parallel.py. Parallel testing is enabled by default (for configurations that support it) but can be skipped using the new option, which I do in the CI (updated the rule accordingly).

Other than that, this PR should be ready in terms of testing functionality. I keep getting some errors when using multigpu with cuquantum, particularly the following script fails:
<details>
<summary>cuquantum multigpu failing example</summary>

```Python
import numpy as np
import qibo
from tqdm import tqdm


nqubits = 3

def create_circuit(accelerators=None):
    circuit = models.Circuit(nqubits, accelerators=accelerators)
    circuit.add(gates.H(0))
    circuit.add(gates.CU1(0, 1, theta=0.1))
    circuit.add(gates.H(1))
    circuit.add(gates.CU1(1, 2, theta=0.1))
    return circuit


np.random.seed(123)
initial_state = np.random.random(2 ** nqubits) + 1j * np.random.random(2 ** nqubits)
initial_state = initial_state / np.sqrt(np.sum(np.abs(initial_state) ** 2))

circuit = create_circuit()
target_state = circuit(initial_state=np.copy(initial_state)).numpy()

qibo.set_backend("qibojit", platform="cuquantum")
accelerators = {"/GPU:0": 1, "/GPU:1": 1, "/GPU:2": 1, "/GPU:3": 1}
for _ in tqdm(range(200)):
    # looping multiple times because the error appears sometimes, not always
    circuit = create_circuit(accelerators)
    final_state = circuit(initial_state=np.copy(initial_state)).numpy()
    np.testing.assert_allclose(final_state, target_state, atol=1e-6)
```

</details>

I am not sure if this issue is related to the memory issues we have with multigpu but it doesn't appear always and only happens with specific circuits. The example has a minimal circuit that reproduces the error.

scarrazza:
@stavros11 thanks, I will have a look.

stavros11:
A few more info regarding the cuquantum multigpu issue: 

I believe the problem is associated with gates that for the cupy backend are applied without matrix (such as X, Y, Z) while for cuquantum we are using `qibo.matrices` as its API provides only matrix methods to apply gates. For example the following script:
<details>
<summary>Apply X gate using two GPUs</summary>

```Python
import cupy as cp
from qibo import K


def main():
    K.set_platform("cuquantum")
    print(K.get_platform())

    nqubits = 27
    dtype = "complex128"
    with cp.cuda.Device(0):
        piece0 = cp.zeros(2 ** (nqubits - 1), dtype=dtype)
        piece0[0] = 1
    with cp.cuda.Device(1):
        piece1 = cp.zeros(2 ** (nqubits - 1), dtype=dtype)

    with cp.cuda.Device(0):
        piece0 = K.apply_x(piece0, nqubits - 1, (0,))
    with cp.cuda.Device(1):
        piece1 = K.apply_x(piece1, nqubits - 1, (0,))


if __name__ == "__main__":
    main()
```

</details>

fails with
```
cuquantum.custatevec.custatevec.cuStateVecError: CUSTATEVEC_STATUS_EXECUTION_FAILED (kernel launch failure)
```

In contrast, applying a gate using a matrix (`K.apply_gate` kernel) works:

<details>
<summary>Apply H gate using two GPUs</summary>

```Python
import numpy as np
import cupy as cp
from qibo import K


def main():
    K.set_platform("cuquantum")
    print(K.get_platform())

    nqubits = 27
    dtype = "complex128"
    hmatrix = np.array([[1, 1], [1, -1]]) / np.sqrt(2)
    with cp.cuda.Device(0):
        piece0 = cp.zeros(2 ** (nqubits - 1), dtype=dtype)
        piece0[0] = 1
        hmatrix0 = cp.asarray(hmatrix, dtype=dtype)
    with cp.cuda.Device(1):
        piece1 = cp.zeros(2 ** (nqubits - 1), dtype=dtype)
        hmatrix1 = cp.asarray(hmatrix, dtype=dtype)

    with cp.cuda.Device(0):
        piece0 = K.apply_gate(piece1, hmatrix0, nqubits - 1, (0,))
    with cp.cuda.Device(1):
        piece1 = K.apply_gate(piece1, hmatrix1, nqubits - 1, (0,))


if __name__ == "__main__":
    main()
```

</details>

Notice that the X gate application fails even without joblib, while the H script works even if I parallelize the `K.apply_gate` part using joblib. So this problem is associated to joblib.

The illegal address error I mentioned in the previous comment is possibly associated to this issue. I could reproduce it with a script that is slightly more complicated than the above as it implements the CPU-GPU distributed scheme we use in qibo, but is still simpler to follow than qibo code:

<details>
<summary>Script to reproduce illegal address error</summary>

```Python
import joblib
import numpy as np
import cupy as cp
from qibo import K


class State:
    """Minimal implementation of a distributed state for two GPUs.

    The full state vector is kept as a numpy array on CPU (``self.cpu``)
    and slices of it are passed to GPUs for computation.

    Args:
        nqubits (int): Total number of qubits in the system.
        dtype (str): Complex type of state.
        device1 (int): Device id for the first GPU.
        device2 (int): Device id for the second GPU.
    """

    def __init__(self, nqubits, dtype, device1, device2):
        self.dtype = dtype
        self.nqubits = nqubits
        self.devices = (device1, device2)
        self.cpu = None

    @classmethod
    def zero(cls, nqubits, dtype, device1, device2):
        """Creates the |00...0> state."""
        state = cls(nqubits, dtype, device1, device2)
        state.cpu = np.zeros(2 ** nqubits, dtype=dtype)
        state.cpu[0] = 1
        state.cpu = np.reshape(state.cpu, nqubits * (2,))
        return state

    @staticmethod
    def slicer(qubit, index):
        """Helper method for slicing the full state to pieces."""
        return qubit * (slice(None),) + (index,)

    def piece(self, qubit, index):
        """State slice to cast to GPU for calculation.

        Args:
            qubit (int): Global qubit id to use for slicing.
            index (int): State of the global qubit (0 or 1).
        """
        slicer = self.slicer(qubit, index)
        return self.cpu[slicer]

    def assign(self, piece, qubit, index):
        """Update the CPU state using a GPU piece.

        Args:
            piece (cp.ndarray): State piece on GPU.
            qubit (int): Global qubit id.
            index (int): State of the global qubit the given piece corresponds to.
        """
        slicer = self.slicer(qubit, index)
        self.cpu[slicer] = np.reshape(piece.get(), (self.nqubits - 1) * (2,))


def apply_device1(state, gate, index):
    ndev = state.nqubits - 1
    with cp.cuda.Device(state.devices[index]):
        piece = K.apply_gate(state.piece(0, index), gate, ndev, (0,))
    state.assign(piece, 0, index)
    del(piece)


def apply_device2(state, index):
    ndev = state.nqubits - 1
    with cp.cuda.Device(state.devices[index]):
        piece = K.apply_x(state.piece(0, index), ndev, (0,))
    state.assign(piece, 0, index)
    del(piece)


def main():
    K.set_platform("cuquantum")
    print(K.get_platform())

    np.random.seed(123)
    nqubits = 27
    dtype = "complex128"
    hmatrix = np.array([[1, 1], [1, -1]]) / np.sqrt(2)
    with cp.cuda.Device(0):
        hmatrix0 = cp.asarray(hmatrix, dtype=dtype)
    with cp.cuda.Device(1):
        hmatrix1 = cp.asarray(hmatrix, dtype=dtype)

    state = State.zero(nqubits, dtype, 0, 1)

    pool = joblib.Parallel(n_jobs=2, prefer="threads")
    pool(joblib.delayed(apply_device1)(state, gate, i)
         for i, gate in enumerate([hmatrix0, hmatrix1]))

    pool = joblib.Parallel(n_jobs=2, prefer="threads")
    pool(joblib.delayed(apply_device2)(state, i) for i in range(2))


if __name__ == "__main__":
    main()
```

</details>

This script fails with
```
cupy_backends.cuda.api.runtime.CUDARuntimeError: cudaErrorIllegalAddress: an illegal memory access was encountered
```
when an X gate is applied in `apply_device2`, but works if I change this to a matrix gate.

This investigation is related to cuquantum and all these scripts work with cupy. I have also not observed the double memory issue we get with cupy in any of these scripts (although I actually started trying to reproduce this).

scarrazza:
@stavros11 I can confirm the issue with cuquantum multi-GPU, on the other hand, I don't understand why you are trying to run this code. I would delegate this feature to cuquantum directly (as soon as they release a new version). Now, if I understand correctly, the only motivation to keep on hold the qibo release is cupy with multiGPU, correct?

stavros11:
> @stavros11 I can confirm the issue with cuquantum multi-GPU, on the other hand, I don't understand why you are trying to run this code. I would delegate this feature to cuquantum directly (as soon as they release a new version). Now, if I understand correctly, the only motivation to keep on hold the qibo release is cupy with multiGPU, correct?

Indeed, I was also thinking that since cuquantum will have its own multigpu implementation in the near future it's not very useful to fix this issue now. Perhaps the best solution is to disable the multigpu tests for cuquantum. As for cupy, it the issue is only memory related and is not captured in tests, but it would still be good to resolve it (if the problem is from our side) before making any release.

For this PR, I will just disable the multigpu tests for cuquantum and it should be ready.

scarrazza:
Fine, I agree.

stavros11:
Using qiboteam/qibojit#67 and the latest changes here, multigpu tests will now only run with qibotf and qibojit-cupy which are the only backends that officially support it. Other than that, I believe this PR is testing automatically all available backends and platforms for the hardware configuration that the tests are executed on. I have tested multiple environments and machines with and without GPU and with cuquantum and all tests are now passing for me, so it should be good to go.

It would be good if others could verify that everything works in your enviroment.

andrea-pasquale (approved):
I've performed tests on two different machines and everything works fine.

PR #542, opened 2022-02-03
Add support for sparse matrices in Hamiltonians

### LINKED ISSUE #540: Energy callback for sparse hamiltonians
I am currently working in a field theory project where we must compute the energy of a ~20 qubit hamiltonian. The problem is that the statevector is 20-qubit long, but the hamiltonian takes the same memory as a 40-qubit state, which does not fit in any RAM. I have observed that `cirq` and `openfermion` are capable to deal with this kind of hamiltonians by letting the matrix be sparse.  This in fact can be done by qibo obtaining the expectation value of a sparse matrix as 

```
E = openfermion.expectation(sparse_matrix, state)
```

or equivalently
```
E = np.conj(state) @ sparse_matrix @ state
```

Now, the problem is to configure a callback for the energy from a sparse matrix as our hamiltonian, for example

```
from qibo import hamiltonians, callbacks
ham = hamiltonians.Hamiltonian(qubits, sparse_matrix) 
energy_callback = callbacks.Energy(ham)

C = Circuit(qubits)
C.add(operations_1)
adiabatic_circuit.add(gates.CallbackGate(energy_callback))
C.add(operations_2)

C.execute()
```

Is it possible? When I run a similar code my error is
`  TypeError: Matrix of invalid type <class 'scipy.sparse.csc.csc_matrix'> given during Hamiltonian initialization `

stavros11:
Fixes #540. Currently is implemented only for qibojit and numpy backends and requires qiboteam/qibojit#68. It uses the `sparse` module of scipy and cupy.

TODO:
- [x] Add tests including constructing Hamiltonians with various sparse formats and testing Hamiltonian operations (add, sub, mul, etc.)
- [x] Implement eigenvalue and eigenvector methods for sparse matrices (`sparse.linalg` is supported by both scipy and cupy).
- [x] Extend sparse support to tensorflow and qibotf backends.

@AdrianPerezSalinas, although this PR is still incomplete, I did a quick check and the current version should be sufficient to run the example you write in #540. If you would like to give a try, please checkout the `sparse` branch for both qibo and qibojit (if you use the qibojit backend).


AdrianPerezSalinas:
Hi @stavros11 , I have checked and now it works. It is nice since now I can go to larger systems. Thanks! Is there anything else you need?

stavros11:
> Hi @stavros11 , I have checked and now it works. It is nice since now I can go to larger systems. Thanks! Is there anything else you need?

No, I just wanted to confirm it works in your case, thanks for checking. I will complete the above TODO list, especially the tests to make sure that everything works properly, before asking for review and merging this.

stavros11:
I extended the Hamiltonian operation tests (addition, multiplication with scalars/vectors) to sparse matrices and implemented the eigenvector methods for sparse using `sparse.linalg`. This PR is complete for numpy and qibojit backends, except the following issues:
* The linear algebra of cupy for sparse tensors seems a bit unstable and also does not support the 'SA' (smallest algebraic) option for eigenvalues, which I believe is relevant for Hamiltonians. My solution currently is to fall back to scipy (CPU) for eigenvalue ops of sparse tensors with cupy.
* I keep getting some segfaults in CI which I cannot reproduce locally.

Note that the `eigh` and `eigvalsh` methods, as well as `Hamiltonian.eigenvalues()` and `Hamiltonian.eigenvectors()` have an additional argument `k` which is the number of eigenvalues to calculate (see [`scipy.sparse.linalg.eigsh`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.eigsh.html#scipy.sparse.linalg.eigsh)). This is ignored if the tensor is not sparse.

I also had a look at tensorflow's [`tf.sparse.SparseTensor`](https://www.tensorflow.org/api_docs/python/tf/sparse/SparseTensor). The main issue is that it does not support even simple operations, such as scalar multiplication. For example, the following works with scipy:
```Python
from scipy import sparse

n = 64
m = sparse.rand(n, n)
print(type(m))
print(type(4 * m))
```
and as expected both tensors are sparse of the same format, while the same with Tensorflow would be:
```Python
import numpy as np
import tensorflow as tf

n = 64
nonzero = int(0.1 * n * n)
indices = np.random.randint(0, n, size=(nonzero, 2))
data = np.random.random(nonzero)
m = tf.sparse.SparseTensor(indices, data, (n, n))
print(type(m))
print(type(4 * m))
```
but it fails with
```
TypeError: unsupported operand type(s) for *: 'int' and 'SparseTensor'
```
This means that if we would like to have the same level of support as numpy/qibojit we would have to implement such operations ourselves. Alternatively, we could only implement the Hamiltonian expectation which is what is required by #540, using [`tf.sparse.sparse_dense_matmul`](https://www.tensorflow.org/api_docs/python/tf/sparse/sparse_dense_matmul). @scarrazza let me know what you think.

scarrazza:
> This means that if we would like to have the same level of support as numpy/qibojit we would have to implement such operations ourselves. 

No, this goes beyond the scope of this feature.

> Alternatively, we could only implement the Hamiltonian expectation which is what is required by https://github.com/qiboteam/qibo/issues/540, using [tf.sparse.sparse_dense_matmul](https://www.tensorflow.org/api_docs/python/tf/sparse/sparse_dense_matmul).

This seems reasonable to me, so please go ahead.

stavros11:
> This seems reasonable to me, so please go ahead.

Thanks for the reply. I have implemented this in the tensorflow backend and added a test for Hamiltonian expectation. I also updated the docs. CI seems to work now even though I did not make any change regarding the segfaults. I believe this is good for review.

Note that I'm checking out the `sparse` branch of qibojit for tests to pass here. Once we merge qiboteam/qibojit#68 (should be merged before this RP) we can revert to the original rules which use qibojit master.

scarrazza:
@stavros11 I think the workflow should be updated before merging, right?

stavros11:
@scarrazza or @andrea-pasquale could you please check if you can reproduce the test failure that appears in the CI here? Make sure to use the latest main for qibojit. 

I tried in a clean environment that follows exactly how `rules.yml` is installing the libraries but all tests pass for me.

andrea-pasquale:
Tests are not passing for me @stavros11.
I also created a clean environment following the rules.yml file.
I get a different error from the CI.
```
Current thread 0x0000000119d9b600 (most recent call first):
  File "<string>", line 43 in initial_state
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/qibotf/custom_operators/python/ops/qibo_tf_custom_operators.py", line 62 in initial_state
  File "/Users/andreapasquale/PhD/QIBO/qibo/src/qibo/tests/test_backends_agreement.py", line 55 in test_backend_methods_list
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/_pytest/python.py", line 192 in pytest_pyfunc_call
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/pluggy/_callers.py", line 39 in _multicall
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/pluggy/_manager.py", line 80 in _hookexec
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/pluggy/_hooks.py", line 265 in __call__
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/_pytest/python.py", line 1718 in runtest
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/_pytest/runner.py", line 168 in pytest_runtest_call
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/pluggy/_callers.py", line 39 in _multicall
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/pluggy/_manager.py", line 80 in _hookexec
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/pluggy/_hooks.py", line 265 in __call__
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/_pytest/runner.py", line 261 in <lambda>
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/_pytest/runner.py", line 340 in from_call
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/_pytest/runner.py", line 260 in call_runtest_hook
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/_pytest/runner.py", line 221 in call_and_report
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/_pytest/runner.py", line 132 in runtestprotocol
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/_pytest/runner.py", line 113 in pytest_runtest_protocol
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/pluggy/_callers.py", line 39 in _multicall
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/pluggy/_manager.py", line 80 in _hookexec
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/pluggy/_hooks.py", line 265 in __call__
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/_pytest/main.py", line 347 in pytest_runtestloop
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/pluggy/_callers.py", line 39 in _multicall
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/pluggy/_manager.py", line 80 in _hookexec
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/pluggy/_hooks.py", line 265 in __call__
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/_pytest/main.py", line 322 in _main
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/_pytest/main.py", line 268 in wrap_session
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/_pytest/main.py", line 315 in pytest_cmdline_main
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/pluggy/_callers.py", line 39 in _multicall
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/pluggy/_manager.py", line 80 in _hookexec
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/pluggy/_hooks.py", line 265 in __call__
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/_pytest/config/__init__.py", line 165 in main
  File "/Users/andreapasquale/anaconda3/envs/sparse/lib/python3.8/site-packages/_pytest/config/__init__.py", line 188 in console_main
  File "/Users/andreapasquale/anaconda3/envs/sparse/bin/pytest", line 8 in <module>
[1]    1994 abort      pytest src/qibo/tests/test_backends_agreement.py
```
However it is the usual abort message of MacOs which doesn't tell much.

scarrazza:
I have tested this locally and indeed depends on the seed. Unfortunately, I cannot find a seed which manually spots this issue. However, I don't believe the eigh follows/guarantees a specific ordering, so maybe including a sort before doing abs could help here.

stavros11:
> I have tested this locally and indeed depends on the seed. Unfortunately, I cannot find a seed which manually spots this issue. However, I don't believe the eigh follows/guarantees a specific ordering, so maybe including a sort before doing abs could help here.

Yes, I also tried executing the test many times and indeed it fails for some instances. I can try sorting, however I would like to also trace what methods are called by the failing test, because if am not wrong, the eigenvectors compared in this case are both produced by `np.linalg.eigh`. This is because both qibojit and numpy backends fall back to `np.linalg.eigh`. So unless `np.linalg.eigh` returns different orderings even when called on the same vector, there may be something else going on too.

stavros11:
@scarrazza, regarding cupy, the following script
```Python
import cupy as cp
x = cp.zeros((2, 2))
print(cp.linalg.eigh(x))
```
returns arrays filled with `nan`, while the same operation on numpy works without returning `nan`.

Other than that, the following script:
```Python
import numpy as np
import cupy as cp
from scipy import sparse


while True:
    m = sparse.rand(10, 10, format="coo")
    m = m + m.T
    m1 = m.toarray()
    m2 = cp.asarray(m.toarray())

    eigvals1, eigvecs1 = np.linalg.eigh(m1)
    eigvals2, eigvecs2 = cp.linalg.eigh(m2)
    eigvals2, eigvecs2 = eigvals2.get(), eigvecs2.get()

    np.testing.assert_allclose(eigvals1, eigvals2, atol=1e-10)
    np.testing.assert_allclose(np.abs(eigvecs1), np.abs(eigvecs2), atol=1e-10)
```
which checks agreement between numpy and cupy for the same sparse matrix usually fails. Note that the matrix is converted to dense form before passed to the eigh method.

scarrazza:
Many thanks @stavros11, could you please open an issue in cupy posting both examples?

stavros11:
> Many thanks @stavros11, could you please open an issue in cupy posting both examples?

Reported in cupy/cupy#6539. I also added a specific example of matrix where cupy disagrees with numpy, instead of using the sparse I wrote above. In this case it is just the order of degenerate eigenvectors and some signs that differ, so both libraries are mathematically correct.

stavros11:
Following the discussion in the cupy issue, it appears that the issue with `nan` values returned by `eigh` is related to cudatoolkit 11.6 and does not appear in earlier versions. Cupy will not do anything about this so I am not sure if we should do something either.

Other than that, I made some updates in the tests so that they work on both CPU and GPU. I found a small issue with qibojit which is fixed in qiboteam/qibojit#75. With this everything should be fine regarding the sparse. I am not sure why the CI is failing in the documentation tests.

PR #543, opened 2022-02-07
Fix message error in distributed circuit

andrea-pasquale:
In this PR I fixed the error message for the distributed circuit according to the renaming of the backends implemented in #533. The error shown now in the case of cuquantum is the following
```
NotImplementedError: Distributed circuit is not supported by the qibojit (cuquantum) backend.
```
The same mechanism is applied when using the numba backend.

### DIFF
--- a/src/qibo/models/circuit.py
+++ b/src/qibo/models/circuit.py
@@ -72,7 +72,7 @@ def QFT(nqubits: int, with_swaps: bool = True,
     return circuit
 
 
-def _DistributedQFT(nqubits, accelerators=None):
+def _DistributedQFT(nqubits, accelerators=None): # pragma: no cover
     """QFT with the order of gates optimized for reduced multi-device communication."""
     from qibo import gates
     circuit = Circuit(nqubits, accelerators)
--- a/doc/source/code-examples/advancedexamples.rst
+++ b/doc/source/code-examples/advancedexamples.rst
@@ -813,7 +813,7 @@ possible to use :meth:`qibo.abstractions.circuit.AbstractCircuit.set_parameters`
 circuit needs to be defined inside the compiled ``tf.GradientTape()``.
 For example:
 
-.. code-block:: python
+.. testcode::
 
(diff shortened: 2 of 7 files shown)


andrea-pasquale:
Apparently tests are failing due to an import error from tensorflow. I don't really know why this happens.

stavros11 (approved):
Thanks for the fix, I also saw it a few days ago and wanted to fix it.

> Apparently tests are failing due to an import error from tensorflow. I don't really know why this happens.

I also noticed this while working on #542. I am not sure why it happens, perhaps the latest release of tensorflow has an issue with windows?! 
An easy solution I could think of was to add E0401 in the pylint exceptions for all lines that cause the issue (see for example [here](https://github.com/qiboteam/qibo/blob/d7bf4c07e537deb7b56217592a24e765e22fd606/src/qibo/models/qgan.py#L72)). I think CI was passing after. Not sure if that's what we should do though.

Btw, the qgan tests fail on my machine when using GPU for some time now but this is not related to the CI issue. I thought it had something to do with my tensorflow installation.

andrea-pasquale:
I think that this PR is ready to merge. I had to disable a few tests with distributed circuit, which are available only on GPU, in order to have 100% coverage. Let me know what you think.

src/qibo/models/circuit.py
```diff
@@ -72,7 +72,7 @@ def QFT(nqubits: int, with_swaps: bool = True,
     return circuit
 
 
-def _DistributedQFT(nqubits, accelerators=None):
+def _DistributedQFT(nqubits, accelerators=None): # pragma: no cover
```
stavros11, line 75: Thanks for the updates. In principle this method and the one in `test_circuit_constructor` should be tested virtually on CPU with qibotf. If it is not, there may be something strange going on. I am checking in another branch if coverage will pass without these exceptions.

scarrazza:
Thanks looks good.

PR #544, opened 2022-02-09
Remove some coverage exceptions

stavros11:
Removes some coverage exceptions for #543. @andrea-pasquale, the issue was that these tests were only testing for the default backend (qibojit numba) for which the multigpu is now disabled. I enabled them for all available backends so that the multigpu code is tested virtually from qibotf. Let me know if you agree.

Btw, if at some point we disable/remove qibotf from the test we will have several coverage issues with the distributed circuit.


andrea-pasquale (approved):
I agree, thanks for the fix @stavros11.

Issue #545, opened 2022-02-10
Wrong results using evolve with qibojit


carlacarov:
I have found that using the evolve function, I obtain different results whether I use the qibojit backend or not. 

`from qibo import models, hamiltonians
from qibo.symbols import X
import numpy as np

plus_state = np.ones(2 ** 2) / np.sqrt(2 ** 2)

symbolic_ham = sum(X(i) for i in range(2))
h = hamiltonians.SymbolicHamiltonian(symbolic_ham)

evolve = models.StateEvolution(h, dt=0.02)
medium_state = evolve(final_time=0.02, 
                        initial_state=plus_state)
final_state = evolve(final_time=0.02, 
                        initial_state=medium_state)
final_exp = h.expectation(final_state)
medium_exp = h.expectation(medium_state)

print('Medium state tensor: ', medium_state.tensor)
print('Expected value of medium state: ', medium_exp)
print('Final state tensor: ', final_state.tensor)
print('Expected value of final state: ', final_exp)`

The stdout obtained without qibojit is the following:

`[Qibo 0.1.6|WARNING|2022-02-09 20:25:40]: numpy backend uses 'np.einsum'
[qibojit_issue.zip](https://github.com/qiboteam/qibo/files/8044044/qibojit_issue.zip)
[qibojit_issue.zip](https://github.com/qiboteam/qibo/files/8044046/qibojit_issue.zip)
 and supports CPU only. Consider installing the qibojit or qibotf backends for increased performance and to enable GPU acceleration.    
Medium state tensor:  [0.49960005-0.01999467j 0.49960005-0.01999467j 0.49960005-0.01999467j
 0.49960005-0.01999467j]
Expected value of medium state:  1.9999999999999991
Final state tensor:  [0.49840085-0.03995735j 0.49840085-0.03995735j 0.49840085-0.03995735j
 0.49840085-0.03995735j]
Expected value of final state:  1.999999999999998`

and then using qibojit:

`[Qibo 0.1.6|INFO|2022-02-09 20:26:33]: Using qibojit backend on /CPU:0
Medium state tensor:  [0.49840085-0.03995735j 0.49840085-0.03995735j 0.49840085-0.03995735j
 0.49840085-0.03995735j]
Expected value of medium state:  1.999999999999998
Final state tensor:  [0.49840085-0.03995735j 0.49840085-0.03995735j 0.49840085-0.03995735j
 0.49840085-0.03995735j]
Expected value of final state:  1.999999999999998`

[...]

PR #549, opened 2022-02-13
Implement state deep copy

stavros11:
Following #545, this implements the option to create a deep copy of a state using `VectorState.copy(deep=True)`. I also added in the docs (States section) that qibojit and qibotf perform in-place updates so the state (and all its shallow copies) will be modified if it is used as initial state in a circuit or time evolution.


scarrazza (approved):
Thanks, looks very good.

PR #550, opened 2022-02-15
Adding channels to circuit.draw()

### LINKED ISSUE #546: Draw Noise Channel Gates
**Is your feature request related to a problem? Please describe.**
I have implement a customised noise model for my quantum circuit. However, I could not use the built in .draw() function to visualise my quantum circuit. It would be difficult to check whether I have placed my quantum channels correctly.

> ~/miniconda3/envs/mypsi4/lib//python3.8/site-packages/qibo/abstractions/circuit.py in draw(self, line_wrap)
>953         for gate in self.queue:
>954             if gate.name not in labels:
>--> 955                 raise_error(NotImplementedError, f"{gate.name} gate is not supported by `circuit.draw`")
>956             gate_name = labels.get(gate.name)
>957             targets = list(gate.target_qubits)
>
>~/miniconda3/envs/mypsi4/lib//python3.8/site-packages/qibo/config.py in raise_error(exception, message, args)
>50         raise exception(message, args)
>51     else:
>---> 52         raise exception(message)
>53 
>54 
> 
> NotImplementedError: PauliNoiseChannel gate is not supported by `circuit.draw`

**Describe the solution you'd like**
Draw a quantum circuit with simple special symbols that represents the single qubit noise channels gates (Thermal, Pauli, etc) for the purpose of checking gates placement quickly.

**Describe alternatives you've considered**
I have place all the gates in a list before adding them to the quantum circuit one by one. By printing out the gate type and qubit index of each gate, one can 'visualise' the quantum circuit with great difficulty.

**Additional context**
None

andrea-pasquale:
This PR fixes #546. The current version of qibo does not allow to draw circuits where there are channel gates.
I've added a few labels to identify these channels. In order to differentiate those from the usuale gates I've added a `c` at the end. For example in the case of the `PauliNoiseChannel` the output should be the following:
```python

c = models.Circuit(2, density_matrix=True)
c.add(gates.X(1))
c.add(gates.PauliNoiseChannel(0, px=0.3))
result = c()
print(c.draw())
# q0: ─Pc─
# q1: ─X──
```
The same mechanism using different labels is applied for all the other channels.
We can also change the name of the labels that I've chosen.
Let me know what you think.

### DIFF
--- a/src/qibo/abstractions/circuit.py
+++ b/src/qibo/abstractions/circuit.py
@@ -943,6 +943,8 @@ class AbstractCircuit(ABC):
         Args:
             line_wrap (int): maximum number of characters per line. This option
                 split the circuit text diagram in chunks of line_wrap characters.
+            legend (bool): If ``True`` prints a legend below the circuit for
+                callbacks and channels. Default is ``False``.
 
         Return:
             String containing text circuit diagram.
--- a/src/qibo/abstractions/circuit.py
+++ b/src/qibo/abstractions/circuit.py
@@ -955,7 +955,10 @@ class AbstractCircuit(ABC):
                   "crx": "RX", "cry": "RY", "crz": "RZ",
                   "cu1": "U1", "cu3": "U3", "ccx": "X",
                   "id": "I", "measure": "M", "fsim": "f",
-                  "generalizedfsim": "gf", "Unitary": "U", "fswap":"fx"}
+                  "generalizedfsim": "gf", "Unitary": "U", "fswap":"fx",
+                  "PauliNoiseChannel": "Pc", "KrausChannel": "Kc",
+                  "UnitaryChannel": "Uc", "ThermalRelaxationChannel": "Tc",
+                  "ResetChannel": "Rc", "PartialTrace": "Pt"}
 
         # build string representation of gates
         matrix = [[] for _ in range(self.nqubits)]
--- a/src/qibo/abstractions/circuit.py
+++ b/src/qibo/abstractions/circuit.py
@@ -1030,7 +1030,7 @@ class AbstractCircuit(ABC):
             table = tabulate([[i, labels[i]] for i in names],
                               headers=['Gate', 'Symbol'],
                               tablefmt='orgtbl')
-            table = '\n Legend: \n' + table
+            table = '\n Legend for callbacks and channels: \n' + table
 
         # line wrap
         if line_wrap:
--- a/src/qibo/abstractions/circuit.py
+++ b/src/qibo/abstractions/circuit.py
@@ -956,11 +956,11 @@ class AbstractCircuit(ABC):
                   "cu1": "U1", "cu3": "U3", "ccx": "X",
                   "id": "I", "measure": "M", "fsim": "f",
                   "generalizedfsim": "gf", "Unitary": "U", "fswap":"fx",
-                  "PauliNoiseChannel": "Pc", "KrausChannel": "Kc",
-                  "UnitaryChannel": "Uc", "ThermalRelaxationChannel": "Tc",
-                  "ResetChannel": "Rc", "PartialTrace": "Pt",
-                  "EntanglementEntropy": "EE", "Norm": "N",
-                  "Overlap": "O", "Energy": "E", "Gap": "G"}
+                  "PauliNoiseChannel": "PNCh", "KrausChannel": "KCh",
+                  "UnitaryChannel": "UCh", "ThermalRelaxationChannel": "TRCh",
+                  "ResetChannel": "RCh", "PartialTrace": "PTCh",
+                  "EntanglementEntropy": "EECb", "Norm": "NCb",
+                  "Overlap": "OCb", "Energy": "ECb"}
 
         # build string representation of gates
         matrix = [[] for _ in range(self.nqubits)]
--- a/src/qibo/abstractions/circuit.py
+++ b/src/qibo/abstractions/circuit.py
@@ -937,7 +937,7 @@ class AbstractCircuit(ABC):
 
         return len(qubits), gate_list
 
-    def draw(self, line_wrap=70) -> str:
+    def draw(self, line_wrap=70, legend=False) -> str:
         """Draw text circuit using unicode symbols.
 
         Args:
@@ -956,11 +956,11 @@ class AbstractCircuit(ABC):
                   "cu1": "U1", "cu3": "U3", "ccx": "X",
                   "id": "I", "measure": "M", "fsim": "f",
                   "generalizedfsim": "gf", "Unitary": "U", "fswap":"fx",
-                  "PauliNoiseChannel": "PNCh", "KrausChannel": "KCh",
-                  "UnitaryChannel": "UCh", "ThermalRelaxationChannel": "TRCh",
-                  "ResetChannel": "RCh", "PartialTrace": "PTCh",
-                  "EntanglementEntropy": "EECb", "Norm": "NCb",
-                  "Overlap": "OCb", "Energy": "ECb"}
+                  "PauliNoiseChannel": "PN", "KrausChannel": "K",
+                  "UnitaryChannel": "U", "ThermalRelaxationChannel": "TR",
+                  "ResetChannel": "R", "PartialTrace": "PT",
+                  "EntanglementEntropy": "EE", "Norm": "N",
+                  "Overlap": "O", "Energy": "E"}
 
         # build string representation of gates
         matrix = [[] for _ in range(self.nqubits)]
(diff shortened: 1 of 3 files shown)


stavros11 (approved):
Looks good to me, thanks. I had a look at the tests and I believe we only have two tests for `circuit.draw()`, which are in test_abstract_circuit.py, and do not test all possible gates. So I am not sure if it is really necessary to add a test for these new gates.

andrea-pasquale:
Thanks for the review @stavros11.
Should we include also callback gates in this PR? I can do that if needed.

andrea-pasquale:
> If you have time you can try this, however I am not sure if it is as simple as the channels and other gates because callback gates are special in the sense that they "act" on all qubits but their qubits lists (target_qubits and control_qubits) are empty.

I've just coded a simple solution but I don't know if this behaviour is correct.
Since the target_qubits and the control_qubits are empty I've added as targets_qubits the list containing all qubits.
For example now if you run the following example from the documentation:
```python
from qibo import models, gates, callbacks

# create entropy callback where qubit 0 is the first subsystem
entropy = callbacks.EntanglementEntropy([0])

# initialize circuit with 2 qubits and add gates
c = models.Circuit(2) # state is |00> (entropy = 0)
c.add(gates.CallbackGate(entropy)) # performs entropy calculation in the initial state
c.add(gates.H(0)) # state is |+0> (entropy = 0)
c.add(gates.CallbackGate(entropy)) # performs entropy calculation after H
c.add(gates.CNOT(0, 1)) # state is |00> + |11> (entropy = 1))
c.add(gates.CallbackGate(entropy)) # performs entropy calculation after CNOT

# execute the circuit using the callback
final_state = c()
print(c.draw())
```
The output is:
```
q0: ─EE─H─EE─o─EE─
q1: ─EE───EE─X─EE─
```
which seems correct to me.
Let me know what you think.

src/qibo/abstractions/circuit.py
```diff
@@ -955,7 +955,12 @@ def draw(self, line_wrap=70) -> str:
                   "crx": "RX", "cry": "RY", "crz": "RZ",
                   "cu1": "U1", "cu3": "U3", "ccx": "X",
                   "id": "I", "measure": "M", "fsim": "f",
-                  "generalizedfsim": "gf", "Unitary": "U", "fswap":"fx"}
+                  "generalizedfsim": "gf", "Unitary": "U", "fswap":"fx",
+                  "PauliNoiseChannel": "Pc", "KrausChannel": "Kc",
+                  "UnitaryChannel": "Uc", "ThermalRelaxationChannel": "Tc",
+                  "ResetChannel": "Rc", "PartialTrace": "Pt",
+                  "EntanglementEntropy": "EE", "Norm": "N",
+                  "Overlap": "O", "Energy": "E", "Gap": "G"}
```
stavros11, line 963: Do you think it would be useful to have a prefix (eg. Cb) for all callback gates, something like (CbEE, CbN, CbE, etc.) so that they can be distinguished are distinguishable from normal gates when calling `circuit.draw()`? 

Also, the gap callback is probably not needed here as this callback can only be used in an adiabatic evolution model, not in a circuit.

andrea-pasquale: Maybe prefix for the callbacks and suffix for the channels, just be consistent with the name of the gates, so for example the `PauliNoiseGate` should be something like `PNCh`.

andrea-pasquale: You mean adding like a legend that explains that `Ch` stands for channel and `Cb` for callback or something more complicated like a full legend of all the gates used in the circuit?

scarrazza: Yes, for this and maybe all other gates, we could have an extra flag `legend` so the user can switch on/off.

andrea-pasquale: I coded small example where I removed the prefix and the suffix and I added a legend at the bottom.
Here are a few examples:
```python
from qibo import gates

c2 = Circuit(2, density_matrix=True)
c2.add(gates.H(0))
c2.add(gates.PauliNoiseChannel(0, 0.1, 0.0, 0.2))
c2.add(gates.H(1))
c2.add(gates.PauliNoiseChannel(1, 0.0, 0.2, 0.1))
c2.add(gates.CNOT(0, 1))
c2.add(gates.PauliNoiseChannel(0, 0.1, 0.0, 0.2))
c2.add(gates.PauliNoiseChannel(1, 0.0, 0.2, 0.1))
print(c2.draw(legend=True))
```
With output:
```
q0: ─H─PN─o─PN─
q1: ─H─PN─X─PN─

 Legend: 
| Gate              | Symbol   |
|-------------------+----------|
| PauliNoiseChannel | PN       |
```
Or 
```python
from qibo import models, gates, callbacks

# create entropy callback where qubit 0 is the first subsystem
entropy = callbacks.EntanglementEntropy([0])

# initialize circuit with 2 qubits and add gates
c = models.Circuit(2) # state is |00> (entropy = 0)
c.add(gates.CallbackGate(entropy)) # performs entropy calculation in the initial state
c.add(gates.H(0)) # state is |+0> (entropy = 0)
c.add(gates.CallbackGate(entropy)) # performs entropy calculation after H
c.add(gates.CNOT(0, 1)) # state is |00> + |11> (entropy = 1))
c.add(gates.CallbackGate(entropy)) # performs entropy calculation after CNOT

# execute the circuit using the callback
final_state = c()
print(c.draw(legend=True))
```
with output:
```
q0: ─EE─H─EE─o─EE─
q1: ─EE───EE─X─EE─

 Legend: 
| Gate                | Symbol   |
|---------------------+----------|
| EntanglementEntropy | EE       |
```
For the time being I think that we can add to the legend only the gates which don't have a standard notation, so the channels and the callbacks.
I made the table using [tabulate](https://pypi.org/project/tabulate/). Let me know if this approach could work.

scarrazza: I like it, @stavros11 what do you think?

andrea-pasquale: Ok, maybe I'll add also a test with the legend.
I should also add the tabulate package as an installation requirement at this point, correct?

stavros11: > I should also add the tabulate package as an installation requirement at this point, correct?

Yes, I believe so, you need to add it to `install_requires` in setup.py. Would be good if you also add a test. You can add a `@pytest.mark.parametrize("legend", [False, True])` in the current test if you don't want to write a new one.

andrea-pasquale: Sure, this is exactly what I'm doing, thanks!

andrea-pasquale:
I think we should be ready to merge. 
@AdrianPerezSalinas do you like this implementation?

scarrazza (approved):
I like, and it seems to be working fine.

src/qibo/abstractions/circuit.py
```diff
@@ -937,7 +937,7 @@ def read_args(args):
 
         return len(qubits), gate_list
 
-    def draw(self, line_wrap=70) -> str:
+    def draw(self, line_wrap=70, legend=False) -> str:
```
scarrazza, line 940: Could you please update the docstring of this function?

andrea-pasquale: Yes.
Before I do that I found a possible issue. If the user calls the draw method using `legend=True` with a circuit that does not contain any channels or callbacks then the result will be to print just an empty legend.
```python
from qibo import gates

c2 = Circuit(2, density_matrix=True)
c2.add(gates.H(0))
#c2.add(gates.PauliNoiseChannel(0, 0.1, 0.0, 0.2))
c2.add(gates.H(1))
#c2.add(gates.PauliNoiseChannel(1, 0.0, 0.2, 0.1))
c2.add(gates.CNOT(0, 1))
#c2.add(gates.PauliNoiseChannel(0, 0.1, 0.0, 0.2))
#c2.add(gates.PauliNoiseChannel(1, 0.0, 0.2, 0.1))
print(c2.draw(legend=True))
#output
#q0: ─H─o─
#q1: ─H─X─

#Legend: 
#| Gate   | Symbol   |
#|--------+----------|
```
I don't know if you are happy with this behavior. Maybe since in the legend we are adding just the callbacks and the channels the draw method should for example draw directly the legend if a circuit uses those particular gates, and not draw it, since it will be empty, for a circuit without those gates. If we choose this option we can drop the legend flag. Otherwise I'm also fine with the current version.
Let me know what you think @scarrazza @stavros11 .

AdrianPerezSalinas:
Hi all, 

I have seen the changes and the implementation and changes. They look fine to me. However, I would like to ask for something else. When implementing a callback, my way to understand it is that there is a transversal measurement on the full circuit, and it would be nice to see this in the graphical implementation, for example
``` 
   ───                ───
─| EE |─H──o─| EE |─
─| EE |────X─| EE |─
   ───                ───
```

Would it be possible to implement something of this kind? If not, I am happy with the current version

scarrazza:
@AdrianPerezSalinas I think this layout is too complicated in terms of comprehension.
@andrea-pasquale I think it is fine to keep the legend empty, but maybe include something like `Legend for callbacks`.

andrea-pasquale:
@AdrianPerezSalinas I agree with @scarrazza the layout is a bit too complicated to implement.
@scarrazza I'll add `Legend for callbacks and channels` since in the legend we are including also the channels.
I will also update the documentation.

AdrianPerezSalinas:
@andrea-pasquale @scarrazza Fine then, I am happy with this layout

Issue #555, opened 2022-02-25
Execute circuit in a loop 


juliabarbera:
I was wondering if there is any way, in Qibo language, to execute a quantum circuit many times (inside a loop) in order to see how the probability to find the result increases with number of times the oracle is applied.

I am aware that when we execute the circuit we are collapsing the wave function so once I have executed it I can't do it again, but I was hoping if there was any way to do so.

igres26:
The proposed solution by @stavros11 would solve the issue but I would like to suggest some alternatives.

**First way**: Run the full circuit every iteration (as you would on a quantum machine)

```Python
probs = []
init_c = # initial circuit (H gates for example)
step_c = # Circuit that needs to be repeated

for i in range(1, nreps+1):
    c = Circuit()
    c += init_c
    for j in range(i)
        c += step_c
    state = c()
    probs.append( # probabilities)
``` 

**Second way**: Run one step at a time reusing the output state (only possible through simulation)

```Python
probs = []
init_c = # initial circuit (H gates for example)
step_c = # Circuit that needs to be repeated

state = init_c()
for i in range(nreps):
    state = step_c(state)
    probs.append( # probabilities)
```

**Third way**: Use callbacks within a circuit

https://qibo.readthedocs.io/en/stable/api-reference/qibo.html?highlight=callbacks#callbacks

Callbacks are special gates that one can apply during a circuit to compute certain properties of the quantum state along the execution of a circuit.

Right now, using the callbacks available, I suggest using the `callbacks.Overlap` and compute the overlap after each step with the target state.

```Python
target state = # the state that you are trying to see the evolution of
overlap = callbacks.Overlap(target_state)

init_c = # initial circuit (H gates for example)
step_c = # Circuit that needs to be repeated
c = Circuit()
c += init_c
for i in range(nreps):
    c += step_c
    c.add(gates.CallbackGate(overlap))

c()

print(overlap)
```

We will try to implement a callback that keeps track of the full quantum state along the circuit for situations like this.

PR #559, opened 2022-03-10
Full state callback

stavros11:
Implements a callback that keeps track of the full state vector or density matrix during circuit simulation. This was requested by @igres26 in #555. Usage is similar to existing callbacks, one can add this to a circuit, in the place which a stamp of the state is required, using a `CallbackGate`.

This PR is complete, I just mark it as do not merge because the test will change if we merge qiboteam/qibojit#74.


stavros11:
@scarrazza I updated the test after the latest merge in qibojit. This should be good to go.

scarrazza:
Thanks!

PR #564, opened 2022-03-24
Fix sampling when using `circuit.add()`

### LINKED ISSUE #563: Sampling circuit with channels not working as expected
I'm using the following toy model for randomized benchmarking:
```python
import numpy as np
gates_set = [gates.X, gates.Y, gates.Z, gates.H, gates.S, gates.SDG, gates.I]
circ = models.Circuit(1)
circ_no_noise = models.Circuit(1)
for gate in range(100):
    new_gate = np.random.choice(gates_set)(0)
    circ.add(gates.PauliNoiseChannel(0, pz=0.01))
    circ.add(new_gate)
    circ_no_noise.add(new_gate)
circ.add(gates.PauliNoiseChannel(0, pz=0.01))
circ += circ_no_noise.invert()
circ.add(gates.M(0))
```
When I perform measurements using:
```python
ex = circ(nshots=10)
print(ex.samples())
# output
# [[0] [0] [0]  [0]  [0]  [0]  [0]  [0]  [0]  [0]]
# or 
#  [[1] [1] [1]  [1]  [1]  [1]  [1]  [1]  [1]  [1]]
```
I get the same results for all the shots. Therefore it seems like that the circuit is executed only once.
Even if according to the documentation when using channels the full circuit should be re-executed for every shot.

If I re-execute the circuit manually 

```python
for _ in range(10):
    ex = circ(nshots=1)
    print(ex.samples())
# output
#  [[1] [0] [0]  [1]  [0]  [1]  [0]  [0]  [1]  [1]]
```
I get the correct output.

andrea-pasquale:
Closes #563.
As already discussed in today's meeting, the problem was related to the `add` method of `AbstractCircuit`.
When the new circuit is created the `repeated_execution` attribute, which allows to re-execute the circuit for every
shot, was set by default to `False`. This is why when adding the noisy circuit with its inverse the sampling was odd.
Now `repeated_execution` is set to:
```python
newcircuit.repeated_execution = old_circuit.repeated_execution or circuit_to_add.repeated_execution
```
Therefore if one of the two circuits requires to re-execute the circuit the new circuit will also be re-executed for every shot. 
I've also added tests about this.
Let me know what you think @igres26 @scarrazza @stavros11.

### DIFF
--- a/src/qibo/tests/test_abstract_circuit.py
+++ b/src/qibo/tests/test_abstract_circuit.py
@@ -117,19 +117,6 @@ def test_circuit_add_nested_generator():
     assert isinstance(c.queue[5], gates.CNOT)
     assert isinstance(c.queue[7], gates.H)
 
-def test_circuit_add_repeated_execution():
-    c = Circuit(1)
-    c.add(gates.H(0))
-    c.add(gates.PauliNoiseChannel(0, px=0.2))
-
-    c1 = Circuit(1)
-    c1.add(gates.X(0))
-
-    c2 = c + c1
-
-    assert c.repeated_execution
-    assert not c1.repeated_execution
-    assert c2.repeated_execution
 
 def test_set_nqubits():
     c = Circuit(2)
--- a/src/qibo/tests/test_abstract_circuit.py
+++ b/src/qibo/tests/test_abstract_circuit.py
@@ -117,6 +117,19 @@ def test_circuit_add_nested_generator():
     assert isinstance(c.queue[5], gates.CNOT)
     assert isinstance(c.queue[7], gates.H)
 
+def test_circuit_add_repeated_execution():
+    c = Circuit(1)
+    c.add(gates.H(0))
+    c.add(gates.PauliNoiseChannel(0, px=0.2))
+
+    c1 = Circuit(1)
+    c1.add(gates.X(0))
+
+    c2 = c + c1
+
+    assert c.repeated_execution
+    assert not c1.repeated_execution
+    assert c2.repeated_execution
 
 def test_set_nqubits():
     c = Circuit(2)
--- a/src/qibo/tests/test_core_circuit_noise.py
+++ b/src/qibo/tests/test_core_circuit_noise.py
@@ -180,12 +180,14 @@ def test_circuit_add_sampling(backend):
     circ.add(gates.M(0))
 
     # Sampling using 10 shots
+    np.random.seed(123)
     K.set_seed(123)
     samples = circ(nshots=10).samples()
 
     # Sampling using 1 shot in for loop
     target_samples = []
     K.set_seed(123)
+    np.random.seed(123)
     for _ in range(10):
         target_samples.append(circ(nshots=1).samples())
 
--- a/src/qibo/tests/test_core_circuit_noise.py
+++ b/src/qibo/tests/test_core_circuit_noise.py
@@ -162,7 +162,7 @@ def test_density_matrix_circuit_errors():
         c.add(gates.KrausChannel([((0,), np.eye(2))]))
 
 
-def test_circuit_add_sampling():
+def test_circuit_add_sampling(backend):
     """Check measurements when simulating added circuits with noise"""
     # Create random noisy circuit and add noiseless inverted circuit
     gates_set = [gates.X, gates.Y, gates.Z, gates.H, gates.S, gates.SDG, gates.I]
@@ -189,4 +189,4 @@ def test_circuit_add_sampling():
     for _ in range(10):
         target_samples.append(circ(nshots=1).samples())
 
-    K.assert_allclose(np.array(target_samples).flatten(), samples.flatten())
+    K.assert_allclose(samples[:, 0], target_samples[:, 0, 0])
--- a/src/qibo/tests/test_core_circuit_noise.py
+++ b/src/qibo/tests/test_core_circuit_noise.py
@@ -160,3 +160,31 @@ def test_density_matrix_circuit_errors():
     c = Circuit(5)
     with pytest.raises(ValueError):
         c.add(gates.KrausChannel([((0,), np.eye(2))]))
+
+
+def test_circuit_add_sampling():
+    """Check measurements when simulating added circuits with noise"""
+    # Create random noisy circuit and add noiseless inverted circuit
+    gates_set = [gates.X, gates.Y, gates.Z, gates.H, gates.S, gates.SDG, gates.I]
+    circ = Circuit(1)
+    circ_no_noise = Circuit(1)
+    for _ in range(100):
+        new_gate = np.random.choice(gates_set)(0)
+        circ.add(gates.PauliNoiseChannel(0, pz=0.01))
+        circ.add(new_gate)
+        circ_no_noise.add(new_gate)
+        circ.add(gates.PauliNoiseChannel(0, pz=0.01))
+    circ += circ_no_noise.invert()
+    circ.add(gates.M(0))
+
+    # Sampling using 10 shots
+    K.set_seed(123)
+    samples = circ(nshots=10).samples()
+
+    # Sampling using 1 shot in for loop
+    target_samples = []
+    K.set_seed(123)
+    for _ in range(10):
+        target_samples.append(circ(nshots=1).samples())
+
+    K.assert_allclose(np.array(target_samples).flatten(), samples.flatten())
--- a/src/qibo/tests/test_core_circuit_noise.py
+++ b/src/qibo/tests/test_core_circuit_noise.py
@@ -168,12 +168,14 @@ def test_circuit_add_sampling():
     gates_set = [gates.X, gates.Y, gates.Z, gates.H, gates.S, gates.SDG, gates.I]
     circ = Circuit(1)
     circ_no_noise = Circuit(1)
-    for _ in range(100):
+
+    for _ in range(10):
         new_gate = np.random.choice(gates_set)(0)
         circ.add(gates.PauliNoiseChannel(0, pz=0.01))
         circ.add(new_gate)
         circ_no_noise.add(new_gate)
-        circ.add(gates.PauliNoiseChannel(0, pz=0.01))
+
+    circ.add(gates.PauliNoiseChannel(0, pz=0.01))
     circ += circ_no_noise.invert()
     circ.add(gates.M(0))
 
--- a/src/qibo/tests/test_core_circuit_noise.py
+++ b/src/qibo/tests/test_core_circuit_noise.py
@@ -189,4 +189,6 @@ def test_circuit_add_sampling(backend):
     for _ in range(10):
         target_samples.append(circ(nshots=1).samples())
 
-    K.assert_allclose(samples[:, 0], target_samples[:, 0, 0])
+    target_samples = np.stack(target_samples)
+
+    K.assert_allclose(samples, target_samples[:, 0])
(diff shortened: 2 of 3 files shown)


src/qibo/tests/test_abstract_circuit.py
```diff
@@ -117,6 +117,19 @@ def gen():
...
+    c.add(gates.PauliNoiseChannel(0, px=0.2))
+
+    c1 = Circuit(1)
+    c1.add(gates.X(0))
+
+    c2 = c + c1
+
+    assert c.repeated_execution
+    assert not c1.repeated_execution
+    assert c2.repeated_execution
```
scarrazza, line 132: Looks okay to me, however maybe one should include an example where you compare the shots directly (after fixing the seed).

src/qibo/tests/test_abstract_circuit.py
```diff
@@ -117,6 +117,19 @@ def gen():
     assert isinstance(c.queue[5], gates.CNOT)
     assert isinstance(c.queue[7], gates.H)
 
+def test_circuit_add_repeated_execution():
```
stavros11, line 120: A new line is missing to have two line spacing between tests.
```suggestion

def test_circuit_add_repeated_execution():
```

src/qibo/tests/test_abstract_circuit.py
stavros11: I agree with adding a test using the example from #563. You can add this to one of the `test_core_circuit_*.py` files (for example `test_core_circuit_noise.py`) because `test_abstract_circuit.py` is not supposed to test execution and is backend free.

stavros11 (commented):
Thanks for finding the issue. I believe this solves the problem, just two comments below.

src/qibo/tests/test_abstract_circuit.py
andrea-pasquale: Sure, I also agree with adding the test to compare shots.
At this point should I remove this test?

scarrazza: >  At this point should I remove this test?

I would say yes.

andrea-pasquale: Done.

src/qibo/tests/test_core_circuit_noise.py
```diff
@@ -160,3 +160,31 @@ def test_density_matrix_circuit_errors():
...
+    # Create random noisy circuit and add noiseless inverted circuit
+    gates_set = [gates.X, gates.Y, gates.Z, gates.H, gates.S, gates.SDG, gates.I]
+    circ = Circuit(1)
+    circ_no_noise = Circuit(1)
+    for _ in range(100):
+        new_gate = np.random.choice(gates_set)(0)
+        circ.add(gates.PauliNoiseChannel(0, pz=0.01))
+        circ.add(new_gate)
+        circ_no_noise.add(new_gate)
+        circ.add(gates.PauliNoiseChannel(0, pz=0.01))
```
andrea-pasquale, line 176: I've implemented a test that checks the sampling after fixing the seed (using `K.set_seed(123)`.
However, I've notice that the parameter `seed` of the `PauliNoiseChannel` does not affect the results of the sampling. That's why here I've decided to omit it. Is this the correct behaviour, @stavros11 ?

stavros11, line 176: It doesn't really matter for testing purposes, but the last noise channel is outside the for loop in #563. Also it is fine to use a smaller circuit for testing (eg. less than 100 gates).

stavros11: Thanks for adding the test.

> However, I've notice that the parameter `seed` of the `PauliNoiseChannel` does not affect the results of the sampling. That's why here I've decided to omit it. Is this the correct behaviour, @stavros11 ?

Can you give an example where it doesn't affect? For example, the following script:
```Python
import numpy as np


circ = models.Circuit(1)
circ_no_noise = models.Circuit(1)

for _ in range(10):
    new_gate = gates.H(0)
    circ.add(gates.PauliNoiseChannel(0, pz=0.1))
    circ.add(new_gate)
    circ_no_noise.add(new_gate)

circ.add(gates.PauliNoiseChannel(0, pz=0.1))
circ += circ_no_noise.invert()
circ.add(gates.M(0))

print(circ(nshots=100).samples()[:, 0])
```
gives me different samples every time I run, while if I add a `seed` in the last noise channel, it gives me the same samples every time.

src/qibo/tests/test_core_circuit_noise.py
```diff
@@ -160,3 +160,31 @@ def test_density_matrix_circuit_errors():
     c = Circuit(5)
     with pytest.raises(ValueError):
         c.add(gates.KrausChannel([((0,), np.eye(2))]))
+
+
+def test_circuit_add_sampling():
```
stavros11, line 165: I would add the backend fixture here
```suggestion
def test_circuit_add_sampling(backend):
```
so that this is tested on all available backends. Right now it is only tested on the default backend and will have different behavior depending on your environment. In fact it fails for some backends (see comment below).

src/qibo/tests/test_core_circuit_noise.py
```diff
@@ -160,3 +160,31 @@ def test_density_matrix_circuit_errors():
...
+    K.set_seed(123)
+    samples = circ(nshots=10).samples()
+
+    # Sampling using 1 shot in for loop
+    target_samples = []
+    K.set_seed(123)
+    for _ in range(10):
+        target_samples.append(circ(nshots=1).samples())
+
+    K.assert_allclose(np.array(target_samples).flatten(), samples.flatten())
```
stavros11, line 190: This fails for:
* tensorflow: `AttributeError: 'tensorflow.python.framework.ops.EagerTensor' object has no attribute 'flatten'`
* qibojit-cupy: `TypeError: Implicit conversion to a NumPy array is not allowed. Please use .get() to construct a NumPy array explicitly.`

I believe it is simpler if you just do the indexing manually, eg. something like:
```suggestion
    K.assert_allclose(samples[:, 0], target_samples[:, 0, 0])
```
as `K.assert_allclose` is supposed to properly handle the conversions to numpy for different backends.

src/qibo/tests/test_core_circuit_noise.py
andrea-pasquale: Yes, I agree. I can lower it down to 10.

src/qibo/tests/test_core_circuit_noise.py
andrea-pasquale: Yep, you are right. It is better to test it for every backend.

src/qibo/tests/test_core_circuit_noise.py
andrea-pasquale: Perfect, thank you!
I will add your suggestions.

andrea-pasquale:
> Can you give an example where it doesn't affect? 

I've noticed it with the test that I've added, but I can reproduce it also using your example.
If I add `K.set_seed` than the seed in the noise is ignored. For example I can choose a random seed for each gate as shown here
```python
from qibo import models, gates, K
import numpy as np

K.set_seed(123)

circ = models.Circuit(1)
circ_no_noise = models.Circuit(1)

for i in range(10):
    new_gate = gates.H(0)
    circ.add(gates.PauliNoiseChannel(0, pz=0.1, seed=i))
    circ.add(new_gate)
    circ_no_noise.add(new_gate)

circ.add(gates.PauliNoiseChannel(0, pz=0.1))
circ += circ_no_noise.invert()
circ.add(gates.M(0))

print(circ(nshots=100).samples()[:, 0])
```
But if I set the seed at the beginning the output remains the same.

stavros11:
The example from the last post gives me
```
[Qibo 0.1.8.dev0|INFO|2022-03-25 15:50:00]: Using qibojit (numba) backend on /CPU:0
[0 1 0 0 0 1 0 0 1 0 0 0 0 1 1 1 1 0 0 0 1 1 1 0 0 0 0 0 1 0 0 0 0 1 0 0 1
 1 0 0 0 0 0 0 0 1 1 0 0 1 1 0 0 0 1 0 0 0 0 1 0 0 0 1 0 0 0 0 0 1 1 1 1 0
 0 0 1 1 0 0 1 1 1 0 1 0 0 0 0 0 0 1 1 0 0 1 0 0 0 0]
```
every time I run the script, regardless of whether I use `K.set_seed(123)` in the beginning. Do you get something different?

If I read the code correctly, using `seed=...` when defining the channel just [sets the numpy random seed](https://github.com/qiboteam/qibo/blob/e66224d1f2792298d792884dd6db82cd0115e834/src/qibo/core/gates.py#L936) at that time. So any seed setting you do before that will not have any effect, only the last seed setting will matter.

Regarding the tests, I still get a failure for qibojit-cupy and tensorflow, but now because the output differs from the target. Have you tested in an environment with tensorflow or cupy?
This is because `K.set_seed` sets the seed only for the active backend (tensorflow or cupy resepectively), while `PauliNoiseChannel` uses `np.random` when deciding which Pauli to apply, regardless of the active backend. Adding a `np.random.seed` before each circuit execution in the test solves the issue for me.

andrea-pasquale:
> Regarding the tests, I still get a failure for qibojit-cupy and tensorflow, but now because the output differs from the target. Have you tested in an environment with tensorflow or cupy?
> This is because `K.set_seed` sets the seed only for the active backend (tensorflow or cupy resepectively), while `PauliNoiseChannel` uses `np.random` when deciding which Pauli to apply, regardless of the active backend. Adding a `np.random.seed` before each circuit execution in the test solves the issue for me.

I understand. I've implemented the fix and now tests are passing for all backends locally on cpu and gpu. Thanks!

>The example from the last post gives me
[Qibo 0.1.8.dev0|INFO|2022-03-25 15:50:00]: Using qibojit (numba) backend on /CPU:0
[0 1 0 0 0 1 0 0 1 0 0 0 0 1 1 1 1 0 0 0 1 1 1 0 0 0 0 0 1 0 0 0 0 1 0 0 1
 1 0 0 0 0 0 0 0 1 1 0 0 1 1 0 0 0 1 0 0 0 0 1 0 0 0 1 0 0 0 0 0 1 1 1 1 0
 0 0 1 1 0 0 1 1 1 0 1 0 0 0 0 0 0 1 1 0 0 1 0 0 0 0]
every time I run the script, regardless of whether I use K.set_seed(123) in the beginning. Do you get something different?

Yes I get the same here. Instead if I do something like this:
```python
from qibo import models, gates, K
import numpy as np

K.set_seed(123)

circ = models.Circuit(1)
circ_no_noise = models.Circuit(1)

for i in range(10):
    new_gate = gates.H(0)
    circ.add(gates.PauliNoiseChannel(0, pz=0.1, seed=np.random.randint(100)))
    circ.add(new_gate)
    circ_no_noise.add(new_gate)

circ.add(gates.PauliNoiseChannel(0, pz=0.1))
circ += circ_no_noise.invert()
circ.add(gates.M(0))

print(circ(nshots=5).samples()[:, 0])
```
The result is always the same only if I put the set_seed at the beginning, which make senses since I am fixing the numpy seed at the beginning with `K.set_seed`.

stavros11 (approved):
Thanks, everything works for me too now.

> The result is always the same only if I put the set_seed at the beginning, which makes sense since I am fixing the numpy seed at the beginning with `K.set_seed`.

Yes, but I believe this is expected because you are using a random seed in the `PauliNoiseChannel` and `K.set_seed` is fixing this. If you don't use a random seed but a fixed number, the channel seed should work and give the same results in every run.

Issue #558, opened 2022-03-09
Using Hamiltonians with different qubit numbers on QAOA


stavros11:
Following from #556, the following script
```Python
import numpy as np
from qibo import hamiltonians, models
from qibo.symbols import X, Z

h_obj = hamiltonians.SymbolicHamiltonian(Z(0)*Z(1)+Z(1)*Z(2)+Z(2)*Z(3) + Z(3)*Z(0))
h_mixer = hamiltonians.SymbolicHamiltonian(X(0)*X(1)*X(2))
qaoa = models.QAOA(h_obj, mixer=h_mixer)

initial_parameters = 0.01 * np.random.uniform(0, 1, 4)
best_energy, final_parameters, extra = qaoa.minimize(initial_parameters, method="BFGS")
```
fails with
```
ValueError: Invalid tensor shape (16,) for state of 3 qubits.
```
because the second Hamiltonian acts on three qubits while the first acts on four. I had a look at the code and I am not sure how easy it is to implement this for Hamiltonians acting on different qubit numbers. 

Right now the number of qubits of a Hamiltonian is fixed:
* if the Hamiltonian is created using a matrix the number of qubits is given by the user and should be consistent with the given matrix size (which should be 2^n x 2^n),
* if the Hamiltonian is created from symbols the number of qubits is inferred from the expression. For example `X(0) * X(1)` will create two qubit Hamiltonian, `X(0) * X(1) * X(3)` four qubit, etc.

QAOA requires applying the exponential of the Hamiltonian to the state.
* if the Hamiltonian is created using matrix, this is calculated using `expm` on the full matrix and is applied to the state,
* if the Hamiltonian is created using symbols, a qibo circuit that implements the exponential via the Trotter decomposition is created and applied to the state.

In both cases, the number of qubits in the state should agree with the number of qubits in the Hamiltonian. In the first case because the matrix dimension is fixed and in the second case because the number of qubits of the Trotter circuit is the number of qubits associated with the Hamiltonian.

To fix the above QAOA issue we could do one of the following:
1. Force the user to use Hamiltonians with the same qubit numbers in the QAOA by raising an error if this condition is not satisfied. This should be simple as one can use the identity symbol implemented in #557 to add qubits in the smaller Hamiltonian.
2. Automatically expand the smaller Hamiltonian using identities within the QAOA (a bit messy solution).
[...]

PR #565, opened 2022-03-25
Raise error if QAOA Hamiltonians have different nqubits

stavros11:
Temporary fix for #558 until we make the model more flexible to allow applying operators of different sizes.


DiegoGM91:
Thanks for implementing this.

I just had a look and looks good to me. The only doubt I have is whether we should raise a `ValueError` or a `NotImplementedError`.

stavros11:
> I just had a look and looks good to me. The only doubt I have is whether we should raise a `ValueError` or a `NotImplementedError`.

Thank you for the review. I am fine either way, given that we plan to implement this feature. Let's see if someone else has a preference.

scarrazza (approved):
Thanks, I think ValueError matches the specific check we perform.

andrea-pasquale (approved):
Thanks for the implementation.
I also agree with `ValueError`.

PR #572, opened 2022-04-11
Add noise model

### LINKED ISSUE #567: Noise model
We need a simple and extensible noise model interface where we can store predefined noise models and eventually customize even more the noise mechanism. One possibility is to create an abstract noise model class:

```python
class Noise:
   ...
   def apply(self, circuit):
      ...
      return noisy_circuit
```

and then document how to store and perform custom noise optimization. This class should handle/store with `circuit.with_noise` and `noise_map`, and provide more flexibility. The final API should look like:
```python
c = Circuit()
noisy = MyCustomNoise()
c_noisy = noise.apply(c)
```

@andrea-pasquale could you please help identifying the initialization structure required for your setups?

andrea-pasquale:
Closes #567.
As already discussed in #567 this PR implement a noise model interface in Qibo.
The noise model implement a map that associate a quantum channel Q to a specific gate G. In particular, the quantum channel Q is added before every instance of the gate G. It is also possible specify for which qubits the noise will be added.
Here is an example 
EDIT: I've updated the current noise model API:
* Noise are added after the gates.
* `noise.add` supports directly the qibo gates no longer the qasm representation.
```python
from qibo.noise import PauliError, ThermalRelaxationError, NoiseModel

# declare quantum errors
pauli = PauliError(0, 0.2, 0.3)
thermal = ThermalRelaxationError(1, 1, 0.3)

# define noise model and add quantum errors
noise = NoiseModel() 
noise.add(pauli, gates.X, 1)
noise.add(pauli, gates.CNOT)
noise.add(thermal, gates.Z, (0,1))

circuit = Circuit(3)
circuit.add(gates.CNOT(0,1))
circuit.add(gates.Z(1))
circuit.add(gates.X(1))
circuit.add(gates.X(2))
circuit.add(gates.Z(2))

# create noisy circuit from noise model
print(noisy_circuit.draw())
```
with output
```
q0: ─o─PN───────────
q1: ─X─PN─Z─TR─X─PN─
q2: ───X──Z─────────
```
I mark this as WIP in case we want to implement other features related to the noise model in this PR.
This PR is also missing documentation.
* [x] documentation

### DIFF
--- a/doc/source/api-reference/qibo.rst
+++ b/doc/source/api-reference/qibo.rst
@@ -560,10 +560,10 @@ Thermal relaxation channel
 
 _______________________
 
-Noise Model
+Noise
 -----------
 
-In Qibo is possible to create a custom noise model using the
+In Qibo it is possible to create a custom noise model using the
 class :class:`qibo.noise.NoiseModel`. This enables the user to create
 circuits where the noise is gate and qubit dependent.
 
--- a/doc/source/api-reference/qibo.rst
+++ b/doc/source/api-reference/qibo.rst
@@ -561,7 +561,7 @@ Thermal relaxation channel
 _______________________
 
 Noise
------------
+-----
 
 In Qibo it is possible to create a custom noise model using the
 class :class:`qibo.noise.NoiseModel`. This enables the user to create
@@ -573,6 +573,9 @@ For more information on the use of :class:`qibo.noise.NoiseModel` see
 .. autoclass:: qibo.noise.NoiseModel
     :members:
     :member-order: bysource
+
+Quantum errors
+^^^^^^^^^^^^^^
 
 The quantum errors available to build a noise model are the following:
 
--- a/doc/source/api-reference/qibo.rst
+++ b/doc/source/api-reference/qibo.rst
@@ -560,6 +560,34 @@ Thermal relaxation channel
 
 _______________________
 
+Noise Model
+-----------
+
+In Qibo is possible to create a custom noise model using the
+class :class:`qibo.noise.NoiseModel`. This enables the user to create
+circuits where the noise is gate and qubit dependent.
+
+For more information on the use of :class:`qibo.noise.NoiseModel` see
+:ref:`How to perform noisy simulation? <noisemodel-example>`
+
+.. autoclass:: qibo.noise.NoiseModel
+    :members:
+    :member-order: bysource
+
+The quantum errors available to build a noise model are the following:
+
+.. autoclass:: qibo.noise.PauliError
+    :members:
+    :member-order: bysource
+
+.. autoclass:: qibo.noise.ThermalRelaxationError
+    :members:
+    :member-order: bysource
+
+.. autoclass:: qibo.noise.ResetError
+    :members:
+    :member-order: bysource
+
 
 .. _Hamiltonians:
 
--- a/src/qibo/noise.py
+++ b/src/qibo/noise.py
@@ -6,6 +6,7 @@ class PauliError():
         Args:
             options (tuple): see :class:`qibo.abstractions.gates.PauliNoiseChannel`
     """
+
     def __init__(self, px=0, py=0, pz=0, seed=None):
         self.options = px, py, pz, seed
         self.channel = gates.PauliNoiseChannel
@@ -17,6 +18,7 @@ class ThermalRelaxationError():
         Args:
             options (tuple): see :class:`qibo.abstractions.gates.ThermalRelaxationChannel`
     """
+
     def __init__(self, t1, t2, time, excited_population=0, seed=None):
         self.options = t1, t2, time, excited_population, seed
         self.channel = gates.ThermalRelaxationChannel
@@ -28,13 +30,35 @@ class ResetError():
         Args:
             options (tuple): see :class:`qibo.abstractions.gates.ResetChannel`
     """
+
     def __init__(self, p0, p1, seed=None):
         self.options = p0, p1, seed
         self.channel = gates.ResetChannel
 
 
 class NoiseModel():
-    """Class for the implementation of a custom noise model."""
+    """Class for the implementation of a custom noise model.
+
+        Example:
+        .. testcode::
+
+            from qibo import models, gates
+            from qibo.noise import NoiseModel, PauliError
+
+            # Build specific noise model with 2 quantum errors:
+            # - Pauli error on H only for qubit 1.
+            # - Pauli error on CNOT for all the qubits.
+            noise = NoiseModel()
+            noise.add(PauliError(px = 0.5), gates.H, 1)
+            noise.add(PauliError(py = 0.5), gates.CNOT)
+
+            # Generate noiseless circuit.
+            c = models.Circuit(2)
+            c.add([gates.H(0), gates.H(1), gates.CNOT(0, 1)])
+
+            # Apply noise to the circuit according to the noise model.
+            noisy_c = noise.apply(c)
+    """
 
     def __init__(self):
         self.errors = {}
@@ -51,6 +75,7 @@ class NoiseModel():
                 qubits (tuple): qubits where the noise will be applied, if None the noise
                                 will be added after every instance of the gate.
         """
+
         if isinstance(qubits, int):
             qubits = (qubits, )
 
@@ -67,6 +92,7 @@ class NoiseModel():
                 to the initial circuit with noise gates added according
                 to the noise model.
         """
+
         circ = circuit.__class__(**circuit.init_kwargs)
         for gate in circuit.queue:
             circ.add(gate)
--- a/src/qibo/noise.py
+++ b/src/qibo/noise.py
@@ -45,9 +45,8 @@ class NoiseModel():
             Return:
                 Circuit with noise.
         """
-        circ = circuit.copy()
-        count = 0
-        for num, gate in enumerate(circuit.queue):
+        circ = circuit.__class__(**circuit.init_kwargs)
+        for gate in circuit.queue:
             if gate.name in self.errors:
                 error, qubits = self.errors.get(gate.name)
                 if qubits is None:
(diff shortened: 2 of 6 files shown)


stavros11 (commented):
Thanks for implementing this. In terms of features, this looks complete to me. As far as I understand #567 is mostly focused on the API of adding noise to circuits which is implemented here. One straightforward addition one could propose is to extend the list of noise types, however this can be done later in collaboration with theorists/experimentalists.

In terms of code, my main concern is the use of `circ.queue.insert`, which may have some unwanted behavior. The issue is that `queue` is not a Python list but a custom `_Queue` object defined in abstractions/circuit.py and inserting elements to it using methods other than `.append` may break some of its features. For example, in your example, if you do
```py
print(circuit.depth)
print(noisy_circuit.depth)
```
you will see that both are 6, even though the noisy circuit is deeper.

Not related to this PR, but to avoid such issues in the future, perhaps we should remove the inheritance of `_Queue` from `list` or rename `circuit.queue` to `circuit._queue` to discourage modifying this outside the circuit.

Regarding this PR, one easy solution is to recreate the noisy circuit in `NoiseModel.apply`, instead of using copy, eg.:
```py
def apply(self, circuit):
    circ = circuit.__class__(**circuit.init_kwargs)
    for gate in circuit.queue:
        circ.add(...) # add channels associated with the gate's noisy
        circ.add(gate) # add the gate
```
Perhaps this would also simplify the code a bit as you do not need to keep track of `num` and `count` to insert the noise in the proper place.

Some other comments below:
The example in the first post misses some imports:
```py
from qibo import gates
from qibo.noise import PauliError, ThermalRelaxationError, NoiseModel
```

src/qibo/noise.py
```diff
@@ -0,0 +1,58 @@
+from qibo import gates
+
+class PauliError():
+  def __init__(self, px=0, py=0, pz=0, seed=None):
+    self.options = px, py, pz
```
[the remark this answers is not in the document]
andrea-pasquale: Fixed.

src/qibo/noise.py
```diff
@@ -0,0 +1,58 @@
...
+                circuit (`qibo.core.circuit.Circuit`): ideal quantum circuit
+
+            Return:
+                Circuit with noise.
+        """
+        circ = circuit.copy()
+        count = 0
+        for num, gate in enumerate(circuit.queue):
+            for noises in self.errors:
+                if gate.name == str(noises):
```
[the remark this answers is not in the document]
andrea-pasquale: Yes, you are right.

src/qibo/noise.py
```diff
@@ -0,0 +1,58 @@
...
+                    if self.errors[gate.name][1] is None:
```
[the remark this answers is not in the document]
andrea-pasquale: Also fixed.

andrea-pasquale:
Thanks for the review @stavros11.
I should have covered all the points that you mentioned. Now the code is much simpler.
Also now tests use both random states and random density matrices.
However, locally tests for me are failing when using qibojit-numba with random density matrices for the `PauliError` and the `ThermalRelaxationError`.
Could you please check if you can reproduce the issue?

stavros11 (commented):
Thanks for the updates. 

I checked the test issue and why it fails only for density matrix and I believe the behavior is expected. If you don't copy the initial state before passing it to the circuits, the three variables: `initial_state`, `final_state` and `target_final_state` will all point to the same array in memory. So
```py
K.assert_allclose(final_state, target_final_state)
```
performs a meaningless comparison that this array equals itself and always passes, even if `target_circuit` is different than `circuit`. This is not true when using density matrix because the `PauliNoiseChannel` [uses numpy addition](https://github.com/qiboteam/qibo/blob/0154fc738020ed32527eb1a0ff82060ef7b62c57/src/qibo/core/gates.py#L950) which silently creates a copy, so `final_state` and `target_final_state` point to different arrays (as they should) and the test fails.

Generally it is a good idea to use a copy when planning to use the same initial state on different circuits on the same script, otherwise the in-place updates may lead to strange behavior as the above.

Regarding this PR, everything looks good to me now. If documentation is added it should be good to go.

src/qibo/tests/test_noise.py
```diff
@@ -0,0 +1,112 @@
+import numpy as np
+import pytest
+from qibo import K, gates
+from qibo.models import Circuit
+from qibo.noise import *
+from qibo.tests.utils import random_density_matrix, random_state
+
```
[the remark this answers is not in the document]
andrea-pasquale: Thanks, I've fixed it.

andrea-pasquale:
I've added the documentation.
If it looks good we should be ready to merge this PR.

doc/source/api-reference/qibo.rst
```diff
@@ -560,6 +560,34 @@ Thermal relaxation channel
 
 _______________________
 
+Noise Model
```
stavros11, line 563: I think
```suggestion
Noise
```
looks better because all other section titles are single word.

doc/source/api-reference/qibo.rst
```diff
@@ -560,6 +560,34 @@ Thermal relaxation channel
...
+-----------
+
+In Qibo is possible to create a custom noise model using the
```
stavros11, line 566:
```suggestion
In Qibo it is possible to create a custom noise model using the
```

doc/source/api-reference/qibo.rst
```diff
@@ -560,6 +560,34 @@ Thermal relaxation channel
...
+circuits where the noise is gate and qubit dependent.
+
+For more information on the use of :class:`qibo.noise.NoiseModel` see
+:ref:`How to perform noisy simulation? <noisemodel-example>`
+
+.. autoclass:: qibo.noise.NoiseModel
+    :members:
+    :member-order: bysource
+
+The quantum errors available to build a noise model are the following:
```
stavros11, line 577: We could have a separate subsection for the quantum errors here:

```suggestion
Quantum errors
^^^^^^^^^^^^^^^

The quantum errors available to build a noise model are the following:
```

src/qibo/noise.py
```diff
@@ -0,0 +1,81 @@
+from qibo import gates
+
+class PauliError():
+    """Quantum error associated with the `qibo.core.gates.PauliNoiseChannel`.
+
+        Args:
+            options (tuple): see :class:`qibo.core.gates.PauliNoiseChannel`
+    """
```
stavros11, line 8: I would use
```suggestion
    """Quantum error associated with the :class:`qibo.abstractions.gates.PauliNoiseChannel`.

        Args:
            options (tuple): see :class:`qibo.abstractions.gates.PauliNoiseChannel`
    """
```
so that the link to the gate docs works properly when viewing the html version that will be online.

src/qibo/noise.py
```diff
@@ -0,0 +1,81 @@
...
+        self.options = px, py, pz, seed
+        self.channel = gates.PauliNoiseChannel
+
+
+class ThermalRelaxationError():
+    """Quantum error associated with the `qibo.core.gates.ThermalRelaxationChannel`.
+
+        Args:
+            options (tuple): see :class:`qibo.core.gates.ThermalRelaxationChannel`
+    """
```
stavros11, line 19: Same here
```suggestion
    """Quantum error associated with the :class:`qibo.abstractions.gates.ThermalRelaxationChannel`.

        Args:
            options (tuple): see :class:`qibo.abstractions.gates.ThermalRelaxationChannel`
    """
```

src/qibo/noise.py
```diff
@@ -0,0 +1,81 @@
...
+        self.options = t1, t2, time, excited_population, seed
+        self.channel = gates.ThermalRelaxationChannel
+
+
+class ResetError():
+    """Quantum error associated with the `qibo.core.gates.ResetChannel`.
+
+        Args:
+            options (tuple): see :class:`qibo.core.gates.ResetChannel`
+    """
```
stavros11, line 30: Same here
```suggestion
    """Quantum error associated with the `qibo.abstractions.gates.ResetChannel`.

        Args:
            options (tuple): see :class:`qibo.abstractions.gates.ResetChannel`
    """
```

src/qibo/noise.py
```diff
@@ -0,0 +1,81 @@
...
+    def __init__(self, p0, p1, seed=None):
+        self.options = p0, p1, seed
+        self.channel = gates.ResetChannel
+
+
+class NoiseModel():
+    """Class for the implementation of a custom noise model."""
```
stavros11, line 37: I would add a short example here, such as the one you have in the first post, particularly to show the usage of `.add` and demonstrate the API more clearly (for example that the `gate` argument should be a class type, eg. `gates.X`, etc.).

src/qibo/noise.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def add(self, error, gate, qubits=None):
+        """Add a quantum error for a specific gate and qubit to the noise model.
+
+            Args:
+                error: quantum error to associate with the gate. Possible choices
+                       are :class:`qibo.noise.PauliError`,
+                       :class:`qibo.noise.ThermalRelaxationError` and
+                       :class:`qibo.noise.ResetError`.
+                gate (:class:`qibo.core.gates`): gate after which the noise will be added.
```
stavros11, line 50:
```suggestion
                gate (:class:`qibo.abstractions.gates.Gate`): gate after which the noise will be added.
```

stavros11 (commented):
Thanks for adding the docs, looks good to me, some minor comments below.

After reading the updated docs, I believe that at this stage we could remove completely the `circuit.with_noise` method, since that the new `NoiseModel` replaces all its functionality and is more clear and flexible. @andrea-pasquale, @scarrazza what do you think?

doc/source/api-reference/qibo.rst
andrea-pasquale: I agree. It is more readable this way.

src/qibo/noise.py
andrea-pasquale: I agree. I've added the same example.

scarrazza:
I agree, we can keep the `with_noise`, in some circumstances this API is quite simple and quick to manage.

PR #573, opened 2022-04-13
TSP class

shangtai:
We introduce a class to generate Hamiltonians associated with TSP instances from https://arxiv.org/pdf/1709.03489.pdf  

within qibo.hamiltonian.py, X, Y, Z are defined there. 

Hence, for now I do the following:

    from qibo.symbols import X as tspX

Feel free to edit and reposition the code.

We have also coded code for circuit folding.


src/qibo/hamiltonians.py
```diff
@@ -186,3 +188,129 @@ def MaxCut(nqubits, dense=True):
     if dense:
         return ham.dense
     return ham
+
+
+from qibo.symbols import X as tspX
+from qibo.symbols import Y as tspY
+from qibo.symbols import Z as tspZ
```
stavros11, line 195: Another way, which avoids the global import, would be to do
```py
from qibo import symbols
self.X = symbols.X
self.Y = symbols.Y
self.Z = symbols.Z
```
inside `__init__` and then use `self.X`, etc. in the code.

src/qibo/hamiltonians.py
```diff
@@ -186,3 +188,129 @@ def MaxCut(nqubits, dense=True):
...
+class TSP:
+    """
+    This is a TSP class that enables us to implement TSP according to
```
stavros11, line 198: I would explain the TSP acronym the first time is used:
```suggestion
    This is a TSP (Travelling Salesman Problem) class that enables us to implement TSP according to
```

src/qibo/hamiltonians.py
```diff
@@ -186,3 +188,129 @@ def MaxCut(nqubits, dense=True):
...
+    https://arxiv.org/pdf/1709.03489.pdf by Hadfield (2017).
```
stavros11, line 199: Please follow the sphinx syntax for links
```suggestion
    `arxiv:1709.03489<https://arxiv.org/abs/1709.03489>`_ by Hadfield (2017).
```
so that this is nicely rendered in the docs.

src/qibo/hamiltonians.py
```diff
@@ -186,3 +188,129 @@ def MaxCut(nqubits, dense=True):
...
+
+    num_cities = 3
+    distance_matrix = np.random.rand(num_cities, num_cities)
+    distance_matrix = distance_matrix.round(1)
+    print(distance_matrix)
+    small_tsp = TSP(distance_matrix)
+    obj_hamil, mixer = small_tsp.TspHamiltonians(dense=False)
+    initial_parameters = np.random.uniform(0, 1, 2)
+    initial_state = small_tsp.PrepareInitialStateTsp([i for i in range(num_cities)])
+    qaoa = QAOA(obj_hamil, mixer=mixer)
```
stavros11, line 210: Same here, the proper syntax is:
```py
Example:
        .. testcode::

             # your example code here
```

src/qibo/hamiltonians.py
```diff
@@ -186,3 +188,129 @@ def MaxCut(nqubits, dense=True):
...
+        self.two_to_one = dict()
+        self.num_cities = distance_matrix.shape[0]
+        counter = 0
+        for i in range(self.num_cities):
+            for j in range(self.num_cities):
+                self.two_to_one[(i, j)] = counter
+                counter += 1
+
+
+    def PrepareObjTsp(self, dense=True):
```
stavros11, line 228: Not really important, but we tend to use lowercase and _ for method names, eg. these would be `prepare_obj_tsp`. Also I am not sure if it is necessary to include `tsp` in all methods, since they are part of the TSP class anyway.

src/qibo/hamiltonians.py
```diff
@@ -186,3 +188,129 @@ def MaxCut(nqubits, dense=True):
...
+
+        Returns: Hamiltonian describing the mixer.
+
+        """
+        form = 0
+        for i in range(self.num_cities):
+            for u in range(self.num_cities):
+                for v in range(self.num_cities):
+                    if u != v:
+                        form += self.SPlusTsp(u, i) * self.SPlusTsp(v, (i+1) % self.num_cities) * self.SNegTsp(u, (i+1)%self.num_cities) * self.SNegTsp(v, i) + self.SNegTsp(u, i) * self.SNegTsp(v, (i+1) % self.num_cities) * self.SPlusTsp(u, (i+1) % self.num_cities) * self.SPlusTsp(v, i)
```
stavros11, line 286: Very long lines usually look better if split to multiple lines.

stavros11 (commented):
Thank you for this contribution, I think it is nice to have. A few comments for discussion:

I am not sure if we really need a class to implement the Hamiltonians. In particular I believe that we could have just two separate functions that create the objective and mixer Hamiltonians, for example:
```py
def calculate_two_to_one(num_cities):
    # calculate ``two_to_one``
    # can also be done with numpy
    return two_to_one


def tsp_objective(distance_matrix):
    num_cities = len(distance_matrix)
    two_to_one = calculate_two_to_one(num_cities)
    # create the ``SymbolicHamiltonian``...


def tsp_mixer(num_cities):
    two_to_one = calculate_two_to_one(num_cities)
    splus = lambda u, i: X(two_to_one[u, i]) + 1j * Y(two_to_one[u, i])
    sminus = lambda u, i: X(two_to_one[u, i]) - 1j * Y(two_to_one[u, i])
    # create the ``SymbolicHamiltonian``...
```

Regarding the initial state preparation circuit, I am not sure if it's a good idea to have it in hamiltonians.py. 

Generally, I would consider following a template similar to [`qibo.models.hep`](https://github.com/qiboteam/qibo/blob/master/src/qibo/models/hep.py). We can create a `qibo.models.tsp` which implements a `TSP` class that contains the initial state preparation and some method to perform the optimization (not implemented here, but could be useful). Then we could have the above methods for constructing the Hamiltonians in the same file, similarly to the [`qpdf_hamiltonian`](https://github.com/qiboteam/qibo/blob/0154fc738020ed32527eb1a0ff82060ef7b62c57/src/qibo/models/hep.py#L80) in the HEP example. That's just a suggestion, let me know your opinion.

A few other more technical comments, some of which may not be relevant if we follow the above structure:

src/qibo/hamiltonians.py
shangtai: Thank you so much for your suggestions and feedbac. Thanks for introducing me to Sphinx as well.  

 I have changed some files (and also introduce some bugs). Let me work on it a little bit more.

shangtai: Hello, I tried running QAOA and I obtained the following state vector output:

`(0.07208-0.29489j)|001010100> + (0.01032-0.13008j)|001100010> + (0.38305-0.14509j)|010001100> + (-0.37556-0.44498j)|010100001> + (0.60235+0.0927j)|100001010> + (-0.10676-0.03421j)|100010001>`

Any advice on  what I can do about it? such as how do I measure it multiple times? or convert it to a dictionary encoding the probability in Qibo?

src/qibo/hamiltonians.py
```diff
@@ -186,3 +188,129 @@ def MaxCut(nqubits, dense=True):
     if dense:
         return ham.dense
     return ham
+
+
+from qibo.symbols import X as tspX
+from qibo.symbols import Y as tspY
+from qibo.symbols import Z as tspZ
+class TSP:
```
scarrazza, line 196: @shangtai thanks for this. Following our discussion, I believe this class should return the hamiltonian object, similarly to the other functions defined in this file. The TSP hamiltonian can then be used inside the tsp model object.

src/qibo/models/error_mitigation.py
```diff
@@ -0,0 +1,35 @@
+'''
```
scarrazza, line 1: @shangtai I believe this file can be removed, right?

src/qibo/models/tsp.py
```diff
@@ -0,0 +1,172 @@
...
+                            i + 1) % num_cities) * sminus(v, i) + sminus(u, i) * sminus(v, (
+                            i + 1) % num_cities) * splus(u, (i + 1) % num_cities) * splus(
+                        v, i)
+    ham = SymbolicHamiltonian(form)
+    if dense:
+        ham = ham.dense
+    return ham
+
+
+class tsp:
```
scarrazza, line 75: Following my comment above, if we could reuse the TSP hamiltonian here, would be great!

src/qibo/models/error_mitigation.py
shangtai: yes, it can be removed.

src/qibo/hamiltonians.py
shangtai: hmm... seems that I have a version control problem. I have deleted the TSP code from hamiltonians.py and it is currently only appear in TSP class under model.

shangtai:
I have removed the files that are not needed. 

Besides TSP code, I have commited code that convert qibo circuit to a Cirq circuit and perform the reverse operations as well. This is motivated by Mitiq doesn't recognize a Qibo circuit. To do so, I also added mitiq as part of the requirement.  (If the TSP code is ready, perhaps I can delete this file momentarily while it is under review to close the TSP part?)

requirements.txt
```diff
@@ -0,0 +1,11 @@
+qibo~=0.1.7
```
scarrazza, line 1: Please remove this file.

src/qibo/models/mitiqinterface.py
```diff
@@ -0,0 +1,51 @@
+from cirq import Circuit as cirqCircuit
```
scarrazza, line 1: Please move this example to the example's folder, otherwise we have to include cirq as dependency.

scarrazza (commented):
@shangtai, we have to rebase this branch from master. 
Could you please open another PR from master directly from this repository (no fork)?

shangtai:
can't figure out how to open another PR for now. Let me close it and see if I can create another PR without forking.

PR #583, opened 2022-05-10
Fix noise model

### LINKED ISSUE #582: Noise model doesn't work with circuits including measurements
Currently when generating the noisy circuit from a `NoiseModel` the measurements gate are removed.

```python
from qibo import gates, models, set_backend
from qibo.noise import PauliError, ThermalRelaxationError, NoiseModel, ResetError

pauli = PauliError(0, 0.2, 0.3)
thermal = ThermalRelaxationError(2, 1.5, 0.3)
noise = NoiseModel() 
noise.add(pauli, gates.X, 1)
noise.add(thermal, gates.Z, (0,1))

circuit = models.Circuit(3)
circuit.add(gates.CNOT(0,1))
circuit.add(gates.Z(1))
circuit.add(gates.X(1))
circuit.add(gates.X(2))
circuit.add(gates.Z(2))
circuit.add(gates.M(0,1,2))

print(noisy_circuit.draw())
# output
# q0: ─o────────
# q1: ─X─Z─X─PN─
# q2: ───X─Z────
```
I'll open a PR shortly to fix this issue.

andrea-pasquale:
Closes #582. 
I've also updated tests and now they also include a comparison of the samples generated.
The samples comparison is only performed with `PauliError` since for the `ThermalRelaxation` and the `ResetError` it may get tricky because if a qubit collapses and then there is a gate applied after the collapse an error is generated, as it should be.

I've also fixed the error that is making tests fail on the master branch, so once we merge this everything should be ok.


src/qibo/noise.py
```diff
@@ -93,16 +93,19 @@ def apply(self, circuit):
...
                     qubits = gate.qubits
                 else:
                     qubits = tuple(set(gate.qubits) & set(qubits))
                 for q in qubits:
-                    circ.add(error.channel(q, *error.options))
-        return circ
\ No newline at end of file
+                    noisy_circuit.add(error.channel(q, *error.options))
+        noisy_circuit.parametrized_gates = list(circuit.parametrized_gates)
+        noisy_circuit.trainable_gates = list(circuit.trainable_gates)
```
stavros11, line 108: I think we should remove these two lines because:
1. In the original circuit these are not lists but [`_ParametrizedGates`](https://github.com/qiboteam/qibo/blob/20c6f3f22effbeccd5d31ed456717f9bee449e0c/src/qibo/abstractions/circuit.py#L13) objects. I am not sure if this makes a big difference but making them lists may break some other functionality (eg. using set_parameters, get_parameters in the noisy circuit).
2. Since we add all original gates properly in the noisy circuit using the `.add` method, this should properly update the `parametrized_gates` and `trainable_gates` lists and we don't need to worry about copying them.

src/qibo/tests/test_abstract_circuit.py
```diff
@@ -266,7 +267,7 @@ def test_circuit_light_cone():
     c.add(gates.CZ(i, i + 1) for i in range(1, nqubits - 1, 2))
     c.add(gates.CZ(0, nqubits - 1))
     sc, qubit_map = c.light_cone(4, 5)
-    target_qasm = """// Generated by QIBO 0.1.8.dev0
+    target_qasm = f"""// Generated by QIBO {__version__}
```
stavros11, line 270: Good catch, thanks!

stavros11 (commented):
Thanks for this fix. It looks good to me, except a single comment below:

src/qibo/noise.py
andrea-pasquale: Ok, I'll remove them. For these lines I took inspiration from the [`with_noise`](https://github.com/qiboteam/qibo/blob/20c6f3f22effbeccd5d31ed456717f9bee449e0c/src/qibo/abstractions/circuit.py#L425-L426) method. I understand that since I am using `add` they will probably get updated automatically. 


> 1. In the original circuit these are not lists but [`_ParametrizedGates`](https://github.com/qiboteam/qibo/blob/20c6f3f22effbeccd5d31ed456717f9bee449e0c/src/qibo/abstractions/circuit.py#L13) objects. I am not sure if this makes a big difference but making them lists may break some other functionality (eg. using set_parameters, get_parameters in the noisy circuit).

Regarding this maybe it is better if we remove `list` in the corresponding lines of the `with_noise` method if you think that they can break some functionalities. Even if it is not strictly related to this PR I can do it here.

stavros11: > Regarding this maybe it is better if we remove `list` in the corresponding lines of the `with_noise` method if you think that they can break some functionalities. Even if it is not strictly related to this PR I can do it here.

You are right, this is not properly implemented in `with_noise` too. Here is an example:
```py
from qibo import gates

c = Circuit(2)
c.add(gates.RX(0, theta=0))
c.add(gates.U3(1, theta=0, phi=0, lam=0))
nc = c.with_noise((0.1, 0.1, 0.1))
```
If you do:
```py
c.set_parameters([0.1, 0.2, 0.3, 0.4])
```
it works, while
```py
nc.set_parameters([0.1, 0.2, 0.3, 0.4])
```
fails with an error that list has no attribute `nparams`.

If you like you can fix this, but remember to also change [`noisy_circuit.queue.append(gate)`](https://github.com/qiboteam/qibo/blob/20c6f3f22effbeccd5d31ed456717f9bee449e0c/src/qibo/abstractions/circuit.py#L422) to `noisy_circuit.add(gate)` so that all gates are properly added in the noisy circuit (as in your method). I see that there is a comment to not use `.add` because the gates were already added in the original circuit but I believe this is outdated now, as I tried switching this to `.add` and it works (all tests pass).

andrea-pasquale: Thanks for the example, now I see why this could be a problem. I can fix it now. 
I am also going to remove the comment since it is outdated.

PR #584, opened 2022-05-12
Backend agnostic circuits

stavros11:
@scarrazza following our discussion, here is a proposal to make circuits backend agnostic and delegate the execution to the engine. I defined a `NumpyEngine` (possibly to be renamed to `NumpyBackend` in order to keep our current conventions and not confuse users) which contains the following methods
* `asmatrix(gate)`: Converts abstract gates to the corresponding matrix representation.
* `apply_gate(gate, state, nqubits)`: Applies an abstract gate to a state vector of nqubits.
* `execute_circuit(circuit, initial_state, nshots)`: Executes a circuit (shots not implemented yet).
and a few other helper methods (some of which could be cached using what you proposed).

The modules `qibo.abstractions.circuit.AbstractCircuit` and `qibo.abstractions.gates` are promoted to `qibo.models.Circuit` and `qibo.gates` as the user will use the current abstract (backend agnostic) to define the circuit and gates. The general idea is to promote objects in `qibo.abstractions` to the main qibo objects and move the implementations of `qibo.core` to the engines.

The following script works with this branch:

```py
from qibo import gates

c = Circuit(2)
c.add(gates.H(0))
c.add(gates.RX(0, 0.1))
c.add(gates.RY(1, 0.2))
c.add(gates.CZ(0, 1))
c.add(gates.Y(0))
c.add(gates.U2(0, 0.1, 0.2))
c.add(gates.U3(0, 0.1, 0.2, 0.3))

# qibo-like approach (uses global engine)
result = c()
print(result)

# qiskit-like approach (uses custom engine)
from qibo.engines import NumpyEngine
engine = NumpyEngine()
result = engine.execute_circuit(c)
print(result)
```
The qibo-like approach will use a global backend which can be selected via `qibo.set_backend`, while the qiskit-like approach is useful in situations where the same circuit is executed using multiple backends.

The next steps would be the following:
1. Add a second non-numpy based engine (eg. cupy) to see what problems arise regarding the state type and possibly a hardware engine. Once a few engines are in place we can identify the `AbstractEngine`.
2. Do some benchmarks to see if we lose simulation performance due to "translating" the circuit on-the-fly.
3. Start re-implementing all qibo features (measurements, density matrices, Hamiltonians, etc.).

### Progress summary 

General features/modules:
- [x] States (#589)
- [x] Hamiltonians (#595)
- [x] Noise simulation (#594)
- [x] Callbacks (#593)
- [x] Parallel execution
- [x] Distributed circuit (multigpu)
- [x] Tensorflow backend (#596)
- [x] Tensorflow backpropagation (#596)

Gates:
- [x] Common gates
- [x] Measurements (#589)
- [x] Collapse (#589)
- [x] Channels (#594)
- [x] Fusion
- [x] PartialTrace (#593)

Models:
- [x] Circuit
- [x] Parameter set/get
- [x] Density matrix circuit
- [x] Repeated execution (#589)
- [x] QFT
- [x] Grover (#589)
- [x] HEP (#595)
- [x] Variational models
- [x] Time evolution (#603)


src/qibo/engines/__init__.py
```diff
@@ -1,6 +1,51 @@
...
+
+    def __new__(cls):
+        """Creates singleton instance."""
+        return cls._instance
+
+    @classmethod
+    def set_backend(cls, backend, platform):
+        if not config.ALLOW_SWITCHERS and backend != cls._instance.name:
+            log.warning("Backend should not be changed after allocating gates.")
+        # TODO: Use ``profiles.yml`` here
```
scarrazza, line 22: Not sure about keeping this `profiles.yml`. Seems to me that a simple env variable should be more than sufficient.

src/qibo/engines/matrices.py
```diff
@@ -0,0 +1,139 @@
+import numpy as np
+
+
+class Matrices:
```
scarrazza, line 4: Is there any special reason to use a class instead of single functions for each matrix?

src/qibo/models/circuit.py
```diff
@@ -786,8 +786,8 @@ def execute(self, initial_state=None, nshots=None):
         See :meth:`qibo.core.circuit.Circuit.execute` for more
         details.
         """
-        from qibo import engine
-        return engine.execute_circuit(self, initial_state, nshots)
+        from qibo.engines import GlobalSimulator
+        return GlobalSimulator().execute_circuit(self, initial_state, nshots)
```
scarrazza, line 790: Looks good.

src/qibo/engines/__init__.py
```diff
@@ -0,0 +1,111 @@
...
+        try:
+            return backend_cls()
+        except (ModuleNotFoundError, ImportError):
+            if show_error:
+                name = backend_cls.__name__
+                raise_error(ModuleNotFoundError, f"{name} is not installed.")
+            return None
+
+    if backend == "qibojit":
+        from qibojit.engines import CupyEngine, NumbaEngine
```
scarrazza, line 18: I believe this line will make the code crash if qibojit is not installed, right?
Maybe this could be moved inside `_initialize_backend` or wrapped around a try catch.

scarrazza:
> Thanks for testing. Yes, this is expected because I have not implemented the latest features in the qibojit engine yet. Qibojit contains only what is needed for simple state vector simulation with common gates to check performance. When I implement the new abstract methods it will work.

I see, so at the moment tests are not using any engine / state comparison.

stavros11:
> I see, so at the moment tests are not using any engine / state comparison.

In the latest push I enabled tests for qibojit-numba too. I had to fix a few things but if you now use the latest version of qibo and qibojit `abstractions` branches the tests should pass for numpy and qibojit-numba. 

I will try adding cupy tomorrow, but this should be easier as its structure is very close to numba. I am not sure what issues we will have with array types. I don't think these will be captured in the tests but it may affect performance. For example `set_parameters` assumes numpy backend and it is not clear to me if it is useful or no to make this backend dependend.

stavros11:
After #611, the following are remaining to conclude the refactoring:

- [] merge latest master and update TSP model (done in a different branch but have trouble with its tests),
- [] add `QIBO_BACKEND` and `QIBO_PLATFORM` environment variables,
- [] fix coverage.

Should be done by tomorrow.

scarrazza (approved):

PR #585, opened 2022-05-12
Circuit to unitary matrix

stavros11:
Adds a `circuit.unitary()` method which multiplies all the gates in the circuit and gives the corresponding 2^N x 2^N unitary that the circuit implements. This feature was requested by several people in the past.

I am not sure why the CI fails but I think it is not related to this PR, eg. see the [checkci](https://github.com/qiboteam/qibo/runs/6409507754?check_suite_focus=true) branch which contains a small change from master. It seems that there is an issue with the tests in macos.


andrea-pasquale (approved):
It looks good and also clean to me.
Tests are failing for the same reason also in another branch that I'm working on. It seems that there is an issue with qibotf on mac os. I'll look more into this on my local machine.

PR #587, opened 2022-05-19
qibotf removal

scarrazza:
Following our discussion this PR fixes workflows and removes qibotf dependencies and docs.

### DIFF
--- a/doc/source/api-reference/qibo.rst
+++ b/doc/source/api-reference/qibo.rst
@@ -883,6 +883,30 @@ variational model.
 .. automodule:: qibo.optimizers
    :members:
    :member-order: bysource
+   :exclude-members: ParallelBFGS
+
+.. _Parallel:
+
+Parallelism
+-----------
+
+We provide CPU multi-processing methods for circuit evaluation for multiple
+input states and multiple parameters for fixed input state.
+
+When using the methods below the ``processes`` option controls the number of
+processes used by the parallel algorithms through the ``multiprocessing``
+library. By default ``processes=None``, in this case the total number of logical
+cores are used. Make sure to select the appropriate number of processes for your
+computer specification, taking in consideration memory and physical cores. In
+order to obtain optimal results you can control the number of threads used by
+each process with the ``qibo.set_threads`` method. For example, for small-medium
+size circuits you may benefit from single thread per process, thus set
+``qibo.set_threads(1)`` before running the optimization.
+
+.. automodule:: qibo.parallel
+   :members:
+   :member-order: bysource
+   :exclude-members: ParallelResources
 
 .. _Backends:
 
--- a/doc/source/api-reference/qibo.rst
+++ b/doc/source/api-reference/qibo.rst
@@ -772,7 +772,7 @@ by the currently active backend (see :ref:`Backends <Backends>` for more informa
 A copy of the state can be created using :meth:`qibo.abstractions.states.AbstractState.copy`.
 The new state will point to the same tensor in memory as the original one unless
 the ``deep=True`` option was used during the ``copy`` call.
-Note that some backends (qibojit) perform in-place updates when the
+Note that the qibojit backend performs in-place updates
 state is used as input to a circuit or time evolution. This will modify the
 state's tensor and the tensor of all shallow copies and the current state vector
 values will be lost. If you intend to keep the current state values,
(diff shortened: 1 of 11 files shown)


stavros11:
Also, I noticed that we are removing the parallel functionality and parallel-BFGS. In principle this could work with qibojit or other backend too, right?

doc/source/api-reference/qibo.rst
```diff
@@ -772,7 +772,7 @@ by the currently active backend (see :ref:`Backends <Backends>` for more informa
 A copy of the state can be created using :meth:`qibo.abstractions.states.AbstractState.copy`.
 The new state will point to the same tensor in memory as the original one unless
 the ``deep=True`` option was used during the ``copy`` call.
-Note that some backends (qibojit, qibotf) perform in-place updates when the
+Note that some backends (qibojit) perform in-place updates when the
```
andrea-pasquale, line 775:
```suggestion
Note that the qibojit backend performs in-place updates
```
Since qibojit is now the only backend which performs in-place updates.

doc/source/getting-started/backends.rst
```diff
@@ -35,21 +35,20 @@ We provide multiple simulation backends for Qibo, which are automatically loaded
...
 
 The active default backend will be printed as an info message the first time
-Qibo is imported in the code. If qibojit and qibotf are not installed,
-an additional warning will appear prompting the user to install one of the two
-for increased performance and multi-threading and/or GPU capabilities.
-The logging level can be controlled using the ``QIBO_LOG_LEVEL`` environment
-variable. This can be set to 3 to hide info messages or 4 to hide both info
-and warning messages. The default value is 1 allowing all messages to appear.
+Qibo is imported in the code. If qibojit is not installed, an additional warning
+will appear prompting the user to install one of the two for increased
```
andrea-pasquale, line 47:
```suggestion
will appear prompting the user to install qibojit for increased
```

src/qibo/core/hamiltonians.py
```diff
@@ -13,7 +13,7 @@ class Hamiltonian(hamiltonians.MatrixHamiltonian):
         matrix (np.ndarray): Matrix representation of the Hamiltonian in the
             computational basis as an array of shape ``(2 ** nqubits, 2 ** nqubits)``.
             Sparse matrices based on ``scipy.sparse`` for numpy/qibojit backends
-            or on ``tf.sparse`` for tensorflow/qibotf backends are also
+            or on ``tf.sparse`` for tensorflow backends are also
```
andrea-pasquale, line 16:
```suggestion
            or on ``tf.sparse`` for the tensorflow backend are also
````

andrea-pasquale (approved):
I've added a few comments to fix some sentences in the docs.

scarrazza:
@stavros11 thanks for pointing the parallel out, indeed it works with qibojit. I also agree with merging this in #584.

@andrea-pasquale I think we should disable the parallel doctest, do you know if there is some configuration file where we can set this explicitly? (instead of removing the testcode tag in the doctstring)

andrea-pasquale:
> @andrea-pasquale I think we should disable the parallel doctest, do you know if there is some configuration file where we can set this explicitly? (instead of removing the testcode tag in the doctstring)

The doctest extension for sphinx provides a functionality to skip tests conditionally. See for example [here](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#skipping-tests-conditionally). However this would require to define the skipping condition on the `conf.py` file and add a `:skipif:` for each test that we want to skip.

Maybe at this point it is easier to remove the testcode tag in the docstring. We should only disable the two tests in https://github.com/qiboteam/qibo/blob/rmqibotf/src/qibo/parallel.py, correct?

scarrazza:
Yes, exactly. I give a try by disabling both tests.

scarrazza:
@stavros11 do you see why the coverage is not happy?

stavros11:
> @stavros11 do you see why the coverage is not happy?

I believe it is because all tests that involve multigpu are skipped and any related code is not covered. In the past we were running these tests using qibotf on CPU but now the only backend that supports multigpu is qibojit-cupy which is not available in the CI.

scarrazza:
Ok, as I imagined. Then lets merge this in your #584 and have the try/catch behaviour which at least allows us to test locally on GPUs.

PR #588, opened 2022-05-23
testing distributed circuits with cupy

scarrazza:
Includes cupy in actions. @stavros11 thanks for pointing this out.


stavros11 (approved):
Thanks for fixing this, I didn't know that cupy can be installed in CI!

PR #589, opened 2022-05-29
Circuit result

stavros11:
Implements a `CircuitResult` object that holds the final state vector and/or measurements. This replaces the current state objects. The API remains the same from the user perspective.

Implements the following features:
- [x] States
- [x] Measurements
- [x] Repeated execution
- [x] Collapse


src/qibo/backends/abstract.py
```diff
@@ -16,7 +18,7 @@ def __repr__(self):
         else:
             return f"{self.name} ({self.platform})"
 
-    @abc.abstractmethod
+    @abc.abstractmethod
```
scarrazza, line 21:
```suggestion
    @abc.abstractmethod
```

scarrazza (approved):

PR #594, opened 2022-06-03
Complete channel refactoring

stavros11:
This completes the refactoring of channels and noise as part of #584. I also fixed an issue with the `ResetChannel` which, as @andrea-pasquale discovered, was returning nan values in some cases due to bad mathematical definition.


scarrazza:
@andrea-pasquale could you please confirms this PR fixes the `ResetChannel` issue?

andrea-pasquale:
> @andrea-pasquale could you please confirms this PR fixes the `ResetChannel` issue?

Yes, now the issue is fixed. The only thing that I've noticed is that now the `ResetChannel` works only with density matrices, even if the `ThermalRelaxationChannel` which includes the `ResetChannel` works with and without density matrices.
@stavros11 is there a reason for this behavior or are you planning to implement the reset channel also for non-density matrix circuits?

andrea-pasquale:
> Thats right, but are you sure that thermal relaxation works without density matrices? I think in this branch both reset and thermal relaxation (for all T1, T2 cases) work only with density matrices, exactly because thermal relaxation uses the reset.

I tested it again and it is working without density matrices. This happens because I believe that we are missing the `apply` method in `ThermalRelaxationChannel`, similar to what you did in the `ResetChannel`. Right now in `ThermalRelaxationChannel` we have only the `apply_density_matrix` which calls directly the backend implementation of the reset channel. That's why it does not raise an error. 
https://github.com/qiboteam/qibo/blob/6564ff708bb41a9859397e3619d9d5b89f63ac8e/src/qibo/gates/channels.py#L295-L314


> I was planning to think about this after we are done with the refactoring as I was not sure how useful this feature is to spend time on it right now. If it is needed for some near term application (eg. randomized benchmarking) we can put a higher priority. We could also open an issue to remember this.

For the time being, I don't think that it is super necessary. If so, as you said, I'll open an issue. 
 
> Mathematically, I am not sure how to express the correct [reset definition](https://qiskit.org/documentation/stubs/qiskit.providers.aer.noise.reset_error.html) using states and repeated execution. In the past we were implementing reset using the "collapse" gate but it turned out wrong, so now I followed the exact definition using the partial trace. Note that Tr(rho) in the qiskit definition is a partial trace over the target qubit, for example if you have 20 qubits, Tr(rho) is a 19-qubit density matrix, not a number (the notation is a bit misleading).
> 
> There are actually two open questions: whether this can be done using collapse and how to do it for state vectors. The current implementation is doing the partial trace using numpy primitives (both for numpy and qibojit) so we lose the custom op speed-up, but at least is (hopefully) correct! :-)

I see, so it is non trivial to implement these channel for the state vector. Thanks for the explanation!

src/qibo/tests/test_noise.py
```diff
@@ -36,24 +36,21 @@ def test_pauli_error(backend, density_matrix, nshots):
...
 
     if nshots is None:
-        K.assert_allclose(final_state, target_final_state)
+        backend.assert_allclose(final_state, target_final_state)
     else:
-        K.assert_allclose(final_state_samples, target_final_state_samples)
+        backend.assert_allclose(final_state_samples, target_final_state_samples)
 
 
 @pytest.mark.parametrize("density_matrix", [False, True])
```
andrea-pasquale, line 52: Since the `ThermalRelaxationChannel` is working only with density matrices we should only test with the flag `density_matrix=True`. Similarly for the `ResetChannel`.

stavros11:
> I tested it again and it is working without density matrices. This happens because I believe that we are missing the `apply` method in `ThermalRelaxationChannel`, similar to what you did in the `ResetChannel`. Right now in `ThermalRelaxationChannel` we have only the `apply_density_matrix` which calls directly the backend implementation of the reset channel. That's why it does not raise an error.

Right, I believe it does nothing because the `self.gates` tuple is empty. I fixed it so that an error is raised and also skipped the noise test. Thanks!

scarrazza (approved):

PR #596, opened 2022-06-07
Add Tensorflow backend

stavros11:
This is an attempt to add the tensorflow backend on top of numpy with as little additional code as possible. It is based on [`tensorflow.experimental.numpy`](https://www.tensorflow.org/guide/tf_numpy) and the idea is to use `self.np` (instead of `np`) in every numpy call used in the numpy backend and then replace this with `self.np = tnp` for the tensorflow backend. It seems to work fairly well and it includes GPU, since the tnp ND array is an alias for `tf.Tensor` and not a numpy array. I have not tested backpropagation but at least according to their docs it is expected to work.

One potential disadvantage with this approach is that it requires to have a reference to the numpy module in the numpy backend, by doing `self.np = np`. This breaks the pickle serialization of the backend object because Python's `pickle` cannot pickle modules. Note that `dill` does not have this limitation ([related stackoverflow discussion](https://stackoverflow.com/questions/2790828/python-cant-pickle-module-objects-error)), however I am not sure how efficient this is, mostly in terms of the pkl file size. Moreover, this breaks parallel.py because multiprocessing uses pickle.


scarrazza:
@stavros11 could you please fix conflicts?

PR #600, opened 2022-06-14
Add Minimum Vertex Cover in example

ianmalcolm:
Add Minimum Vertex Cover (MVC) in example. The example includes the problem formulation of MVC problem as qubo or as hamiltonian. A demo markdown is included. Peripheral functions, such as the estimation of the penalty weight, checking the feasibility of the solution, a quick fix to constraints violation, are also included.

This is a squash of multiple commits in the past, to make the history clean. The complete list of commits is listed below.

> Add utils to convert between binary and spin models
> 
> Ensure all decision variables appear in the linear terms. In rare cases, a variable could be ignored if it has zero bias and iteractions and would be missing in final solutions and causes problems.
> 
> Add a negative sign to hamiltonian
> 
> Add the usage of converting mvc problem to hamiltonian and an example of adiabatic time evolution.
> 
> Add tests
> 
> Add test entry for mvc
> 
> Add hamiltonian_mvc, which directly creates a hamiltonian from a networkx graph
> 
> Remove duplicated documentaiton
> 
> Add test on hamiltonian_mvc
> 
> Revert "Remove duplicated documentaiton"
> 
> This reverts commit 1decc38ad68ee0b3062c372ae523676ef78f4436.
> 
> Remove duplicated documentation


scarrazza:
Closing in favour of #1002.

PR #574, opened 2022-04-18
Add Minimum Vertex Cover in example

ianmalcolm:
Hi @scarrazza, 

May I have your comments on this pull request, please?

Many thanks!

PR #601, opened 2022-06-14
Add Quadratic Assignment Problem (QAP) example

ianmalcolm:
Add quadratic assignment problem example. This commit is rebased to the most recent master branch.


scarrazza:
@ianmalcolm we are planning to renew the documentation rendering mechanism, so it would be great to understand if this and PR #574 are going to be included asap or not. Could you please advise?

PR #606, opened 2022-06-30
Variational models refactor

andrea-pasquale:
Refactor of the variational models.
Tests are passing on CPU, I only need to take care of tests on GPU.
@stavros11, now that all the pylint issues are fixed we could update the CI for all the PRs concerning the refactor, correct?
Maybe we should only avoid testing the docs since we still need to update it.

- [ ] GPU test


src/qibo/models/variational.py
```diff
@@ -67,24 +66,27 @@ def minimize(self, initial_state, method='Powell', jac=None, hess=None,
...
         def _loss(params, circuit, hamiltonian):
             circuit.set_parameters(params)
-            final_state = circuit()
+            final_state = circuit().state()
             return hamiltonian.expectation(final_state)
 
         if compile:
-            if K.is_custom:
+            #TODO: introduce is_custom attribute or maybe just compare with qibojit
+            if not self.hamiltonian.backend == 'qibojit':
```
stavros11, line 74: I think this causes some issues in the tensorflow tests. I don't think we need the `is_custom` attribute anymore since qibotf is removed, instead I would do one of:

1. Change this to
```suggestion
            if not self.hamiltonian.backend == 'tensorflow':
```
since tensorflow is the only backend that supports compilation.

2. Remove this check completely. If `backend.compile` is used with numpy or qibojit it will still work but will have no effect.

src/qibo/backends/tensorflow.py
```diff
@@ -212,6 +212,18 @@ def zero_density_matrix(self, nqubits):
...
+        state = self.tf.ones((2 ** nqubits,), dtype=self.dtype)
+        norm = self.tf.cast(self.np.sqrt(2 ** nqubits), dtype=self.dtype)
+        state = self.tf.math.scalar_mul(1/norm, state)
+        return state
+
+    def plus_density_matrix(self, nqubits):
+        state = self.tf.zeros(2 * (2 ** nqubits,), dtype=self.dtype)
+        norm = self.tf.cast(2 ** nqubits, dtype=self.dtype)
+        state = self.tf.math.scalar_mul(1/norm, state)
+        return state
```
stavros11, line 225: I have not tested this, but it may be possible to completely remove these methods if you use `self.np` in the respective methods in the numpy backend. This is because `self.np` becomes `tnp` in the tensorflow backend which is new feature in tensorflow and implements many numpy methods for `tf.Tensor` automatically.

stavros11 (commented):
Thanks for implementing this. I left some comments below, I think the missing points are the following:

* make sure that for all models the circuit is executed using the same backend as the given Hamiltonian (first comment). Otherwise the circuit will be executed using the `GlobalBackend()` and if this is different than the Hamiltonian backend (such as in the tests), it will create problems.
* I get several test failures when using tensorflow (even on CPU), some possibly associated with the point above. Have you tried to run pytest in an environment with tensorflow?

> @stavros11, now that all the pylint issues are fixed we could update the CI for all the PRs concerning the refactor, correct? Maybe we should only avoid testing the docs since we still need to update it.

Yes, I fixed CI excluding the doctests in the `abstrtnp` branch (included here). I was planning to open another PR with a final cleanup, updating the docs and fixing the CI and coverage. I like the idea of disabling the doctests and doing this in two steps. The order I would do that is the following:
1. Finalize this PR.
2. Rebase `abstrdistributed` on this and implement the missing accelerator features in models.
3. Fix CI (without doctests) in the latest branch.
4. Update docs/doctests.

I think it will be easier to do 2 before 3 otherwise pylint might complain about missing features.

andrea-pasquale:
Thanks for the quick review.

> * make sure that for all models the circuit is executed using the same backend as the given Hamiltonian (first comment). Otherwise the circuit will be executed using the `GlobalBackend()` and if this is different than the Hamiltonian backend (such as in the tests), it will create problems.

I agree, I'll fix this issue.

> * I get several test failures when using tensorflow (even on CPU), some possibly associated with the point above. Have you tried to run pytest in an environment with tensorflow?

What version of tensorflow are you using? I am using python 3.10 and tensorflow 2.9 and tests are passing on CPU (with and without implementing the previous fix). I know for sure that there are some issues with tensorflow on GPU. As soon as you push the CI fix in the `abstrtnp` we can run tests to see what is directly the issue with tf on CPU. 


> Yes, I fixed CI excluding the doctests in the `abstrtnp` branch (included here). I was planning to open another PR with a final cleanup, updating the docs and fixing the CI and coverage. I like the idea of disabling the doctests and doing this in two steps. The order I would do that is the following:
> 
> 1. Finalize this PR.
> 2. Rebase `abstrdistributed` on this and implement the missing accelerator features in models.
> 3. Fix CI (without doctests) in the latest branch.
> 4. Update docs/doctests.
> 
> I think it will be easier to do 2 before 3 otherwise pylint might complain about missing features.

Perfect, I agree with this plan!

src/qibo/models/variational.py
```diff
@@ -67,24 +66,27 @@ def minimize(self, initial_state, method='Powell', jac=None, hess=None,
         """
         def _loss(params, circuit, hamiltonian):
             circuit.set_parameters(params)
-            final_state = circuit()
+            final_state = circuit().state()
```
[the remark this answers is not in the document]
andrea-pasquale: I've noticed that this could cause problems when I was implementing tests. I forgot to change this line. Thanks!

stavros11:
> What version of tensorflow are you using? I am using python 3.10 and tensorflow 2.9 and tests are passing on CPU (with and without implementing the previous fix). I know for sure that there are some issues with tensorflow on GPU. As soon as you push the CI fix in the `abstrtnp` branch we can run tests to see what is directly the issue with tf on CPU.

With tensorflow 2.9.1 and Python 3.8 I get the following errors:
* RuntimeError: Cannot compile VQE that uses custom operators. Set the compile flag to False.
* TypeError: Cannot calculate Hamiltonian expectation value for state of type <class 'tensorflow.python.framework.ops.EagerTensor'>
* RuntimeError: SGD optimizer requires Tensorflow backend.

in different tests of `test_models_variational.py`. I also tried with Python 3.10 and I get the same.

I fixed the CI in the `abstrevol` branch by disabling the doctests and example tests temporarily. If you merge this here it should be possible to run the tests for this branch and see what happens for different versions.

src/qibo/models/variational.py
andrea-pasquale: I've ended up remove this check.

src/qibo/backends/tensorflow.py
andrea-pasquale: Yes, I agree. I've removed these methods. Thanks!

src/qibo/models/circuit.py
```diff
@@ -630,8 +630,11 @@ def set_parameters(self, parameters):
                 params = [0.123, 0.456, 0.789, 0.321]
                 c.set_parameters(params)
         """
+        import tensorflow as tf
         if isinstance(parameters, (list, tuple, np.ndarray)):
             self._set_parameters_list(parameters, len(parameters))
+        elif isinstance(parameters, (tf.Tensor, tf.Variable)):
+            self._set_parameters_list(parameters, int(parameters.shape[0]))
```
andrea-pasquale, line 637: There was an issue with tests regarding this.
Probably because we were missing this from the master branch.
I don't know if this is the best solution.
Let me know what you think @stavros11.

andrea-pasquale: Perfect, I also didn't like my first solution for the same reason.

scarrazza (approved):

PR #612, opened 2022-07-05
Fixing parallel

scarrazza:


src/qibo/parallel.py
```diff
@@ -177,46 +97,18 @@ def parallel_parametrized_execution(circuit, parameters, initial_state=None, pro
...
+    if not isinstance(parameters, list):
         from qibo.config import raise_error
         raise_error(RuntimeError, "parameters must be a list.")
 
-    _check_parallel_configuration(processes, backend)
-
-    def operation(params, circuit, state): # pragma: no cover
+    def operation(params, circuit, state):
+        if state is not None:
+            state = backend.cast(state)
```
stavros11, line 106: I would think that we need
```suggestion
            state = backend.cast(state, copy=True)
```
here but strangely the test passes as it is now, even if I change the test state to complex type.

src/qibo/tests/test_parallel.py
```diff
@@ -46,12 +30,6 @@ def test_parallel_circuit_evaluation(backend):  # pragma: no cover
 
 def test_parallel_parametrized_circuit(backend):  # pragma: no cover
```
stavros11, line 31: I believe we can re-enable coverage for the parallel tests.

PR #614, opened 2022-07-07
Fix coverage

stavros11:
Brings the coverage after refactoring back to 100%. The multigpu related code is excluded from coverage as it cannot be tested anymore.

This also removes the `Simulator` abstract class since the qibolab backend will inherit `NumpyBackend`. Depending on future backend developments we may add the `Simulator` abstraction if needed.


scarrazza (approved):
Great job, looks good to me.

PR #615, opened 2022-07-07
Add refactored TSP model

stavros11:
This merges the latest master to abstractions to fix the conflicts and refactors the TSP model. The related test is failing.

I tried running the same test from master several times in the cluster and it seems to fail with a similar error, so I am not sure if the failure is associated to the refactoring.


src/qibo/tests/test_models_tsp.py
```diff
@@ -0,0 +1,64 @@
...
+        standard_cauchy_key = convert_to_standard_Cauchy(freq_key)
+        cauchy_dict[standard_cauchy_key] += freq_counter[freq_key]
+    max_key = max(cauchy_dict, key=cauchy_dict.get)
+    # Given a permutation of 0 to n-1, we compute the distance of the tour
+    m = len(max_key)
+    return sum(distance_matrix[max_key[i]][max_key[(i + 1) % m]] for i in range(m))
+
+
+@pytest.mark.parametrize("test_layer, expected", [(4, 1.0), (6, 1.0), (8, 1.9)])
+def test_tsp(test_layer, expected):
```
scarrazza, line 60: I believe it is safe to add `backend` here, but see my comment above.

scarrazza:
I have tried to compare the output from this branch and the main, after applying my comments, it looks like the `quantum_state` object is different, and seems that master has some indexing issues.

@shangtai could you please check if the last configuration should really output 1.9 instead of 1.0?

src/qibo/tests/test_models_tsp.py
stavros11: Thanks for checking. Indeed it is okay to add `backend` but I believe some tests will still fail. The last configuration sometimes fails with numpy too so I do not think it is a copy issue.

shangtai:
Hello. Theoretically, it can take either value. To fix the value, I tried to fix the random seed. Can you check If I have fixed the random seed correctly? If not how should I do so to fix the randomness in QAOA?

shangtai:
The quantum measurement of QAOA would give random outcomes right? is there a way to fix it for example by fixing the backend? or during purest, we don't perform the measurement but by comparing the quantum states?

src/qibo/models/tsp.py
```diff
@@ -0,0 +1,160 @@
...
+
+        Returns:
+            An initial state that is used to start TSP QAOA.
+
+        """
+        c = Circuit(len(ordering) ** 2)
+        for i in range(len(ordering)):
+            c.add(gates.X(int(self.two_to_one[ordering[i], i])))
+        result = self.backend.execute_circuit(c)
+        return result.state(numpy=True)
```
scarrazza, line 160: I would suggest to drop numpy here:
```suggestion
        return result.state()
```

shangtai:
Hi @stavros11 ya, I have not found a seed that works for all the backend and  we might have more backends in the future. I think comparing the quantum state is a more robust solution in the long run.

scarrazza:
@stavros11 are you able to reproduce the sphinx compilation error spotted by the CI?

stavros11:
> @stavros11 are you able to reproduce the sphinx compilation error spotted by the CI?

I gave some tries but I was not able to reproduce yet. When I compile the docs locally, I skip the pandoc installation and the related file (qPDF notebook) because installing pandoc on Arch is a bit annoying (requires many dependencies). However it seems that the issue here is independent of pandoc. For example [here](https://github.com/qiboteam/qibo/actions/runs/2690483916) I disabled the pandoc installation and it stills fails with the same error when trying to compile the docs. The same configuration works on my laptop.

stavros11:
> Thanks, I think `sphinx_markdown_tables` has been updated (or some of its deps), we may try to downgrade the version and see what happens...

Indeed, it's the markdown library. My old environment had version 3.3.7 and it works. I can reproduce the error in a new environment with markdown version 3.4.1. Markdown 3.4 was released three days ago so that's why this issue appeared now.

PR #620, opened 2022-07-13
Fix TSP test

stavros11:
@shangtai this implements the TSP test by comparing the state vector, instead of doing the measurements, thus avoiding the issues with the seed. Please have a look and let me know if you agree.

I also added a copy in the loss function of the QAOA as it was causing the numba backend to fail, most likely due to in-place updates. Perhaps this was the reason we could not get it to work with the measurements either.


scarrazza:
@shangtai could you please have a look and confirm if this solution works for you?

shangtai:
Hi @stavros11 I think the change is great since we can now avoid the issue with the random number seed.

PR #621, opened 2022-07-15
Add dill tests

stavros11:
Adds tests which check that various qibo objects can be pickled using dill. I had to to make the following changes in the code to make this work:

* Remove the `DistributedCircuit` object and move its methods in `Circuit`, to avoid implementing the `Circuit.__new__` constructor which does not work well with dill.
* Remove the `solvers.Exponential.__new__` constructor and delegate the solver construction to a `get_solver` method.
* For qibo classes that inherit `sympy.Symbol` I could not find a way to avoid defining `__new__` and keeping the same behavior, so I implemented the `__getstate__` and `__setstate__` methods which fix the pickle.

I also did some basic tests of this using the notebook magics and it seems to work.


scarrazza (approved):
LGTM, thank you.

PR #622, opened 2022-07-18
Adding CuQuantum version in docs

andrea-pasquale:
In this PR I've added in the docs which CuQuantum versions work with qibojit as suggested in qiboteam/qibojit#78.
@stavros11 in the docs I've put both the 22.03 and the 22.05, I'll check more deeply if everything is working with the refactor.
I think so, since I don't think that they've made significant changes from one version to the other.
With which version did you last run tests?

### DIFF
--- a/.github/workflows/rules.yml
+++ b/.github/workflows/rules.yml
@@ -19,6 +19,7 @@ jobs:
     - name: Install package
       run: |
         python -m pip install --upgrade pip
+        pip install markdown==3.3.7 # docs fail with >=3.4
         pip install pytest-cov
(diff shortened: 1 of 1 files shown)


stavros11 (approved):
Thanks for updating the docs.

> With which version did you last run tests?

Currently 7 tests are failing with qibojit GPU. I think they are associated with #621, I probably forgot to test on GPU when I did this. The behavior of v22.03 and v22.05 seems to be the same so what you write here is true.

scarrazza (approved):

PR #623, opened 2022-07-18
Fix docs compilation

stavros11:
I pinned the version of markdown to 3.3.7 to fix the issue with the compilation of documentation. @scarrazza, I am not sure if that's the best solution but it works for now.

### DIFF
--- a/.github/workflows/rules.yml
+++ b/.github/workflows/rules.yml
@@ -19,6 +19,7 @@ jobs:
     - name: Install package
       run: |
         python -m pip install --upgrade pip
+        pip install markdown==3.3.7 # docs fail with >=3.4
         pip install pytest-cov
         pip install .[tests,docs]
         pip install git+https://github.com/qiboteam/qibojit


scarrazza:
@stavros11 thanks for this. I was playing with markdown 3.4.1 and it seems that we don't really need `sphinx_markdown_tables`, so if I remove this package sphinx compiles fine. Could you please give a try and let me know if that works for you?

stavros11:
> @stavros11 thanks for this. I was playing with markdown 3.4.1 and it seems that we don't really need `sphinx_markdown_tables`, so if I remove this package sphinx compiles fine. Could you please give a try and let me know if that works for you?

I tried removing `sphinx_markdown_tables` and sphinx compilation fails. See for example [here](https://github.com/qiboteam/qibo/runs/7406820926?check_suite_focus=true). I believe this package is used for rendering the tables in the benchmark results section of the docs. Are you using a different library or configuration?

scarrazza:
@stavros11 you have to remove it from `doc/source/conf.py` too.

stavros11:
Indeed, that works, however I think that the bechmark results tables are then not rendered properly:

Screenshot from 2022-07-19 19-41-50

scarrazza:
Good point, then let's merge this PR and possibly think about removing this section from the website.

scarrazza (approved):

Issue #625, opened 2022-07-21
docs build are failing


scarrazza:
Readthedocs is using the latest sphinx and plugins. The compilation is failing with:
```
Extension error (sphinx_markdown_tables):
Handler <function process_tables at 0x7f705b63fcb0> for event 'source-read' threw an exception (exception: __init__() missing 1 required positional argument: 'config')
```

stavros11:
This will cause CI issues in all new developments and releases. @scarrazza do you think that we should remove the old tables from the docs and remove this dependency?

scarrazza:
I think so, at this point they are obsolete.

PR #627, opened 2022-08-04
removing benchmark section

scarrazza:
Removes benchmark section and solves issue #625.


stavros11 (approved):
Thanks for fixing this. Looks good to me.

PR #628, opened 2022-08-04
Adding pre-commit

scarrazza:
Adding a basic pre-commit hook and applying black to all files.


scarrazza:
@stavros11 this is failing with a tsp+tensorflow issue, which looks familiar to me, do you remember if this is related to seeds?

andrea-pasquale:
Thanks for implementing this. 
I have a few comments:
- For now, are you happy with just adding black? Additionally we can include [isort](https://pycqa.github.io/isort/) to fix the imports order.
- Since this repository is public we can even add [precommit](https://pre-commit.ci/) to CI. With this we can either: (1) make CI fail if the user doesn't use the precommit hooks and/or (2) let the action automatically fix any issues with this (in this case we'll have an extra commit that follows the user commits).
- I feel like we need to add somewhere in the docs that we are using pre-commit, under `Developer guides` probably.

scarrazza:
@andrea-pasquale thanks for the comments.

> For now, are you happy with just adding black? Additionally we can include [isort](https://pycqa.github.io/isort/) to fix the imports order.

yes, but if you try here, you will see that qibo failes, due to circular import, so I think we can do that in another PR.

>    Since this repository is public we can even add [precommit](https://pre-commit.ci/) to CI. With this we can either: (1) make CI fail if the user doesn't use the precommit hooks and/or (2) let the action automatically fix any issues with this (in this case we'll have an extra commit that follows the user commits).

yes, I fully agree.

> I feel like we need to add somewhere in the docs that we are using pre-commit, under Developer guides probably.

yes, absolutely.

stavros11:
Thanks for implementing this. I checked locally and it works for me. 

> * Since this repository is public we can even add [precommit](https://pre-commit.ci/) to CI. With this we can either: (1) make CI fail if the user doesn't use the precommit hooks and/or (2) let the action automatically fix any issues with this (in this case we'll have an extra commit that follows the user commits).

I also agree with this. I think option (2) is more friendly for external developers, but we may get many additional commits if people push frequently (not sure if this is an issue).

> * I feel like we need to add somewhere in the docs that we are using pre-commit, under `Developer guides` probably.

I agree, we should add a related section in "How to contribute". We could also add the pre-commit libarary in setup.py as a requirement for the [tests] package, which is oriented towards developers.

> @stavros11 this is failing with a tsp+tensorflow issue, which looks familiar to me, do you remember if this is related to seeds?

Yes, I was aware of this from other branches and I can get reproduce it locally. An easy fix is to set `atol=1e-5` for this test, but before doing so I would like to understand better where this precision issue comes from and why it happens only with tensorflow, because it is a bit high to classify as numerical error. It should not be related to seeds as this test is not random since we removed the measurements and we are checking the state vector directly.

andrea-pasquale:
I've fixed the documentation and I've also added the pre-commit action.
@scarrazza I don't know if it will be enabled when we open another PR or if you as admin will need to install the Github action. I tried it to install the action but I think that I don't have the permissions.

scarrazza:
Thanks, I have granted permission, please try again now.

andrea-pasquale:
Now the action is there. I believe that it was failing because there were some conflicts.
If you want I can open a dummy PR and perform some checks.

scarrazza:
Sure, please go ahead, and indeed conflicts should not fail.

doc/source/developer-guides/contributing.rst
```diff
@@ -45,6 +45,7 @@ The current code standards that are applied to any new changes:
 - **Tests**: We use pytest to run our tests that must continue to pass when new changes are integrated in the code. Regression tests, which are run by the continuous integration workflow are stored in ``qibo/tests``. These tests contain several examples about how to use Qibo.
 - **Coverage**: Test coverage should be maintained / be at least at the same level when new features are implemented.
 - **Pylint**: Test code for anomalies, such as bad coding practices, missing documentation, unused variables.
+- **Pre commit**: We use pre-commit to enforce automatation and to format the code.
```
stavros11, line 48: Do you mean
```suggestion
- **Pre commit**: We use pre-commit to enforce automation and to format the code.
```
?

Perhaps we could also say that this is enforced automatically for all PRs.

stavros11 (approved):
Thank you for all the updates and tests @andrea-pasquale. I think this can be merged.

doc/source/developer-guides/contributing.rst
andrea-pasquale: Thanks for catching this.
Yes that's what I meant :)
Now that we've enabled the pre-commit action I should probably update this line.

scarrazza:
@stavros11 please let me know when you are happy with the atol increase so we can merge this.

PR #630, opened 2022-08-04
Add atol in TSP test

stavros11:
Hopefully CI will pass with this. 

After some quick tests it seems that the final parameters found by the BFGS optimization are slightly different for tensorflow which causes the final state to also be slightly different. If I fix the final parameters then the final states agree. I am still not sure what is causing the discrepancy in optimization.


scarrazza:
Thanks, I will merge this now, but one possibility is that tf is using math operations in a order which introduces rounding differences when compared to numpy / qibojit.

scarrazza (approved):

PR #632, opened 2022-08-05
add isort

scarrazza:
Adding isort to pre commit hook. @stavros11 please have a look at the `matrices` import, I had to explicitly change the imports otherwise we get circular import errors.


andrea-pasquale (approved):

stavros11:
Thanks for adding this, it works for me.

> @stavros11 please have a look at the `matrices` import, I had to explicitly change the imports otherwise we get circular import errors.

Do you mean that you changed 
```py
from qibo import matrices
```
to
```py
from qibo.backends import matrices
```
in a few places? I guess this was needed because isort imports the modules (symbols, hamiltonians, etc.) before matrices in __init__.py. Is there any issue with keeping it this way? From my side it's okay and tests pass, also on GPU.

scarrazza:
Yes, exactly, thanks for checking GPU tests.

stavros11 (approved):
Great, that's fine with me.

PR #636, opened 2022-08-25
Customize qaoa loss

shangtai:
I introduce general loss functions in QAOA.  

This is due to it has been [shown](https://arxiv.org/pdf/1907.04769.pdf) that general loss functions can improve the performance of the optimization.


stavros11 (commented):
Thank you for implementing this. I left some comments below, including how to execute the circuit with the proper backend in the Hamiltonian methods you have added. Could you please add some tests for all your new methods?

src/qibo/models/variational.py
```diff
@@ -507,13 +523,29 @@ def _loss(params, qaoa, hamiltonian, state):
             qaoa.set_parameters(params)
             state = qaoa(state)
             return hamiltonian.expectation(state)
-
-        if method == "sgd":
-            loss = lambda p, c, h, s: _loss(self.hamiltonian.backend.cast(p), c, h, s)
+            # elif loss == "cvar":
+            #    return hamiltonian.cvar(state)
+            # elif loss == "gibbs":
+            #    return hamiltonian.gibbs(state)
```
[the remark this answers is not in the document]
shangtai: I have removed these, initially I intend to only allow several methods.

src/qibo/models/variational.py
```diff
@@ -336,6 +336,20 @@ class QAOA(object):
 
     from qibo import hamiltonians, optimizers
 
+    def _cvar_loss(params, qaoa, hamiltonian, state):
+        if state is not None:
+            state = hamiltonian.backend.cast(state, copy=True)
+        qaoa.set_parameters(params)
+        state = qaoa(state)
+        return hamiltonian.expectation(state)
```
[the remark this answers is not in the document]
shangtai: thanks for spotting the mistake. :) I will use self as the first argument.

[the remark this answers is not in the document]
shangtai: As of now, my main concern is that cvar and gibbs would involve measurement of the quantum state. Hence it can be backend-dependent.  Any idea how can we bypass this challenge?

shangtai:
Hi, I am trying to pass loss function to my QAOA. 

I coded a dummy function,  

```
def _dummy_zero(params, qaoa, hamiltonian, state):
    return 0
```

and then I call the minimization as follows:

from qibo import hamiltonians

```
h = hamiltonians.XXZ(3)
qaoa = QAOA(h)
initial_p = [0.05, 0.06, 0.07, 0.08]
best, params, _ = qaoa.minimize(initial_p, method="BFGS", loss=_dummy_zero)
```

I get the following message:

```
    loss(p, c, h, s)
  [Previous line repeated 957 more times]
RecursionError: maximum recursion depth exceeded
```

Can you advise me on how should I change it?

scarrazza:
@shangtai I am afraid you have to debug your code carefully. There are multiple strategies for that, e.g. using pdb.

src/qibo/models/variational.py
```diff
@@ -473,6 +572,7 @@ def minimize(
             method (str): the desired minimization method.
                 See :meth:`qibo.optimizers.optimize` for available optimization
                 methods.
+            mode (str): the desired loss function.
```
scarrazza, line 575: Please include the list of possible options so the user does not have to read the code, e.g. `cvar`, `gibbs`.

src/qibo/models/variational.py
```diff
@@ -301,6 +379,26 @@ def minimize(
...
 
+def _cvar_loss(params, qaoa, hamiltonian, state):
+    if state is not None:
+        state = hamiltonian.backend.cast(state, copy=True)
+    qaoa.set_parameters(params)
+    state = qaoa(state)
+    return cvar(hamiltonian, state)
+
+
+def _dummy_zero(params, qaoa, hamiltonian, state, mode):
```
scarrazza, line 390: Do you really need this function?

shangtai: Thanks. not really, I was debugging and I forgot to remove it.

src/qibo/models/variational.py
```diff
@@ -1,9 +1,87 @@
...
+    avg = 0
+    sum_count = 0
+    # common_metric = float("inf")
+    for bitstring, count in counts.items():
+        obj = convert_bit_to_energy(hamiltonian, bitstring)
+        avg += np.exp(-eta * obj)
+        sum_count += count
+    return -np.log(avg / sum_count)
+
+
```
scarrazza, line 84: Please document all these functions. I think we should move these methods in another place, e.g. as class functions or just in another file/utils.

scarrazza:
@shangtai could you please have a look at coverage?

src/qibo/models/utils.py
```diff
@@ -0,0 +1,91 @@
+# -*- coding: utf-8 -*-
+import numpy as np
+
+from qibo import gates as maingates
+from qibo.gates import gates
+from qibo.models.circuit import Circuit
+
+
+class Loss_Utils(object):
```
stavros11, line 9: @shangtai since you moved the utils methods to a new file, it is simpler to remove this class completely and define them as standalone methods. Then you can import them directly in other modules, for example:
```py
from qibo.models.utils import cvar
```

src/qibo/models/variational.py
```diff
@@ -506,7 +511,12 @@ def _loss(params, qaoa, hamiltonian, state):
                 state = hamiltonian.backend.cast(state, copy=True)
             qaoa.set_parameters(params)
             state = qaoa(state)
-            return hamiltonian.expectation(state)
+            if mode is None:
+                return hamiltonian.expectation(state)
+            elif mode == "cvar":
+                return Loss_Utils.cvar(hamiltonian, state)
```
stavros11, line 517: Then here you can do:
```suggestion
                from qibo.models.utils import cvar
                
                return cvar(hamiltonian, state)
```

stavros11 (changes requested):
@shangtai please have a look at the comments above.

src/qibo/models/utils.py
shangtai: that would make the code cleaner :) thanks.

src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import numpy as np
```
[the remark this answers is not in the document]
shangtai: thanks. :)

src/qibo/models/utils.py
```diff
@@ -0,0 +1,90 @@
...
+
+    Args:
+        probabilities (list): list/array of probabilities
+        values (list): list/array of corresponding values
+        alpha (float): confidence level
+
+    Returns:
+        CVaR
+    """
+    sorted_indices = np.argsort(values)
```
stavros11, line 47: Are you sure that this should be sorted according to values and not probabilities (from high to low)? Because the loop below has a termination rule according to the probability.

src/qibo/models/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    cvar /= total_prob
+    return cvar
+
+
+def cvar(hamiltonian, state, alpha=0.1):
+    """
+    Given the hamiltonian and state, this function estimate the
+    corresponding cvar function
+    """
+    counts = convert_state_to_count(state)
```
stavros11, line 68: Here you are calculating the probabilities via a non-deterministic measurement (like emulating quantum hardware). Is this on purpose? Because you could also extract the probabilities deterministically from the state vector without the need of measurements.

stavros11 (commented):
Indeed, it looks cleaner now, thanks.

In terms of structure this looks okay. I was just looking at the code more carefully and I have some questions/things that don't look correct to me. Could you please have a look at the comments below.

src/qibo/models/utils.py
```diff
@@ -0,0 +1,90 @@
...
+
+def convert_bit_to_energy(hamiltonian, bitstring):
+    """
+    Given a binary string and a hamiltonian, we compute the corresponding energy.
+    """
+    n = len(bitstring)
+    c = Circuit(n)
+    for i in range(n):
+        c.add(gates.X(int(i)))
+    result = c()  # this is an execution result, a quantum state
```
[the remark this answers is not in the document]
shangtai: I made a mistake, thanks.

src/qibo/models/utils.py
shangtai: yes, we sort according to the values.

src/qibo/models/utils.py
shangtai: thanks. let me change the code to make it deterministically.

shangtai: Hi Stavros, it is easy to compute the energy with the corresponding probabilities in Qibo?

stavros11: If I understand correctly there are two parts of the calculation:
1. `convert_state_to_count`: takes a statevector and returns frequencies. Then these frequencies are used to calculate probabilities. You could avoid this step by calculating probabilities directly as `np.abs(state) ** 2`. However these probabilities will be exact (as if nshots -> infinity), while in your calculation there will be some statistical error due to using finite nshots. I am not sure if you need this error in your calculation for some reason.
2. `convert_bit_to_energy`: calculates the energy of a product state: <bitstring|H|bitstring>. By definition this is a diagonal element in the matrix representation of the Hamiltonian in the computational basis. So if you have the matrix, it is probably easier and faster if you just convert the bitstring to decimal and index the matrix diagonal: `hamiltonian.matrix[decimal, decimal]`. If you don't have the matrix (eg. you are using `SymbolicHamiltonian`) what you are doing may be faster.

shangtai: I changed my `convert_state_to_count` code to compute the probability exactly. Having some macos 3.10 installation issues. Currently, I am trying to figure out

stavros11: > Having some macos 3.10 installation issues. Currently, I am trying to figure out

Don't worry about this, it is not related to this PR. There are some issues with dill tests and mac which are associated with new releases of some dependencies and we will fix them in another PR.

shangtai: Is this pull request alright? Anymore changes that I should make?

stavros11: @shangtai I believe so, yes. Could you please fix the conflicts? I believe they are due to some pre-commit update.

shangtai: @stavros11  Thanks. I have fixed the conflicts.

stavros11 (approved):
Should be fine, thanks.

PR #640, opened 2022-09-08
Fix expectation value of symbolic Hamiltonian

### LINKED ISSUE #639: SymbolicHamiltonian linear algebra functions form the full dense matrix
### Summary

In Qibo 0.1.8, for the SymbolicHamiltonian linear algebra functions, is there a need to form the full dense matrix for these functions?

#### More details

I was testing my notebooks after updating my conda environment (Qibo 0.1.7 -> Qibo 0.1.8), and I found that the calculation of expectation values slowed down by quite a lot, accompanied by this warning:

```
[Qibo 0.1.8|WARNING|2022-09-07 11:45:31]: Calculating the dense form of a symbolic Hamiltonian. This operation is memory inefficient.
```

Digging through `hamiltonians.py`, it seems that the `expectation` function inherited from the parent Hamiltonian class calls `self.matrix`, which forms the dense matrix.

This holds for some of the other linear algebra functions in the SymbolicHamiltonian class as well, e.g. the eigenvalues function returns `self.dense.eigenvalues`, which brings up my question above: Why the need to form the dense matrix?

As a workaround, I'm currently following the calculate_expectation_state function exactly, which doesn't form the dense matrix:

```
result = circuit()
ket = result.state()
bra = np.conj(ket)
expectation = np.real(np.sum(bra * (symbolic_ham @ ket)))
```

stavros11:
Fix #639.

@chmwzc if you have time please give a try if this solves your issue.

The following script can be used to compare with master
```py
import numpy as np
from qibo.hamiltonians import XXZ

nqubits = 20
state = np.random.random(2 ** nqubits) + 1j * np.random.random(2 ** nqubits)
state = state / np.sqrt(np.sum(np.abs(state) ** 2))

ham = XXZ(nqubits, dense=False)

expectation = np.real(np.dot(np.conj(state), ham @ state))
print(expectation)
print(ham.expectation(state))
```

### DIFF
--- a/src/qibo/backends/abstract.py
+++ b/src/qibo/backends/abstract.py
@@ -349,13 +349,15 @@ class Backend(abc.ABC):
         raise_error(NotImplementedError)
 
     @abc.abstractmethod
-    def calculate_expectation_state(self, matrix, state, normalize):  # pragma: no cover
+    def calculate_expectation_state(
+        self, hamiltonian, state, normalize
+    ):  # pragma: no cover
         """Calculate expectation value of a state vector given the observable matrix."""
         raise_error(NotImplementedError)
 
     @abc.abstractmethod
     def calculate_expectation_density_matrix(
-        self, matrix, state, normalize
+        self, hamiltonian, state, normalize
     ):  # pragma: no cover
         """Calculate expectation value of a density matrix given the observable matrix."""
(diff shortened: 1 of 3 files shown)


chmwzc:
Tested the script, all OK for me👍

scarrazza (approved):

PR #643, opened 2022-09-19
Add version attribute in backends

andrea-pasquale:
As suggested by @stavros11 in qiboteam/qcvv#66, I am adding an attribute containing the backend version
Given that the qibo version is already stored in the report, I choose to save directly the version of the library used by the backend.


scarrazza (approved):

stavros11 (approved):
Looks good to me, thanks.

PR #647, opened 2022-09-21
Measurement fixes and improvements

stavros11:
This PR implements various improvements in measurements, particularly:

* Eliminates the `circuit.measurement_gate` and adds all measurements (with and without collapse) in the gate queue.
* Makes possible to reuse (add more gates to) qubits that are measured. If the `collapse=True` option was not used when measuring a qubit that is reused then it is enabled automatically, as I am not aware of any other physical way that a qubit can be reused without collapsing it.
* Fixes an issue with `result = c.add(gates.M(0, 1, collapse=True))` which was out of sync with previous versions and the docs (thanks @igres26).
Now the `result` is a `MeasurementResult` object which has `result.samples()`, `result.frequencies()` and also `result.symbols` which is a list of `sympy.Symbols` that can be used to condition later gates on the measurement outcomes.
* The `result` object returned by `c.add` is now available for non-collapse measurements too.
For example
```py
from qibo import gates

c = Circuit(2)
c.add(gates.H(0))
c.add(gates.H(1))
m0 = c.add(gates.M(0))
m1 = c.add(gates.M(1))
r = c()
```
then
* `r` is a `CircuitResult` object that contains the final state vector and all measurements.
* `m0` is a `MeasurementResult` object that contains the measurements of qubit 0.
* `m1` is a `MeasurementResult` object that contains the measurements of qubit 1.

This is complete in terms of developments but I am labeling it as "do not merge" as I would like to make sure it plays well with qibolab before merging.

### DIFF
--- a/doc/source/code-examples/advancedexamples.rst
+++ b/doc/source/code-examples/advancedexamples.rst
@@ -436,18 +436,19 @@ any parametrized gate as follows:
     c = Circuit(2)
     c.add(gates.H(0))
     output = c.add(gates.M(0, collapse=True))
-    c.add(gates.RX(1, theta=np.pi * output / 4))
+    c.add(gates.RX(1, theta=np.pi * output.symbols[0] / 4))
     result = c()
 
 In this case the first qubit will be measured and if 1 is found a pi/4 X-rotation
 will be applied to the second qubit, otherwise no rotation. Qibo allows to
-use ``output`` as a parameter during circuit creation by representing it using
-a ``sympy.Symbol``. The symbol acquires a numerical value later during execution
-when the measurement is performed. As explained above, if a ``nshots > 1`` is
-given during circuit execution the execution is repeated using a loop.
+use ``output`` as a parameter during circuit creation through the use of
+``sympy.Symbol`` objects. These symbols can be accessed through the ``output.symbols``
+list and they acquire a numerical value during execution when the measurement
+is performed. As explained above, if ``nshots > 1`` is given during circuit
+execution the execution is repeated using a loop.
 
 If more than one qubits are used in a ``collapse=True`` measurement gate the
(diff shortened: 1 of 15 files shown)


andrea-pasquale (approved):
Thanks for implementing this @stavros11.
Everything looks good to me. Good job!

wilkensJ:
The unitary() function  of a circuit object does not work anymore if a measurement is added.
```python
circuit = models.Circuit(1)
circuit.add(gates.X(0))
print(circuit.unitary())
# array([[0.+0.j, 1.+0.j],
#        [1.+0.j, 0.+0.j]])
circuit.add(gates.M(0))
print(circuit.unitary())
```
raises
``` txt
  File "<stdin>", line 1, in <module>
  File "/nfs/users/jadwiga.wilkens/qibo/src/qibo/models/circuit.py", line 902, in unitary
    return fgate.asmatrix(backend)
  File "/nfs/users/jadwiga.wilkens/qibo/src/qibo/gates/special.py", line 102, in asmatrix
    return backend.asmatrix_fused(self)
  File "/nfs/users/jadwiga.wilkens/qibo/src/qibo/backends/numpy.py", line 117, in asmatrix_fused
    gmatrix = gate.asmatrix(self)
  File "/nfs/users/jadwiga.wilkens/qibo/src/qibo/gates/measurements.py", line 140, in asmatrix
    raise_error(
  File "/nfs/users/jadwiga.wilkens/qibo/src/qibo/config.py", line 46, in raise_error
    raise exception(message)
NotImplementedError: Measurement gates do not have matrix representation.
```

stavros11:
> The unitary() function of a circuit object does not work anymore if a measurement is added.

Thanks for reporting this. It was fixed and tested in the latest push.

PR #659, opened 2022-10-16
Initial commit of metrics module

renatomello:


scarrazza:
@renatomello thank you for this. Are you planning to merge to main or to another branch (as it is now you are pointing to `qiboteam:metrics`)?

scarrazza:
@stavros11 the workflow is not loading most likely because we need to specify `on: pull_request`, as we did with #475. However, we did not include those changes because `on: [push, pull_request]` was generating 2 set of jobs at the time. Could you please have a look at the github docs and check if we can simply filter on pull_request to master or find a way to execute only once?

renatomello:
> @renatomello thank you for this. Are you planning to merge to main or to another branch (as it is now you are pointing to `qiboteam:metrics`)?

I was planning on merging to `qiboteam:metrics` first and then to `qiboteam:main` once I include other functionalities

andrea-pasquale:
@renatomello to avoid to format the code manually you can install [pre-commit](https://pre-commit.com/#install) as indicated in Qibo [documentation](https://qibo.readthedocs.io/en/stable/developer-guides/contributing.html#tests).

scarrazza:
> I was planning on merging to `qiboteam:metrics` first and then to `qiboteam:main` once I include other functionalities

Ok, them please open a PR for the `qiboteam:metrics` too so the reviewer can check your changes also there.

stavros11:
> I was planning on merging to `qiboteam:metrics` first and then to `qiboteam:main` once I include other functionalities

@renatomello thank you very much for this. It looks good and well documented, you only need to add some tests. Why are you planning to use different branches for this? It is simpler if you point your PR directly to master, either from the branch in the base repository or from your fork. If you are planning to add more functionalities you can either convert this PR to draft until it is ready or we could even review and merge this and you keep adding with a new PR.

> @stavros11 the workflow is not loading most likely because we need to specify `on: pull_request`, as we did with #475. However, we did not include those changes because `on: [push, pull_request]` was generating 2 set of jobs at the time. Could you please have a look at the github docs and check if we can simply filter on pull_request to master or find a way to execute only once?

Filtering on pull_request to master [seems simple](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-filters-to-target-specific-branches-for-pull-request-events), but if we use this together with push, it will trigger the workflow twice for all PRs pointing to master, so I believe the issue remains.

Another approach is to filter on pull_request to labeled PRs and create a fork label that we manually put on all PRs from forks to trigger this workflow. The problem with this is that first it is not fully automatic and second I could not find a way to filter on specific label. So the workflow will be triggered twice for all PRs that have any label, but we can use an `if: ${{ github.event.label.name == 'fork' }}` condition to skip one of the runs.

There should be a better solution but @scarrazza let me know what you think.

src/qibo/metrics/metrics.py
```diff
@@ -0,0 +1,152 @@
...
+            f"State has dims {state.shape} while target has dims {target.shape}."
+        )
+    elif len(state.shape) >= 3:
+        raise TypeError(
+            f"Both objects must have dims either (k,) or (k,l), but have dims {state.shape} and {target.shape}"
+        )
+
+    if len(state.shape) == 1:
+        state = np.outer(state.conj(), state)
+        target = np.outer(target.conj(), target)
```
stavros11, line 34: Not 100% sure but I think it is safer to use `np.conj` here:
```suggestion
        state = np.outer(np.conj(state), state)
        target = np.outer(np.conj(target), target)
```
and in other places.

The reason is that if you are using qibo to get the state, for example from circuit simulation, then depending on the backend the state may not be `np.ndarray`. 

With the current backends the state could also be cupy array (stored on GPU) or `tf.Tensor`. From my experience calling numpy methods on these two objects usually works, but calling it as an object method (eg. `state.conj()`) may not. So I would suggest to stick to the first option, but we can check this when we add tests.

renatomello: Thank you. just made the changes.

PR #660, opened 2022-10-16
Metrics

renatomello:
Checklist:
- [x] Tests are passing.


renatomello:
I was looking into how to add the Diamond Norm to the metrics. Both [Qiskit](https://qiskit.org/documentation/_modules/qiskit/quantum_info/operators/measures.html#diamond_norm) and [QuTip](https://qutip.org/docs/4.0.2/modules/qutip/metrics.html) use the convex optimization [CVXPY](https://www.cvxpy.org/) library to implement it.
I didn't move forward with it because I'm not aware if this is the best way to do it and because it would introduce an extra dependency to Qibo and I don't know how's the standard way of doing that.

Glad to hear opinions about it.

renatomello:
@stavros11 could you help me here? the tests are failing only for macOS and not even in the tests I created. Do you know what it could be?

src/qibo/quantum_info/utils.py
```diff
@@ -0,0 +1,124 @@
...
+    entropy = -np.sum(probability_array * log_prob)
+
+    # absolute value if entropy == 0.0 to avoid returning -0.0
+    entropy = np.abs(entropy) if entropy == 0.0 else entropy
+
+    return entropy
+
+
+def hellinger_distance(prob_dist_p, prob_dist_q, validate: bool = False):
+    """Calculate the Hellinger ditance :math:`H(p, q)` between
```
stavros11, line 85:
```suggestion
    """Calculate the Hellinger distance :math:`H(p, q)` between
```

src/qibo/quantum_info/utils.py
```diff
@@ -0,0 +1,124 @@
...
+    two discrete probability distributions, :math:`p` and :math:`q`.
+
+    Args:
+        prob_dist_p: (discrete) probability distribution :math:`p`.
+        prob_dist_q: (discrete) probability distribution :math:`q`.
+        validate (bool): if True, checks if :math:`p` and :math:`q` are proper
+            probability distributions. Default: False.
+
+    Returns:
+        Hellinger ditance :math:`H(p, q)`.
```
stavros11, line 95:
```suggestion
        Hellinger distance :math:`H(p, q)`.
```

src/qibo/quantum_info/metrics.py
```diff
@@ -0,0 +1,216 @@
...
+    """
+
+    if (
+        (len(state.shape) >= 3)
+        or (len(state) == 0)
+        or (len(state.shape) == 2 and state.shape[0] != state.shape[1])
+    ):
+        raise TypeError(
+            f"Object must have dims either (k,) or (k,k), but have dims {state.shape}."
+        )
```
stavros11, line 26: In qibo we have a custom way of raising errors which is handled by the logger, so that the version and datetime is logged with the error. This can be used as
```py

raise_error(TypeError, "error text")
```
If you decide to use this all errors need to be replaced.

src/qibo/tests/test_quantum_info.py
```diff
@@ -0,0 +1,184 @@
...
+        target = np.random.rand(d**2, d**2)
+        process_fidelity(channel, target, validate=True)
+
+    channel = np.eye(d**2)
+    backend.assert_allclose(process_fidelity(channel), 1.0)
+    backend.assert_allclose(process_fidelity(channel, channel), 1.0)
+
+
+def test_average_fidelity():
+    average_gate_fidelity(np.eye(4))
```
stavros11, line 184: I saw that the `average_gate_fidelity` differs from `process_fidelity` only by a rescaling. How about removing this test and checking both in the test above, where you can also assert both results?

stavros11 (commented):
Thanks a lot for implementing these @renatomello. It looks good to me including documentation and tests. 

My only comment is regarding the backend. In mosts tests you are using the `backend` argument which automatically loops over all the available backends, however both the code and tests use numpy, so only this will be really tested. If you would like to test different backends we need a backend agnostic way of initializing the arrays in the tests, so that we also test for tf.Tensor and cupy array. Depending on how compatible these are with the numpy methods you are using the tests may or may not pass (I think some linalg methods won't be compatible). The main question there is whether you think that a GPU (particularly cupy) would offer any advantage in these calculations. If yes, we could try generalizing these, otherwise it is sufficient to delegate everything to numpy but then it is not necessary to test all backends.

> @stavros11 could you help me here? the tests are failing only for macOS and not even in the tests I created. Do you know what it could be?

I rerun the tests and there is an issue but is not related to macos. Regardless, it does not seem related to what you are doing here so don't worry about it, we will try to debug and fix in a different PR. Tests, including yours, are passing on my machine.

src/qibo/quantum_info/metrics.py
renatomello: Changed it. Thanks.

src/qibo/tests/test_quantum_info.py
renatomello: Done.

src/qibo/tests/test_quantum_info.py
```diff
@@ -0,0 +1,182 @@
...
+    with pytest.raises(ValueError):
+        p = np.asarray([0.5, 0.4999999])
+        shannon_entropy(p)
+
+
+@pytest.mark.parametrize("base", [2, 10, np.e, 5])
+def test_shannon_entropy(backend, base):
+    prob_array = np.asarray([1.0, 0.0])
+    result = shannon_entropy(prob_array, base)
+    backend.assert_allclose(result, 0.0)
```
stavros11, line 33:
```suggestion
    prob_array = backend.cast(np.asarray([1.0, 0.0]))
    result = shannon_entropy(prob_array, base)
    backend.assert_allclose(result, 0.0)
```

stavros11 (commented):
> If yes, we could try generalizing these, otherwise it is sufficient to delegate everything to numpy but then it is not necessary to test all backends.

Thanks for the updates. Following up on this point, an easy way to test if your methods work with different backends is to use `backend.cast(x)` on all the arrays, before calling each method in the tests (see comment below for example). If we then run the tests in a machine with GPU and cupy, we can see if the code will crash.

src/qibo/tests/test_quantum_info.py
renatomello: Added `backend.cast()` to all arrays. Also added `dtype` to them, otherwise `backend.cast()` would default to `complex128` and throw exceptions on TensorFlow.

Even though I do have cupy installed locally, I'm not sure if it was tested with no errors or if it wasn't tested.

stavros11:
Thanks for adding the `backend.cast`. 

> Even though I do have cupy installed locally, I'm not sure if it was tested with no errors or if it wasn't tested.

I tried and I get some failures for cupy, so I guess it is not properly installed in your environment. This may have to do with the GPU driver or CUDA, which is not always trivial to install. If you are interested in playing with this on your laptop (if it has GPU), we could give a try together when I am back. If you are using conda environment, you could also try to [install cupy using conda](https://docs.cupy.dev/en/stable/install.html#installing-cupy-from-conda-forge), which automatically installs the proper cudatoolkit.

Regaring failures, they are due to the following:
1. `np.asarray` used in a few places in `shannon_entropy`,
2. `entropy` fails due to `np.linalg.eig`, so as I was expecting the eigenvalue does not work well with cupy,
3. `trace_distance` fails due to `sqrtm`.

For 1, I see that you are mainly using it to force 0 * log0 = 0. There may be more efficient ways to do this than looping through every element. For example see [here](https://stackoverflow.com/questions/21752989/numpy-efficiently-avoid-0s-when-taking-logmatrix) or just adding a small epsilon constant could work.

2 and 3 may be more complicated, because in order to solve them we would need access to the backend within the methods, and this would change the API. I am not sure if it is worth going to this direction given that most methods work for free with cupy as they are now.

renatomello:
> I tried and I get some failures for cupy, so I guess it is not properly installed in your environment. This may have to do with the GPU driver or CUDA, which is not always trivial to install. If you are interested in playing with this on your laptop (if it has GPU), we could give a try together when I am back. If you are using conda environment, you could also try to [install cupy using conda](https://docs.cupy.dev/en/stable/install.html#installing-cupy-from-conda-forge), which automatically installs the proper cudatoolkit.

I have the lastest version of nvidia-driver, cuda, and cupy. However, still tests on this backed aren't triggered. Will have to do some debugging somehow.

> Regaring failures, they are due to the following:
> 
>     1. `np.asarray` used in a few places in `shannon_entropy`,
> 
>     2. `entropy` fails due to `np.linalg.eig`, so as I was expecting the eigenvalue does not work well with cupy,
> 
>     3. `trace_distance` fails due to `sqrtm`.
> 
> 
> For 1, I see that you are mainly using it to force 0 * log0 = 0. There may be more efficient ways to do this than looping through every element. For example see [here](https://stackoverflow.com/questions/21752989/numpy-efficiently-avoid-0s-when-taking-logmatrix) or just adding a small epsilon constant could work.

I'll change this one.

> 2 and 3 may be more complicated, because in order to solve them we would need access to the backend within the methods, and this would change the API. I am not sure if it is worth going to this direction given that most methods work for free with cupy as they are now.

Yeah, there are no cupy equivalents to `numpy.linalg.eig` and `scipy.linalg.sqrtm`.

stavros11 (commented):
Thank you for the updates, all tests are passing locally for all backends for me too. If you can fix the coverage issues this should be ready.

> EDIT 3: changing numpy.linalg.eig to numpy.linalg.eigh solves incompatibility issues with cupy locally, but only yield the correct results under the assumption that input matrices are Hermitian.

Indeed, I think cupy is still missing the implementation of `eig` for general matrices (cupy/cupy#3255). Are these methods relevant for non-Hermitian matrices? If yes we should consider changing back to numpy's eig and ignoring cupy in these tests.

renatomello:
> Indeed, I think cupy is still missing the implementation of `eig` for general matrices ([cupy/cupy#3255](https://github.com/cupy/cupy/issues/3255)). Are these methods relevant for non-Hermitian matrices? If yes we should consider changing back to numpy's eig and ignoring cupy in these tests.

There's always the possibility the user inputs a matrix that is not necessarily Hermitian. To be more general, we could drop cupy compatibility or add something to validate if the inputs are proper quantum states.

stavros11:
> There's always the possibility the user inputs a matrix that is not necessarily Hermitian. To make the functions more strict, we could drop cupy compatibility or add something to validate if the inputs are proper quantum states.

If non-Hermitian input happens only by mistake from the user then you can just raise an error if the matrix is not Hermitian. 

But if it is actually useful for some application to use these methods on non-Hermitian matrices, we could check the input and if it is Hermitian we use `eigh`, otherwise we fall back to `eig`. Then if someone uses non-Hermitian with cupy they will get an error but at the end of the day this is a cupy missing feature so we should not worry about it. All the other cases will work. 

There is a way to resolve the cupy issue from our side (cast to numpy, use `eig` and cast back to `cupy`) but it is messy and not really useful.

renatomello:
> > There's always the possibility the user inputs a matrix that is not necessarily Hermitian. To make the functions more strict, we could drop cupy compatibility or add something to validate if the inputs are proper quantum states.
> 
> If non-Hermitian input happens only by mistake from the user then you can just raise an error if the matrix is not Hermitian.
> 
> But if it is actually useful for some application to use these methods on non-Hermitian matrices, we could check the input and if it is Hermitian we use `eigh`, otherwise we fall back to `eig`. Then if someone uses non-Hermitian with cupy they will get an error but at the end of the day this is a cupy missing feature so we should not worry about it. All the other cases will work.
> 
> There is a way to resolve the cupy issue from our side (cast to numpy, use `eig` and cast back to `cupy`) but it is messy and not really useful.

Added a `validate` flag that allows the user to choose if the function will check for Hermiticity or not. I left the default option as `validate=False` because 1) it's faster and 2) the functions are intended to be used on Hermitian inputs.

stavros11 (approved):
Great, thanks. If you are not planning to add anything else, I believe this is ready.

renatomello:
I am planning on adding other things, but I think we can do that later.

scarrazza:
@stavros11 could you please have a look at the workflow in master after the merge of this PR, looks like a pre-commit issue.

stavros11:
> @stavros11 could you please have a look at the workflow in master after the merge of this PR, looks like a pre-commit issue.

I am not sure what went wrong, I do not get this issue locally with the latest master. I would wait and see if it disappears when we merge #684 which should happen soon.

liangwang0734:
Sorry to dig this out; I'm curious if `eig`  is in the plan?

renatomello:
> Sorry to dig this out; I'm curious if `eig` is in the plan?

In the end, this whole module ended up written to be `numpy` specific. However, that means it also works with the `tensorflow` backend (and probably `numba` backend, but I'm not sure).  You could use `backend.cast()` to cast any return of these functions to the backend you want.

renatomello:
> > Sorry to dig this out; I'm curious if `eig` is in the plan?
> 
> In the end, this whole module ended up written to be `numpy` specific. However, that means it also works with the `tensorflow` backend (and probably `numba` backend, but I'm not sure). You could use `backend.cast()` to cast any return of these functions to the backend you want.

A direct implementation of `eig` on cupy is a long-standing request for the library (e.g. https://github.com/cupy/cupy/issues/3255).

Issue #663, opened 2022-10-17
Gradients using parameter shift


stavros11:
Following my comment in #657, it would be useful to have one (or more) implementations of the parameter shift rule. We had a discussion with @DiegoGM91 and a potential way to do this would be a gradient module which contains methods with the following format:
```py
def parameter_shift(circuit, loss, parameters, *args):
    """Calculates gradients of the loss on the given point in parameter space.
        
        Args:
            circuit (qibo.models.Circuit): Variational circuit model.
            loss (Callable): Function implementing the loss on a given parameters point.
            parameters (np.ndarray): Array of parameter values.
            args: Additional optional arguments for the loss function.

        Returns
            Array with the size of `parameters` holding the gradient with respect to each parameter.
"""
```
This could also be exposed as a circuit method `circuit.gradient(loss, parameters)`, however a module would be better to add different variations of the parameter shift rule.

The basic approach which requires 2 * len(parameters) full evaluations of the circuit with different values for the parameters is certainly needed because it is hardware compatible. For simulation we could explore various shortcuts and optimization, including caching or parallelization of the circuit evaluations (for small circuits).

MatteoRobbiati:
Hi @stavros11 , I like this approach. 
Instead of a loss function I would run the PSR to calculate the derivative of a generic observable with respect to the target parameter. If one then wants to do this for a loss function it will be sufficient to use the PSR appropriately in the algorithm (as I did in the tutorial of the [vqregressor](https://github.com/qiboteam/qibo/tree/master/examples/vqregressor)).

DiegoGM91:
I agree with @MatteoRobbiati about the derivative of Hermitian observables, since this is actually what the parameter shift rule gives you. The user can then apply the chain rule for more general loss functions. I discussed with him today, and he's going to open a PR implementing the basic parameter shift rule in a new module that we might call `gradient`. Regarding the observables, I suggested using qibo `Hamiltonians`.

@stavros11 please let us now what you think.

stavros11:
Yes, I agree with implementing it on an observable instead of the loss. It is more flexible and we could even consider using it in conjuntion with other autograd libraries such as tensorflow that implement the classical chain rule.

Regarding using `Hamiltonian` as the observable, I believe it should work. In the future we could consider implementing a more primitive `Operator` object if it is needed for other uses.

To summarize, I guess the interface will change to:
```py
def parameter_shift(circuit, hamiltonian, paramters, *args)
```

[...]

PR #631, opened 2022-08-04
RXX, RYY and RZZ gates

damarkian:
Added 2-qubit RXX, RYY and RZZ gates for #591 in https://github.com/qiboteam/qibo/commit/a75ab7a2c4e11d4aa56b793b85886cbaccce0020

PR #667, opened 2022-10-21
First implementation of the parameter shift rule

MatteoRobbiati:
In this PR I present the first implementation of the parameter shift rule, which you can find in `qibo/src/qibo/derivative.py`. The function is called `parameter_shift`.

After an exchange of ideas had with @DiegoGM91  and @stavros11, this is the form of the function I  implemented:

```
def parameter_shift(
    circuit,
    observable,
    parameter_index,
    gate_eigenv,
    initial_state = None
):
```

Where `circuit` is the custom circuit $\mathcal{U}(\vec{\theta})$ which acts on the `initial_state` $| q_i \rangle$ (default `None` and we execute starting from the zeros), providing the final state $|q_f \rangle$ on which the expected value of `observable` is evaluated. 
`observable` is chosen to be a `qibo.hamiltonians.Hamiltonian` object. 
`parameter_index` identifies the specific parameter of the circuit with respect to which the derivative is made. 
Finally, `gate_eigenv` is useful to implement the shift formula: $\mu^{\pm} = \mu \pm s$, where `s` depends on the absolute value of the observable's eigenvalues $\pm r$ as $s = \pi / (4 r)$.

In the `parameter_shift` documentation, I have written a rather long example: my idea is to provide the user with a dummy PSR code along with numerical benchmarking with the gradients' values obtained by calculating the derivatives via `tensorflow.GradientTape()`.

This PR is linked to #663 .


DiegoGM91 (commented):
@MatteoRobbiati thanks for implementing this. It looks good to me. I left some comments/suggestions, please let me know if you agree with them. Let us see what @stavros11 thinks.

The function you implemented here can be called from another function that computes the gradient. 

One thing I'm not sure about is that here you are requiring the absolute value of the eigenvalue of the generator of the gate to be passed to the function. I guess this may be okay because it provides flexibility, and we can automatize the extraction of this eigenvalue for the gates already implemented in qibo somewhere else. But I'm not really sure.

DiegoGM91:
Could you also please link issue #663 to this PR so that the issue is automatically closed when the PR is merged?

src/qibo/derivative.py
```diff
@@ -0,0 +1,125 @@
...
+    # saving original parameters and making a copy
+    original = np.asarray(circuit.get_parameters()).copy()
+    shifted = original.copy()
+
+    # forward shift and evaluation
+    shifted[parameter_index] += s
+    circuit.set_parameters(shifted)
+
+    if initial_state is None:
+        forward = observable.expectation(circuit.execute().state())
```
stavros11, line 105: In the core qibo code I avoid using the `GlobalBackend` because it makes testing harder and also global objects are generally dangerous. Here `circuit.execute()` will fall back to the `GlobalBackend` so if someone defines the `observable` Hamiltonian with a different backend we may have issues, particularly if cupy is mixed with numpy.

Since `observable` is a `Hamiltonian` object, it already has an underlying backend, so the easiest would be to use this backend for circuit execution. This can be done with:
```suggestion
        backend = observable.backend
        forward = observable.expectation(backend.execute_circuit(circuit).state())
```
or you may prefer to define the `backend` variable in the beginning of the function since it will be reused multiple times.

src/qibo/derivative.py
```diff
@@ -0,0 +1,125 @@
...
+    shifted[parameter_index] -= 2 * s
+    circuit.set_parameters(shifted)
+
+    if initial_state is None:
+        backward = observable.expectation(circuit.execute().state())
+    else:
+        backward = observable.expectation(
+            circuit.execute(initial_state=initial_state).state()
+        )
+
```
stavros11, line 121: Diego is right, this `if`/`else` is not needed. You can pass the `initial_state` in the circuit execution even if it is `None`. If `None` the circuit will automatically use |00...0> as you want:
```suggestion
    backward = observable.expectation(backend.execute_circuit(circuit, initial_state=initial_state).state())
```
Same for the `forward` propagation above.

src/qibo/derivative.py
```diff
@@ -0,0 +1,125 @@
...
+    Original references:
+        `https://arxiv.org/abs/1811.11184`;
+        `https://arxiv.org/abs/1803.00745`.
+
+    Args:
+        circuit (:class:`qibo.models.circuit.Circuit`): custom quantum circuit.
+        observable (:class: `qibo.hamiltonians.Hamiltonian`): the hamiltonian here is the target observable.
+        parameter_index (int): the index which identify the targe parameter in the circuit.get_parameters list
+        gate_eigenv (float): abs(eigenvalue) of H. In case of Pauli 1/2{sigmas} observable r = 0.5
```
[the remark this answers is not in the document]
stavros11: I am not sure if this argument is valid here, but I generally don't like duplication in function arguments. Here there is some duplication because one has to pass the observable and the generator eigenvalues seperately, while both can be calculated if only the generator is given to begin with.

If this algorithm only works for observables of the form exp(-iG), maybe it would be cleaner if we take G as an argument and calculate the observable and eigenvalues from it? Note that G can also be a `Hamiltonian` and one can use `ham.exp()` and `ham.eigenvalues()` for these calculations.

That being said, I understand why the observable may be more natural argument as this is what appears in the loss. @DiegoGM91 and other potential users of this, let us know your opinion. In this case we could even think a way to derive the eigenvalues from the observable directly.

What I am thinking is to make `gate_eigenv` optional, something like:
```py
if gate_eigenv is None:
    gate_eigenv = # calculate the eigenvalues automatically if user did not give them
```

src/qibo/derivative.py
```diff
@@ -0,0 +1,125 @@
...
+        initial_state ((1, 2**nqubits) matrix): initial state on which we act with the circuit.
+
+    Returns:
+        np.float value of the derivative of the expected value of H on the final state obtained applying U
+        to the initial state with respect to the target variational parameter.
+
+    Example:
+        .. testcode::
```
stavros11, line 26: I think this example is quite extensive here. We should certainly write a test using this approach (comparing with tensorflow) and maybe we could also use it in the "Advanced Examples" sections in the docs. For the docstring I would use a more minimal example that only shows how to use qibo's PSR, not the comparison with tensorflow.

src/qibo/derivative.py
```diff
@@ -0,0 +1,125 @@
...
+            print('Test gradient with respect params[0] with tf:  ', tf_grads[0][0].numpy())
+            print('Test gradient with respect params[1] with PSR: ', grad_1.numpy())
+            print('Test gradient with respect params[1] with tf:  ', tf_grads[0][1].numpy())
+
+    """
+
+    if parameter_index > len(circuit.get_parameters()):
+        raise ValueError(
+            """This index doesn't exist! Use one of the circuit's parameters."""
+        )
```
stavros11, line 91: In qibo we have a custom way of raising errors in order to be captured by our logger. This is used as 
```suggestion
        from qibo.config import raise_error
        raise_error(ValueError, "This index doesn't exist! Use one of the circuit's parameters.")
```

stavros11 (commented):
Thanks for implementing this, looks good to me. The next step would be to add tests to make sure that the code works. For example now an `import numpy as np` is missing, which is why pylint is complaining. A good suggestion for test would be to use the code you currently have in the example, which compares the gradient calculation with tensorflow's autograd.

I agree with Diego's comments, a few more below:

src/qibo/derivative.py
DiegoGM91: @stavros11 regarding the previous comment about duplication of arguments (I don't know why GitHub doesn't let me write the reply up there):

There is no such duplication between `observable` and `gate_eigenv`, the eigenvalue refers to the generator of the parametrized gate, which is completely independent of the observable measured at the end of the circuit.

However, it is true that there is some duplication: from the `parameter_index` we could get the type of gate the parameter belongs to, and from this get the eigenvalue of the generator of the gate. I guess this is what we should do, right?

src/qibo/derivative.py
DiegoGM91: Out of curiosity: what is this logger used to?

src/qibo/derivative.py
DiegoGM91: Moreover, the latter would also allow us to raise an `NotImplementedError` if the current parameter shift rule does not apply to the gate in question.

src/qibo/derivative.py
```diff
@@ -0,0 +1,125 @@
...
+
```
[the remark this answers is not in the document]
MatteoRobbiati: Added the `TypeError` and changed the `ValueError` according to your suggestion.

src/qibo/derivative.py
```diff
@@ -0,0 +1,125 @@
+# -*- coding: utf-8 -*-
+def parameter_shift(
+    circuit, observable, parameter_index, gate_eigenv, initial_state=None
+):
```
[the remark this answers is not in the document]
MatteoRobbiati: Done, i agree, it is more clear.

src/qibo/derivative.py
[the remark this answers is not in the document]
MatteoRobbiati: I removed all the `if/else`, according to @stavros11 suggestion in a comment below.

src/qibo/derivative.py
```diff
@@ -0,0 +1,125 @@
...
+    """In this method the parameter shift rule (PSR) is implemented.
+    Given a circuit U and an observable H, the PSR allows to calculate the derivative
+    of the expected value of H on the final state obtained by application of U with
+    respect to a variational parameter of the circuit.
+
```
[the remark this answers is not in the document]
MatteoRobbiati: Done, thank you!

src/qibo/derivative.py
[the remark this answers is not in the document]
MatteoRobbiati: This specific implementation is valid only for circuits in which the variational parameter appears in a gate whose generator has at most two eigenvalues. With this in mind, perhaps it would converge:

1. add a `raise_error` that rejects ports that do not meet these prerequisites;
2. make a distinction between generic gates and rotations. I say this because a large proportion of the variational quantum algorithms use rotations, and can be useful to have a smart way for involving gates like them (without calculating the eigenvals every time).

What do you think @DiegoGM91 and @stavros11 ?

src/qibo/derivative.py
```diff
@@ -0,0 +1,125 @@
...
+    Original references:
+        `https://arxiv.org/abs/1811.11184`;
+        `https://arxiv.org/abs/1803.00745`.
+
+    Args:
+        circuit (:class:`qibo.models.circuit.Circuit`): custom quantum circuit.
+        observable (:class: `qibo.hamiltonians.Hamiltonian`): the hamiltonian here is the target observable.
+        parameter_index (int): the index which identify the targe parameter in the circuit.get_parameters list
```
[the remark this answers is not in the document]
MatteoRobbiati: Done, thank you!

src/qibo/derivative.py
MatteoRobbiati: Before doing this, I want to be sure to have understood. 
How do I import the generic backend object?

src/qibo/derivative.py
MatteoRobbiati: Done, thank you!

src/qibo/derivative.py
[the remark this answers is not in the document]
DiegoGM91: I agree with both points. 

i) For the time being (before we implement generalized parameter shift rules valid for all our gates), we can raise an error. As I commented below, we could remove the `gate_eigenv` argument and use the `parameter_index` to get the type of gate from the queue of gates of the `circuit`. Then, depending on the type of gate, we raise the error.

ii) We already know what the eigenvalues of Pauli matrices are, so I see no point in calculating them every single time.


By the way, point ii) applies not just to single-qubit Paulis, but also to any tensor product of Paulis; this made me think that perhaps we could implement a gate that is the exponential of a tensor product of Paulis times a variational parameter, I think it would be useful. These gates are used, for instance in the Hamiltonian Variational Ansatz. Currently, to implement such a unitary we can use a `hamiltonian` and then the `StateEvolution` model. But what if I want to easily combine such exponential with other built-in gates in `qibo`? Another advantage of having these gates would be that the parameter-shift rule implemented in this PR would apply to them as well.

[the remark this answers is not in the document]
stavros11: Thank you for the replies and @DiegoGM91 for explaining that the eigenvalues are related to the circuit, not the observable. 

I also agree with raising an error, we should certainly do that for all circuits not supported by the current implementation. If we implement more general rules in the future we will replace the error by delegating to the appropriate rule.

> ii) We already know what the eigenvalues of Pauli matrices are, so I see no point in calculating them every single time.
s well.

What I wrote above is one way (maybe not the best) of implementing this, as one can pass the eigenvalues when calling the function. If they do not pass anything, `generator_eigenval` will be `None` and qibo will calculate it by diagonalizing the matrix. Alternatively we could somehow hardcode the eigenvalues for Paulis in the code but I am not sure what is an elegant way to do this.

> By the way, point ii) applies not just to single-qubit Paulis, but also to any tensor product of Paulis; this made me think that perhaps we could implement a gate that is the exponential of a tensor product of Paulis times a variational parameter, I think it would be useful. 

I haven't done the math to ensure they are exactly the same but someone implemented such gates in #631 (already merged).

src/qibo/derivative.py
stavros11: In practice it provides a uniform format for errors and standard output, which looks like `[Qibo 0.1.9.dev0|ERROR|2022-10-24 12:54:00]`. This may help when reporting issues as you can see the date and version. There may be some additional usage that I am not aware of.

The implementation of this format is done [here](https://github.com/qiboteam/qibo/blob/8b0e0d49cbab6eae7d550ae55081af9c0c0f448e/src/qibo/config.py#L82).

src/qibo/derivative.py
stavros11: After the last large refactoring, there are two ways to execute circuits: one is to explicitly define a backend object and use it to execute a circuit:
```py
from qibolab.backends import NumpyBackend

circuit = Circuit(...)
backend = NumpyBackend() # can be replaced by TensorflowBackend
# or NumbaBackend() / CupyBackend() if you have qibojit installed
result = backend.execute_circuit(circuit)
```
With this approach is clear what backend you are using. The other approach is to just execute the circuit directly:
```py
circuit = Circuit(...)
result = circuit()
```
This will [automatically instantiate](https://github.com/qiboteam/qibo/blob/8b0e0d49cbab6eae7d550ae55081af9c0c0f448e/src/qibo/models/circuit.py#L994)  a `GlobalBackend` object which is a reference to one of the available backends (numpy, tensorflow, etc.). `qibo.set_backend` controls to which actual backend `GlobalBackend` is refering to. If you do not use `set_backend`, qibo will automatically fall back to the first available backend from the following list: cupy, numba, tensorflow, numpy (the first available in your environment).

The idea is that in the core code and tests we use the first approach because it is more explicit, know exactly which backend you are using. We just provide the `result = circuit()` approach to simplify circuit execution for users who do not wish to go through the backend details and also keep compatibility with older qibo versions where backends were handled differently. 

Note that the `GlobalBackend` is used internally to make this work, the user is not supposed to import it or use it directly. If they wish to get exposed to the different backends, they should import them directly:
```py
from qibo.backends import NumpyBackend, TensorflowBackend
from qibojit.backends import NumbaBackend, CupyBackend
```

In the current PSR code, you don't need to worry about this because the backend is specified (either to the global one or something else) when defining the hamiltonian and you can access it via `hamiltonian.backend`. This is because `Hamiltonian` is backend dependent (requires a backend in order to be instantiated), unlike `Circuit` which is backend agnostic (the same circuit object can be executed with different backends).

MatteoRobbiati: Thank you very much @stavros11 for the explanation. I've modified the code as you suggested. Do you think it is okay now?

src/qibo/derivative.py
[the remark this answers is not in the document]
MatteoRobbiati: As for (i), I agree with the solution that @stavros11  proposes for now. If the eigenvalues are explicitly given better, otherwise we can calculate them. Another course of action might be to have a dictionary that saves some notable generator-eigenvalue combinations. 

With your (ii) comment you mean a sequence of rotations in which the same parameter can be involved several times? e.g. `RY(theta) -> RZ -> RY -> RZ(theta)`  @DiegoGM91

src/qibo/derivative.py
DiegoGM91: I see, thanks!

src/qibo/derivative.py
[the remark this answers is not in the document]
DiegoGM91: > I haven't done the math to ensure they are exactly the same but someone implemented such gates in https://github.com/qiboteam/qibo/pull/631 (already merged).

I see. This is certainly useful. I think this should be included in the documentation, I couldn't find any reference to them in the latest version of the documentation. Perhaps we could generalize these gates to general  imaginary exponentials of Pauli operators.

Also, it is easy to check that the eigenvalues of any tensor product of Pauli operators are always $\pm 1$, so the current parameter-shift rule can be applied to (imaginary) exponentials of them.


> With your (ii) comment you mean a sequence of rotations in which the same parameter can be involved several times? e.g. `RY(theta) -> RZ -> RY -> RZ(theta)` @DiegoGM91

Nope, I mean $e^{-i\theta P}$, where $P$ is a tensor product of Pauli gates, e.g. $P=\sigma^Z_1 \otimes \sigma^X_2 \otimes \sigma^Z_3$ where the subscripts indicate the qubit acted upon by each single-qubit Pauli. Note that in this example, this exponential would be a three-qubit gate.

src/qibo/derivative.py
stavros11: Yes, it is okay now, thanks.

DiegoGM91:
Following our discussion yesterday, I have computed the generators of controlled Pauli rotations. They are of the form $\frac{(\sigma^Z-I)}{2}\otimes \frac{\sigma^\mu}{2}$, where $\mu\in\{X,Y,Z\}$ for each controlled Pauli rotation respectively. Hence, now we have an extra eigenvalue (equal to zero), in comparison to the single-qubit Pauli rotation. This means that the basic parameter-shift rule does not apply in this case. However, I have checked the known results and it is not difficult to obtain the corresponding parameter-shift rule, although it requires more circuit evaluations.

MatteoRobbiati:
### Changes in the last push

I have just implemented some new features in `qibo`'s `gates` and `circuit`.

- `circuit.associate_gates_with_parameters()` is a method which returns a $N_{params}$-long `flatlist`. The idea is that the $i$-th element of this list is the gate affected by the $i$-th parameter in the parameters' `flatlist` obtained through `circuit.get_parameters(format='flatlist')`;
- `gate.generator_eigenvalue()` is a method which raises a `NotImplementedError` for each gate different from the rotations and returns `0.5` if the gate is a rotation. Now we can freely implement the customized returns for other types of gate.


#### How to use the PSR now
After these recent modifications the derivative strategy is the following:

1. the `parameter_shift(circuit, hamiltonian, parameter_index)` function is called;
2. an object of type `gate` is obtained by calling the new `associate_gates_with_parameters` method explained above;
3. the generator eigenvalue is obtained using `gate.generator_eigenvalue()`.

Point 2 and 3 can be implemented as follows:

```python
    # getting the gate's type
    gate = circuit.associate_gates_with_parameters()[parameter_index]

    # getting the generator_eigenvalue
    generator_eigenval = gate.generator_eigenvalue()
```

What do you think about it @DiegoGM91 and @stavros11?

### Some issues
- This strategy is more general, but I suggest to recall the `parameter_shift` function with a new name which clarify that this one can be used only for gates whose generator presents at most two eigenvalue (alternatively we could differentiate the various cases within a generic `parameter_shift` function).
- It would be great to cache the `gates_associated_with_parameters` instead of calculate them every time (and also the parameters maybe?). How this can be done?

src/qibo/derivative.py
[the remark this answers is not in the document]
MatteoRobbiati: Do you think we can  open a new issue/PR combo for this further implementation and merge this PR before of that?
I purpose this in order to make it lighter. 
In this case, we can resolve this conversation and link it to the new issue/PR.

stavros11 (commented):
Thanks for the updates. I believe this is complete in terms of code, but tests should be added.

src/qibo/derivative.py
```diff
@@ -0,0 +1,117 @@
...
+    Args:
+        circuit (:class:`qibo.models.circuit.Circuit`): custom quantum circuit
+        hamiltonian (:class: `qibo.hamiltonians.Hamiltonian`): target observable
+        parameter_index (int): the index which identifies the target parameter in the circuit.get_parameters() list
+        initial_state ((1, 2**nqubits) matrix): initial state on which we act with the circuit.
+
+    Returns:
+        np.float value of the derivative of the expected value of H on the final state obtained applying U
+        to the initial state with respect to the target variational parameter.
+
```
DiegoGM91, line 31: Suggestion: I think "derivative of the expectation value of the `hamiltonian` with respect to the target variational parameter." should be enough

src/qibo/gates/abstract.py
```diff
@@ -258,6 +258,20 @@ def decompose(self, *free) -> List["Gate"]:
     def asmatrix(self, backend):
         return backend.asmatrix(self)
 
+    def generator_eigenvalue(self):
+        """
+        This function returns the absolute value of the eigenvalues of the
+        operator which generates the unitary target one.
+
```
DiegoGM91, line 265: Suggestion: "eigenvalues of the gate's generator."

src/qibo/gates/abstract.py
```diff
@@ -258,6 +258,20 @@ def decompose(self, *free) -> List["Gate"]:
...
+        Returns:
+            generator's eigenvalue value or raise an error if not implemented.
```
DiegoGM91, line 267: Suggestion: "np.float generator's eigenvalue or raise..."

src/qibo/models/circuit.py
```diff
@@ -828,6 +828,21 @@ def traverse(x):
             )
         return params
 
+    def associate_gates_with_parameters(self):
+        """Associates to each parameter its gate.
+
+        Returns:
+            A nparams-long flatlist whose i-th element is the gate affected by the
+            i-th parameter.
+        """
```
DiegoGM91, line 837: Suggestion: "affected -> parametrized"

src/qibo/derivative.py
```diff
@@ -0,0 +1,117 @@
...
+            test_hamiltonian = hamiltonian()
+
+            # running the psr with respect to the two parameters
+            grad_0 = parameter_shift(circuit = c, hamiltonian = test_hamiltonian, parameter_index = 0)
+            grad_1 = parameter_shift(circuit = c, hamiltonian = test_hamiltonian, parameter_index = 1)
+
+            print('Test gradient with respect params[0]: ', grad_0.numpy())
+            print('Test gradient with respect params[0]: ', grad_1.numpy())
+    """
```
DiegoGM91, line 70: Question: I haven't tested this example, but will `.numpy()` work with all backends?

DiegoGM91 (commented):
Thanks @MatteoRobbiati for implementing this. It looks good to me, although I didn't test it yet.

I like the solution adopted regarding the eigenvalues of the generators.

I agree with both of you that it would be good to add some sort of memoization.

I left some comments, mostly regarding the docstrings.

src/qibo/derivative.py
DiegoGM91: Also I think it should be "test_params". In particular, "test_params[1]" in the second line.

MatteoRobbiati: You are right, thank you. I put it in the first version of the example (when I used the tensorflow backend) but now it is not needed.

MatteoRobbiati:
I have written some tests @stavros11, since this is my first time writing tests, can you tell me if they are done properly?
I also added @DiegoGM91 suggestions to the docstrings.

Let me know if i must to some modification before merging.

src/qibo/tests/test_derivative.py
```diff
@@ -0,0 +1,51 @@
...
+        grad_0 = parameter_shift(
+            circuit=c, hamiltonian=test_hamiltonian, parameter_index=5
+        )
+
+    # testing hamiltonian type
+    with pytest.raises(TypeError):
+        grad_0 = parameter_shift(circuit=c, hamiltonian=c, parameter_shift=0)
+
+    grad_0 = parameter_shift(circuit=c, hamiltonian=test_hamiltonian, parameter_index=0)
+    grad_1 = parameter_shift(circuit=c, hamiltonian=test_hamiltonian, parameter_index=1)
```
stavros11, line 49: The structure of the tests looks fine, however ideally we would like to have an assertion that the calculated values are the expected ones. You can either hardcode the correct results and do something like
```py
assert grad_0 == 0.2
```
where 0.2 is whatever the actual result should be, or you can calculate the gradient using a different way (eg. tensorflow backpropagation) and check that results agree.

src/qibo/tests/test_derivative.py
```diff
@@ -0,0 +1,51 @@
...
+    c = Circuit(nqubits=1)
+    # all gates for which generator eigenvalue is implemented
+    c.add(gates.RX(q=0, theta=0))
+    c.add(gates.RY(q=0, theta=0))
+    c.add(gates.RZ(q=0, theta=0))
+    c.add(gates.M(0))
+    return c
+
+
+def test_derivative():
```
stavros11, line 27: In order to test all different backends, you can use the backend fixture here
```suggestion
def test_derivative(backend):
```
You don't need to define it, just put it as an argument. Its values are automatically generated from the existing conftest.py.

src/qibo/tests/test_derivative.py
```diff
@@ -0,0 +1,51 @@
...
+
+    # initializing the circuit
+    c = circuit(nqubits=1)
+
+    # some parameters
+    test_params = np.random.randn(3)
+    c.set_parameters(test_params)
+
+    test_hamiltonian = hamiltonian()
```
stavros11, line 36: and then use the `backend` when defining the hamiltonian here
```suggestion
    test_hamiltonian = hamiltonian(backend)
```

src/qibo/tests/test_derivative.py
```diff
@@ -0,0 +1,51 @@
...
+from qibo import gates, hamiltonians
+from qibo.derivative import parameter_shift
+from qibo.models import Circuit
+
+
+# defining an observable
+def hamiltonian(nqubits=1):
+    m0 = (1 / nqubits) * hamiltonians.Z(nqubits).matrix
+    ham = hamiltonians.Hamiltonian(nqubits, m0)
+    return ham
```
stavros11, line 13: This also needs to be adjusted for the backend
```suggestion
def hamiltonian(nqubits, backend):
    m0 = (1 / nqubits) * hamiltonians.Z(nqubits, backend=backend).matrix
    ham = hamiltonians.Hamiltonian(nqubits, m0, backend=backend)
    return ham
```
Also why you do not return
```py
return (1 / nqubits) * hamiltonians.Z(nqubits, backend=backend)
```
directly? I believe it will have exactly the same effect.

MatteoRobbiati:
Thank you for your suggestions @stavros11.
I have done what you purposed. I needed to use `round(grad_0)` because of the different precision of the PSR and `tf.GradientTape`: do you agree with this?

src/qibo/tests/test_derivative.py
```diff
@@ -0,0 +1,57 @@
...
+    grad_0 = parameter_shift(circuit=c, hamiltonian=test_hamiltonian, parameter_index=0)
+    grad_1 = parameter_shift(circuit=c, hamiltonian=test_hamiltonian, parameter_index=1)
+    grad_2 = parameter_shift(circuit=c, hamiltonian=test_hamiltonian, parameter_index=2)
+
+    assert isinstance(test_hamiltonian, hamiltonians.AbstractHamiltonian)
+    # check of known values
+    # calculated using tf.GradientTape
+    assert round(grad_0, 10) == 8.51104358e-02
+    assert round(grad_1, 8) == 5.20075970e-01
+    assert round(grad_2, 10) == 0.0000000000
```
stavros11, line 57: I needed to use `round(grad_0)` because of the different precision of the PSR and `tf.GradientTape`: do you agree with this?

For precision issues you can also use the atol/rtol arguments of [`np.testing.assert_allclose`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html), for example in our case:

```suggestion
    backend.assert_allclose(grad_0, 8.51104358e-02, atol=1e-10)
    backend.assert_allclose(grad_1, 5.20075970e-01, atol=1e-10)
    backend.assert_allclose(grad_2, 0, atol=1e-10)
```
It may be a bit easier to see the tolerance this way.

Also, can you merge the latest master to this branch? The failure in CI is due to a bug we fixed long time ago so it should be fixed if you merge master here.

MatteoRobbiati: Yes, right. I have done it now.

src/qibo/derivative.py
```diff
@@ -0,0 +1,117 @@
...
+    Original references:
+        `https://arxiv.org/abs/1811.11184`;
+        `https://arxiv.org/abs/1803.00745`.
+
+    Args:
+        circuit (:class:`qibo.models.circuit.Circuit`): custom quantum circuit.
+        hamiltonian (:class: `qibo.hamiltonians.Hamiltonian`): target observable.
+        parameter_index (int): the index which identifies the target parameter in the circuit.get_parameters() list
+        initial_state ((1, 2**nqubits) matrix): initial state on which the circuit acts.
+
```
DiegoGM91, line 27: In principle the `initial_state` needs not be a `(1,2**nqubits)` matrix, it can simply be a `(2**nqubits)` vector, right?

DiegoGM91 (commented):
Looks good to me. 

I just left a couple of very small comments.

Also, a final comment is that now it would be extremely simple to add the generator eigenvalue to the `RXX, RYY, RZZ` gates, right? The eigenvalue is the same as for the `RX, RY, RZ` rotations. I would add this and include these gates in the tests. I can also open a new PR after this one is merged, if you prefer.

src/qibo/tests/test_derivative.py
```diff
@@ -52,6 +52,6 @@ def test_derivative(backend):
     assert isinstance(test_hamiltonian, hamiltonians.AbstractHamiltonian)
     # check of known values
```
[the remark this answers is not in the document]
stavros11: It is not really needed.

src/qibo/derivative.py
MatteoRobbiati: Right, it works also with a vector. I have just changed the docstrings.

src/qibo/tests/test_derivative.py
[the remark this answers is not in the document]
MatteoRobbiati: Thank you very much @DiegoGM91 for your review. 
I proceed with these modifications.

I purpose to open a new PR, because this one is starting to get pretty big.
I have already opened a second PR related to the RDP: the #735. 
I'm going to add also you as reviewer.

DiegoGM91 (approved):

stavros11 (approved):

PR #672, opened 2022-10-25
adding realistic_noise_model module

LucaZilli:
The function takes a generic circuit and the noise channels' parameters and returns the noisy circuit padded with the quantum channels.

Co-Authored-By @AlejandroSopena 

Checklist:
- [ ] Tests are passing.


src/qibo/noise_model.py
```diff
@@ -0,0 +1,136 @@
+# -*- coding: utf-8 -*-
+from qibo import gates, models
+from qibo.noise import DepolarizingError, NoiseModel, ThermalRelaxationError
+
+
+def noise_model(Circuit, params):
```
scarrazza, line 6: I don't understand why you are passing as first argument the `Circuit` class definition instead of a simple class instance.

src/qibo/noise_model.py
```diff
@@ -0,0 +1,136 @@
...
+        is a tuple which contains the parameter of the :class:`qibo.gates.DepolarizingChannel` for 1-qubit gate and 2-qubit gate, respectively.
+        The third element is a dictionary "error map" for the bit flips.
+        nshots (int): number of final samples.
+
+    Returns:
+        Circuit (qibo.models.Circuit) padded with noise channels.
+
+
+    """
+    Noisy_Circuit = Circuit.__class__(**Circuit.init_kwargs)
```
scarrazza, line 30: No need for that if you pass a circuit instance. If needed you can always make a copy of the input circuit.

src/qibo/noise_model.py
```diff
@@ -0,0 +1,136 @@
...
+
+    time_steps = max(Circuit.queue.moment_index)
+    time = time_steps * [
```
scarrazza, line 33: `np.zeros(time_steps)`?

src/qibo/noise_model.py
```diff
@@ -0,0 +1,136 @@
...
+    The function applies a :class:`qibo.gates.ThermalRelaxationChannel` after each step of the circuit
+    and, after each gate, a :class:`qibo.gates.DepolarizingChannel`, whose parameter depends on whether the
+    gate applies on one or two qubits. In the end on the samples are applied bitflips errors using
+    the :class:`qibo.gates.PauliNoiseChannel`.
+
+
+    Args:
+        Circuit (qibo.models.Circuit): Circuit on which noise will be applied. Since in the end are
+        applied bitflips, measurement gates are required.
+        Params (tuple): object which contains the parameters of the channels organized as follow
```
scarrazza, line 18: Please update the parameter name, no capital letters.

src/qibo/noise_model.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+    """
+    Noisy_Circuit = Circuit.__class__(**Circuit.init_kwargs)
+
+    time_steps = max(Circuit.queue.moment_index)
+    time = time_steps * [
+        0,
+    ]
+    current_time = Circuit.nqubits * [
```
scarrazza, line 36: `np.zeros(nqubits)`?

src/qibo/noise_model.py
scarrazza, line 30: Usually local variables are written in low letters: `noisy_circuit`.

src/qibo/noise_model.py
```diff
@@ -0,0 +1,136 @@
...
+        0,
+    ]
+    t1 = params[0][1][0]
```
scarrazza, line 39: I find this mechanism quite complex to understand. You might prefer to use a dictionary or multiple parameters to avoid 2 or 4 parameters.

scarrazza (changes requested):
@LucaZilli @AlejandroSopena thank you for this implementation. Please remove the .DS_Store files from this PR.

scarrazza:
@AlejandroSopena and @LucaZilli may I ask you to write down here, the next steps you are planning to perform, given that now we have operative qpus? Thanks.

AlejandroSopena:
Now we need to focus on testing the model with different circuits. I think it is a good idea to consider circuits producing highly entangled states as they are more prone to noise. For example, W-like states and absolutely maximally entangled states. These states can be prepared following the algorithms proposed in [2202.04673](https://arxiv.org/abs/2202.04673) and [1904.07955](https://arxiv.org/abs/1904.07955), respectively.
Since we do not have yet two-qubit gates available, we can get data from ibm computers for testing.
The test consists of verifying that our model is able to prepare a state that minimises the hellinger fidelity with the hardware state with parameters close to those obtained by characterisation. We expect these parameters to depend little on the circuit under consideration. However, the hellinger fidelity $f$ we typically achieve is a little less than 1 due to shot noise $\epsilon_f$, i.e. $1-f<=\epsilon_f$. This is one of the reasons for the variations in the parameters. Another possible reason is that we are only measuring in the Z-basis. If we wanted to capture the full effect of the noise by calculating the complete state, we would have to do tomography.

src/qibo/noise_model.py
```diff
@@ -0,0 +1,398 @@
...
+        return hellinger_fid
+
+
+class CompositeNoiseModel:
+    """Class associated with a realistic representation of a noisy circuit modeled by the :func:`qibo.noise_model.noisy_circuit`.
+    This class is able to fit the parameters of the noise model to reproduce an experimental realization of the circuit
+    you want to simulate.
+
+    Args:
+        noisy_circuit (qibo.models.Circuit): the noisy circuit. See :func:`qibo.noise_model.noisy_circuit`.
```
MatteoRobbiati, line 292:
```suggestion
        noisy_circuit (:class:`qibo.models.circuit.Circuit`): the noisy circuit. See :func:`qibo.noise_model.noisy_circuit`.
```

src/qibo/noise_model.py
```diff
@@ -0,0 +1,398 @@
...
+        self.noisy_circuit = noisy_circuit(circuit, self.params)
+
+    def fit(
+        self,
+        target_result,
+        bounds=True,
+        f_min_rtol=None,
+        backend=None,
+    ):
+        r"""Performes the fitting procedure of the noise model parameters, using the method nlopt.opt from the library nplot. The fitting procedure is implemented to maximize the hellinger fidelity calculated using the :func:`qibo.noise_model.loss` between the probability distribution function estimated by the noise model and the one measured experimentally. Since, we are using probability distribution functions estimated using a finite number of shots, the hellinger fidelity is going to have an error caused by an imperfect estimation of the probabilities. This method takes into account this effect and stops when the fidelity reaches a corrected maximum $1-\epsilon$, with $\epsilon$=:func:`qibo.noise_model.hellinger_shot_error`.
```
MatteoRobbiati, line 318:
```suggestion
        r"""Performs the fitting procedure of the noise model parameters, using the method `nlopt.opt` from the library NLopt. The fitting procedure is implemented to maximize the hellinger fidelity calculated using the :func:`qibo.noise_model.loss` between the probability distribution function estimated by the noise model and the one measured experimentally. Since we are using probability distribution functions estimated using a finite number of shots, the hellinger fidelity is going to have an error caused by an imperfect estimation of the probabilities. This method takes into account this effect and stops when the fidelity reaches a corrected maximum $1-\epsilon$, with $\epsilon$=:func:`qibo.noise_model.hellinger_shot_error`.
```

src/qibo/noise_model.py
```diff
@@ -0,0 +1,398 @@
...
+
+        Args:
+            target_result (qibo.states.CircuitResult): the circuit result with frequencies you want to emulate.
+            bounds: If True are given the default bounds for the depolarizing and thermal relaxation channels' parameters.
```
MatteoRobbiati, line 322:
```suggestion
            bounds: if True are given the default bounds for the depolarizing and thermal relaxation channels' parameters.
```

src/qibo/noise_model.py
```diff
@@ -0,0 +1,398 @@
...
+            Otherwise it's possible to pass a matrix of size (2, 4 * nqubits + 4), where bounds[0] and bounds[1]
+            will be respectively the lower and the upper bounds for the parameters. The first 2 * nqubit columns are related
+            to the :math:`T_1` and :math:`T_2` parameters; the subsequent 2 columns are related to the gate time parameters; the other subsequent 2 columns are related depolarizing error parameters; the last 2 * nqubit columns are related to bitflips errors.
```
MatteoRobbiati, line 325:
```suggestion
            to the :math:`T_1` and :math:`T_2` parameters; the subsequent 2 columns are related to the gate time parameters; the other subsequent 2 columns are related to depolarizing error parameters; the last 2 * nqubit columns are related to bitflips errors.
```

MatteoRobbiati (commented):
Thank you for this, it is in a very good shape in my opinion.
Some small suggestions (related to docstrings) follow.

src/qibo/noise_model.py
AlejandroSopena: Thanks @MatteoRobbiati. I have added your suggestions.

scarrazza (approved):
Thank you @AlejandroSopena and @LucaZilli, this is good to merge.

PR #674, opened 2022-10-25
Rename matrices to qmatrices

stavros11:
The latest version of `dill` released yesterday causes some of our [tests to fail](https://github.com/qiboteam/qibo/actions/runs/3318916439/jobs/5483421300). While debugging this issue, I realized that there is a variable name collision in our code: in backends/__init__.py we define a `matrices` variable, while there is also a matrices.py under backends. Therefore I am not sure what happens when one imports qibo.backends.matrices.

Here I renamed the `matrices` variable to `qmatrices`. Maybe it would be cleaner to rename matrices.py, however this is also used in qibojit so we would need to update it and create a version incompatibility. @scarrazza let me know what you think.

Another solution to fix the CI is to pin dill's version to the previous one, but the problem with the variable names may appear again in the future.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/.github/workflows/rules.yml
+++ b/.github/workflows/rules.yml
@@ -11,6 +11,7 @@ jobs:
   build:
     if: contains(github.event.pull_request.labels.*.name, 'run-workflow') || github.event_name == 'push'
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: [3.7, 3.8, 3.9, '3.10']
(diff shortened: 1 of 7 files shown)


scarrazza (commented):
@stavros11 thank you so much. I also think that renaming matrices in qmatrices is a consistent choice, so please go ahead with that choice. Anyway, the next release of qibo and qibojit should be published together.

scarrazza:
Btw, I believe joblib has also updated something that breaks macos.

scarrazza:
I have tested this PR successfully, i.e. no crashes for joblib, on macos M1/M2 and x86_64 with Catalina, Monterey and Ventura.
This issue should be related to github runners, so we could disable parallel tests on macos.

scarrazza:
Closing, implemented in #677.

PR #675, opened 2022-10-26
Expectation values from samples

AlejandroSopena:
In this PR I implemented the computation of expectation values of `Zs` operators from counts. The input is the frequencies `result.frequencies(binary=True)` and a mapping for the qubits indices. For example, suppose we have a 10 qubit circuit but we are measuring 5 of them `[1,3,5,7,9]` and we want to compute an observable that acts on those or less qubits `Z(5)*Z(3)*Z(9)`. Then, `qubit_map=[1,3,5,7,9]`.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/hamiltonians/hamiltonians.py
+++ b/src/qibo/hamiltonians/hamiltonians.py
@@ -140,10 +140,10 @@ class Hamiltonian(AbstractHamiltonian):
             raise_error(NotImplementedError, "Observable is not diagonal.")
         keys=list(freq.keys())
         if qubit_map is None:
-            qubit_map=list(range(len(keys[0])))
+            qubit_map=list(range(int(np.log2(len(Obs)))))
         counts=np.array(list(freq.values()))/sum(freq.values())
         O=0
-        kl=len(keys[0])
+        kl=len(qubit_map)
         for j, k in enumerate(keys):
             index=0
             for i in qubit_map:
@@ -579,7 +579,7 @@ class SymbolicHamiltonian(AbstractHamiltonian):
                 if subk.count(1)%2==1:
                     o_k=-1
                 O_q+=o_k*counts[i]
-            O+=O_q*coeff[j]
+            O+=O_q*float(coeff[j])
         return O
         
     def __add__(self, o):
--- a/src/qibo/hamiltonians/hamiltonians.py
+++ b/src/qibo/hamiltonians/hamiltonians.py
@@ -137,21 +137,21 @@ class Hamiltonian(AbstractHamiltonian):
             )
 
     def expectation_from_samples(self, freq, qubit_map=None):
-        Obs = self.matrix
-        if np.count_nonzero(Obs - np.diag(np.diagonal(Obs))) != 0:
+        obs = self.matrix
+        if np.count_nonzero(obs - np.diag(np.diagonal(obs))) != 0:
             raise_error(NotImplementedError, "Observable is not diagonal.")
         keys = list(freq.keys())
         if qubit_map is None:
-            qubit_map = list(range(int(np.log2(len(Obs)))))
+            qubit_map = list(range(int(np.log2(len(obs)))))
         counts = np.array(list(freq.values())) / sum(freq.values())
-        O = 0
+        expval = 0
         kl = len(qubit_map)
         for j, k in enumerate(keys):
             index = 0
             for i in qubit_map:
                 index += int(k[qubit_map.index(i)]) * 2 ** (kl - 1 - i)
-            O += Obs[index, index] * counts[j]
-        return O
+            expval += obs[index, index] * counts[j]
+        return expval
 
     def eye(self, n=None):
         if n is None:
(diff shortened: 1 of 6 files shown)


src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -134,6 +136,23 @@ def expectation(self, state, normalize=False):
...
+        keys = list(freq.keys())
+        if qubit_map is None:
+            qubit_map = list(range(int(np.log2(len(Obs)))))
+        counts = np.array(list(freq.values())) / sum(freq.values())
+        O = 0
+        kl = len(qubit_map)
+        for j, k in enumerate(keys):
+            index = 0
+            for i in qubit_map:
+                index += int(k[qubit_map.index(i)]) * 2 ** (kl - 1 - i)
```
stavros11, line 152: Here you can avoid the conversion from binary to decimal if `freq` is given in decimal. If these frequencies are result of some circuit execution, the user can do `result.frequencies(binary=False)` to get this without much effort.

Even if you get the frequencies in binary, I believe you can still avoid the for loops if the operation is vectorized. For example something like
```py
ids = tuple(map(lambda x: int(x, 2), freq.keys())) # covert indices to decimal
O = np.diag(Obs)[ids] * counts
```
I would also find a better name for the variable `O` because it does not look very nice in the code. I generally avoid single letter names and capitals in local variables. `O` is even more confusing because it's very close to `0`.

src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -535,6 +554,38 @@ def calculate_dense(self):
...
         return Hamiltonian.expectation(self, state, normalize)
 
+    def expectation_from_samples(self, freq, qubit_map=None):
+        obs = self.terms
+        for term in obs:
+            for factor in term.factors:
+                if isinstance(factor, Z) == False:
+                    raise_error(
+                        NotImplementedError, "Observable is not a Z Pauli string."
+                    )
```
stavros11, line 564: I am not sure if this and other symbol manipulations below will fail for more complicated expressions. For example what happens if you have a constant coefficient, eg. `2 * Z(0) * Z(1)` or powers, eg. `Z(0) ** 3` (not very meaningful in terms of physics, but it may appear somehow)? The easiest way to check is to add a few more tests with different hamiltonian forms. 

I remember with other symbolic manipulations we have found several bugs as people were using the code, so these are a bit hard to control.

stavros11 (commented):
Thanks for implementing this, it's a useful feature. I wrote two comments below, let me know what you think.

src/qibo/hamiltonians/hamiltonians.py
AlejandroSopena: Thanks for the suggestion. The user can include a `qubit_map` that indicates which qubit each position in the counts refers to. For example, let's consider a 6 qubit circuit where we only measure qubits 1, 3 and 5 because we want to get an expected value of an observable that only acts on those qubits. We will then get counts for the states 000, 001, 010 ...111. When converting these numbers to decimal we have to take into account the `qubit_map` because the first position indicates qubit 1, the second position indicates qubit 3 and the third position indicates qubit 5. With this in mind, I do not see a clear way to do it in the way you propose. If you have any suggestion, let me know and I will implement it.

I agree that the variable names are not appropriate so I have made some changes. Let me know what you think.

scarrazza:
@stavros11 please let me know if you agree with the new changes so we can merge.

src/qibo/tests/test_hamiltonians.py
```diff
@@ -246,6 +248,38 @@ def test_hamiltonian_expectation_errors(backend):
...
+    c.add(gates.RX(3, np.random.rand()))
+    c.add(gates.M(0, 1, 2, 3))
+    nshots = 10**5
+    result = c(nshots=nshots)
+    state = c()
+    freq = result.frequencies(binary=True)
+    Obs0 = hamiltonians.Hamiltonian(3, matrix).expectation_from_samples(
+        freq, qubit_map=None
+    )
+    Obs1 = h1.expectation(c().state())
```
stavros11, line 271: Here the `state` variable is not used anywhere. 

Also you don't need to execute the circuit twice to get the state, you can just do `result.state()` directly.

src/qibo/hamiltonians/hamiltonians.py
stavros11: Thank you for the explanation, you are right. I think some part of this calculation can still be vectorized, even with the `qubit_map`, but I am not sure if you gain anything.

If you expect `freq` to be a result of circuit execution, in principle you could instead take the the [`CircuitResult` object](https://github.com/qiboteam/qibo/blob/c99e296693aa3ee5b54de6833bbdfe97f86c193d/src/qibo/states.py#L7), which is what circuit execution returns and contains both the frequencies and the qubit_map. This could be simpler in some cases as the user would not need to specify the qubit map twice, it would be read directly from the measurement gate in the circuit. However, what you are doing here is more flexible because it also works when frequencies do not come from a circuit.

A way to have both approaches is to define an additional method `CircuitResult.expectation_from_samples(hamiltonian)` which calls the method you defined here. So one can do
```py
result = c(nshots=100)
result.expectation_from_samples(hamiltonian)
```
I am not sure if this is useful, let me know what you think.

> I agree that the variable names are not appropriate so I have made some changes. Let me know what you think

Thanks, looks better now.

src/qibo/tests/test_hamiltonians.py
AlejandroSopena: Thanks for spotting this. Corrected.

AlejandroSopena: I think adding an aditional method to `CircuitResult`  is useful so I have implemented it.

stavros11 (approved):
Thanks for the updates. If you are not planning to add anything else, I believe this is ready.

PR #685, opened 2022-11-03
workflow publish.yml

Edoardo-Pedicillo:
***publish.yml*** is the workflow to deploy static content to GitHub Pages
Checklist:
- [ ] Tests are passing.


scarrazza:
If I understand correctly, you are planning to start a workflow that upload the documentation website as an artifact, however this workflow should start another workflow in the website repository which should download the artifact, regenerate the webpage and publish, right?

PR #688, opened 2022-11-09
Fix doc bug from automodule to autofunction

renatomello:
Checklist:
- [ ] Tests are passing.


stavros11 (approved):

stavros11:
@scarrazza we should also merge this before releasing.

PR #690, opened 2022-11-11
From Standard to Furo doc's theme

MatteoRobbiati:
I have modified the `conf.py` in the doc folder selecting the Furo sphinx theme and setting the colors according to the web site.
When we are ready with the site we can merge this branch.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 # Configuration file for the Sphinx documentation builder.
 #
 # This file only contains a selection of the most common options. For a full
@@ -26,7 +25,7 @@ copyright = "2020-2022 by the Qibo team"
 author = "The Qibo team"
 
 # The full version, including alpha/beta/rc tags
-#release = qibo.__version__
+# release = qibo.__version__
 
 
 # -- General configuration ---------------------------------------------------
@@ -111,6 +110,6 @@ def setup(app):
     app.add_css_file("css/style.css")
 
 
-#html_logo = "logo.png"
+# html_logo = "logo.png"
 
 html_show_sourcelink = False
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -20,7 +20,7 @@ import qibo
 
 # -- Project information -----------------------------------------------------
 
-project = "Qibo"
+project = "qibo"
 copyright = "2020-2022 by the Qibo team"
 author = "The Qibo team"
 
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -110,6 +110,6 @@ def setup(app):
     app.add_css_file("css/style.css")
 
 
-# html_logo = "logo.png"
+html_logo = "logo.png"
 
 html_show_sourcelink = False
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 # Configuration file for the Sphinx documentation builder.
 #
 # This file only contains a selection of the most common options. For a full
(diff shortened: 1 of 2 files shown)


scarrazza:
@MatteoRobbiati thank you for this. My only suggestion is to keep the icon and version until we complete the related issues in the webpage mechanism. In principle we would like to have the new docs in the [next milestone](https://github.com/qiboteam/qibo/milestone/3), but if it doesn't fit we should be able to maintain the logo/text as it is. Let me know, and in case you agree just open an issue for the removal in the future.

MatteoRobbiati:
@scarrazza Yes, I agree. I will remove the comment to the string related to the logo and open an issue for the removal in the future. Is it okay?
Thank you for the feedback.

scarrazza:
> @scarrazza Yes, I agree. I will remove the comment to the string related to the logo and open an issue for the removal in the future. Is it okay?

yes

> Do you want I do this also in qibocal and qibolab?

no, I would like to see the first release of both codes and the new website together.

stavros11:
Thanks for this. I tried to compile the docs locally but I think there is a dependency required (`pip install furo`). If that is the case, maybe we should add furo to the docs requirements in setup.py?

Other than that, some design changes I noticed:
* The white qibo logo does not look good in the bright version, because the background is also very light color. The dark version looks okay.
* The sections level in the left side bar is not as "deep" as rtd. For example, if I remember correctly, in rtd each gate under "Gates" was appearing as a separate entry which is not the case now.
The first point will probably be solved in the website, for the second personally I do not have strong preference.

MatteoRobbiati:
> Thanks for this. I tried to compile the docs locally but I think there is a dependency required (`pip install furo`). If that is the case, maybe we should add furo to the docs requirements in setup.py?
> 
> Other than that, some design changes I noticed:
> 
>     * The white qibo logo does not look good in the bright version, because the background is also very light color. The dark version looks okay.
>

do you mean the `favicon` above the search bar? I had decided to place the black Q of Qibo on a squared white background to make it at least readable in the two cases but I welcome suggestions because it is a very naive solution.

>     * The sections level in the left side bar is not as "deep" as rtd. For example, if I remember correctly, in rtd each gate under "Gates" was appearing as a separate entry which is not the case now.
>       The first point will probably be solved in the website, for the second personally I do not have strong preference.

This is a feature of the `Furo` theme, I think. With Furo we have a double arrangement of labels (left and right) and a different font. I do not know how much action is possible in changing the layout of the `sphinx` theme. In this case I think it is more a matter of figuring out which theme you prefer and best fits the layout of your site.

stavros11:
> do you mean the `favicon` above the search bar? I had decided to place the black Q of Qibo on a squared white background to make it at least readable in the two cases but I welcome suggestions because it is a very naive solution.

No, I meant the Qibo logo on top left of the page, which is white on an almost white background (at least locally). But I guess in the final website you will have the top navbar so this logo will disappear.

> This is a feature of the `Furo` theme, I think. With Furo we have a double arrangement of labels (left and right) and a different font. I do not know how much action is possible in changing the layout of the `sphinx` theme. In this case I think it is more a matter of figuring out which theme you prefer and best fits the layout of your site.

You are right, I did not check the right sidebar, the subsections appear there. Thanks.

MatteoRobbiati:
> No, I meant the Qibo logo on top left of the page, which is white on an almost white background (at least locally). But I guess in the final website you will have the top navbar so this logo will disappear.

Oh, sorry! Now I have understood. I have created an issue related to this topic, the #691. It is exactly what you said: the logo is well visible now (with `read the docs` theme) but we are going to remove it in the future, because the site will have the white background of Furo and we will already have the logo in the navbar.

> You are right, I did not check the right sidebar, the subsections appear there. Thanks.

You're welcome!

PR #699, opened 2022-11-15
Add FusedGate to circuit

### LINKED ISSUE #676: Fuse of inverse of multiple gates gives multiple gates
When building the inverse gate of all gates in a circuit, the gates are not fused but applied one by one, meaning they are not fused.
This is problematic for simulating noise for example.
``` python
circuit = models.Circuit(1)
circuit.add(gates.H(0))
circuit.add(gates.X(0))
inverse = circuit.invert().fuse().queue[0]
circuit.add(inverse)
print(circuit.draw()) # q0: ─H─X─X─H─
```
Right now, the single fused inverse gate has to be build by hand.
``` python
circuit = models.Circuit(1)
circuit.add(gates.H(0))
circuit.add(gates.X(0))
inverse = circuit.invert().fuse().queue[0]
circuit.add(gates.Unitary(inverse.matrix,0))
print(circuit.draw()) # q0: ─H─X─U─
```

stavros11:
Fixes #676.

Checklist:
- [x] Tests are passing.


scarrazza (approved):
LGTM, thanks.

wilkensJ:
Hello Stavros,
thank you for fixing this, now my code from the issue throws an Error:
``` python
circuit = models.Circuit(1)
circuit.add(gates.H(0))
circuit.add(gates.X(0))
inverse = circuit.invert().fuse().queue[0]
circuit.add(inverse)
print(circuit.draw()) # q0: ─H─X─X─H─
```
gives
```txt
NotImplementedError: fused gate is not supported by `circuit.draw`
```

stavros11:
Thanks for checking. I guess this is expected. Given that our drawing mechanism is quite simple, I think there are to ways to draw the fused gate:
1. Decide on a symbol, such as FG or something similar, and use it on all participating qubits. I think this is what we do for unitary gates or channels. The disadvantage of this is that you don't know what the fused gate contains.
2. Draw the gates contained in the fused gate. Then you don't know that these gates are actually.

I believe that option 1 is more in line with our current drawing mechanism. @wilkensJ @scarrazza let me know what you prefer.

wilkensJ:
Is something possible that this:
``` python
circuit = models.Circuit(1)
circuit.add(gates.H(0))
circuit.add(gates.X(0))
inverse = circuit.invert().fuse().queue[0]
circuit.add(inverse)
print(circuit.draw())
```
Will print something like this:
```txt
 q0: ─H─X─[X─H]─
```
If not I would prefer option 1.
Thank you Stavros!

stavros11:
> ```
>  q0: ─H─X─[X─H]─
> ```

@wilkensJ thanks for the proposal, looks good to me. I have implemented it in another PR (#708) because the changes in the code are non-trivial so it is better to review seperately. Let me know if it works for you.

Issue #698, opened 2022-11-15
Negative probabilities error when computing frequencies


AlejandroSopena:
Running circuits representing the identity with density_matrx=True and a computational basis state as an initial state gives some negative probabilities of order $~10^{-18}$ which should be zero. Since they are negative, it is not posible to get the frequencies.
``` python
nqubits=2
c = models.Circuit(nqubits, density_matrix = True)
c.add((gates.RZ(q,theta=-np.pi/2) for q in range(nqubits)))
c.add((gates.RX(q,theta=np.pi/2) for q in range(nqubits)))
c.add((gates.RZ(q,theta=np.pi) for q in range(nqubits)))
c.add((gates.RX(q,theta=np.pi/2) for q in range(nqubits)))
c.add((gates.RZ(q,theta=-np.pi/2) for q in range(nqubits)))
c.add(gates.M(q) for q in range(nqubits))
result=c(nshots=10000)
```
``` python
print(result)
(1+0j)|00><00|

print(result.probabilities())
[ 1.00000000e+00 -3.92523115e-17 -9.81307787e-18 -9.81307787e-18]

print(result.frequencies())
---------------------------------------------------------------------------
File mtrand.pyx:937, in numpy.random.mtrand.RandomState.choice()
ValueError: probabilities are not non-negative

PR #701, opened 2022-11-15
Absolute value in probabilities

AlejandroSopena:
This PR implements a possible fix to #698 

Checklist:
- [ ] Tests are passing.


AlejandroSopena:
Thanks for the suggestions. I have removed the `rtype` variable since it is no longer needed as you said. I could add a test but I think it is not necessary because it seems to me that it is clear from the changes in the code that now the probabilities will always be positive. I have checked that it is now posible to get the frequencies.

stavros11 (approved):
Sounds good to me, thanks.

Issue #652, opened 2022-09-30
Error mitigation methods


AlejandroSopena:
It would be useful to implement error mitigation methods in Qibo. Here I propose some of them:

**Priority**

- [ ]  Zero noise extrapolation [(ZNE)](https://arxiv.org/abs/1612.02058v3)
- [ ] Clifford data regression [(CDR)](https://arxiv.org/abs/2005.10189)
- [ ] Variable-noise Clifford data regression [(vnCDR)](https://arxiv.org/abs/2011.01157)

**Exploratory**

- [ ] Probabilistic error cancellation (PEC) [[1]](https://arxiv.org/abs/1612.02058v3), [[2]](https://arxiv.org/abs/1712.09271)
- [ ] Dynamical decoupling [(DD)](https://arxiv.org/abs/1807.08768)

yjmaxpayne:
@matthew0021: maybe it will be good for the local QAOA QEM efforts to be incorporated into the overall efforts?

PR #702, opened 2022-11-15
Error mitigation

BrunoLiegiBastonLiegi:
Implementation of several error mitigation methods as requested in #652 

Checklist:
- [x] Tests are passing.


src/qibo/tests/test_error_mitigation_cdr.py
```diff
@@ -0,0 +1,84 @@
...
+    return noise
+
+
+@pytest.mark.parametrize("nqubits", [3])
+@pytest.mark.parametrize(
+    "noise",
+    [get_noise_model(DepolarizingError(0.1))],
+)
+@pytest.mark.parametrize("full_output", [False, True])
+def test_cdr(nqubits, noise, full_output):
```
stavros11, line 24: Could you please add a `backend` argument in all tests that are executing circuits:
```suggestion
def test_cdr(backend, nqubits, noise, full_output):
```
Note that you do not need to define this argument. It is a global fixture so pytest will pass the proper values automatically.

src/qibo/tests/test_error_mitigation_cdr.py
```diff
@@ -0,0 +1,84 @@
...
+    c.add(gates.CNOT(q, q + 1) for q in range(0, nqubits - 1, 2))
+    c.add(gates.RZ(q + 1, theta=-2 * dt) for q in range(0, nqubits - 1, 2))
+    c.add(gates.CNOT(q, q + 1) for q in range(0, nqubits - 1, 2))
+    c.add(gates.CNOT(q, q + 1) for q in range(1, nqubits, 2))
+    c.add(gates.RZ(q + 1, theta=-2 * dt) for q in range(1, nqubits, 2))
+    c.add(gates.CNOT(q, q + 1) for q in range(1, nqubits, 2))
+    c.add(gates.M(q) for q in range(nqubits))
+    # Define the observable
+    obs = np.prod([Z(i) for i in range(nqubits)])
+    obs = SymbolicHamiltonian(obs)
```
stavros11, line 45: Then change this to use the proper backend
```suggestion
    obs = SymbolicHamiltonian(obs, backend=backend)
```

src/qibo/tests/test_error_mitigation_cdr.py
```diff
@@ -0,0 +1,84 @@
...
+    # Noise-free expected value
+    exact = obs.expectation(c().state())
```
stavros11, line 47: and this to
```suggestion
    exact = obs.expectation(backend.execute_circuit(c).state())
```

src/qibo/tests/test_error_mitigation_cdr.py
```diff
@@ -0,0 +1,84 @@
...
+    # Noisy expected value without mitigation
+    noisy = obs.expectation(noise.apply(c)().state())
+    # Mitigated expected value
+    estimate = CDR(
+        circuit=c,
+        observable=obs,
+        backend=None,
```
stavros11, line 54: and this to
```suggestion
        backend=backend,
```
to make sure that all backends work properly.

src/qibo/tests/test_error_mitigation_cdr.py
```diff
@@ -0,0 +1,84 @@
...
+    c.add(gates.RZ(q + 1, theta=-2 * dt) for q in range(1, nqubits, 2))
+    c.add(gates.CNOT(q, q + 1) for q in range(1, nqubits, 2))
+    c.add(gates.M(q) for q in range(nqubits))
+    # Define the observable
+    obs = np.prod([Z(i) for i in range(nqubits)])
+    obs = SymbolicHamiltonian(obs)
+    # Noise-free expected value
+    exact = obs.expectation(c().state())
+    # Noisy expected value without mitigation
+    noisy = obs.expectation(noise.apply(c)().state())
```
stavros11, line 49:
```suggestion
    noisy = obs.expectation(backend.execute_circuit(noise.apply(c)).state())
```

src/qibo/tests/test_error_mitigation_vncdr.py
```diff
@@ -0,0 +1,62 @@
+import numpy as np
```
stavros11, line 1: I think it would be simpler if you use a single file for all tests. Then you would need to do the imports and define `get_noise_model` only once. It would also be in correspondance with the code where all error mitigation methods are defined in a single file.

stavros11 (commented):
Thank you for implementing this. I left a few comments below, let me know what you think.

BrunoLiegiBastonLiegi:
Hi @stavros11, thanks for the suggestions! I wasn't aware of the problem of using mutable objects as default arguments, good to know for the future.
I implemented all the suggested modifications. However, eventually I had to define the ``backend`` argument with:

> @pytest.mark.parametrize("backend", [None, GlobalBackend()])

and then inside the tests do:

> b = GlobalBackend() if backend is None else backend
    obs = SymbolicHamiltonian(obs, backend=b)
    # Noise-free expected value
    exact = obs.expectation(b.execute_circuit(c).state())
    # Noisy expected value without mitigation
    noisy = obs.expectation(b.execute_circuit(noise.apply(c)).state())

since we needed the ``backend`` argument of the various ``ZNE``, ``CDR`` and ``vnCDR`` methods to take also the value ``None`` in order to have 100% coverage.
I don't know if this is the cleanest solution, but apparently it is working fine. The only problem I see, is that now we have twice as much test parameters combinations compared to before.

stavros11:
Thank you for the updates.

> However, eventually I had to define the `backend` argument with:
> 
> > @pytest.mark.parametrize("backend", [None, GlobalBackend()])

If you do this it will run the same test twice with the same backend, which is indeed pointless. I would suggest reverting the last commit (d543610ad89a3ca0132e9ff2cea5cb3dcbd46bac) and using the existing pytest fixture for backend as you did in the previous commit (do not use parametrize). This will automatically test all the available backends and the GlobalBackend is always one of them, so it will also be tested.

As you say, coverage will complain because you will not be testing the backend=None case. There are two ways to resolve this: either put some dummy calls at the end of each test where you execute the function with backend=None, or just disable the coverage for these lines (you can do this with `# pragma: no cover`). Here I would go with the second solution (disable coverage), because these functions do non-trivial computations so if you add dummy calls this will add redundant execution time to the tests. 

Although it is generally not recommended to disable coverage or pylint, here it is acceptable, because as I wrote above all backends (including the GlobalBackend) will be tested when you use the backend fixture. In fact we are disabling the coverage in other similar cases in the code, see for example [hamiltonians](https://github.com/qiboteam/qibo/blob/7b9ca94bbe3919177fdda08433ccaf219184e25e/src/qibo/hamiltonians/hamiltonians.py#L21).

BrunoLiegiBastonLiegi:
Thanks for the detailed explanation. I excluded those lines from the coverage calculation as you suggested.
Everything should be fine now.

andrea-pasquale (commented):
Thanks @BrunoLiegiBastonLiegi and @AlejandroSopena for implementing this.
Everything looks good from my side including tests and documentation.
I would only remove the `.DS_Store` file inside the `doc` folder since it is not necessary.
I think that we should add a short tutorial on `How to perform error mitigation?` in the `Advanced examples` section. In this way it should be more easy for a generic user to understand that it is possible to perform error mitigation in Qibo.

BrunoLiegiBastonLiegi:
Thanks for the feedback @andrea-pasquale. I added a section for error mitigation under advanced code examples, let me know what you think about it.

BrunoLiegiBastonLiegi:
Ok, I think the problem is that I inserted some prints in the testcode, but I don't have any output to match them with in testoutput, since the results are random. I could just get rid of the prints, but I liked the idea to print here and there to make it more clear what I am doing. Isn' t there any way to tell sphinx to ignore the output?

doc/source/code-examples/advancedexamples.rst
```diff
@@ -1113,6 +1113,208 @@ and the measurement gate. If ``p1`` is not specified the value of ``p0`` will
 be used for both errors.
 
 
+How to perform error mitigation?
+--------------------------------
+
+Noise and errors in circuits are on of the biggest obstacles to face in quantum computing.
```
andrea-pasquale, line 1119:
```suggestion
Noise and errors in circuits are one of the biggest obstacles to face in quantum computing.
```

andrea-pasquale (approved):
Thanks @BrunoLiegiBastonLiegi.
You can find below a small typo.
The new example is exactly what I had in mind.
Also the fix for the sphinx issue looks good to me.

scarrazza:
Looks good to me too. @stavros11 I remember some comments about the backend, what is the status at this point?

src/qibo/tests/test_models_error_mitigation.py
```diff
@@ -0,0 +1,180 @@
...
+
+@pytest.mark.parametrize("nqubits", [3])
+@pytest.mark.parametrize(
+    "noise",
+    [get_noise_model(DepolarizingError(0.1))],
+)
+@pytest.mark.parametrize("solve", [False, True])
+def test_zne(backend, nqubits, noise, solve):
+    """Test that ZNE reduces the noise."""
+    qibo.set_threads(1)
```
stavros11, line 27: Thanks for setting the threads, I believe the tests are faster now, do you also observe some difference? 

One small detail, it may be better to use
```suggestion
    backend.set_threads(1)
```
so that we only set the threads for the local backend and we don't affect other tests that run after this.

AlejandroSopena: Thanks. By using `backend.set_threads(1)` the times drop from 1min 35sec to 0.5sec for ZNE, from 4min 12sec to 1.9sec for CDR and from 12min 36sec to 3.9sec for vnCDR.

PR #716, opened 2022-11-29
fixing typo in docs appendix

MatteoRobbiati:
This is a very tiny PR, I saw a typo into the `qibojit` reference.

Perhaps this branch can be kept open to correct any errors like this that we find.


scarrazza (approved):
Thanks.

PR #721, opened 2022-12-05
 Liouville representation of quantum channels

renatomello:
Checklist:
- [x] Tests are passing.


renatomello:
This PR intends to implement a solution to #721

renatomello:
> @renatomello thanks for this. Is this ready for review?

Not yet, there's still more things to be added.

renatomello:
This is ready for review @stavros11 @scarrazza

src/qibo/quantum_info/basis.py
```diff
@@ -0,0 +1,211 @@
...
+        from qibo.backends import GlobalBackend
+
+        backend = GlobalBackend()
+
+    basis = [
+        I(0).asmatrix(backend),
+        X(0).asmatrix(backend),
+        Y(0).asmatrix(backend),
+        Z(0).asmatrix(backend),
+    ]
```
stavros11, line 125: Here you can also use
```suggestion
    from qibo.matrices import I, X, Y, Z
    basis = [I, X, Y, Z]
```
Note that these matrices are always numpy arrays.

src/qibo/quantum_info/basis.py
```diff
@@ -0,0 +1,211 @@
...
+    return backend.cast(state, dtype=state.dtype)
+
+
+def pauli_basis(
+    nqubits: int, normalize: bool = False, vectorize: bool = False, backend=None
+):
+    """Creates the ``nqubits``-qubit Pauli basis.
+
+    Args:
+        qubits (int): number of qubits.
```
stavros11, line 87:
```suggestion
        nqubits (int): number of qubits.
```

stavros11 (commented):
Some tests are failing with
```py
E   TypeError: Unsupported type <class 'numpy.ndarray'>

cupy/_core/_kernel.pyx:124: TypeError
```
if you run in a machine with GPU. I suspect that `GlobalBackend` is used somewhere is not supposed to because even the numpy and numba tests are failing with the cupy error.

Generally is there a motivation for using the `backend` object in basis.py? As far as I can see everything is done in numpy and you are only casting the result in the end. If that is the case, I guess you could remove backend from everywhere similar to the rest of methods under quantum_info.

src/qibo/quantum_info/basis.py
renatomello: Done

src/qibo/quantum_info/basis.py
renatomello: I couldn't import `qibo.matrices`. Do you mean `from qibo.backends.npmatrices import NumpyMatrices` ?

renatomello: changed that import and removed backend from all `basis.py` functions

renatomello:
@wilkensJ found a bug in the `PauliNoiseChannel` class that was fixed in commit https://github.com/qiboteam/qibo/pull/721/commits/9b32f56f2b3feb34b42db38d5db04de8fa746fe3 and the fix was generalized to other classes in commit https://github.com/qiboteam/qibo/pull/721/commits/4f07525636f3d1972583f8e450f6a9dcc095e6b5

renatomello:
> Some tests are failing with
> 
> ```python
> E   TypeError: Unsupported type <class 'numpy.ndarray'>
> 
> cupy/_core/_kernel.pyx:124: TypeError
> ```
> 
> if you run in a machine with GPU. I suspect that `GlobalBackend` is used somewhere is not supposed to because even the numpy and numba tests are failing with the cupy error.
> 
> Generally is there a motivation for using the `backend` object in basis.py? As far as I can see everything is done in numpy and you are only casting the result in the end. If that is the case, I guess you could remove backend from everywhere similar to the rest of methods under quantum_info.

Made this proposed change. All tests are passing.

stavros11 (commented):
> Made this proposed change. All tests are passing.

Thanks for the updates. Basis tests are passing, I still get some failures with cupy for test_gates_channels. I believe the following will fix this:

src/qibo/tests/test_gates_channels.py
```diff
@@ -25,9 +26,26 @@ def test_general_channel(backend):
 
 def test_krauss_channel_errors():
```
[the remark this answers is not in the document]
renatomello: Done

src/qibo/tests/test_gates_channels.py
```diff
@@ -25,9 +26,26 @@ def test_general_channel(backend):
...
+            [0.4 + 0.0j, 0.0 + 0.0j, 0.0 + 0.0j, 0.6 + 0.0j],
+        ]
+    )
+
+    test_pauli = np.diag([2.0, -0.4, -2.0, 0.4])
+
+    channel = gates.KrausChannel([((0,), a1), ((0,), a2)])
+
+    assert np.linalg.norm(channel.to_superop() - test_superop) < PRECISION_TOL
+    assert np.linalg.norm(channel.to_pauli_liouville() - test_pauli) < PRECISION_TOL
```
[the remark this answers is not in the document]
renatomello: Also done

stavros11 (approved):
I pushed the fix for cupy tests, let me know if you agree. If CI passes this should be ready. Maybe @wilkensJ would also like to have a look before we merge.

renatomello (commented):
Thank you for these. I fixed my cupy install and can confirm the tests are passing

PR #724, opened 2022-12-06
Created Custom, Unitary and Kraus errors for NoiseModel

vodovozovaliza:
Checklist:
- [x] Tests are passing.


src/qibo/noise.py
```diff
@@ -1,4 +1,32 @@
...
+    Args:
+        channel (:class:`qibo.gates.Channel`): any channel
+
+    Example:
+
+    .. testcode::
+
+        import numpy as np
+        from qibo.gates import KrausChannel
+        from qibo.noise import Error
```
andrea-pasquale, line 17: If you called the class `CustomError` also here you should use `CustomError` instead of `Error`.

src/qibo/noise.py
```diff
@@ -1,4 +1,32 @@
...
+
+        # define |0><0|
+        a1 = np.array([[1, 0], [0, 0]])
+        # define |0><1|
+        a2 = np.array([[0, 1], [0, 0]])
+
+        # Create an Error associated with Kraus Channel rho -> |0><0| rho |0><0| + |0><1| rho |0><1|
+        error = Error(gates.KrausChannel([((0,), a1), ((0,), a2)]))
```
andrea-pasquale, line 25: Same thing here.

andrea-pasquale (commented):
Thank you very much @vodovozovaliza.
You can find below a few comments.
This PR is almost ready to merge in my opinion.
Could you please add the new errors in the documentation?
You need to modify this file in order to add the new errors. https://github.com/qiboteam/qibo/blob/master/doc/source/api-reference/qibo.rst#quantum-errors

andrea-pasquale (approved):
Thanks @vodovozovaliza.
Everything looks good to me.

stavros11 (approved):
Looks good to me, thanks. Regarding the KrausChannel that is not available for state vectors, the error I am getting is:
```py
NotImplementedError: KrausChannel cannot be applied to state vector.
```
which is acceptable. So I do not think anything else is needed regarding this.

scarrazza:
@stavros11 after merging master here `test_expectation_from_samples` is failing on mac. Do you remember if we had something similar recently?

stavros11:
> @stavros11 after merging master here `test_expectation_from_samples` is failing on mac. Do you remember if we had something similar recently?

I get it in my laptop with master so it is not related to this PR and mac. It happens randomly, not in every execution and is more likely a tolerance issue because different samples are used every time and there is some statistical error since the shots are not infinite. If I set the seed using
```py
np.random.seed(123)
```
in the beginning of the test it seems to fix the issue. Shall I push this?

scarrazza:
Thanks for checking. Indeed, fixing the seed is probably sufficient, please go ahead.

scarrazza (approved):

PR #729, opened 2022-12-12
Qcnn2

ihpcdingwj:
Checklist:
- [ ] Tests are passing.


src/qibo/models/qcnn.py
```diff
@@ -0,0 +1,299 @@
...
+        Args:
+            theta: list or numpy.array with the biases to be used in the circuit
+            init_state: numpy.array with the quantum state to be classified
+            nshots: int number of runs of the circuit during the sampling process (default=10000)
+        Returns:
+            numpy.array() with predictions for each qubit, for the initial state
+        """
+        bias = np.array(theta[0 : self.measured_qubits])
+        circuit = circuit(init_state, nshots)
+        result = circuit.frequencies(binary=False)
```
MatteoRobbiati, line 204: Another possibility here can be to use directly [`probabilities`](https://qibo.readthedocs.io/en/stable/api-reference/qibo.html#qibo.states.CircuitResult.probabilities), but both ways are effective.

src/qibo/models/qcnn.py
```diff
@@ -0,0 +1,299 @@
...
+        Args:
+            labels: list or numpy.array with the qubit labels of the quantum states to be classified
+            predictions: list or numpy.array with the qubit predictions for the quantum states to be classified
+        Returns:
+            numpy.float32 with the value of the square-loss function
+        """
+        loss = 0
+        for l, p in zip(labels, predictions):
+            for qubit in range(self.measured_qubits):
+                loss += (l[qubit] - p[qubit]) ** 2
```
MatteoRobbiati, line 225: Here you use an MSE: ere the labels associated with the training data supposed to be continuous? I am used to encoding classes in discrete variables or in the components of a $N_{classes}$ dimensional vector, so I am interested in understanding this choice.

MatteoRobbiati (changes requested):
Thank you for this model, I find it personally very interesting. 
However, I would like to better understand some steps, which are currently unclear to me.

- [ ] you use some conventions here that I am not familiar with (e.g., the use of the list of symbols in which you collect rotations, or some magic numbers whose origin I do not understand). I think it would be helpful to better explain these choices, perhaps through a `README.md` file, or by linking some references that help understand the model, or writing a tutorial about it.
- [ ] Need an example in the main docstring.
- [ ] Docstrings should be completed and conformed to the general convention we use in `qibo`. This would help users in understanding the formalism and immediately grasp how the classifier modules work.
- [x] There are some steps where the file is in "developing mode." These lines need to be fixed before merge (possibly we collect suggestions in appropriate issues).  
- [ ] There are some new features in this model compared to the re-uploading classifier we already have in `qibo`. I think it is important to point them out (embedding of data in amplitudes, pooling strategy, etc.). This is also why I suggest again to build a reference list and write a tutorial (showing some results you have obtained). Regarding the tutorial, if you agree, a new PR could be opened. 



I left some comments in the main file, much of them due to my curiosity in understanding this mechanism.

src/qibo/models/qcnn.py
```diff
@@ -0,0 +1,299 @@
...
+
+import qibo
+from qibo import gates, matrices
+from qibo.models import Circuit
+
+
+class QuantumCNN:
+    def __init__(
+        self, nqubits, nlayers, nclasses=2, params=None
+    ):  ### TODO: modify this to build the QCNN circuit
```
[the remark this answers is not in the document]
jf-kong: yes

src/qibo/models/qcnn.py
```diff
@@ -0,0 +1,299 @@
...
+
+    def two_qubit_unitary(self, bits, symbols):
+        """Make a circuit that creates an arbitrary two qubit unitary.
+        Symbols should be a list of length 15.
+        """
+        c = Circuit(self.nqubits)
+        c += self.one_qubit_unitary(bits[0], symbols[0:3])
+        c += self.one_qubit_unitary(bits[1], symbols[3:6])
+        # to improve: to define new gates of XX YY and ZZ outside.
+        """matrixXX = K.np.kron(matrices.X,matrices.X)
```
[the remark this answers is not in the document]
jf-kong: Thanks for suggestions. Have removed the comments and to leave dedicated XX, YY, ZZ gates for future improvement, as the current implementation wth RXX, RYY, RZZ gates works too.

src/qibo/models/qcnn.py
```diff
@@ -0,0 +1,299 @@
...
+        Class for a multi-task variational quantum classifier
+        Args:
+            nclasses: int number of classes to be classified. Default setting of 2 (phases).
+            nqubits: int number of qubits employed in the quantum circuit
+        """
+        self.nclasses = nclasses
+        self.nqubits = nqubits
+        self.nlayers = nlayers
+
+        # NOTE: magic number: 21=15+6 params per layer of conv and pool
```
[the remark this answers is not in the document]
jf-kong: Currently this "magic" number is the number of variational parameters in each convolution and pooling layer and is based on the ansatz chosen. The ansatz here is inspired by the one shown here: https://www.tensorflow.org/quantum/tutorials/qcnn. The comment was a reminder and has been removed.

src/qibo/models/qcnn.py
```diff
@@ -0,0 +1,299 @@
...
+
+        self.set_circuit_params(angles)
+        return self._circuit
+
+    def Predictions(self, circuit, theta, init_state, nshots=10000):
+        """
+        Args:
+            theta: list or numpy.array with the biases to be used in the circuit
+            init_state: numpy.array with the quantum state to be classified
+            nshots: int number of runs of the circuit during the sampling process (default=10000)
```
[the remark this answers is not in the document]
jf-kong: Not particularly at the moment. A relatively large number was chosen as a default to reduce sampling errors during development/debugging. Should be adjusted during usage for specific problems.

src/qibo/models/qcnn.py
```diff
@@ -0,0 +1,299 @@
...
+        expanded_params = []
+        nbits = self.nqubits
+        for layer in range(self.nlayers):
+            nleft = nbits / (2**layer)
+            param_start = layer * self.nparams_layer
+            conv_params = params[param_start : param_start + self.nparams_conv]
+            pool_params = params[
+                param_start + self.nparams_conv : param_start + self.nparams_layer
+            ]
+            # pool_params += pool_params[2::-1]
```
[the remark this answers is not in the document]
jf-kong: No longer needed and removed, thanks for catching it.

src/qibo/models/qcnn.py
jf-kong: The labels are $N_{\textrm{measured qubits}}$ dimensional vectors with each component being -1 or 1. Currently MSE is chosen because the predictions are "continuous". This choice is largely inspired by the variational classifier example, but we could possibly provide user the choice to specify loss function in the future.

MatteoRobbiati: Okay, I understand. Maybe it might be useful to use the sign of the expected value as a class predictor.
In any case, I do not think that this choice will have a considerable impact on the results. 

By the way, I was wondering if you had saved any results or applications of this model: so that I could compare the accuracy of this QCNN with other models (I am curious about this).

MatteoRobbiati:
I have modified my first comment to a checklist. It can be useful for keeping track of changes to be made. Please let me know if you agree with my suggestions or if we would like to discuss any particular point. If so, we can also schedule a meeting to discuss it.

jf-kong:
Thanks @MatteoRobbiati for the comments and suggestions. For the first and fourth points, a separate tutorial might be useful as you suggested. Should that be in a separate PR?

MatteoRobbiati:
> Thanks @MatteoRobbiati for the comments and suggestions. For the first and fourth points, a separate tutorial might be useful as you suggested. Should that be in a separate PR?

Yes, I agree with you. We can move the points 1 and 4 to a new PR, where we can better explain the potential of the model and show some results. In doing so, I would suggest setting the tutorial to be compatible with those shared on our [tutorials webpage](https://qibo.science/tutorials).

At this point I am going to do a new review and I think @Edoardo-Pedicillo can help me with the review of the tests. 

Once this is done, I think it should be ready.

Thank you for the job!

src/qibo/models/qcnn.py
```diff
@@ -0,0 +1,307 @@
...
+
+        s = self.square_loss(labels, predictions)
+
+        return s
+
+    def minimize(
+        self, init_theta, data=None, labels=None, nshots=10000, method="Powell"
+    ):
+        """
+        Args:
```
MatteoRobbiati, line 255: I think these docstring are wrong (e.g. `RY` is not passed here)! Can you have a look please?

MatteoRobbiati (changes requested):
Thank you for your modifications. I think this shape is close to be ready.
I would suggest to choose one way to write docstrings: 

```
"""
Brief description.

Args:
      arg1: ..
      arg2: ..

Returns: ...
"""
```

At the moment, we have some different shapes in our `qcnn` docstrings.

I tried to run the example code and it fails on my jupyter notebook. It seems to be a problem related to the reshaping of `data`. Is this a problem related to my tensorflow installation or do you have the same error?

src/qibo/models/qcnn.py
jf-kong: Thanks for the catch! Have corrected the docstrings.

src/qibo/models/qcnn.py
```diff
@@ -0,0 +1,307 @@
...
+        self.nparams_layer = self.nparams_conv + self.nparams_pool
+        self.measured_qubits = int(np.ceil(np.log2(self.nclasses)))
+
+        if self.nqubits <= 1:
+            raise ValueError("nqubits must be larger than 1")
+
+        self._circuit = self.ansatz(nlayers, params=params)
+
+    def quantum_conv_circuit(self, bits, symbols):
+        c = Circuit(self.nqubits)
```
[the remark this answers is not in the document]
jf-kong: Thanks for the suggestion. Have added the docstrings. Just to also add a comment that these are primarily meant to be internal functions.

jykhoo1987:
> 

Yes so by default without specifying the backend, I believe qibo uses qibo tensorflow and it has issues running. I had no problems running with qibo.set_backend("numpy") or using qibojit.

jykhoo1987:
> and it fails on my jupyter notebook. It seems to be a problem related to the reshaping of `data`. Is this a problem related to my

Yes so by default without specifying the backend, I believe qibo uses qibo tensorflow and I also encountered issues while running the code. I had no problems running with qibo.set_backend("numpy") or using qibojit. Perhaps you could try running with numpy or qibojit?

MatteoRobbiati:
> > and it fails on my jupyter notebook. It seems to be a problem related to the reshaping of `data`. Is this a problem related to my
> 
> @MatteoRobbiati Yes so by default without specifying the backend, I believe qibo uses qibo tensorflow and I also encountered issues while running the code. I had no problems running with qibo.set_backend("numpy") or using qibojit. Perhaps you could try running with numpy or qibojit?

I did, what I get is:

```bash
ValueError: cannot reshape array of size 1 into shape (2,2,2,2)
```
Also changing the number of qubits the error raised is similar, in particular the `(2, ..., 2)` array is asked to be `nqubits` long. I see that `data.shape` is `(16,)`, so each data is here a scalar. If I have understood, the idea is to embed these data into the initial state. Maybe the only needed thing is to do a brief data pre-processing in order to have shapes like `(2, nqubits)`? 

Let me know, and sorry if it is my misunderstanding!

ihpcdingwj:
>

jykhoo1987:
> > > and it fails on my jupyter notebook. It seems to be a problem related to the reshaping of `data`. Is this a problem related to my
> > 
> > 
> > @MatteoRobbiati Yes so by default without specifying the backend, I believe qibo uses qibo tensorflow and I also encountered issues while running the code. I had no problems running with qibo.set_backend("numpy") or using qibojit. Perhaps you could try running with numpy or qibojit?
> 
> I did, what I get is:
> 
> ```shell
> ValueError: cannot reshape array of size 1 into shape (2,2,2,2)
> ```
> 
> Also changing the number of qubits the error raised is similar, in particular the `(2, ..., 2)` array is asked to be `nqubits` long. I see that `data.shape` is `(16,)`, so each data is here a scalar. If I have understood, the idea is to embed these data into the initial state. Maybe the only needed thing is to do a brief data pre-processing in order to have shapes like `(2, nqubits)`?
> 
> Let me know, and sorry if it is my misunderstanding!

Hi @MatteoRobbiati,

Sorry for the delay. We have added the explicit use of numpy backend and also modified the data format. We have tested on multiple machines and it should work now. Please try again, thanks!

src/qibo/models/qcnn.py
```diff
@@ -0,0 +1,366 @@
...
+        nqubits (int): number of qubits of the input states. Currently supports powers of 2.
+        nlayers (int): number of convolutional and pooling layers.
+        nclasses (int): number of classes to be classified. Default setting of 2 (phases).
+        params: initial list of variational parameters. If not provided, all parameters
+            will be initialized to zero.
+    Example:
+        .. testcode::
+            import qibo
+            from qibo.models.qcnn import QuantumCNN
+            from qibo import gates, matrices
```
MatteoRobbiati, line 24:
```suggestion
```

src/qibo/models/qcnn.py
```diff
@@ -0,0 +1,366 @@
...
+            from qibo.models import Circuit
```
MatteoRobbiati, line 25:
```suggestion
```

src/qibo/models/qcnn.py
```diff
@@ -0,0 +1,366 @@
...
+            import math
```
MatteoRobbiati, line 26:
```suggestion
            import qibo
```

MatteoRobbiati (changes requested):
Thank you for the modifications. I confirm that now the code is working.
The expected result is the prediction for a single data, right?

Some last suggestions follow: some unused imports have been removed from the example and `qibo` is imported in order to make `set_backend` work.

After this, I think we can proceed.

Issue #731, opened 2022-12-14
add phase shift rule into optimizers


scarrazza:
Following the discussion today, I believe we should make #663 available for the optimizer method so a family of models/circuits can benefit from it.

PR #735, opened 2022-12-20
Rescaled parameter shift rule

MatteoRobbiati:
In this PR a rescaled version of the parameter shift rule (PSR) is implemented.
I purpose to add this feature to the `derivative.py` module because in the optimization context (this is also linked to #731) we often make use of the re-uploading strategy. 
Doing this, we act with rotations whose angles are built up multiplying a variational parameter with a feature: $\theta' = \theta \cdot x$.

Thus:

- in order to perform the derivative of the $\langle H \rangle$ ( $H$ hamiltonian) with respect to the variational parameters only (we optimize the parameters, not the variables), we need to rescale the output of the PSR using $x$ itself;
- in order to perform the derivative of the same object with respect to $x$ (it can be useful in some situation), we need to rescale the PSR output using $\theta$.

This simple modification can be useful for these reasons and can be used in either case, specifying which scaling factor to apply to the result ( $\theta$ if we derive with respect to $x$ and vice versa).


MatteoRobbiati:
@DiegoGM91 this is the PR I mentioned in #667. 
I don't know why, but I can't add you as reviewer, so I mentioned you here.

src/qibo/tests/test_derivative.py
```diff
@@ -54,3 +56,39 @@ def test_derivative(backend):
     backend.assert_allclose(grad_0, 8.51104358e-02, atol=1e-10)
     backend.assert_allclose(grad_1, 5.20075970e-01, atol=1e-10)
     backend.assert_allclose(grad_2, 0, atol=1e-10)
+
+    # ------------------- rescaled parameter shift rule ------------------------
+
```
stavros11, line 61: I would suggest to split this to two test methods `test_parameter_shift` and `test_rescaled_parameter_shift`, instead of seperating with the comment. This way it will be easier to read and they will appear seperately in the pytest report, so it will be easier to identify issues in case one of them fails.

src/qibo/derivative.py
```diff
@@ -115,3 +115,85 @@ def circuit(nqubits = 1):
...
+    variational parameters we need to "free" this procedure from the x depencency.
+    This is the aim of this method.
+
+    Original references:
+        `https://arxiv.org/abs/1811.11184`;
+        `https://arxiv.org/abs/2210.10787`.
+
+    Args:
+        circuit (:class:`qibo.models.circuit.Circuit`): custom quantum circuit.
+        hamiltonian (:class: `qibo.hamiltonians.Hamiltonian`): target observable.
```
stavros11, line 144:
```suggestion
        hamiltonian (:class:`qibo.hamiltonians.Hamiltonian`): target observable.
```

src/qibo/derivative.py
```diff
@@ -115,3 +115,85 @@ def circuit(nqubits = 1):
...
+        parameter_index (int): the index which identifies the target parameter in the circuit.get_parameters() list
+        scale_factor (float): parameter scale factor;
+        initial_state ((1, 2**nqubits) matrix): initial state on which the circuit acts.
```
stavros11, line 147:
```suggestion
        initial_state ((2**nqubits,) matrix): initial state on which the circuit acts.
```
as before.

stavros11 (commented):
Thanks for implementing this. Maybe I missed something but the new function looks identical with the parameter shift, just with the result multiplied by the scale factor. If that's true, then how about just calling `parameter_shift` instead of rewriting everything, eg:
```py
return parameter_shift() * scale_factor
```

MatteoRobbiati:
> Thanks for implementing this. Maybe I missed something but the new function looks identical with the parameter shift, just with the result multiplied by the scale factor. If that's true, then how about just calling `parameter_shift` instead of rewriting everything, eg:
> 
> ```python
> return parameter_shift() * scale_factor
> ```

Thank you Stavros for this. 
You are right, I have re-built the method as suggested yesterday during the meeting: now the "standard" PSR method has a new input parameter: `scale_factor=None`. 
On this way, if a value is passed, we use it. Otherwise, we impose `scale_factor=1` and it doesn't affect the calculation. 
Do you agree with the solution? You can find it into the code uploaded with the last push.

src/qibo/derivative.py
MatteoRobbiati: With the new code, and following @DiegoGM91 suggestion, this parameter is now represented as `(2**qubits) vector`.

src/qibo/tests/test_derivative.py
MatteoRobbiati: Since I have removed the new method and modified the old one, this split is no longer necessary.

src/qibo/derivative.py
```diff
@@ -9,21 +9,35 @@ def parameter_shift(
...
+    whose angles are in the form: theta' = theta * x, where theta is a variational
+    parameter and x an input variable. The PSR allows to calculate the derivative
+    with respect of theta' but, if we want to optimize a system with respect its
+    variational parameters we need to "free" this procedure from the x depencency.
+    If the `scale_factor` is not provided, it is set equal to one and doesn't
+    affect the calculation.
 
     Original references:
         `https://arxiv.org/abs/1811.11184`;
         `https://arxiv.org/abs/1803.00745`.
```
stavros11, line 32: More relevant for the docs added in #741 but if you like there is a way in Sphinx to make these appear as actual links in the docs page (so that one can press and go directly to the arxiv page). Now one needs to copy-paste.

stavros11 (commented):
> Thank you Stavros for this. You are right, I have re-built the method as suggested yesterday during the meeting: now the "standard" PSR method has a new input parameter: `scale_factor=None`. On this way, if a value is passed, we use it. Otherwise, we impose `scale_factor=1` and it doesn't affect the calculation. Do you agree with the solution? You can find it into the code uploaded with the last push.

Yes, I think that is the best solution. You can even directly use `scale_factor=1` as the default value in the function definition, so that you avoid `if scale_factor is None`. It is fine to use floats as default values.

In the tests, I still think it is better to use two different methods to check with or without rescaling or alternatively you could use a single test with
```py
@pytest.mark.parametrize("scale_factor", [1, 0.5])
```

Issue #689, opened 2022-11-10
Plot density matrices or distributions automatically with Qibo


bpcarlos:
Adding this issue as a possible enhancement for the future.
After a discussion with @stavros11, we thought that could be a good idea to have the possibility to plot density matrices or distributions automatically from Qibo, in a visually appealing manner.

It could be something similar as it is done, for instance, in [Qiskit](https://qiskit.org/documentation/stubs/qiskit.visualization.plot_state_city.html) or [QuTiP](https://qutip.org/docs/3.1.0/guide/guide-visualization.html).

renatomello:
@sergiomtzlosa would you be interested in doing this?

sergiomtzlosa:
~~sorry, but I am busy right now and I cannot do this, thanks a lot for the offer!~~....well, let me try, I will take some time slot to try this implementation

sergiomtzlosa:
see pull request #1672

PR #737, opened 2022-12-21
Added a function for plotting density as a 3d histogram

DaniloKolesnyk:
Checklist:
- [ ] Tests are passing.

### DIFF
--- a/histogram_density.py
+++ b/histogram_density.py
@@ -1,0 +1,89 @@
+import numpy  as np
+import matplotlib.pyplot as plt
+from qibo import * 
+from matplotlib import cm
+ 
+
+def plot_density_hist(circuit):
+    """Function that plots density matrix as a 3d histogram
+
+    Args:
+        circuit (:class:`qibo.models.Circuit`): a circuit with either a state vector or density
+    """
+
+    state = circuit.execute().state() 
+    # Create a density matrix from state vector
+    if not circuit.density_matrix:  
+        ket = np.asarray(state)
+        state = np.outer(ket, ket.conj())  
+
+    # Plot real values of density matrix
+    matrix = state.real 
+
+    fig = plt.figure()
+    ax = fig.add_subplot(111, projection='3d')
+
+    xpos = [range(matrix.shape[0])]
+    ypos = [range(matrix.shape[1])]
+    xpos, ypos = np.meshgrid(xpos, ypos)
+    xpos = xpos.flatten('F')
+    ypos = ypos.flatten('F')
+    zpos = np.zeros_like(xpos)
+
+    dx = 0.5 * np.ones_like(zpos)
+    dy = dx.copy()
+    dz = matrix.flatten()
+    
+    # Ticks length
+    number_of_states_x = len([bin(i)[2:].zfill(circuit.nqubits) for i in range(state.shape[0])])
+    number_of_states_y = len([bin(i)[2:].zfill(circuit.nqubits) for i in range(2 ** circuit.nqubits)])
+    ax.set_xticks(0.5 + np.arange( number_of_states_x))
+    ax.set_yticks(0.5 + np.arange(number_of_states_y))
+    
+    # Ticklabels
+    ax.set_xticklabels([bin(i)[2:].zfill(circuit.nqubits) for i in range(state.shape[0])])
+    ax.set_yticklabels([bin(i)[2:].zfill(circuit.nqubits) for i in range(2 ** circuit.nqubits)])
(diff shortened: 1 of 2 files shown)


scarrazza:
@DaniloKolesnyk thank you for this. I like the idea of including this feature.

Let me open a discussion about pre-defined plotting. In qibo we never implemented a module for matplotlib figure generation and we motivate this choice with two arguments: (a) users might have issues when running the code in a shell environment and (b) usually plots require post-processing cosmetic changes.

Now, back to this PR, I see two options:
1. (a) include a method in the circuit class which returns the 3d histogram data, (b) move the explicit plotting code to a tutorial so the user can implement his own version.
2. instead of 1.b, we define a new plotting module in qibo where all matplotlib specific code can go.

@stavros11 @andrea-pasquale opinions about this?

andrea-pasquale:
I also remember having this discussion in the past. I agree with the fact that we shouldn't put plotting methods in Qibo, however especially when it comes to show how Qibo (and Quantum Computing in general) works it may be useful to have our own plotting methods stored somewhere. Especially for people that are using Qibo and that would like to include plots in their publications (see also issue #689).

Regarding the two options, personally I like the most option 2, which is actually the same approach that qiskit is following with the [visualization module](https://qiskit.org/documentation/apidoc/visualization.html).

This module could simply take a Qibo object, like a circuit or a state vector, and generate the appropriate plots.

If we see that this approach works in the future we could also move the plotting methods from qibocal and store everything in this new module. It may also be useful for the new webiste (?).

stavros11:
I also agree with option 2. In the past we have received similar requests, particularly for the plot implemented in this PR, so I think having some basic visualization module that can be used easily would help many users. This will require some effort from our side to develop and maintain but it's a new feature so there's no way to avoid this.

The counter arguments are valid, however for (a) the plotting module will be seperate so people using shell may decide to not use it. For (b) if users require something special they will have to write their own code. We can think of our plotting module as very basic which is still useful for research even if it is not the plots one would put in a paper. That being said, we can design the module in a way that returns a basic plot that can be printed directly with `plt.show()` or `plt.savefig()` (for new users) but also gives access to data that allow further customization (for advanced users).

Regarding if this would help for the website, maybe @MatteoRobbiati @wilkensJ have more input.

If we decide to go with the module, I think the next steps for this PR would be the following:
1. Create a new folder `visualization` under src/qibo and move this plot method there.
2. Currently matplotlib is in the requirements (I am not sure why). We can remove it and in the plotting module check if available and if not raise an error that prompts the user to install it.
3. Check whether the plot method can be more modular and return data that can be customized later.
4. Optionally we could add a shortcut `.plot()` method in `CircuitResult` that uses the plotting module. This will allow doing `result.plot()` and getting the histogram right after circuit execution which could be useful for many users, especially in notebooks.

histogram_density.py
```diff
@@ -0,0 +1,103 @@
+import matplotlib.pyplot as plt
+import numpy as np
+from matplotlib import cm
+
+from qibo import *
+
+
+def plot_density_hist(circuit):
```
stavros11, line 8: I believe this will be more flexible if it takes a state (as array) or a `CircuitResult` object. This would delegate the circuit execution to the user, instead of the plotting method:
```py
result = circuit.execute()
plot_density_hist(result)
```
which is useful in cases where the execution is customized, for example if one wants to pass initial state, nshots, etc.

Moreover, the state to be visualized may come from someting that is not a circuit, such variational models, etc..

scarrazza:
Thanks for the feedback, we can go for option 2.

@DaniloKolesnyk could you please perform the changes requested by @stavros11 and then move your plotting function to a subfolder called `plotting`?

Concerning tests, we can use https://pypi.org/project/pytest-mpl/ to generate fixtures and then compare plots.

MatteoRobbiati:
Thank you @DaniloKolesnyk for this.

I have been following the discussion and I like the idea of having a visualization module. I think, as suggested by @andrea-pasquale and @stavros11, this could help in the development of the tutorials in our documentation (just think of the choice of colors, which in a visualization module would be set by us and would be convenient if it was the same adopted throughout the documentation). In addition, I think that using this kind of approach could allow us to reduce the number of images located "outside" the tutorial itself: we would show them as notebook (through one-line commands) output and could involve them in the process of conversion to `*.json` without output and reconstruction in rendering web pages.

scarrazza:
@DaniloKolesnyk please let us know how things are going for this implementation so we can help.

scarrazza:
I am closing this and reporting the discussion for a future implementation based on #689.

PR #739, opened 2022-12-22
Update `gate.init_kwargs` when setting new parameters

stavros11:
When changing the parameters of a parametrized gate the `gate.init_kwargs` dictionary is not updated. Therefore if a new gate is created after it will have the old parameter values. For example
```py
gate = gates.RX(0, theta=0)
gate.parameters = 0.5
gate2 = gates.__class__(*gate.init_args, **gate.init_kwargs)
# gate2.parameters is still 0
```
This is relevant among other things for the [transpiler](https://github.com/qiboteam/qibolab/blob/d64cc98afbdc23adc1e472d70d8c1827d5f8da96/src/qibolab/transpilers/connectivity.py#L100), which is using this when creating new gates that respect hardware connectivity.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/gates/abstract.py
+++ b/src/qibo/gates/abstract.py
@@ -316,6 +316,7 @@ class ParametrizedGate(Gate):
         """Updates the values of gate's parameters."""
         if isinstance(self.parameter_names, str):
             nparams = 1
+            names = [self.parameter_names]
             if not isinstance(x, collections.abc.Iterable):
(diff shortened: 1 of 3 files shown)


andrea-pasquale (approved):
Thanks @stavros11.
LGTM.

scarrazza (approved):

PR #741, opened 2022-12-23
Add derivative to qibo docs

MatteoRobbiati:
in this PR, the description of the `derivative` module has been added to the documentation.


stavros11 (commented):
Looks good. I would suggest to merge #735 here and fix the doctest so that we can merge.

MatteoRobbiati:
I have just modified the module following your suggestions @stavros11. 
Thank you for the review.

scarrazza:
@MatteoRobbiati looks good to me, however there is a documentation test failing, i.e. you have introduced a short code example in the documentation and when executed there is something that raises an error, could you please check so we can proceed with the merge?

MatteoRobbiati:
> @MatteoRobbiati looks good to me, however there is a documentation test failing, i.e. you have introduced a short code example in the documentation and when executed there is something that raises an error, could you please check so we can proceed with the merge?

I think the problem is now solved.

PR #745, opened 2022-12-27
Moved "qibo release" to the doc's side-bar

MatteoRobbiati:
In this PR the software's release is added to the doc's side-bar (in the front-page) and removed from the introductory text. 

In order to show e.g.: `Release 0.1.10dev0`, I set "Release" where the package name should be placed. I know it is conceptually wrong, but it seemed the best solution to avoid repetition (Qibo is already mentioned in the navbar) and to clarify that the string `0.1.10dev0` refers to the release.

If there is a more elegant way, it is certainly preferable.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -20,12 +20,12 @@ import qibo
 
 # -- Project information -----------------------------------------------------
 
-project = "qibo"
+project = "Release"
 copyright = "2020-2022 by the Qibo team"
 author = "The Qibo team"
 
 # The full version, including alpha/beta/rc tags
-# release = qibo.__version__
+release = qibo.__version__
 
 
 # -- General configuration ---------------------------------------------------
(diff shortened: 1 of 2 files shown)


scarrazza:
@MatteoRobbiati please do not overcomplicate things: https://pradyunsg.me/furo/customisation/sidebar-title/

MatteoRobbiati:
> @MatteoRobbiati please do not overcomplicate things: https://pradyunsg.me/furo/customisation/sidebar-title/

Thank you.

doc/source/conf.py
```diff
@@ -70,9 +70,12 @@
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-#
+
 html_theme = "furo"
 
+# custom title
+html_title = "Software version " + release
```
AleCandido, line 77: You can simplify even more:
```suggestion
html_title = "Version " + release
```
"Software" sounds a bit verbose and generic ("package" would be more specific, but just "version" is already enough).

PR #751, opened 2022-12-29
rm Ubuntu font from furo configuration

MatteoRobbiati:
After a discussion I had with the @qiboteam/coredevs, I propose to return to the original Furo font for the time being. 
Right now it is important to have a stable and presentable version of the documentation by the release (in a few days). We propose to address the restoration of the style in more detail at a later date.

If you agree with this, I propagate this choice in `qibolab` and `qibocal` too.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -81,8 +81,6 @@ html_theme_options = {
         "color-brand-primary": "#6400FF",
         "color-brand-secondary": "#6400FF",
         "color-brand-content": "#6400FF",
-        "font-stack": "Ubuntu",
-        "font-stack--monospace": "Courier, monospace",
     }
 }
 


scarrazza (approved):

scarrazza:
Correct me if wrong, but the https://qibo.science/docs/qibo/latest is still using ubuntu font?

AleCandido:
> Correct me if wrong, but the https://qibo.science/docs/qibo/latest is still using ubuntu font?

That one is also still using the rtd site, while GitHub Pages is somehow not working https://qibo.science/qibo/, even though the workflow is successful https://github.com/qiboteam/qibo/actions/runs/3799592100...

scarrazza:
Sorry, but then I am confused, the docs are rendered in 2 places: https://qibo.science/qibo/latest/ and https://qibo.science/docs/qibo/latest. Shall we get rid of the second?

AleCandido:
> Shall we get rid of the second?

The second is the "official" one (the one with the navbar), while the first is the navbar-less.
We can't get rid of the first one, because the second is an iframe to the first. We could give up on the second, but we loose the navbar.

The proper solution would to directly apply the global style also to the sphinx docs (by compiling the site altogether or sharing components and layouts), but we will work on this later on.

scarrazza:
This is clear, I was considering the link update just opened here: https://github.com/qiboteam/qiboteam.github.io/pull/15

Issue #718, opened 2022-11-30
Backend issues with DepolarizingChannel


stavros11:
There are two issues with the `DepolarizingChannel`.

First, it is not backend agnostic. The [unitary of a circuit](https://github.com/qiboteam/qibo/blob/513d302ca20d01f49a6d9073198b7261a99e5dc5/src/qibo/gates/channels.py#L236) is calculated during initialization and this operation initializes the `GlobalBackend`. This is not in line with the fact that gates and circuits should be backend agnostic so that we can easily switch from simulation to hardware (we did a large refactoring for this). For this particular case, I think we can delegate the calculation done with the circuit to numpy which is the fastest choice when it comes to small array operations and is used for all other gate preparations.

Second, there is some performance issue when using multi-threading with qibojit. This can be benchmarked with the following script:
```py
import time
import qibo
from qibo import gates
from qibo.models import Circuit

nc = Circuit(3, density_matrix=True)
nc.add(gates.DepolarizingChannel((0, 1), 0.1))

qibo.set_backend("numpy")
start_time = time.time()
result = nc()
print(time.time() - start_time)
print()

qibo.set_backend("qibojit", platform="numba")
start_time = time.time()
result = nc()
print(time.time() - start_time)
```
In my laptop numpy takes 0.0005sec while qibojit around 0.7sec. If I force qibojit to use single thread using the enrvironment flag `OMP_NUM_THREADS=1` the time drops to 0.02sec. This issue appears only with `density_matrix=True`.

Multi-threading and qibojit are expected to be slower than numpy when simulating small circuits (less than 10 qubits) but not that slower. This issue affects #702 because it makes qibojit tests noticeably slow.

stavros11:
Reopening because the first point (backend independence) of this was solved in #725, however the performance issues with qibojit multithreading should remain. I do not think it is very urgent though.

PR #756, opened 2023-01-05
Depolarizing channel

AlejandroSopena:
This PR implements a fix for issue #718. Debugging the code, I found that the bottleneck was in
https://github.com/qiboteam/qibojit/blob/df5f2360bf89dace3bf6833cde7f0d68999da94f/src/qibojit/backends/cpu.py#L182
The depolarizing channel is given by 
$\mathcal{E}(\rho)=(1-\lambda)\rho+\frac{\lambda}{2^n}I=(1-\frac{(4^n-1)\lambda}{4^n})\rho+\frac{\lambda}{4^n}{\sum}_{j=2}^{4^n}P_j\rho P_j^t$
 where $P_j$ denotes all products of $n$ Pauli matrices and the identity $(P_1=I^{\otimes n})$.
The channel is currently implemented as a sum of Kraus operators which allows it to be used with state vectors and density matrices. However, when running density matrix circuits with qibojit this implementation turns out to be too slow.
So I have implemented the channel as $(1-\lambda)\rho+\frac{\lambda}{2^n}I$ when using density matrices and as a sum of kraus operators for state vectors. This allows the channel to be used in both cases while solving the performance issue with qibojit. This can be benchmarked with:

```py
import time
import qibo
from qibo import gates

nc = Circuit(3, density_matrix=True)
nc.add(gates.DepolarizingChannel((0,1,), 0.1))

qibo.set_backend("numpy")
start_time = time.time()
result = nc()

qibo.set_backend("qibojit", platform="numba")
start_time = time.time()
result = nc()

qibo.set_backend("qibojit", platform="numba")
qibo.set_threads(1)
start_time = time.time()
result = nc()
```
In my computer numpy takes 0.0002 sec, qibojit (multi-threading) 0.001 sec, qibojit (single-thread) 0.0001 sec.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/gates/channels.py
+++ b/src/qibo/gates/channels.py
@@ -303,21 +303,16 @@ class DepolarizingChannel(Channel):
 
     def apply_density_matrix(self, backend, state, nqubits):
         lam = self.init_kwargs["lam"]
-        return (1 - lam) * backend.cast(
-            state
-        ) + lam * backend.apply_gate_density_matrix(
-            I(*self.target_qubits), state, nqubits
-        )
+        return (1 - lam) * backend.cast(state) + lam / 2**nqubits * I(
+            *range(nqubits)
+        ).asmatrix(backend)
 
     def apply(self, backend, state, nqubits):
-        from qibo.backends import NumpyBackend
-
         num_qubits = len(self.target_qubits)
         num_terms = 4**num_qubits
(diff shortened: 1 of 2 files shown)


scarrazza:
@AlejandroSopena thank you, works for me and looks good.

PR #759, opened 2023-01-11
Fix to adiabatic example

scarlehoff:
I was trying to use the code in https://github.com/qiboteam/qibo/tree/master/examples/adiabatic and wasn't working due to the usage of `.numpy()` and how the callbacks were being called. I guess this went unnoticed because it is missing from the [examples](https://github.com/qiboteam/qibo/blob/master/examples/test_examples.py)

I made some minimal changes so that it works for me but let me know if I'm using the correct functions/options (for instance, the `callback.apply` looks a bit cumbersome to me, but reading the code seemed like the right one for the use case).

Checklist:
- [x] Tests are passing.

### DIFF
--- a/examples/adiabatic/linear.py
+++ b/examples/adiabatic/linear.py
@@ -46,7 +46,7 @@ def main(nqubits, hfield, T, dt, solver, save):
     target_energy = bac.to_numpy(h1.eigenvalues()[0]).real
 
     # Check ground state
-    state_energy = bac.to_numpy(callbacks.Energy(h1).apply(bac, target_state)).real
+    state_energy = bac.to_numpy(h1.expectation(target_state)).real
     np.testing.assert_allclose(state_energy.real, target_energy)
 
     energy = callbacks.Energy(h1)
--- a/examples/adiabatic/linear.py
+++ b/examples/adiabatic/linear.py
@@ -1,5 +1,6 @@
 """Adiabatic evolution for the Ising Hamiltonian using linear scaling."""
 import argparse
+from pathlib import Path
 
 import matplotlib
 import matplotlib.pyplot as plt
@@ -38,13 +39,14 @@ def main(nqubits, hfield, T, dt, solver, save):
     """
     h0 = hamiltonians.X(nqubits)
     h1 = hamiltonians.TFIM(nqubits, h=hfield)
+    bac = h1.backend
 
     # Calculate target values (H1 ground state)
     target_state = h1.ground_state()
-    target_energy = h1.eigenvalues()[0].numpy().real
+    target_energy = bac.to_numpy(h1.eigenvalues()[0]).real
 
     # Check ground state
-    state_energy = callbacks.Energy(h1)(target_state).numpy()
+    state_energy = bac.to_numpy(callbacks.Energy(h1).apply(bac, target_state)).real
     np.testing.assert_allclose(state_energy.real, target_energy)
 
     energy = callbacks.Energy(h1)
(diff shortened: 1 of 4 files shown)


scarrazza:
@stavros11 could you please have a look at this? This example should match the code we have in the docs.

examples/adiabatic/linear.py
```diff
@@ -38,13 +39,14 @@ def main(nqubits, hfield, T, dt, solver, save):
...
+    bac = h1.backend
 
     # Calculate target values (H1 ground state)
     target_state = h1.ground_state()
-    target_energy = h1.eigenvalues()[0].numpy().real
+    target_energy = bac.to_numpy(h1.eigenvalues()[0]).real
 
     # Check ground state
-    state_energy = callbacks.Energy(h1)(target_state).numpy()
+    state_energy = bac.to_numpy(callbacks.Energy(h1).apply(bac, target_state)).real
```
stavros11, line 49: > for instance, the `callback.apply` looks a bit cumbersome to me

I agree, it is not very good. I believe 
```suggestion
    state_energy = bac.to_numpy(h1.expectation(target_state)).real
```
would work in this case, right?

examples/adiabatic/linear.py
```diff
@@ -38,13 +39,14 @@ def main(nqubits, hfield, T, dt, solver, save):
...
     np.testing.assert_allclose(state_energy.real, target_energy)
```
stavros11, line 50: Also, if you change this to
```suggestion
    h1.backend.assert_allclose(state_energy, target_energy)
```
I think you can avoid the `bac.to_numpy` calls.

stavros11 (commented):
Thanks for catching this, we must have missed it during the backend refactoring.

> I guess this went unnoticed because it is missing from the [examples](https://github.com/qiboteam/qibo/blob/master/examples/test_examples.py)

In that regard, could you please add a test to make sure we don't break it again in the future? I think you can follow the pattern of other example tests.

> @stavros11 could you please have a look at this? This example should match the code we have in the docs.

This is an example so in our current docs only the [README](https://qibo.readthedocs.io/en/stable/code-examples/tutorials/adiabatic/README.html) appears. There is a link to the code at the top of the page but this redirects back to github, which will be updated once we merge this. I don't think we need to update something else.

examples/adiabatic/linear.py
scarlehoff: How can I do the same thing for the `Overlap` callback?

examples/adiabatic/linear.py
scarlehoff: While for the `state_energy` I can do without, the `target_energy` will need the `to_numpy`/`.real` anyway due to it being used later on by matplotlib so that it works for any backend.

Any suggestions? (I can do `state_energy` without `to_numpy` and `target_energy` with for instance).

stavros11: It should be fine to leave `to_numpy` then. Indeed the plotting might fail for the cupy backend if we don't convert to numpy first, so it should be safe to use `to_numpy`.

examples/adiabatic/linear.py
stavros11: There is no, so I think for the overlap only the current solution with `apply` works. Alternatively, you could also just calculate the overlap directly:
```py
np.dot(np.conj(state1), state2)
```
but if the current solution works there is no issue.

stavros11:
Sorry for the delay in looking at this. I believe it should be ready to be merged, right?

scarlehoff:
> I believe it should be ready to be merged, right?

As far as I am aware, yes (i.e., the example works and @MatteoRobbiati and I are right now using what is basically an extension of the example)

MatteoRobbiati:
> As far as I am aware, yes (i.e., the example works and @MatteoRobbiati and I are right now using what is basically an extension of the example) 

Yes, I agree. It can be merged.

scarrazza (approved):
@scarlehoff @MatteoRobbiati thank you.

PR #770, opened 2023-01-24
Workflow migration

Edoardo-Pedicillo:
This PR modify the workflows in order to use the ones stored in `qiboteam/workflows` (see qiboteam/workflows#23).

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/.github/workflows/rules.yml
+++ b/.github/workflows/rules.yml
@@ -19,6 +19,6 @@ jobs:
       os: ${{ matrix.os }}
       python-version: ${{ matrix.python-version }}
       environment: "qibo"
-      pip-flags: "tests,docs"
-      test-flag: ${{ matrix.os == 'ubuntu-latest'}}
+      pip-extras: "tests,docs"
+      doctests: ${{ matrix.os == 'ubuntu-latest'}}
     secrets: inherit
--- a/.github/workflows/rules.yml
+++ b/.github/workflows/rules.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   build:
-    if: contains(github.event.pull_request.labels.*.name, 'run-workflow') || github.event_name == 'push'
+    #if: contains(github.event.pull_request.labels.*.name, 'run-workflow') || github.event_name == 'push'
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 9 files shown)


Edoardo-Pedicillo:
@AleCandido the coverage is dramatically decreased. I supposed the problem is with `pylint `implementation: in the original workflow there were two `pylint` calls and here there is only one: can we solve the problem with a branching ?

.github/workflows/rules.yml
```diff
@@ -2,55 +2,23 @@
...
+  workflow_dispatch:
   push:
   pull_request:
     types: [labeled]
 
-
 jobs:
   build:
-    if: contains(github.event.pull_request.labels.*.name, 'run-workflow') || github.event_name == 'push'
+    #if: contains(github.event.pull_request.labels.*.name, 'run-workflow') || github.event_name == 'push'
```
AleCandido, line 12: If this is for debugging purpose, add a TODO comment to remove before merging

AleCandido (changes requested):

.github/workflows/rules.yml
Edoardo-Pedicillo: I commented this line to let the workflow dispatch work. Now this if condition is restored and the workflow dispatch trigger deleted.

AleCandido:
#775 should solve the issue with `pre-commit`

AleCandido:
I rebased on current `master`, the `pre-commit` problem should be now solved.

AleCandido:
> @AleCandido the coverage is dramatically decreased. I supposed the problem is with `pylint `implementation: in the original workflow there were two `pylint` calls and here there is only one: can we solve the problem with a branching ?

Pylint does not generate coverage, so that is not the problem. However, concerning this you should add back the second call, without branching.
This means there will be a further call for `qibolab`, but this is fine and even a good thing on its own.

pyproject.toml
```diff
@@ -0,0 +1,19 @@
+[tool.pylint.master]
+
+exit-zero = true
+errors-only = true
```
AleCandido, line 4: Drop this section, this will be included as CLI argument in the workflow, and it is the content of the two separate calls to Pylint:
1. the first one is `errors-only`, and it should make the CI crash in case there is an actual error
2. the second one includes also the warnings, but it has to be `exit-zero`, since you do not want to make tests failing for a Pylint warning

AleCandido (changes requested):

pyproject.toml
AleCandido:
```suggestion
```

AleCandido:
Just waiting on #776

scarrazza:
Looks good to me, shall we merge?

AleCandido (approved):

PR #778, opened 2023-02-03
Fix hamiltonian models neighbour and noise model tests/lint

### LINKED ISSUE #774: `hamiltonians.TFIM` must involve more than 1 qubit
I think it should be great to raise an error in case a user decide to use a `hamiltonians.TFIM` model with only one qubit. I think the transverse field Ising model requires the nearest neighbour interactions to be implemented.

scarrazza:
Closes #774 and fixes pylint and failing tests for noise model (after #672 merge).

Checklist:
- [ ] Tests are passing.


AlejandroSopena (approved):
Thanks Stefano. I did not realise that with the new workflows pylint was failing for the noise model test. Everything looks good.

Issue #788, opened 2023-02-13
Adding conditions to NoiseModel


vodovozovaliza:
**Is your feature request related to a problem? Please describe.**
The gate-dependent `NoiseModel` only works for gates with the same gate type (Unitary, RX, RY etc.), but sometimes this is not enough.

**Describe the solution you'd like**
I would like to propose to add a condition to the `NoiseModel` class which would add an error only if a condition specified by a user is satisfied.

**Example**
Current `add()` method does not allow to add different errors to $R_X(\pi/2)$ and $R_X(3\pi/2)$. A user could create a function `condition(gate)` that would return `True` if the noise should be added to a specific gate in the circuit.

PR #789, opened 2023-02-13
Noise conditions

vodovozovaliza:
This PR would fix the following issues:

- the conditional error described in #788.

- the possibility to add multiple errors to the same gate.

- the possibility to add noise after each gate in the circuit (gate-independent noise).

Checklist:
- [x] Tests are passing.


vodovozovaliza:
This is related to issue #788. I added a `condition` parameter to the NoiseModel `add()` method and modified `self.errors` so that a user could add multiple errors to the same gate class.

src/qibo/noise.py
```diff
@@ -173,12 +175,31 @@ def add(self, error, gate, qubits=None):
...
+                raise TypeError(
+                    "condition should be callable. Got {} instead."
+                    "".format(type(condition))
+                )
+        else:
+            if gate not in self.errors:
+                self.errors[gate] = ([error], [qubits])
+            else:
+                self.errors[gate][0].append(error)
+                self.errors[gate][1].append(qubits)
```
stavros11, line 202: Here something like:
```py
self.errors[gate].append((error, qubits))
```
Btw, in this approach if you initialize as `self.errors = collections.defaultdict(list)` instead of the standard dict you also don't need the `if gate not in self.errors` check.

src/qibo/noise.py
```diff
@@ -222,25 +243,58 @@ def apply(self, circuit):
...
-                    elif isinstance(error, DepolarizingError) and qubits:
-                        noisy_circuit.add(error.channel(qubits, *error.options))
-                    elif isinstance(error, UnitaryError) or isinstance(
-                        error, KrausError
-                    ):
-                        if error.rank == 2:
+                    error_list, qubits_list = self.errors.get(gate.__class__)
+                    for i in range(len(error_list)):
+                        error = error_list[i]
+                        qubits = qubits_list[i]
```
stavros11, line 251: And here
```py
for error, qubits in self.errors.get(gate.__class__)
```

stavros11 (commented):
Thanks for opening the issue and this PR. @andrea-pasquale @vodovozovaliza correct me if I am wrong, because I am not very familiar with the `NoiseModel`, but if I get it correctly this PR fixes two issues:
* the possibility to add multiple errors to the same gate (am I right that this is not possible now?)
* the conditional error described in #788.

Do you think that the code would be simpler if instead of a tuple of lists you use a list of tuples for `errors` and `conditions`. See comments below for example.

Either way, it would be useful to add an example of how the new condition argument is used, maybe in the how-to examples in the docs.

vodovozovaliza:
@stavros11 thank you for your comments, it is a good idea to use `collections.defaultdict(list)`, I changed `errors` and `conditions` adding lists of tuples and added an example of a condition in `add()` method. 
Yes, this PR would fix both issues that you mentioned:
- Multiple errors for the same gate.
- Conditional errors described in #788.

src/qibo/noise.py
```diff
@@ -173,12 +177,50 @@ def add(self, error, gate, qubits=None):
...
+            noise.add(PauliError(px = 0.5), gates.RX, condition=is_sqrt_x)
+
+            # Generate a noiseless circuit.
+            circuit = Circuit(1)
+            circuit.add(gates.RX(0, np.pi / 2))
+            circuit.add(gates.RX(0, 3 * np.pi / 2))
+            circuit.add(gates.X(0))
+
+            # Apply noise to the circuit.
+            circuit = noise.apply(circuit)
```
andrea-pasquale, line 207:
```suggestion
            noisy_circuit = noise.apply(circuit)
```
I would rename this variable `noisy_circuit` just to avoid losing the original circuit.

src/qibo/noise.py
```diff
@@ -158,10 +161,11 @@ class NoiseModel:
     """
 
     def __init__(self):
-        self.errors = {}
+        self.errors = collections.defaultdict(list)
+        self.conditions = collections.defaultdict(list)
```
andrea-pasquale, line 165: I think that you can do everything just by using a single collection (for example `errors`). 
See comment below.

src/qibo/noise.py
```diff
@@ -173,12 +177,50 @@ def add(self, error, gate, qubits=None):
...
+        if condition is not None:
+            if callable(condition):
+                self.conditions[gate].append((condition, error, qubits))
+            else:
+                raise TypeError(
+                    "condition should be callable. Got {} instead."
+                    "".format(type(condition))
+                )
+        else:
+            self.errors[gate].append((error, qubits))
```
andrea-pasquale, line 223:
```suggestion
        if condition is not None and not callable(condition):
            raise TypeError(
                    "condition should be callable. Got {} instead."
                    "".format(type(condition))
                )
        else:
            self.errors[gate].append((condition, error, qubits))
```

src/qibo/noise.py
```diff
@@ -222,25 +264,50 @@ def apply(self, circuit):
...
+                            error, KrausError
+                        ):
+                            if error.rank == 2:
+                                for q in qubits:
+                                    noisy_circuit.add(error.channel([q]))
+                            elif error.rank == 2 ** len(qubits):
+                                noisy_circuit.add(error.channel(qubits))
+                        else:
+                            for q in qubits:
+                                noisy_circuit.add(error.channel(q, *error.options))
```
andrea-pasquale, line 288:
```suggestion
                    for condition, error, qubits in self.errors.get(gate.__class__):
                        if condition is None or condition(gate):
                            if qubits is None:
                                qubits = gate.qubits
                            else:
                                qubits = tuple(set(gate.qubits) & set(qubits))
                            if isinstance(error, CustomError) and qubits:
                                noisy_circuit.add(error.channel)
                            elif isinstance(error, DepolarizingError) and qubits:
                                noisy_circuit.add(error.channel(qubits, *error.options))
                            elif isinstance(error, UnitaryError) or isinstance(
                                error, KrausError
                            ):
                                if error.rank == 2:
                                    for q in qubits:
                                        noisy_circuit.add(error.channel([q]))
                                elif error.rank == 2 ** len(qubits):
                                    noisy_circuit.add(error.channel(qubits))
                            else:
                                for q in qubits:
                                    noisy_circuit.add(error.channel(q, *error.options))
```
 If you do this then you can remove the following if statement.

andrea-pasquale (commented):
Thanks @vodovozovaliza for implementing this.
Overall this PR looks in a good shape to me.
You can find some comments below on how to use a single collection implement this new feature. I think that this should also fix coverage.
Let me know if with this approach everything works as expected.

As @stavros11 was suggesting I think that it would be useful to add an example directly in the documentation describing how to use this new feature.
You will need to edit this file https://github.com/qiboteam/qibo/blob/noise_conditions/doc/source/code-examples/advancedexamples.rst#using-a-noise-model under the section `Using a noise model`.

vodovozovaliza:
Thank you @andrea-pasquale, I added your suggestions and an example to the documentation.

vodovozovaliza:
@andrea-pasquale @stavros11 this PR is ready for review

PR #797, opened 2023-02-17
Fix `test_choi_to_kraus()` randomly crashing

renatomello:
Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/tests/test_quantum_info_superoperator_transformations.py
+++ b/src/qibo/tests/test_quantum_info_superoperator_transformations.py
@@ -216,8 +216,8 @@ def test_choi_to_kraus(order, validate_CP):
     test_evolution_a0 = test_a0 @ state @ test_a0.T.conj()
     test_evolution_a1 = test_a1 @ state @ test_a1.T.conj()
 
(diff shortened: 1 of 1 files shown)


renatomello:
Since `choi_to_kraus()` depends on a spectral decomposition (or SVD), sometimes the numerical precision becomes 1-3% above `qibo.config.PRECISION_TOL`. Therefore, I increased precision for those tests and they should be fine now.

renatomello:
This is ready for merge after review @AleCandido @andrea-pasquale

AleCandido (commented):
Fine by me, but are you sure is it okay with increasing the tolerance by a factor of 10?

In particular, if raising tolerance is the chosen solution, a factor of 2 would have been sufficient.

renatomello:
> Fine by me, but are you sure is it okay with increasing the tolerance by a factor of 10?
> 
> In particular, if raising tolerance is the chosen solution, a factor of 2 would have been sufficient.

Changed it.

andrea-pasquale (approved):
Thanks @renatomello.

PR #798, opened 2023-02-17
Move tests top-level

AleCandido:
~~This PR is not meant to be merged, but it is here to probe how the workflows react to a few changes.~~

Even though this started differently, now some potentially useful work has been done. It is still not the current focus, but I would consider merging this at a later stage.

Checklist:
- [x] Tests are passing.


AleCandido:
It seems undocumented, but `--pyargs` somehow affects the coverage
https://pytest-cov.readthedocs.io/en/latest/search.html?q=pyargs

In particular, the `--pyargs qibo` combination, which means:
1. interpret what to run as Python packages https://docs.pytest.org/en/stable/example/pythoncollection.html#interpreting-cmdline-arguments-as-python-packages
2. use the `qibo` package

I have the feeling that point 2. means to go recursively inside the `qibo` package and discover tests, but since they are all inside `src/qibo/tests/` they should be already found by the `testpaths` argument.

It is true that having a look at the logs the tests run are not the same:

https://github.com/qiboteam/qibo/actions/runs/4206201628/jobs/7299362861#step:6:239
https://app.codecov.io/gh/qiboteam/qibo/commit/5bd3a09f157a564b62f59a7d286e6cd1b0768f6c
https://github.com/qiboteam/qibo/actions/runs/4204926452/jobs/7296284208#step:6:244
https://app.codecov.io/gh/qiboteam/qibo/commit/b5389e5d2132f17b86ac6c05e93d9fd48c4eafce

So, I'm now trying to understand which test have been skipped in one that are not in the other.

AleCandido:
These are the further tests skipped, between one and the other:

https://github.com/qiboteam/qibo/actions/runs/4204926452/jobs/7296284208#step:6:69
https://github.com/qiboteam/qibo/actions/runs/4204926452/jobs/7296284208#step:6:98

AleCandido:
I take it back: I messed up, the difference in the number of tests was only because I was looking at MacOS in one of the two cases

full coverage:
https://github.com/qiboteam/qibo/actions/runs/4206201628/jobs/7299362861#step:6:239
[app.codecov.io/gh/qiboteam/qibo/commit/5bd3a09f157a564b62f59a7d286e6cd1b0768f6c](https://app.codecov.io/gh/qiboteam/qibo/commit/5bd3a09f157a564b62f59a7d286e6cd1b0768f6c)

20% coverage:
https://github.com/qiboteam/qibo/actions/runs/4204926452/jobs/7296285544#step:6:245
[app.codecov.io/gh/qiboteam/qibo/commit/b5389e5d2132f17b86ac6c05e93d9fd48c4eafce](https://app.codecov.io/gh/qiboteam/qibo/commit/b5389e5d2132f17b86ac6c05e93d9fd48c4eafce)

AleCandido:
Just to make it clear, keeping `--pyargs qibo` is not an option: we would not be able to run `pytest <path>` any longer, since it would always run on the full `qibo` package.

AleCandido:
I tried to move tests top-level, to see if I could get something closer to `qibolab`.
However, the answer is that the coverage raised only because it is not counting tests themselves any longer (which make sense, but it was not the main goal).

@scarrazza I have the feeling that the coverage we see is the one coming from the `examples/`, since they are the last ran in the workflows (it is similar to, but it is not, the one from the doctests).
So, my current guess is that the `coverage.xml` file gets overwritten in the following order:
1. by tests
2. by doctests
3. by examples

And only the final one is uploaded.

The same thing is not happening in `qibolab` because by default deselected:
https://github.com/qiboteam/qibo/blob/3557e4442b9306bc3d764d614eddce255b9b53f9/.github/workflows/rules.yml#L23
https://github.com/qiboteam/workflows/blob/87b215c3ce55ca4bc8a5f385ac7a9ed9c27ed7b1/.github/workflows/rules.yml#L19-L22
https://github.com/qiboteam/qibolab/blob/1ae75a8a21aef417f589cf95325edcac0b9d135a/.github/workflows/rules.yml#L15-L19

I still do not have a full understanding of why it is not happening in `qibo@master`, but it is clear to be related to `--pyargs`.

There are three possible solutions now:
1. restore `--pyargs` and `qibo` arguments, and accept that is not possible to run test individually with `pytest <path>`
2. shuffle the workflow, such that the last report produced is the one from unit tests, and that one will be uploaded (equivalently: save it somewhere, and move back just before the upload)
3. upload multiple separate reports for the three sources
    - in particular, we can run the three of them in different jobs in the same workflow, after a common setup, and conclude each job with an upload
    - we could also interleave Codecov uploads, but since they are prone to fail, I prefer to keep them at the end of each job, and set `fail-fast: false`, such that if an upload fails we still collect the full information

stavros11:
Just intervening to say that

> 1. restore `--pyargs` and `qibo` arguments, and accept that is not possible to run test individually with `pytest <path>`

I would like to avoid this option, because running individual tests is very useful when developing. Qibo tests are very slow as a whole, but when I am making a change I usually know which tests it is more likely to affect, so I begin with checking this. This speeds up the development process a lot in my opinion. In fact, right now I have the pyproject.toml modified locally (without commiting) so that I can run individual tests.

> 2. shuffle the workflow, such that the last report produced is the one from unit tests, and that one will be uploaded (equivalently: save it somewhere, and move back just before the upload)
> 3. upload multiple separate reports for the three sources

Other than not liking 1, I don't have a strong opinion. However, are you sure that the coverage.xml is overriden? I had the impression that if you run tests in a directory that already has coverage.xml, then the two reports are appended automatically and you get the total coverage, but I have not explored this much so I may be wrong.

renatomello:
> I would like to avoid this option, because running individual tests is very useful when developing. Qibo tests are very slow as a whole, but when I am making a change I usually know which tests it is more likely to affect, so I begin with these. This speeds up the development process a lot in my opinion. In fact, right now I have the pyproject.toml modified locally (without commiting) so that I can run individual tests.

I second all of the above. I'm also having to maintain a local modified version of `pyproject.toml`.

AleCandido:
> I would like to avoid this option, because running individual tests is very useful when developing. Qibo tests are very slow as a whole, but when I am making a change I usually know which tests it is more likely to affect, so I begin with these. This speeds up the development process a lot in my opinion. In fact, right now I have the pyproject.toml modified locally (without commiting) so that I can run individual tests.

Sorry, I later modified the answer, but it has not been uploaded (maybe I was disconnected...).

In the update, I was saying that 1. is not really a solution, but just a conclusion, giving up on the original issue. I'm perfectly aware it was not ideal (often I sort the alternatives in reverse-preference order, this was the case...), but at some point we need a conclusion, and of course it is also an option...

About 2. and 3., I was saying that I'm not sure about them. I tested locally, and that's the explanation I found.
This is confirmed by the existence of [`--cov-append`](https://pytest-cov.readthedocs.io/en/latest/config.html#reference) (Default: False) that I just discovered, and I'm going to try immediately :)

renatomello:
If the issue was with the lack of `--cov-append`, do we still need to move the location of the `tests` folder?

renatomello (approved):

AleCandido:
> If the issue was with the lack of `--cov-append` in `addopts`, do we still need to move the location of the `tests` folder?
> Though I do think it looks cleaner this way.
> 
> Either way, I'd merge this and #796 to `master` as soon as possible.

In principle we didn't, that's why I kept as a separate PR, and we could have even merged after merging the other one in master (but it's also fine this way).

A couple of reasons to motivate this change are:
- we avoid distributing tests in the wheels (packages) since the user doesn't need them
- we avoid measuring the coverage on the tests themselves, that was artificially increasing the percentage (of course with 100% it is mildly relevant)

renatomello:
I agree with everything

PR #799, opened 2023-02-17
Measurement in Pauli basis

### LINKED ISSUE #720: Mesurement and initialization in Pauli basis
I would like to suggest including a feature for measuring in a different Pauli basis besides the computational basis, i.e. include an argument in the Measurement class `qibo.gates.M(..., basis,...)` such that

> If basis = 'Z', pass
> If basis = 'X', add a Hadamard gate automatically to rotate the state to the X basis
> If basis = 'Y', add a Haddamard + S gate automatically to rotate the state to the Y basis

That would be very useful for several applications: computing expectation values "manually" (useful for educational purposes, closer to the experimental treatment of the quantum computing output) or quantum circuit cutting (which requires measuring and preparing the states on a particular basis)

Following this idea, I would also suggest automatically initializing all qubits in the |+>, |->, |1>, |+i>, |-i>, etc states. e.g. including some function `initialize(basis=X, Y, Z, eigenstate = plus, minus)` or similar.

stavros11:
Fixes #720. I am copying here the TODO list from https://github.com/qiboteam/qibo/issues/720#issuecomment-1344077029

- [X] add a `basis` argument for `gates.M` which accepts gate objects (instead of strings).
- [x] modify circuit execution to accept initial states of two types: arrays and circuits (to be able to execute on hardware).
- [ ] provide a set of function initializers for most common states, e.g. |+>, |->, |1>, |+i>, |-i> among others, using circuits.

I implemented the first point. @andrea-pasquale @Edoardo-Pedicillo let me know if you would like to have a look at any of the others.

Note that as discussed using the `basis` argument in the measurement gate will add additional gates that rotate to this basis before the measurement. For example:
```py
c = Circuit(1)
c.add(gates.M(0, basis=gates.X)
```
will result to
```
0: --H--M
```

Also, I implemented the basis rotation only for the Pauli gates (X, Y, Z) as this was requested in the original issue, however it may be possible to do it for any gate. Essentially for a single qubit gate G we need to solve U^dagger Z U = G for U.

Checklist:
- [x] Tests are passing.


andrea-pasquale (approved):

renatomello:
Does this also work for when a multi-qubit measurement `M` object is created?

Also, it would be good to pass just a string instead of the `gate` object, as in
```python3
c = Circuit(1)
c.add(gates.M(0, basis="X")
```

stavros11:
After discussing with @renatomello, extending this to support a different basis for each qubit measured could be useful. In summary this could accept:
* a single gate class to be applied on all qubits (already implemented),
* a list of gate objects that has the same length as the measured qubits and specifies the basis for each qubit (TODO).

The string representation proposed in the comment above could also be good as it is more compact, however I am a bit against supporting many different types as inputs to a function as it is harder to implement, maintain and document and as a result could also be more confusing for users. With the above addition we already accept two types (gate class or list) and if we add strings we go up to four which could be complicated to handle.

 @scarrazza let me know what you think.

scarrazza:
I would suggest to keep the gate object and list, but avoid the string because we don't use this interfacing mechanism (string-gate) anywhere in qibo, and it is easier to debug for the user.

scarrazza:
Following the discussion today, we agreed to add tests and merge.

stavros11:
> Following the discussion today, we agreed to add tests and merge.

I fixed the conflicts and I also implemented the list for basis as I realized it was a quick update.

@renatomello if you like give a try and let me know if it is what you had in mind. Your example should be modified to
```py
c = Circuit(2)
c.add(gates.M(0, 1, basis=[gates.X, gates.Y])
```
since we decided to not accept lists.

renatomello:
> > Following the discussion today, we agreed to add tests and merge.
> 
> I fixed the conflicts and I also implemented the list for basis as I realized it was a quick update.
> 
> @renatomello if you like give a try and let me know if it is what you had in mind. Your example should be modified to
> 
> ```python
> c = Circuit(2)
> c.add(gates.M(0, 1, basis=[gates.X, gates.Y])
> ```
> 
> since we decided to not accept strings.

I tried this code
```python3
from qibo import gates

# %%
nqubits = 2
circuit = Circuit(nqubits)

basis = "XY"
basis = [getattr(gates, pauli) for pauli in basis]
circuit.add(gates.M(*range(nqubits), basis=basis))

print(circuit.draw())
```
and the gates are being added after the measurement.

stavros11:
Thanks for checking, should be fixed now.

renatomello:
It works now.

renatomello (approved):

Issue #784, opened 2023-02-06
Another circuit .copy() bug


wilkensJ:
After copying the original circuit, when executing the copy, the original circuit is altered somehow.
The pickle error 
``` txt
TypeError: cannot pickle 'module' object
```
occurs after the copy is executed. That indicates that a module was added to the original circuit, even though it was not touched.

``` python
from qibo import gates, models
import pickle
nqubits = 3
circuit = models.Circuit(nqubits)
circuit.add(gates.X(0))
circuit.add(gates.H(1))
circuit.add(gates.M(*range(nqubits)))
with open('test_deletethis.pkl', "wb") as f:
    pickle.dump(circuit, f)
print('------ copy circuit --------')
copied_circuit = circuit.copy(deep = True)
copied_circuit.execute(nshots = 10)
with open('test_deletethis2.pkl', "wb") as f:
    pickle.dump(circuit, f) # -> Error
```

wilkensJ:
This bug makes it not possible to execute the copy of a circuit, retrieve the results and then store the original circuit (which was not execute but its deep copy). In `qibocal` this is necessary for some protocols to post process the execution result via `qq`.

PR #804, opened 2023-02-24
Fix pickle of backend objects

stavros11:
As indirectly mentioned in #784, currently it is not possible to pickle circuits after they have been executed. For example
```py
import pickle
from qibo import gates

circuit = Circuit(2)
circuit.add(gates.H(0))
circuit.add(gates.H(1))

result = circuit()

with open('test_deletethis.pkl', "wb") as file:
    pickle.dump(circuit, file)
```
will fail with 
```
TypeError: cannot pickle 'module' object
```
This happens because when the circuit is executed, the `_final_state` attribute is set to a `CircuitResult` object which contains a reference to a `Backend` object and all backends contain a reference to numpy which is a module and cannot be pickled. 

Here I fixed the pickle of `Backend` objects by implementing [`__reduce__`](https://docs.python.org/3/library/pickle.html#object.__reduce__). I am not sure if that is a good solution but it seems to work.

@wilkensJ note that this does not fix the issues related to the `.copy` you mention in your issues, but it should fix the issues related to pickles.

Checklist:
- [ ] Tests are passing.


AleCandido:
Ideally, I would like more to use some kind of serialization library, to save attributes with a *simple* type in a readable format (you can think JSON/YAML/...), and array as binary blobs.

A couple of these libraries are:
- https://docs.pydantic.dev/usage/exporting_models/
- https://catt.rs/en/latest/index.html

Both of them require using a specific class definition, but they are both very compatible with dataclasses ([`attrs`](https://www.attrs.org/) is where dataclasses [come from](https://www.attrs.org/en/stable/why.html#data-classes), and Pydantic models can be defined as [`pydantic.dataclasses`](https://docs.pydantic.dev/usage/dataclasses/)), so this will be easier to obtain at a later stage.

For the time being, Pickle is fine, and your solution should be good enough (but in the longer term it is better to be dropped, especially if you want to allow people to exchange circuits, since Pickle shouldn't be used with untrusted input).
I'm assuming that the back-end has no state, it is just a convenient way to collect some methods.

AleCandido (approved):

AleCandido:
@stavros11 to improve consistency, in principle you should allow for arguments to be passed to some backends, in order to restore the minimal state they have, e.g.
https://github.com/qiboteam/qibo/blob/892c2b8ef4161ae9d1c9c36a020c9a27ffe479a0/src/qibo/backends/numpy.py#L13-L14
https://github.com/qiboteam/qibo/blob/892c2b8ef4161ae9d1c9c36a020c9a27ffe479a0/src/qibo/backends/numpy.py#L33

So it should become:
```py
class NumpyBackend(Backend):
    def __init__(self, precision, ...):
		...

    def __reduce__(self):
        """Allow pickling backend objects that have references to modules."""
        return self.__class__, tuple(self.precision, ...)
```
such that Pickle will be able to restore the precision while loading.

However, as I argued above, according to me Pickle should not be the preferred long-term way, so it might not be worth to spend too much time on this.

scarrazza (approved):

PR #810, opened 2023-03-01
`ReadoutErrorChannel` class + move `KrausChannel` methods to `Channel`

renatomello:
Checklist:
- [x] Tests are passing.


renatomello:
This is necessary so all channel classes inherit these methods, which wasn't the case before for `DepolarizingChannel` and `ThermalRelaxationChannel`

renatomello:
This is failing tests because of a bug in `qibojit` that is fixed by https://github.com/qiboteam/qibojit/pull/110 and both should be merged together. @scarrazza

renatomello:
> This is failing tests because of a bug in `qibojit` that is fixed by [qiboteam/qibojit#110](https://github.com/qiboteam/qibojit/pull/110) and both should be merged together. All tests as passing locally. @scarrazza

scarrazza:
@renatomello could you please resolve the conflict?

renatomello:
> @renatomello could you please resolve the conflict?

Done

renatomello:
@scarrazza Tests are failing because they're installing `qibojit` version from this commit 5a4125a8b1cc7373228e4967f2276c81cc5d00da from Feb 7

renatomello:
> @scarrazza Tests are failing because they're installing `qibojit` from this commit [qiboteam/qibojit@5a4125a](https://github.com/qiboteam/qibojit/commit/5a4125a8b1cc7373228e4967f2276c81cc5d00da) from Feb 7 instead of the latest

scarrazza:
@AleCandido @Edoardo-Pedicillo is the workflow using the latest qibojit main or something else?

AleCandido:
> @AleCandido @Edoardo-Pedicillo is the workflow using the latest qibojit main or something else?

I believe the answer to be "yes", it is using the latest Qibojit `main`:

https://github.com/qiboteam/qibo/blob/86697d5210aa2f4caa55d833700c1f38978230e9/pyproject.toml#L56

Even though is a circular dependency, so I'm surprised there is no error...

Edoardo-Pedicillo:
The only "mistake" that I see is that in the `pyproject.toml` should be
`qibojit = { git = "https://github.com/qiboteam/qibojit.git" } ` 
I am trying to change it and verify if this is the problem.

renatomello:
> The only "mistake" that I see is that in the `pyproject.toml` should be `qibojit = { git = "https://github.com/qiboteam/qibojit.git" } ` I am trying to change it and verify if this is the problem.

This worked.

AleCandido:
> This worked, now it installed from the latest commit [qiboteam/qibojit@cb3dc89](https://github.com/qiboteam/qibojit/commit/cb3dc89e7936298dea79ba88b778e77e7555f408).

Ah, great...

I don't understand the logic, but good to know :)

Edoardo-Pedicillo:
Solved: see #825

renatomello:
> Solved: see #825

Already implemented the change here, so this closes #825

Edoardo-Pedicillo:
In [95967eb](https://github.com/qiboteam/qibo/pull/810/commits/95967eb41702135a76f7d5d3f1d38281021d7c4d)  the `toml` file is updated, but not the `lock` one, you have to run `poetry update` in your local repository and push the changes.

AleCandido:
Update: I now understood the logic, and the *problem* is not gone at all

https://github.com/qiboteam/qibo/pull/825/files#diff-f53a023eedfa3fbf2925ec7dc76eecdc954ea94b7e47065393dbad519613dc89R3144

I believe that `.git` makes no difference, the difference is actually in the lockfile: since @renatomello regenerated it, it recomputed the exact commit to use as the most recent one.
The workflow actually uses the lockfile for reproducibility, so if you do not update it, it will be stuck to the last time it has been generated.

Luckily, there are two proper solutions to this problem:
1. stop using Git dependencies, since is much more reliable to work with actual releases - if Qibo and Qibojit should always work together a "monorepo" would be a better choice\*
2. use Poetry in development: many people are still using Conda environments, while Poetry has its own; if you manually install with `pip` or with `conda`, your environment won't be reflected by the lockfile, and thus it is from hard to impossible for the workflow to reproduce it - please, use Poetry :)

Maybe, since we introduced Poetry "silently", we might need to discuss at a meeting how to use it.

\* a proper monorepo is not currently supported by Poetry, but we can distribute `qibojit` import-package in the same PyPI-package of Qibo, using extras for dependencies

renatomello:
> 2\. use Poetry in development: many people are still using Conda environments, while Poetry has its own; if you manually install with `pip` or with `conda`, your environment won't be reflected by the lockfile, and thus it is from hard to impossible for the workflow to reproduce it - please, use Poetry :)

this is weird because, incidentally and recklessly, I am using neither `conda` nor `poetry`

AleCandido:
> this is weird because, incidentally and recklessly, I am using neither `conda` nor `poetry` ( I know, I know...)

Using Conda is just one way of not using Poetry: if you're using `virtualenv` manually (that I quite like, and it was my favorite option before Poetry) or your global system environment (most likely the worst option), and installing manually with `pip`, still the lockfile doesn't get updated.

However, don't get me wrong: I don't want to scold anyone, apart from myself.
We introduced Poetry for a reason, but we omitted to explain it, and more important we didn't explain how to use it. That was definitely not the best course of action...

vodovozovaliza (approved):

PR #816, opened 2023-03-03
Poetry v2

AleCandido:
This branch has already been merged in #806. But the target branch was https://github.com/qiboteam/qibo/tree/pytest_one_file, and it happened after #796 was already merged.

In practice, there is nothing different from #806, but from the point of view of https://github.com/qiboteam/qibo/tree/master #806 never happened.

Checklist:
- [ ] Tests are passing.


AleCandido:
Since it has been merged, I thought it was ready. But actually tests are not passing.

The reason is not very clear, so it should be further investigated...
https://github.com/qiboteam/qibo/actions/runs/4321162932/jobs/7542108512#step:9:14

I'm trying to rerun to check if it was a GitHub problem.

AleCandido:
It seems it was more a GitHub problem, and now it is simply working, as it were in #806

PR #819, opened 2023-03-06
Depolarizing channel

AlejandroSopena:
This PR fixes a bug in the multi-qubit depolarizing channel when using density matrices. PR [109 ](https://github.com/qiboteam/qibojit/pull/109)is required to use the channel with `cupy`. The depolarizing channel is equivalent to a uniform Pauli channel when $\lambda=4n/(4n-1)$ where $n$ is the number of qubits. I have verified this result using the `GeneralizedPauliNoiseChannel`.

Checklist:
- [x] Tests are passing.


src/qibo/backends/numpy.py
```diff
@@ -85,6 +85,11 @@ def zero_density_matrix(self, nqubits):
         state[0, 0] = 1
         return state
 
+    def identity_density_matrix(self, nqubits):
+        state = self.np.eye(2**nqubits, 2**nqubits, dtype=self.dtype)
```
renatomello, line 89: the second `2**nqubits` is unnecessary here

src/qibo/backends/numpy.py
```diff
@@ -328,6 +333,36 @@ def thermal_error_density_matrix(self, gate, state, nqubits):
         state = self.apply_gate(gate, state.ravel(), 2 * nqubits)
         return self.np.reshape(state, shape)
 
+    def depolarizing_error_density_matrix(self, gate, state, nqubits):
```
renatomello, line 336: Maybe add one-line comments to explain what some of the variables are (e.g. `qubits_1`, `qubits_2`, `qs`) and why they are needed.

renatomello (commented):
Some comments on the code.

src/qibo/backends/numpy.py
```diff
@@ -85,6 +85,11 @@ def zero_density_matrix(self, nqubits):
         state[0, 0] = 1
         return state
 
+    def identity_density_matrix(self, nqubits):
```
stavros11, line 88: Since you are adding this as a backend method, it would be good to also add and document it in the `AbstractBackend`.

src/qibo/backends/numpy.py
AlejandroSopena: Those variables are used to write the tensor product $Tr_q(\rho)\otimes I^{\otimes n}$ in the correct order. First, I calculate the tensor product using `np.tensordot()` as if the depolarization channel acted on the last `len(q)` qubits, and then I change the order.

AlejandroSopena:
This should be merged together with [qiboteam/qibojit/#109](https://github.com/qiboteam/qibojit/pull/109).

renatomello (approved):

Issue #814, opened 2023-03-01
Review Qibo dependencies


AleCandido:
Starting with #806 we are treating dependencies more carefully, e.g. resolving compatible versions.

In particular, the following dependencies should be reconsidered:

- [`tabulate`](https://github.com/gregbanks/python-tabulate)
    - just a pretty printing library
    - not maintained since 2014
    - if needed, consider using [`rich`](https://github.com/Textualize/rich) and its [`rich.table.Table`](https://rich.readthedocs.io/en/latest/tables.html) object
- [`cirq`](https://github.com/quantumlib/Cirq)
    - used in past for benchmarks and QASM comparison
    - benchmarks: move to qibojit-benchmarks
    - QASM: just apply closure tests, i.e. add a bunch of valid QASM scripts, load and dump, and check that they are equivalent, or just check equivalence at the level of loaded circuit
- [`sphinx-markdown-tables`](https://github.com/ryanfox/sphinx-markdown-tables)
    - not sure where it is used, but this dependency has 27 commits, and it is stale since July 2022 - in case, just use [ReStructuredText tables](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#tables)
- [`ply`](https://github.com/dabeaz/ply)
    - not used anywhere https://github.com/search?q=repo%3Aqiboteam%2Fqibo%20%2F%5Cbply%2F&type=code
- [`nlopt`](https://github.com/DanielBok/nlopt-python)
    - another not that big project, consider alternatives

scarrazza:
Btw, is sphinx-build missing https://github.com/qiboteam/qibo/actions/runs/4342411482/jobs/7583613358?

scarrazza:
@AleCandido I believe we have solved half of your points, could you please update the list?

AleCandido:
> @AleCandido I believe we have solved half of your points, could you please update the list?

I only see one removed, and I ticked as done in the OP. The others seem to be still there as before:
- https://github.com/search?q=repo%3Aqiboteam%2Fqibo+%2F%5Cbsphinx-markdown-tables%2F&type=code
  - this one is only in the deps, but I haven't tested without, so I don't know if it has side-effects (e.g. being used by Sphinx implicitly)
- https://github.com/search?q=repo%3Aqiboteam%2Fqibo+%2F%5Cbcirq%2F&type=code
- https://github.com/search?q=repo%3Aqiboteam%2Fqibo%20%2F\btabulate%2F&type=code
[...]

PR #821, opened 2023-03-08
Removed nlop and added direct method from scipy.

### LINKED ISSUE #820: `Qibo` cannot be installed on an M1 Mac
**Describe the bug**
Running `pip install qibo` on an M1 Mac raises the following error:
```
note: This error originates from a subprocess, and is likely not a problem with pip.
error: legacy-install-failure

× Encountered error while trying to install package.
╰─> nlopt

note: This is an issue with the package mentioned above, not pip.
hint: See above for output from the failure.
```

**Additional context**
This problem arises due to the dependency with the `nlopt` package ([see link](https://github.com/DanielBok/nlopt-python/issues/13)).

I realised that this package is only used in one place: https://github.com/qiboteam/qibo/blob/e70426cdbb67dd4dbdf530b69681ad517254cfcf/src/qibo/noise_model.py#L333

It would be nice to consider making `nlopt` an optional dependency.

AlejandroSopena:
This PR removes `nlopt` as suggested in #814. This dependency was previously used to implement the DIRECT optimization method. Now, it is already implemented in `scipy==1.10.1` so is no longer needed. This resolves #820.
Checklist:
- [x] Tests are passing.


scarrazza:
I believe the poetry lock file should be updated too.

andrea-pasquale:
> I believe the poetry lock file should be updated too.

Yes, @AlejandroSopena you can update the poetry lock by running `poetry update`.
Let me know if you have any issues.

scarrazza (approved):

PR #826, opened 2023-03-10
Track evolved states

MatteoRobbiati:
In this PR a new feature is implemented in the evolution model. 
By setting `track_state=True` the evolution objects return a list of states instead of the last state only. 
Default value is False. 

I was thinking about the idea to add a `savestep` variable, in order to decide how often the state is saved.

Checklist:
- [ ] Tests are passing.


src/qibo/models/evolution.py
```diff
@@ -31,6 +31,7 @@ class StateEvolution:
         accelerators (dict): Dictionary of devices to use for distributed
             execution. This option is available only when the Trotter
             decomposition is used for the time evolution.
+        track_state (bool):
```
stavros11, line 34: This docstring is missing.

src/qibo/models/evolution.py
```diff
@@ -129,8 +143,15 @@ def execute(self, final_time, start_time=0.0, initial_state=None):
             if self.callbacks:
                 state = self.normalize_state(state)
                 self.calculate_callbacks(state)
+
+            if self.track_state:
+                state_hist.append(state)
```
stavros11, line 148: This will not work properly if a qibojit operation is used, which can happen if you use `SymbolicHamiltonian` and Trotterization, which falls back to circuit execution. Qibojit updates the state in-place so in the end you'll get a list of references to the same array, which will be the final state only.

stavros11 (commented):
Thank you for implementing this. I did not test my self but two remarks:

1. Could you please try to test it using qibojit as a backend, a `SymbolicHamiltonian` and trotterized evolution? I suspect that it will not work as expected because of the in-place updates of the state.
2. I think you have asked me about this in the past, but now looking at the code, it may be possible to do the same using the `State` callback. Have you tried this? Even if it works it is fine to keep your approach too as it is simpler, but it would be good to know if the callback also works or if it fails, why. If you see in the `execute` method the callback is evaluated for every step so you should be able to track the state with it.

PR #835, opened 2023-03-21
Hardware compatible parameter shift rule

MatteoRobbiati:
In this PR a new feature has been added to the `parameter_shift` function. 
In this suggestion it accepts two more arguments: 

- `nshots`: if `None`, the simulation is executed, if not `None`, the `expectation_from_samples` function is called for calculating the expectation value of the target observable.
- `nruns`: since we have NISQ deivces, it can be useful to have a more stable estimation of the derivative. By setting `nruns`, one can decide to calculate the derivative as mean of `nruns` estimations.  

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/derivative.py
+++ b/src/qibo/derivative.py
@@ -11,7 +11,6 @@ def parameter_shift(
     initial_state=None,
     scale_factor=1,
     nshots=None,
-    nruns=1,
 ):
     """In this method the parameter shift rule (PSR) is implemented.
     Given a circuit U and an observable H, the PSR allows to calculate the derivative
@@ -42,10 +41,6 @@ def parameter_shift(
         scale_factor (float): parameter scale factor (default None).
         nshots (int): number of shots if derivative is evaluated on hardware. If
         `None`, the simulation mode is executed (default None).
-        nruns (int): number of times the derivative is evaluated on hardware for
-        a more stable evaluation. The final result is the mean of the values in this case
-        (default 1).
-
 
     Returns:
         np.float value of the derivative of the expectation value of the hamiltonian
@@ -114,14 +109,12 @@ def parameter_shift(
     original = np.asarray(circuit.get_parameters()).copy()
     shifted = original.copy()
 
-    # forward shift and evaluation
+    # forward shift
     shifted[parameter_index] += s
     circuit.set_parameters(shifted)
 
-    forward = 0
-    backward = 0
-
-    if nshots == None:
+    if nshots is None:
+        # forward evaluation
         forward = hamiltonian.expectation(
             backend.execute_circuit(
                 circuit=circuit, initial_state=initial_state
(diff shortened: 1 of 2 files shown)


src/qibo/derivative.py
```diff
@@ -106,19 +118,49 @@ def circuit(nqubits = 1):
...
+        circuit.set_parameters(shifted)
+
+        backward = hamiltonian.expectation(
+            backend.execute_circuit(
+                circuit=circuit, initial_state=initial_state
+            ).state()
+        )
+
+    else:
+        copied = shifted.copy()
```
stavros11, line 142: Is there any reason you need an additional copy of the parameters when `nshots` are given? In principle you can use the same approach as before, having a single copy on which you do the shifts.

src/qibo/derivative.py
```diff
@@ -106,19 +118,49 @@ def circuit(nqubits = 1):
...
-    forward = hamiltonian.expectation(
-        backend.execute_circuit(circuit=circuit, initial_state=initial_state).state()
-    )
+    forward = 0
+    backward = 0
 
-    # backward shift and evaluation
-    shifted[parameter_index] -= 2 * s
-    circuit.set_parameters(shifted)
+    if nshots == None:
```
stavros11, line 124:
```suggestion
    if nshots is None:
```

src/qibo/derivative.py
```diff
@@ -26,13 +28,23 @@ def parameter_shift(
...
+        if you want to execute on hardware, a symbolic hamiltonian must be provided
+        as follows (example with Pauli Z and `nqubits=1`): `SymbolicHamiltonian(np.prod([ Z(i) for i in range(1) ]))`.
+        parameter_index (int): the index which identifies the target parameter in the `circuit.get_parameters()` list.
         initial_state ((2**nqubits) vector): initial state on which the circuit acts (default None).
         scale_factor (float): parameter scale factor (default None).
+        nshots (int): number of shots if derivative is evaluated on hardware. If
+        `None`, the simulation mode is executed (default None).
+        nruns (int): number of times the derivative is evaluated on hardware for
+        a more stable evaluation. The final result is the mean of the values in this case
+        (default 1).
```
stavros11, line 47: This is fine, but I think if I were to deploy this on hardware using multiple runs (software averages), maybe I would do a loop over the whole method myself, eg:
```py
grads = [parameter_shift(...) for _ in range(nruns)]
```
because this way I get each individual calculation so I can calculate more statistics than just the mean (eg. `np.std(grads)` could give an estimate for the error / error bars). If I use the `nruns` I only get the mean and I don't have access to each individual sample.

That being said, having it there does not really hurt, as I can still use `nruns=1` and do the above. At the end it is the same discussion with `software_averages` argument in qibocal.

stavros11 (commented):
Thank you for implementing this. The code looks good to me but can you please add some tests where you pass `nshots` and/or `nruns` so that the new lines are covered?

AlejandroSopena:
Hi, I have taken a look at this and I think the results are consistent. The error due to the number of shots `N` when calculating `<O>` can be written as $\text{error}=\sqrt{(Tr(O^2\rho)-Tr(O\rho)^2)/N}$ (hence the typical scaling with $\sqrt{N}$). In this case, $O = Z$ so $O^2=I$ and $Tr(\rho)=1$. I have calculated the numerator for the parameters you use in the test. With `scale_factor` 1, the values for the three gradients are 1.42, 1.32, 1.14, and with `scale_factor` 0.5 we have 0.96, 0.90, 0.25. The maximum error with `1e4` shots is 1.4e-2 and the minimum 2.5e-3.

AlejandroSopena:
Hi, I have taken a look at this and I think the results are consistent. The error due to the number of shots `N` when calculating `<O>` can be written as $\text{error}=\sqrt{(Tr(O^2\rho)-Tr(O\rho)^2)/N}$ (hence the typical scaling with $\sqrt{N}$). In this case, $O = Z$ so $O^2=I$ and $Tr(\rho)=1$. I have calculated the numerator for the parameters you use in the test. With `scale_factor` 1, the values for the three gradients are 1.42, 1.32, 1.14, and with `scale_factor` 0.5 we have 0.96, 0.90, 0.25. The maximum error with `1e4` shots is 1.4e-2 and the minimum 2.5e-3.

AlejandroSopena (approved):

MatteoRobbiati:
> Hi @MatteoRobbiati , I have taken a look at this and I think the results are consistent. The error due to the number of shots `N` when calculating `<O>` can be written as error=(Tr(O2ρ)−Tr(Oρ)2)/N (hence the typical scaling with N). In this case, O=Z so O2=I and Tr(ρ)=1. I have calculated the numerator for the parameters you use in the test. With `scale_factor` 1, the values for the three gradients are 1.42, 1.32, 1.14, and with `scale_factor` 0.5 we have 0.96, 0.90, 0.25. The maximum error with `1e4` shots is 1.4e-2 and the minimum 2.5e-3.

Thank you very much for this analysis @AlejandroSopena. All makes sense now.

MatteoRobbiati:
Now all tests are passing, if you agree @scarrazza we can merge.

stavros11 (approved):

scarrazza (approved):
Thanks, looks good to me.

PR #838, opened 2023-03-22
Add sparse representation to `quantum_info.basis`

renatomello:
Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/quantum_info/basis.py
+++ b/src/qibo/quantum_info/basis.py
@@ -70,35 +70,24 @@ def pauli_basis(
 
     basis_single = [matrices.I, matrices.X, matrices.Y, matrices.Z]
 
-    if nqubits >= 2:
+    if nqubits > 1:
         basis_full = list(product(basis_single, repeat=nqubits))
-        basis, indexes = [], []
-        if vectorize and sparse:
-            for row in basis_full:
-                row = vectorization(reduce(np.kron, row), order=order)
-                row_indexes = list(np.flatnonzero(row))
-                indexes.append(row_indexes)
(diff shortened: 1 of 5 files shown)


src/qibo/quantum_info/basis.py
```diff
@@ -45,33 +53,69 @@ def pauli_basis(
...
         raise_error(ValueError, "when vectorize=True, order must be specified.")
 
-    basis = [matrices.I, matrices.X, matrices.Y, matrices.Z]
+    if sparse and not vectorize:
+        raise_error(
+            NotImplementedError,
+            "sparse representation is not implemented for unvectorized Pauli basis.",
+        )
+
+    basis_single = [matrices.I, matrices.X, matrices.Y, matrices.Z]
```
vodovozovaliza, line 72: @renatomello, maybe it would be better to create a `basis_full` in advance depending on the number of qubits and use it for both cases?
```python
if nqubits > 1:
    basis_full = list(product(basis_single, repeat=nqubits)) 
    basis_full = [reduce(np.kron, row) for row in basis_full]
else:
    basis_full = basis_single
    
if vectorize and sparse:
    basis, indexes = [], []
    for row in basis_full:
        row = vectorization(row, order=order)
        row_indexes = list(np.flatnonzero(row))
        indexes.append(row_indexes)
        basis.append(row[row_indexes])
        del row
elif vectorize and not sparse:
    basis = [vectorization(matrix, order=order) for matrix in basis_full]
else:
    basis = basis_full
```

vodovozovaliza (approved):

PR #844, opened 2023-03-31
updating version and lock

scarrazza:
Checklist:
- [ ] Tests are passing.


scarrazza:
@andrea-pasquale thanks. @Edoardo-Pedicillo @AleCandido tests are failing due to a missing tensorflow dependency. If I remember correctly, you had a similar issue in qicocal, correct?

Edoardo-Pedicillo:
Yes, I can fix this.

Edoardo-Pedicillo:
One question for @AleCandido: Can we solve the problem by regenerating the lock file in Windows ?

AleCandido:
> One question for @AleCandido: Can we solve the problem by regenerating the lock file in Windows ?

It seems that is not possible, not even conceptually (see my issue on Poetry, linked on the Qibocal issue/PR).

Poetry generates lockfiles taking into account available versions, not the platform.
So, if it's available on PyPI, and there are no extra constraints, it will resolve to that version, even if it were empty (i.e. it is up to PyPI to avoid empty be releases).
Checking the availability would be more complicate, and possibly prone to various limitations.

So, here the problem is that we'd like to use a library that is not uniformly available on various platforms.
So we rely on TensorFlow availability, and TensorFlow relies on its sibling lib, but it falls short.
So, `pip` would resolve a different version for Windows and Linux, and the lock file ensures that the same version would be used. Thus it can only fail.

In general is simple enough, and the deps we use are reliable enough on all platforms.
But I was reconsidering if it's reasonable to use lockfiles in the CI...
Maybe, we should support a single development platform, and test it with locks, but then all the others we're testing for our users, so we should test from their perspective, i.e.:
- we build the wheel, as we're doing
- we use `pip` to install the built wheel

In this way we'd have a single workflow for testing and releasing (conditionally, as it is right now), and it would adopt the user perspective, including deps resolution.
Poetry would be used anyhow, but only to build.

PR #846, opened 2023-04-04
circuit initialization

Edoardo-Pedicillo:
This PR fixes the circuit initialization requested in #720, with the deployment of the `initialize` and `eigenstate` inputs, that will add the gates to perform the basis changing of the initial state.
It follows a little example,
```python
circuit = Circuit(5, initialize="X", eigenstate="-")
print(circuit.draw())
# Output:
# q0: ─X─H─
# q1: ─X─H─
# q2: ─X─H─
# q3: ─X─H─
# q4: ─X─H─
```
@scarrazza what do you think about it ? 
Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/models/circuit.py
+++ b/src/qibo/models/circuit.py
@@ -186,7 +186,7 @@ class Circuit:
     def _basis_initialization(self, basis: str, eigenstate: str):
         """This function appends some gates at the beginning of the
         circuit's queue in order to initialize all the qubits in a specific
-        basis eigenstate.
+        eigenstate of the operator defined in `basis`:
 
             - if eigenstate is  '+', no gate added
             - if eigenstate is '-', add an X gate
--- a/src/qibo/models/circuit.py
+++ b/src/qibo/models/circuit.py
@@ -205,7 +205,7 @@ class Circuit:
             initial_queue.append(gates.H)
             initial_queue.append(gates.S)
         elif basis != "Z":
-            raise NotImplementedError(f"Invalid basis {basis}")
+            raise_error(NotImplementedError, f"Invalid basis {basis}")
         for gate in initial_queue:
             for qubit in range(self.nqubits):
                 self.queue.append(gate(qubit))
--- a/src/qibo/models/circuit.py
+++ b/src/qibo/models/circuit.py
@@ -129,14 +129,7 @@ class Circuit:
         queues (DistributedQueues): Gate queues for each accelerator device. Defaults to `None`.
     """
 
-    def __init__(
-        self,
-        nqubits,
-        accelerators=None,
-        density_matrix=False,
-        basis="Z",
-        eigenstate="+",
-    ):
+    def __init__(self, nqubits, accelerators=None, density_matrix=False):
         if not isinstance(nqubits, int):
             raise_error(
                 TypeError,
--- a/src/qibo/models/circuit.py
+++ b/src/qibo/models/circuit.py
@@ -128,7 +128,7 @@ class Circuit:
         nlocal (int): Total number of available qubits in each device. Defaults to `None`.
         queues (DistributedQueues): Gate queues for each accelerator device. Defaults to `None`.
         initialize (str): Basis to define the circuit inizial state.
-        eigenstate (str): Eigenstates that will belong to the basis. 
+        eigenstate (str): Eigenstates that will belong to the basis.
     """
 
     def __init__(
--- a/src/qibo/models/circuit.py
+++ b/src/qibo/models/circuit.py
@@ -185,12 +185,15 @@ class Circuit:
 
         if eigenstate == "-":
             initial_queue.append(gates.X)
+        elif eigenstate != "+":
+            raise NotImplementedError("Invalid eigenstate")
         if initialize == "X":
             initial_queue.append(gates.H)
         elif initialize == "Y":
             initial_queue.append(gates.H)
             initial_queue.append(gates.S)
-
+        elif initialize != "Z":
+            raise NotImplementedError("Invalid initialize")
         for gate in initial_queue:
             for qubit in range(nqubits):
                 self.queue.append(gate(qubit))
--- a/src/qibo/models/circuit.py
+++ b/src/qibo/models/circuit.py
@@ -186,7 +186,7 @@ class Circuit:
         if eigenstate == "-":
             initial_queue.append(gates.X)
         elif eigenstate != "+":
-            raise NotImplementedError("Invalid eigenstate")
+            raise NotImplementedError(f"Invalid eigenstate {eigenstate}")
         if initialize == "X":
             initial_queue.append(gates.H)
         elif initialize == "Y":
--- a/src/qibo/models/circuit.py
+++ b/src/qibo/models/circuit.py
@@ -245,21 +245,28 @@ class Circuit:
                 )
 
         newcircuit = self.__class__(**self.init_kwargs)
-        # Add gates from `self` to `newcircuit` (including measurements)
-        for gate in self.queue:
-            if isinstance(gate, gates.M):
-                gate.basis = []
-            newcircuit.add(gate)
-        # Add gates from `circuit` to `newcircuit` (including measurements)
-        for gate in circuit.queue:
-            if isinstance(gate, gates.M):
-                gate.basis = []
-            newcircuit.add(gate)
-        # Re-execute full circuit when sampling if one of the circuit has repeated_execution ``True``
+        Circuit._fill_new_circuit(self, newcircuit)
+        Circuit._fill_new_circuit(circuit, newcircuit)
+        # Re-execute full circuit when sampling if one of the circuit has
+        # repeated_execution ``True``
         newcircuit.repeated_execution = (
             self.repeated_execution or circuit.repeated_execution
         )
         return newcircuit
+
+    @staticmethod
+    def _fill_new_circuit(circuit, newcircuit):
+        # this method appends all the `circuit`'s gate in new_circuit
+        excluded_gates = []
+        # all the gates in the basis of the measure gates should not
+        # be added to the new circuit, otherwise once the measure gate is added in the circuit
+        # there will be two of the same.
+        for gate in circuit.queue:
+            if isinstance(gate, gates.M):
+                excluded_gates.extend(gate.basis)
+        for gate in circuit.queue:
+            if gate not in excluded_gates:
+                newcircuit.add(gate)
 
     def on_qubits(self, *qubits):
         """Generator of gates contained in the circuit acting on specified qubits.
--- a/src/qibo/models/circuit.py
+++ b/src/qibo/models/circuit.py
@@ -127,6 +127,8 @@ class Circuit:
         nglobal (int): Base two logarithm of the number of devices. Defaults to `None`.
         nlocal (int): Total number of available qubits in each device. Defaults to `None`.
         queues (DistributedQueues): Gate queues for each accelerator device. Defaults to `None`.
+        initialize (str): Basis to define the circuit inizial state.
+        eigenstate (str): Eigenstates that will belong to the basis. 
     """
 
     def __init__(
--- a/src/qibo/models/circuit.py
+++ b/src/qibo/models/circuit.py
@@ -248,15 +248,13 @@ class Circuit:
         # Add gates from `self` to `newcircuit` (including measurements)
         for gate in self.queue:
             if isinstance(gate, gates.M):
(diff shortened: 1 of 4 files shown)


src/qibo/models/circuit.py
```diff
@@ -127,9 +127,18 @@ class Circuit:
...
+        eigenstate (str): Eigenstates that will belong to the basis.
     """
 
-    def __init__(self, nqubits, accelerators=None, density_matrix=False):
+    def __init__(
+        self,
+        nqubits,
+        accelerators=None,
+        density_matrix=False,
+        initialize="Z",
```
andrea-pasquale, line 139: You could also use `basis` instead of `initialize` if I understood correctly from #720. Or is there a particular reason on why you didn't use `basis`?

src/qibo/models/circuit.py
```diff
@@ -172,6 +181,23 @@ def __init__(self, nqubits, accelerators=None, density_matrix=False):
...
+            initial_queue.append(gates.X)
+        elif eigenstate != "+":
+            raise NotImplementedError("Invalid eigenstate")
+        if initialize == "X":
+            initial_queue.append(gates.H)
+        elif initialize == "Y":
+            initial_queue.append(gates.H)
+            initial_queue.append(gates.S)
+        elif initialize != "Z":
+            raise NotImplementedError("Invalid initialize")
```
andrea-pasquale, line 196: I think that it is better to have a few comments to explain exactly what you are doing here. You could even create a separate method
```py
def _basis_initialization(self, basis/initialize : str, eigenstate : str):
```

src/qibo/models/circuit.py
```diff
@@ -172,6 +181,23 @@ def __init__(self, nqubits, accelerators=None, density_matrix=False):
                 )
             self._distributed_init(nqubits, accelerators)
 
+        initial_queue = []
+
+        if eigenstate == "-":
+            initial_queue.append(gates.X)
+        elif eigenstate != "+":
+            raise NotImplementedError("Invalid eigenstate")
```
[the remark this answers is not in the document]
andrea-pasquale: We could even define an `Enum` for both `initialize` and `eigenstate` but maybe it is a bit of an overkill...

src/qibo/models/circuit.py
Edoardo-Pedicillo: No particular reason, I think `basis` is better.

andrea-pasquale (approved):
Thanks @Edoardo-Pedicillo.

src/qibo/models/circuit.py
```diff
@@ -172,6 +181,35 @@ def __init__(self, nqubits, accelerators=None, density_matrix=False):
...
+            - if eigenstate is '-', add an X gate
+            - if basis is 'Z', no gate added
+            - if basis is 'X', add a Hadamard gate
+            - if basis is 'Y', add a Hadamard and an S gate
+        """
+        initial_queue = []
+        if eigenstate == "-":
+            initial_queue.append(gates.X)
+        elif eigenstate != "+":
+            raise NotImplementedError(f"Invalid eigenstate {eigenstate}")
```
stavros11, line 201: We usually use a custom error
```suggestion
            raise_error(NotImplementedError, f"Invalid eigenstate {eigenstate}")
```
which I kind of dislike, but since the rest of code is like this it is better to not break the symmetry.

src/qibo/models/circuit.py
```diff
@@ -172,6 +181,35 @@ def __init__(self, nqubits, accelerators=None, density_matrix=False):
...
+        if basis == "X":
+            initial_queue.append(gates.H)
+        elif basis == "Y":
+            initial_queue.append(gates.H)
+            initial_queue.append(gates.S)
+        elif basis != "Z":
+            raise NotImplementedError(f"Invalid basis {basis}")
```
stavros11, line 208: Same here.

stavros11 (commented):
Thanks for implementing this, in terms of code it looks good. 

My main concern is from a usability perspective, personally I would prefer the `basis` and `eigenstate` in `Circuit.execute` (and `__call__`) instead of `Circuit.__init__`. This would allow the user to execute the same circuit with different initialization basis. We also already have an `initial_state` argument in `execute` so it may be confusing that we can inject initial states from two different place.

That's my personal opinion, since this is an interface issue it may be better to also ask others and see what is preferrable.

scarrazza:
Looks good to me, please go ahead.

AleCandido:
As initially proposed in https://github.com/qiboteam/qibo/issues/720#issue-1476258930, and agreed on in https://github.com/qiboteam/qibo/issues/720#issuecomment-1344077029, I'd suggest implementing this writing an `initialize(...) -> Circuit` function, and passing the result of the function as the `initial_state`, without any modification of the `Circuit` class.

src/qibo/models/circuit.py
```diff
@@ -129,7 +129,14 @@ class Circuit:
...
     """
 
-    def __init__(self, nqubits, accelerators=None, density_matrix=False):
+    def __init__(
+        self,
+        nqubits,
+        accelerators=None,
+        density_matrix=False,
+        basis="Z",
+        eigenstate="+",
```
AleCandido, line 138: Since they are not used any longer, remember to drop them.

src/qibo/models/circuit.py
```diff
@@ -240,16 +247,20 @@ def __add__(self, circuit):
         newcircuit = self.__class__(**self.init_kwargs)
         # Add gates from `self` to `newcircuit` (including measurements)
         for gate in self.queue:
-            newcircuit.add(gate)
+            if isinstance(gate, gates.M):
+                newcircuit.add(gates.M(*list(gate._target_qubits)))
+            else:
+                newcircuit.add(gate)
```
AleCandido, line 253: Why is this now needed? Is it associated in any way to the initialization?

src/qibo/models/circuit.py
```diff
@@ -240,16 +247,20 @@ def __add__(self, circuit):
...
         # Add gates from `circuit` to `newcircuit` (including measurements)
         for gate in circuit.queue:
-            newcircuit.add(gate)
-
+            if isinstance(gate, gates.M):
+                newcircuit.add(gates.M(*list(gate._target_qubits)))
+            else:
+                newcircuit.add(gate)
```
AleCandido, line 259: This is repeated, make a separate function and call it twice.

src/qibo/models/circuit.py
Edoardo-Pedicillo: These two lines should solve the following issue:
```python
from qibo import gates
c1 = Circuit(3)
c1.add(gates.M(0,1,2,basis=gates.Y))
c2 = Circuit(3)
c = c1+c2
print(c.draw())
# Output:
# q0: ─U─U─M─
# q1: ─U─U─M─
# q2: ─U─U─M─
```
I expect to find as output something similar to the circuit `c1`, but the `U` gate is added to the circuit `c` twice because it is both in the queue of `c1`, and in the basis of `M`, so it is added also when `M` is added to `c`. It was only a temporary solution ( I forgot to put the PR in draft).

AleCandido: I saw fef6a8a, and I was having hard time understanding why you wanted to avoid appending some gates.

But actually, I see nothing wrong in the snippet above: to me `+` should do the most intuitive operation, that is concatenating all the gates in `c1` to `c2`.

The only bit I do not understand is why measurement gates are floating at the end, I would have expected:
```
Output:
q0: ─U-M─U─
q1: ─U─M─U─
q2: ─U─M─U─
```

AleCandido: Ah, no, I see: it is not `Circuit(3)` generating the `U`s, but `M`. Ok, now I understand the issue (why is it drawn as `U` if above you set to `Y`?)

AleCandido: This practice of "adding gates" however is harmful: `M` should act as the gate, instead of adding it.
You can draw the way you prefer, but it is really weird that you ask to add a gate, and he do add two instead.

Edoardo-Pedicillo: > why is it drawn as U if above you set to Y?

Because (if I understood correctly)`basis = gate.Y` means adding to the circuit the unitary transformation into the basis of the eigenstates of `Y`, that is defined here https://github.com/qiboteam/qibo/blob/379886f075f207d81b85399ed55e11e839c87fba/src/qibo/gates/gates.py#L235-L239

AleCandido: Ok, at least that is clear now.

I guess it is added to standardize the `M` gate, such that it always measures in the Z basis, simply you rotate before such that the basis as a different meaning.
But I really believe this should be internal to the `M` gate, and even drawing it is a bit confusing...

Edoardo-Pedicillo: > This practice of "adding gates" however is harmful: `M` should act as the gate, instead of adding it. You can draw the way you prefer, but it is really weird that you ask to add a gate, and he does add two instead.

I agree, maybe we could define a function `change_basis` that return the circuit to perform the basis transformation from `Z` basis into the others, in this way we could call it to evaluate `circuit_basis` (in `initialize`), and when we want to measure in another basis we append before the measurement the inverse of `change_basis`.

Edoardo-Pedicillo: > This practice of "adding gates" however is harmful: `M` should act as the gate, instead of adding it. You can draw the way you prefer, but it is really weird that you ask to add a gate, and he does add two instead.

I agree, maybe we could define a function similar to `initialize`, that returns the circuit to append before the measurements and drop the automatic appending of the unitary gates.

scarrazza:
@Edoardo-Pedicillo could you please fix the conflict?

src/qibo/models/circuit.py
```diff
@@ -575,7 +574,15 @@ def add(self, gate):
...
 
             if isinstance(gate, gates.M):
-                self.add(gate.basis)
+                # The following loop is useful when two circuits are added together:
+                # all the gates in the basis of the measure gates should not
+                # be added to the new circuit, otherwise once the measure gate is added in the circuit
+                # there will be two of the same.
+
+                for base in gate.basis:
+                    if base not in self.queue:
```
AleCandido, line 583: Is `base` compared by some sort of ID? Because I believe `self.queue` to be just a list and not implementing explicitly `__contains__`:
https://github.com/qiboteam/qibo/blob/87ed54f203db477af604429d8a898e41b96d5056/src/qibo/models/circuit.py#L32

So, I believe `base` won't be added *if and only if* exactly the same object is not already part of `self.queue`. If this is fine, that's the correct way, otherwise we need to find a better comparison.

tests/test_models_circuit.py
```diff
@@ -42,6 +43,28 @@ def test_circuit_init():
...
 
 
+def test_eigenstate():
+    nqubits = 3
+    c = Circuit(nqubits)
+    c.add(gates.M(*list(range(nqubits))))
+    c2 = initialize(nqubits, eigenstate="-")
+    assert c(nshots=100, initial_state=c2).frequencies() == {"111": 100}
+    c2 = initialize(nqubits, eigenstate="+")
+    assert c(nshots=100, initial_state=c2).frequencies() == {"000": 100}
```
AleCandido, line 53: Out of personal curiosity: where is this mapping between `+/-` and `0/1` defined?

AleCandido: In case it is supposed to be defined by the `initialize` function, please document in the related docstring.

If it is a more general Qibo convention, consider referencing in the said docstring the place where this convention is declared.

Otherwise, there is also the option of dropping `+/-` in favor of explicit `0/1`, and drop one translation layer.

AleCandido: I agree that `+` and `-` are conventional, but `0` and `1` are not, so the mapping is still arbitrary.

stavros11 (approved):
Thanks for the updates, looks good to me. Regarding the `0/1` or `+/-` I don't have strong opinion but I agree that `+/-` is more general for all different basis.

Many thanks for fixing the bug with circuit addition with the basis measurements, I did not see it when I implemented this feature. I agree that it is not the best to "secretly" add gates through the measurement but I did it this way following the first point of https://github.com/qiboteam/qibo/issues/720#issuecomment-1344077029.

tests/test_models_circuit.py
Edoardo-Pedicillo: Indeed, the mapping is not conventional (at least I don't find it easily online), it was my bias due to the QM lectures I followed as a student. I will update the docs.

AleCandido (approved):

Issue #850, opened 2023-04-10
Adding labels to Qibo gates


vodovozovaliza:
It would be great to have a `label` property in Qibo gates that can be specified by the user and independent of the gate's `name` property.

For example,
```python
gate_rx1 = gates.RX(0, np.pi/2, label="rx1")
gate_rx2 = gates.RX(0, 3*np.pi/2, label="rx3")
```
This property can be used for NoiseModels or for controlling parameters of the quantum hardware.

PR #853, opened 2023-04-14
Add gate labels

vodovozovaliza:
Related to issue #850. This PR allows to change `name` parameter of `qibo.gates.Gate` and 2 other labels are added:
- `name`: label of the gate. Can be changed by a user.
- `draw_label`: a symbol for drawing a circuit. If `None` or empty, `gate.name[:4]` is used.
    This parameter allows to remove `qibo.gates.DRAW_LABELS` dictionary.
- `@property qasm_label`: for gates supported by OpenQASM.

For example:
```python
from math import pi 
from qibo import gates

rx1 = gates.RX(0, pi/2)
rx1.name = "RX90"
rx1.draw_label = "RX1"
rx2 = gates.RX(1, 3 * pi/2)
rx2.name = "RX270"
rx2. draw_label = "RX2"

circuit = Circuit(2)
circuit.add(rx1)
circuit.add(rx2)
print(circuit.draw())
```
prints
```python
q0: ─RX1─
q1: ─RX2─
```
instead of 
```python
q0: ─RX─
q1: ─RX─
```
As mentioned in #850, the `name` can be used for NoiseModels or for controlling parameters of the quantum hardware.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/models/circuit.py
+++ b/src/qibo/models/circuit.py
@@ -133,12 +133,12 @@ class Circuit:
         if not isinstance(nqubits, int):
             raise_error(
                 TypeError,
-                "Number of qubits must be an integer but " "is {}.".format(nqubits),
+                f"Number of qubits must be an integer but is {nqubits}.",
             )
         if nqubits < 1:
             raise_error(
                 ValueError,
-                "Number of qubits must be positive but is " "{}.".format(nqubits),
+                f"Number of qubits must be positive but is {nqubits}.",
             )
         self.nqubits = nqubits
         self.init_kwargs = {
@@ -168,7 +168,7 @@ class Circuit:
             if density_matrix:
                 raise_error(
                     NotImplementedError,
-                    "Distributed circuit is not implemented " "for density matrices.",
+                    "Distributed circuit is not implemented for density matrices.",
                 )
             self._distributed_init(nqubits, accelerators)
 
@@ -209,7 +209,7 @@ class Circuit:
             raise_error(
                 ValueError,
                 "Number of calculation devices should be a power "
-                "of 2 but is {}.".format(self.ndevices),
+                + f"of 2 but is {self.ndevices}.",
             )
         self.nglobal = int(self.nglobal)
         self.nlocal = self.nqubits - self.nglobal
@@ -233,8 +233,8 @@ class Circuit:
                 raise_error(
                     ValueError,
                     "Cannot add circuits with different kwargs. "
-                    "{} is {} for first circuit and {} for the "
-                    "second.".format(k, kwarg1, kwarg2),
+                    + f"{k} is {kwarg1} for first circuit and {kwarg2} "
+                    + "for the second.",
                 )
 
         newcircuit = self.__class__(**self.init_kwargs)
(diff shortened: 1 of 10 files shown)


src/qibo/models/circuit.py
```diff
@@ -1272,12 +1272,15 @@ def read_args(args):
     def _update_draw_matrix(self, matrix, idx, gate, gate_symbol=None):
         """Helper method for :meth:`qibo.models.circuit.Circuit.draw`."""
         if gate_symbol is None:
-            if gate.name not in gates.DRAW_LABELS:  # pragma: no cover
+            if gate.label is not None and gate.label != gate.name:
+                gate_symbol = gate.label[:4]
+            elif gate.name in gates.DRAW_LABELS:  # pragma: no cover
+                gate_symbol = gates.DRAW_LABELS.get(gate.name)
+            else:
                 raise_error(
```
renatomello, line 1280: a test is missing for this line

renatomello:
the parameters `label` and `name` seem to be somewhat redundant. Maybe there's a way to eliminate the redundancy by using class type instead of `name`, and then using `name` as `label` is being intended. 

For instance, any time there's a `if self.name  == "<insert-name>"` appears, we could use `if isinstance(self, "insert class type") is True`. Then, `self.name` is free to be used as a label.

I haven't looked at every instance in which `self.name` is being used, so I don't know if this suggestion even works.

stavros11:
> the parameters `label` and `name` seem to be somewhat redundant. Maybe there's a way to eliminate the redundancy by using class type instead of `name`, and then using `name` as `label` is being intended.

Looking at the changes, I also agree that we could keep only one attribute for `label` and `name` (feel free to choose whichever you prefer).

> For instance, any time that `if self.name == "<insert-name>"` appears, we could use `if isinstance(self, "insert class type") is True`. Then, `self.name` is free to be used as a label.
> 
> I haven't looked at every instance in which `self.name` is used, so I don't know if this suggestion even works and/or would break other things.

I also do not recall all the places where `name` is used, but most likely should be possible to replace it with `isinstance`. @vodovozovaliza if you also agree with us, try to have a look if you can make this change and use the `name` as `label`. I think you can easily find all lines that `name` appears in the code using some `grep` command in linux.

vodovozovaliza:
@renatomello @stavros11 thank you for the comments, I will look into replacing `name` and `label` with just one of them.

vodovozovaliza:
Thank you @stavros11. Turning `qasm_label` into `@property` is a very good idea.

renatomello:
is this dictionary https://github.com/qiboteam/qibo/blob/e6afcd583a21b63b5a52b5a105d0e1074a764c7a/src/qibo/gates/gates.py#L8 still needed?

vodovozovaliza:
> is this dictionary
> 
> https://github.com/qiboteam/qibo/blob/e6afcd583a21b63b5a52b5a105d0e1074a764c7a/src/qibo/gates/gates.py#L8
> 
> still needed?

Yes, this dictionary is needed for `Circuit.from_qasm` method

renatomello (approved):

renatomello:
@stavros11 could you also check this before we merge?

renatomello:
@vodovozovaliza since `name` became the de facto `label`, could you rename this PR?

stavros11 (approved):
Thanks for the updates, looks good to me too. I just left a comment regarding removing the `PARAMETRIZED_GATES` set. This is not strictly required to merge but since you already cleaned the other dictionaries in gates.py, it may be a good idea to do so.

If understand correctly now the user is able to control both the `name` and the `draw_label` seperately. Is this okay for your applications or you prefer the `draw_label` to change every time the `name` changes? If that's the case we may need to have a setter or drop completely the `draw_label` and use only the `name` (but then we lose the flexibility of adjusting them seperately).

Issue #843, opened 2023-03-29
Add backends to `quantum_info`


vodovozovaliza:
At the moment, most functions in `qibo.quantum_info` use `numpy`. 
All methods should be adapted to work with different backends.

scarrazza:
Following our discussion, the suggestion is the following:
- test all backends locally using pytest
- if fixes require complex changes, raise errors informing that a specific function is not available for backends...

Issue #761, opened 2023-01-12
Backend selection


AleCandido:
I see two problems with the current backend selection:

1. it gets a string
2. it is global

Concerning the first point there is an easy solution: maintain inside Qibo an enum with all the options (it makes easier also to discover them). Enums are part of the [standard library](https://docs.python.org/3/library/enum.html).

About the second one, I understand `qibo` is not the only library doing this, e.g. matplotlib also has global configs (and even a global state) and Tensorflow does something similar.
But learning from Tensorflow, `qibo` could use at least a context manager (like `tf.device`).
An option I'd like more (but I'm not sure about feasibility, for lack of familiarity) would be to pass the backend explicitly to a `run`/`execute` function. I.e. you first build the abstract circuit, in a completely independent way from the backend, and finally you compile and run on the specific platform, getting back an output object; something like:
```py
backend = qibo.Backend.numpy
circuit = qibo.Circuit(nqubits)
...  # build the circuit
frequencies = circuit(nshots=1000, backend=backend).frequencies(...)

# and I can run the same circuit on a different backend
lab = qibolab.backend(platform=MyPlatform())
frequencies = circuit(nshots=1000, backend=lab).frequencies(...)
```

Usually, you can always do the same also using globals, but: it hides the flow, and makes it more complex to reproduce results, since there are parameters that affect the output that are not passed as input (this is a *purely functional* objection), and it makes it impossible to work in a concurrent environment (e.g. a dispatcher thread for various GPUs, sending further computations while waiting for the results).

scarrazza:
Thanks @AleCandido, just a comment concerning global. We have considered the API proposed here, the main limitation is that the code becomes platform specific, in the sense that you must change the flag manually, while with a global object we can have a default which is controlled by an env flag of the node where the job is executing.

AleCandido:
> you must change the flag manually, while with a global object we can have a default which is controlled by an env flag of the node where the job is executing.

Well, this can be solved by adding a "fallback" value, like, "if it's `None`, look for the flag":
```py
# explicitly
frequencies = circuit(nshots=1000, backend=None).frequencies(...)
# or you can also decide to have it as default value
frequencies = circuit(nshots=1000).frequencies(...)
```
On the other hand, maybe it could be even more explicit with a special value `qibo.Backend.FromEnvironment`. But here we really enter details about DX, people will have their own subjective preferences...

scarrazza:
Closing this, since 0.3.x we have now a more robust backend mechanism.

PR #855, opened 2023-04-21
backend-agnostic quantum info module

renatomello:
Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/gates/channels.py
+++ b/src/qibo/gates/channels.py
@@ -90,16 +90,17 @@ class Channel(Gate):
                 self.coefficients += (p0,)
                 self.gates += (I(*self.target_qubits),)
 
-        super_op = np.zeros((4**self.nqubits, 4**self.nqubits), dtype="complex")
+        super_op = np.zeros((4**self.nqubits, 4**self.nqubits), dtype=complex)
+        super_op = backend.cast(super_op, dtype=super_op.dtype)
         for coeff, gate in zip(self.coefficients, self.gates):
             kraus_op = FusedGate(*range(self.nqubits))
             kraus_op.append(gate)
             kraus_op = kraus_op.asmatrix(backend)
-            kraus_op = vectorization(kraus_op, order=order)
+            kraus_op = vectorization(kraus_op, order=order, backend=backend)
             super_op += coeff * np.outer(kraus_op, np.conj(kraus_op))
             del kraus_op
 
-        super_op = backend.cast(super_op, dtype=super_op.dtype)
+        # super_op = backend.cast(super_op, dtype=super_op.dtype)
 
         return super_op
 
@@ -128,8 +129,8 @@ class Channel(Gate):
             backend = GlobalBackend()
 
         super_op = self.to_choi(order=order, backend=backend)
-        super_op = choi_to_liouville(super_op, order=order)
-        super_op = backend.cast(super_op, dtype=super_op.dtype)
+        super_op = choi_to_liouville(super_op, order=order, backend=backend)
+        # super_op = backend.cast(super_op, dtype=super_op.dtype)
 
         return super_op
 
@@ -159,11 +160,11 @@ class Channel(Gate):
         super_op = self.to_liouville(backend=backend)
 
         # unitary that transforms from comp basis to pauli basis
-        unitary = comp_basis_to_pauli(self.nqubits, normalize)
+        unitary = comp_basis_to_pauli(self.nqubits, normalize, backend=backend)
         unitary = backend.cast(unitary, dtype=unitary.dtype)
 
         super_op = unitary @ super_op @ np.transpose(np.conj(unitary))
-        super_op = backend.cast(super_op, dtype=super_op.dtype)
+        # super_op = backend.cast(super_op, dtype=super_op.dtype)
 
         return super_op
 
--- a/src/qibo/gates/channels.py
+++ b/src/qibo/gates/channels.py
@@ -59,7 +59,6 @@ class Channel(Gate):
         Returns:
             Choi representation of the channel.
         """
-        # import numpy as np
 
         from qibo.quantum_info.superoperator_transformations import vectorization
 
@@ -100,8 +99,6 @@ class Channel(Gate):
             super_op += coeff * np.outer(kraus_op, np.conj(kraus_op))
             del kraus_op
 
-        # super_op = backend.cast(super_op, dtype=super_op.dtype)
-
         return super_op
 
     def to_liouville(self, order: str = "row", backend=None):
@@ -119,7 +116,6 @@ class Channel(Gate):
         Returns:
             Liouville representation of the channel.
         """
-        # import numpy as np
 
         from qibo.quantum_info.superoperator_transformations import choi_to_liouville
 
@@ -130,7 +126,6 @@ class Channel(Gate):
 
         super_op = self.to_choi(order=order, backend=backend)
         super_op = choi_to_liouville(super_op, order=order, backend=backend)
-        # super_op = backend.cast(super_op, dtype=super_op.dtype)
 
         return super_op
 
@@ -148,7 +143,6 @@ class Channel(Gate):
         Returns:
             Pauli-Liouville representation of the channel.
         """
-        # import numpy as np
 
         from qibo.quantum_info.basis import comp_basis_to_pauli
 
@@ -164,7 +158,6 @@ class Channel(Gate):
         unitary = backend.cast(unitary, dtype=unitary.dtype)
 
         super_op = unitary @ super_op @ np.transpose(np.conj(unitary))
-        # super_op = backend.cast(super_op, dtype=super_op.dtype)
 
         return super_op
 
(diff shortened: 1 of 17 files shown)


renatomello:
This goes together with https://github.com/qiboteam/qibojit/pull/115 and closes #843

renatomello:
This is ready @vodovozovaliza @stavros11

src/qibo/gates/channels.py
```diff
@@ -91,16 +91,17 @@ def to_choi(self, order: str = "row", backend=None):
...
             kraus_op = FusedGate(*range(self.nqubits))
             kraus_op.append(gate)
             kraus_op = kraus_op.asmatrix(backend)
-            kraus_op = vectorization(kraus_op, order=order)
+            kraus_op = vectorization(kraus_op, order=order, backend=backend)
             super_op += coeff * np.outer(kraus_op, np.conj(kraus_op))
             del kraus_op
 
-        super_op = backend.cast(super_op, dtype=super_op.dtype)
+        # super_op = backend.cast(super_op, dtype=super_op.dtype)
```
stavros11, line 104: Maybe these comments (and same in the following lines) are not needed.

src/qibo/quantum_info/basis.py
```diff
@@ -143,33 +157,50 @@ def comp_basis_to_pauli(
...
 
     unitary = pauli_basis(
-        nqubits, normalize, vectorize=True, sparse=sparse, order=order
+        nqubits, normalize, vectorize=True, sparse=sparse, order=order, backend=backend
     )
 
     unitary = np.conj(unitary)
 
+    if backend.__class__.__name__ == "TensorflowBackend":
+        unitary = backend.cast(unitary, dtype=unitary.dtype)
```
stavros11, line 193: If you do this `cast` for all backends, so that you don't need to check explicitly if it is tensorflow, does it break something?

tests/test_cirq.py
```diff
@@ -166,7 +158,8 @@ def test_one_qubit_gates(backend, gate_name, nqubits, ndevices):
     targets = random_active_qubits(nqubits, nactive=1)
     qibo_gate = getattr(gates, gate_name)(*targets)
     cirq_gate = [(getattr(cirq, gate_name), targets)]
-    assert_gates_equivalent(backend, qibo_gate, cirq_gate, nqubits, ndevices)
+    if backend.__class__.__name__ != "CupyBackend":
```
stavros11, line 161: What is the problem with the `CupyBackend` here? I believe these tests used to pass with it.

src/qibo/quantum_info/metrics.py
```diff
@@ -29,10 +33,12 @@ def purity(state):
     else:
         pur = np.real(np.trace(np.dot(state, state)))
 
+    pur = float(pur)
```
stavros11, line 36: I see that this is done in many functions, why is it needed?

stavros11 (commented):
Thanks for implementing this. I left some comments below which are mostly questions. My main concern is why you need to skip the Cirq tests for cupy (and some for tensorflow) as I believe they were working before and I think what they are testing is not very related to this PR. If they are failing now, it may be worth checking why.

Other than that it looks good to me. I was thinking whether we could simplify this mechanism of
```py
if backend is None: 
    backend = GlobalBackend()
```
which is extensively used here but also in other places in the code but I am not sure how. Some of the ideas in #761 are worth checking and could help in this direction, but this may take more time so we should not block this PR for it.

src/qibo/quantum_info/basis.py
renatomello: It doesn't break but it's redundant. I opted to check if because for bigger systems it might be a slow process to recast to the same backend even though the array is already at the right backend.

tests/test_cirq.py
renatomello: There are some `Cirq` fucntions that don't work with `cupy.ndarray` inputs due to the way they're implemented. Same thing for `EagerTensor`s. I even proposed a change to `Cirq` [here](https://github.com/quantumlib/Cirq/pull/6076) to fix some of these issues.

src/qibo/quantum_info/metrics.py
renatomello: Otherwise it returns a 0-dim array with a float inside.

src/qibo/gates/channels.py
renatomello: Thanks

renatomello:
> Thanks for implementing this. I left some comments below which are mostly questions. My main concern is why you need to skip the Cirq tests for cupy (and some for tensorflow) as I believe they were working before and I think what they are testing is not very related to this PR. If they are failing now, it may be worth checking why.
> 
> Other than that it looks good to me. I was thinking whether we could simplify this mechanism of
> 
> ```python
> if backend is None: 
>     backend = GlobalBackend()
> ```
> 
> which is extensively used here but also in other places in the code but I am not sure how. Some of the ideas in #761 are worth checking and could help in this direction, but this may take more time so we should not block this PR for it.

I also prefer to leave this to another PR.

src/qibo/quantum_info/metrics.py
```diff
@@ -81,11 +76,16 @@ def entropy(state, base: float = 2, validate: bool = False, backend=None):
                 backend.calculate_norm(np.transpose(np.conj(state)) - state)
                 <= PRECISION_TOL
             )
-            eigenvalues, _ = (
-                np.linalg.eigh(state) if hermitian else np.linalg.eig(state)
+            if not hermitian and backend.__class__.__name__ == "CupyBackend":
+                raise_error(
+                    NotImplementedError,
+                    f"CupyBackend does not support `np.linalg.eigvals` for non-Hermitian `state`.",
```
renatomello, line 82: Here, its non-Hermitian `state - target`.

src/qibo/quantum_info/basis.py
renatomello: Changed where arrays are cast to the general backend, and now this is working

stavros11 (approved):
Thanks for the updates and explanations.

Regarding `backend.cast`, if you see the implementation in the `NumpyBackend`
https://github.com/qiboteam/qibo/blob/2753c4fe4a59cccc8f42f4c23f2ca6cb0ef71bde/src/qibo/backends/numpy.py#L56
if the input `x` is already a numpy array, it will return `x.astype(dtype, copy=False)`, which according to the [numpy documentation](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html#numpy.ndarray.astype) will do nothing if `x` already has type `dtype`. So there should be no performance deterioration even if you use `cast` when is not really needed. Note that `NumbaBackend` inherits `cast` from `NumpyBackend`. `if backend.__class__.__name__` checks is not a blocking issue for me, it just looks a bit annoying because it breaks the symmetry between different backends. One thing to point out though is that if you call `backend.cast` without specifying the `dtype`, it will use the default type (which is `complex128`) and this may create a copy of the array while casting to a different type, or even cast a real array to complex which you may not want.

renatomello:
> Thanks for the updates and explanations.
> 
> Regarding `backend.cast`, if you see the implementation in the `NumpyBackend`
> 
> https://github.com/qiboteam/qibo/blob/2753c4fe4a59cccc8f42f4c23f2ca6cb0ef71bde/src/qibo/backends/numpy.py#L56
> 
> if the input `x` is already a numpy array, it will return `x.astype(dtype, copy=False)`, which according to the [numpy documentation](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html#numpy.ndarray.astype) will do nothing if `x` already has type `dtype`. So there should be no performance deterioration even if you use `cast` when is not really needed. Note that `NumbaBackend` inherits `cast` from `NumpyBackend`. `if backend.__class__.__name__` checks is not a blocking issue for me, it just looks a bit annoying because it breaks the symmetry between different backends. One thing to point out though is that if you call `backend.cast` without specifying the `dtype`, it will use the default type (which is `complex128`) and this may create a copy of the array while casting to a different type, or even cast a real array to complex which you may not want.

Thanks for that explanation. Regarding `dtype`, from my experience, it is needed because `Tensorflow` does not accept `complex128`.

tests/test_gates_channels.py
```diff
@@ -140,13 +142,13 @@ def test_unitary_channel_errors():
...
-    initial_rho = random_density_matrix(2**2)
+    initial_rho = random_density_matrix(2**2, backend=backend)
     channel = gates.PauliNoiseChannel(1, px=0.3)
     final_rho = backend.apply_channel_density_matrix(channel, np.copy(initial_rho), 2)
     gate = gates.X(1)
     target_rho = backend.apply_gate_density_matrix(gate, np.copy(initial_rho), 2)
-    target_rho = 0.3 * backend.to_numpy(target_rho)
-    target_rho += 0.7 * initial_rho
+    # target_rho = 0.3 * backend.to_numpy(target_rho)
+    target_rho = 0.3 * target_rho + 0.7 * initial_rho
```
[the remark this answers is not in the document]
renatomello: Done

scarrazza:
@renatomello are you sure this will work without pinning the latest qibojit version in the toml?

renatomello:
> @renatomello are you sure this will work without pinning the latest qibojit version in the toml?

`pyproject.toml` is installing `qibojit` from `master`. Since the changes are already merged there, it is supposed to work (e.g. tests are passing here). 

As I understand, the tests failed on main after merge in an issue that seemed unrelated to this PR.

PR #862, opened 2023-04-25
add docs to noisy circuit probabilities

### LINKED ISSUE #666: Probabilities method with noisy circuits
The execution result for a circuit with density_matrix=False in form of the probabilities of a circuit vs. the same circuit with noise is the same. Could the probabilities() method be overwritten or could at least a warning be printed? Since it does not reflect the probabilities anymore.

The circuit without noise:
```python
import numpy as np
circuit = models.Circuit(1)
circuit.add(gates.X(0))
circuit.add(gates.M(0))
executed = circuit(nshots=1024)
print(executed.probabilities()) # [0. 1.]
print(np.average(executed.samples())) # 1.0
```
The circuit with noise:
```python
from qibo.noise import PauliError, NoiseModel
pauliflip = PauliError(0.1, 0.1, 0.1)
noise = NoiseModel()
noise.add(pauliflip, gates.X)
executed = noisy_circuit(nshots=1024)
print(executed.probabilities()) # [0. 1.]
print(np.average(executed.samples())) # 0.8271484375
```

scarrazza:
Closes #666. I prefer to just update the documentation, otherwise we have to introduce a new attribute to the Circuit class and complicate the code just to raise a warning.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/states.py
+++ b/src/qibo/states.py
@@ -299,10 +299,11 @@ class CircuitResult:
         return self.state(numpy=True)
 
     def probabilities(self, qubits=None):
-        """Calculates measurement probabilities by tracing out qubits. Please
-        note that this method computes probabilities using the exact
-        noise-free state vector output simulation therefore noise applied to
-        circuits will not have impact on the output.
+        """Calculates measurement probabilities by tracing out qubits. When
+        a noise model is applied to the circuit this method will return the
+        probabilities from the last execution shot, which may or may not
+        correspond to the noise-free state depending on whether the noise was
+        applied or not in that specific execution.
 
         Args:
             qubits (list, set): Set of qubits that are measured.
--- a/src/qibo/states.py
+++ b/src/qibo/states.py
@@ -299,7 +299,10 @@ class CircuitResult:
         return self.state(numpy=True)
 
     def probabilities(self, qubits=None):
-        """Calculates measurement probabilities by tracing out qubits. Please note that this method computes probabilities using the exact noise-free state vector output simulation therefore noise applied to circuits will not affect the computation.
+        """Calculates measurement probabilities by tracing out qubits. Please
+        note that this method computes probabilities using the exact
+        noise-free state vector output simulation therefore noise applied to
+        circuits will not have impact on the output.
 
         Args:
             qubits (list, set): Set of qubits that are measured.
--- a/src/qibo/states.py
+++ b/src/qibo/states.py
@@ -299,7 +299,7 @@ class CircuitResult:
         return self.state(numpy=True)
 
     def probabilities(self, qubits=None):
-        """Calculates measurement probabilities by tracing out qubits.
+        """Calculates measurement probabilities by tracing out qubits. Please note that this method computes probabilities using the exact noise-free state vector output simulation therefore noise applied to circuits will not affect the computation.
 
         Args:
             qubits (list, set): Set of qubits that are measured.


stavros11 (commented):
I think this docstring is not correct because if noise is applied to the circuit, it may actually impact the output of `probabilities()`. In the example from the issue:
```py
import numpy as np
from qibo.noise import PauliError, NoiseModel

circuit = models.Circuit(1)
circuit.add(gates.X(0))
circuit.add(gates.M(0))

pauliflip = PauliError(0.1, 0.1, 0.1)
noise = NoiseModel()
noise.add(pauliflip, gates.X)
executed = noisy_circuit(nshots=1024)

print(executed.probabilities()) # [0. 1.]
print(np.average(executed.samples())) # 0.8271484375
```
the `noisy_circuit` will be executed 1024 times with the noise applied randomly (according to the given probabilities) in each execution. `executed.probabilities()` will return the probabilities from the last execution which may or may not correspond to the noise-free state, depending on whether the noise was applied or no in the last execution.

stavros11 (approved):
Looks good to me now, thanks.

wilkensJ:
Hello @scarrazza and @stavros11 ,

thank you for updating the documentation. I would like to make clear why I think that disabling `.probabilities()` and .frequencies()` can be important.

My main point is that the resulting circuit, after executing, does not distinguish between a circuit where a noise model was added or not. So either

1. the fact that a noise model was used has to be passed through the whole protocol until the acquired data is analyzed or 
2. the data which is passed is possibly wrong.

I will sketch the structure of a possible protocol where both of these functions can be called without having knowledge of each other.

```python
def my_executer(circuit, nshots, noise_model=None) -> dict:
    datadict = {'circuit': circuit.copy(deep=True)}
    if noise_model is not None:
        circuit = noise_model.apply(circuit)
    datadict['circuit_result'] = circuit(nshots=nshots)
    return datadict

def process_result(datadict: dict) -> dict:
    # No way of seeing that the executed circuit was executed with a noise model.
    circuit_result = datadict['circuit_result']
    # ... do some fancy post processing of the circuit result ...
    datadict['probabilities'] = circuit_result.probabilities()
    datadict['frequencies'] = circuit_result.frequencies()
    datadict['samples'] = circuit_result.samples()
    return datadict
```

This can be used to quickly execute a circuit, store the dictionary somehow, and move the post processing to a later stage of the protocol or make the rerunning of the post processing possible. 


Now in this (a bit of an obvious) example, if the knowledge of the noise model is not passed, the data dictionaries can not know that `probabilities` and `frequencies` are wrong.

```python
circuit = models.Circuit(1)
circuit.add([gates.X(0),gates.M(0)])

datadict1 = process_result(my_executer(circuit.copy(deep = True), 1024, None))
print(datadict1['probabilities']) # [0. 1.]
print(datadict1['frequencies']) # Counter({'1': 1024})
print(np.average(datadict1['samples'])) # 1.0

noise = NoiseModel()
noise.add(PauliError(0.1, 0.1, 0.1), gates.X)

datadict2 = process_result(my_executer(circuit.copy(deep = True), 1024, noise))
print(datadict2['probabilities']) # [0. 1.]
print(datadict2['frequencies']) # Counter({'1': 1024})
print(np.average(datadict2['samples'])) # 0.8232421875
```

Concluding, I am still very much in favor of either

1. disabling `probabilities` and `frequencies` or
2. calculating them with the samples from the result,  making them approximate of course.

Issue #839, opened 2023-03-23
Quantum annealing with a sixth order Hamiltonian


lucaanutricati:
I'm doing simulated quantum annealing with qibo and I have an issue with an Hamiltonian with 6 coupled qubits (I don't have any issue if the Hamiltonian is at most quintic). Here is a very simple code 

```from qibo import hamiltonians
from qibo import models
from qibo.symbols import X, Z
import numpy as np


time_steps = 2    # number of time-steps 
time_final = 20
delta_t= time_final/time_steps 
dt = 0.01

N= 6 #number of qubits

 #  #  #  #  #  #  #  #  #  #  #  #  #  #  #  #  #  #  #  #
 
if __name__ == '__main__':
  
  SH1 = Z(0)*Z(1)*Z(2)*Z(3)*Z(4)*Z(5) #### Sextic Hamiltonian

  # Initial Hamiltonian
  SH0 = sum( 2* (0.5-0.5*X(i)) for i in range(N))
  
  # Define a Hamiltonian using the above form
  H0 = hamiltonians.SymbolicHamiltonian(SH0)
  H1 = hamiltonians.SymbolicHamiltonian(SH1)

  #Anneal schedule
  def s(t):
      return 0.5 + 0.5 * ((t - 0.5)/0.5)**3
  
  #Now the evolution
  evolve = models.AdiabaticEvolution(H0, H1, s, dt=dt)

  #definition of the initial states
  initial_state = np.ones(2**N) / np.sqrt(2**N)
 
  def out_state(t,in_state):
      return evolve(final_time=t, initial_state=in_state)

  in_state = initial_state
    
  for k in range(time_steps):
    time_k = (k+1)* delta_t
    fs = out_state(time_k, initial_state )  
```

The error I get is 

```
TypeError: No matching definition for argument type(s) array(complex128, 1d, C), array(complex128, 2d, F), array(int32, 1d, C), int64, array(int64, 1d, C)
```

when it calls the ``evolve`` function which in turn is defined through ``models.AdiabaticEvolution``.

Everything goes fine if I change the Hamiltonian to a quintic one. Is this related to QIBO?

MatteoRobbiati:
Thank you for opening this. 
I tried this code and it worked. 

Anyway, this problem seems to me related to a mismatch of dimension of size in variables types, probably related to `numba` in `qibojit` (it should work on `numpy` backend). 
@andrea-pasquale maybe you can better understand this.

[...]

PR #863, opened 2023-04-25
Copy ground state before passing to adiabatic evolution

### LINKED ISSUE #824: The adiabatic evolution model alters the h0 object when using symbolic hamiltonians
It seems like the `AdiabaticEvolution` model replaces things on the initial Hamiltonian, making repeated adiabatic evolutions without re-defining this initial Hamiltonian fail. 

This should reproduce the issue:

```python

import numpy as np
from qibo import hamiltonians
from qibo.models.evolution import AdiabaticEvolution

nqubits = 5

h_x = hamiltonians.X(nqubits, dense=False)
print(h_x.ground_state())

h_ising = hamiltonians.TFIM(nqubits, h=1.0, dense=False)
print(h_ising.ground_state())

s = lambda t: t
evolve = AdiabaticEvolution(h_x, h_ising, s, dt=0.01, solver='exp')
final_state = evolve(final_time=10)
print(final_state)

print(h_x.ground_state()) #<--- This is now changed!
print(h_ising.ground_state())

evolve = AdiabaticEvolution(h_x, h_ising, s, dt=0.01, solver='exp')
final_state = evolve(final_time=10)
print(final_state) #<--- This has now changed too then!

```

You will notice how the h_x has changed.

This is not the case with `dense=True` Hamiltonians.

Why does this happen? Should this be fixed, or just properly documented?

stavros11:
Fixes #824. @MatteoRobbiati indeed you are right, it is a qibojit issue.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/models/evolution.py
+++ b/src/qibo/models/evolution.py
@@ -236,7 +236,9 @@ class AdiabaticEvolution(StateEvolution):
             )
         self.hamiltonian.total_time = final_time - start_time
         if initial_state is None:
-            initial_state = self.hamiltonian.ground_state()
+            initial_state = self.backend.cast(
+                self.hamiltonian.ground_state(), copy=True
+            )
         return super().execute(final_time, start_time, initial_state)
 
     @staticmethod


scarrazza (approved):
Good point, thanks.

scarrazza:
@MatteoRobbiati please let us know if this works also for you.

MatteoRobbiati (approved):
I confirm it works as expected also using the `qibojit` backend.
Thank you @stavros11!

MatteoRobbiati:
Only to be sure it is a problem of mine: is it normal that the execution time of the #824 code requires 1sec using `numpy` and 2min using `qibojit`?

stavros11:
> Only to be sure it is a problem of mine: is it normal that the execution time of the #824 code requires 1sec using `numpy` and 2min using `qibojit` @stavros11?

With @andrea-pasquale we have observed this issue in the past for the `NumbaBackend`, see for example 
https://github.com/qiboteam/qibo/issues/839#issuecomment-1482683282
We suspect something is wrong but we did not investigate. I will try to have a look when I have time. This problem seems to appear only with numba, not cupy.

MatteoRobbiati:
> > Only to be sure it is a problem of mine: is it normal that the execution time of the #824 code requires 1sec using `numpy` and 2min using `qibojit` @stavros11?
> 
> With @andrea-pasquale we have observed this issue in the past for the `NumbaBackend`, see for example [#839 (comment)](https://github.com/qiboteam/qibo/issues/839#issuecomment-1482683282) We suspect something is wrong but we did not investigate. I will try to have a look when I have time. This problem seems to appear only with numba, not cupy.

Ok perfect. I will have a look too with some `python` profiler.

PR #868, opened 2023-04-29
Calibration matrix for readout mitigation

BrunoLiegiBastonLiegi:
This PR implements two methods to mitigate readout errors. 
The first method involves estimating the columns of the calibration matrix by measuring the output frequencies  for all posible bit strings, and then applying its inverse to mitigate the noise.
The second method randomizes the circuit in such a way that it diagonalizes the measurement channel. This means that now the effect of readout noise on the expected value of a Pauli observable is a multiplicative factor.

An option to apply readout mitigation is added to all the error mitigation methods currently implemented.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/models/error_mitigation.py
+++ b/src/qibo/models/error_mitigation.py
@@ -145,7 +145,7 @@ def ZNE(
         )
         if "ncircuits" in readout.keys():
             circuit_result, circuit_result_cal = apply_randomized_readout_mitigation(
-                noisy_circuit, backend, noise_model, nshots, readout["ncircuits"]
+                noisy_circuit, noise_model, nshots, readout["ncircuits"], backend
             )
         else:
             if noise_model is not None and backend.name != "qibolab":
@@ -299,7 +299,7 @@ def CDR(
         train_val["noise-free"].append(val)
         if "ncircuits" in readout.keys():
             circuit_result, circuit_result_cal = apply_randomized_readout_mitigation(
-                c, backend, noise_model, nshots, readout["ncircuits"]
+                c, noise_model, nshots, readout["ncircuits"], backend
             )
         else:
             if noise_model is not None and backend.name != "qibolab":
@@ -318,7 +318,7 @@ def CDR(
     # Run the input circuit
     if "ncircuits" in readout.keys():
         circuit_result, circuit_result_cal = apply_randomized_readout_mitigation(
-            circuit, backend, noise_model, nshots, readout["ncircuits"]
+            circuit, noise_model, nshots, readout["ncircuits"], backend
         )
     else:
         if noise_model is not None and backend.name != "qibolab":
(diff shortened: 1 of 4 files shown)


src/qibo/models/error_mitigation.py
```diff
@@ -323,10 +344,79 @@ def vnCDR(
...
+
+    Returns:
+        qibo.states.CircuitResult : The input state with the updated frequencies.
+    """
+    freq = np.zeros(2**state.nqubits)
+    for k, v in state.frequencies().items():
+        f = v / state.nshots
+        freq[int(k, 2)] = f
+    freq = freq.reshape(-1, 1)
+    for i, val in enumerate(calibration_matrix @ freq * state.nshots):
```
AlejandroSopena, line 420: Here ``nshots`` appears multiplying and dividing so I would remove it
``` python
        freq[int(k, 2)] = v
    freq = freq.reshape(-1, 1)
    for i, val in enumerate(calibration_matrix @ freq):
```

AlejandroSopena (commented):
Thanks for implementing this. Everything looks good. Just one small comment:

BrunoLiegiBastonLiegi:
Thanks for the reference @renatomello, and thanks to @AlejandroSopena for implementing the other method.
@igres26 rightfully suggested not to substitute the original frequencies from the `CircuitResult`. Should we add a second variable to `CircuitResult`, e.g. `mitigated_frequencies`, such that both the frequencies remain available after readout mitigation?

src/qibo/models/error_mitigation.py
```diff
@@ -320,13 +372,144 @@ def vnCDR(
...
+    freq = np.zeros(2**state.nqubits)
+    for k, v in state.frequencies().items():
+        freq[int(k, 2)] = v
+    freq = freq.reshape(-1, 1)
+    for i, val in enumerate(calibration_matrix @ freq):
+        state._frequencies[i] = float(val)
+    return state
+
+
+def apply_temme_readout_mitigation(
```
renatomello, line 457: I would rename this as something like `apply_randomized_readout_mitigation`, for a lack of a better name

src/qibo/models/error_mitigation.py
```diff
@@ -320,13 +372,144 @@ def vnCDR(
...
+
+    string = "{0:0" + str(nqubits) + "b}"
+    for i in range(2**nqubits):
+        state = string.format(i)
+        circuit = Circuit(nqubits, density_matrix=True)
+        for q, bit in enumerate(state):
+            if bit == "1":
+                circuit.add(gates.X(q))
+        circuit.add(gates.M(*range(nqubits)))
+        if noise_model != None and backend.name != "qibolab":
```
renatomello, line 427: `noise_model is not None`

src/qibo/models/error_mitigation.py
```diff
@@ -320,13 +372,144 @@ def vnCDR(
...
+    """
+
+    if backend is None:  # pragma: no cover
+        from qibo.backends import GlobalBackend
+
+        backend = GlobalBackend()
+
+    matrix = np.zeros((2**nqubits, 2**nqubits))
+
+    string = "{0:0" + str(nqubits) + "b}"
```
renatomello, line 419: you could do directly `f"{0:0{nqubits}b}"`

AlejandroSopena (approved):

renatomello (approved):

PR #878, opened 2023-05-03
Display correct probabilities when `circuit.density_matrix=False` and `circuit.repeated_execution=True`

renatomello:
Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/backends/numpy.py
+++ b/src/qibo/backends/numpy.py
@@ -436,6 +436,11 @@ class NumpyBackend(Backend):
 
         results = []
         nqubits = circuit.nqubits
+
+        if not circuit.density_matrix:
+            probabilities = np.zeros(2**nqubits, dtype=float)
+            probabilities = self.cast(probabilities, dtype=probabilities.dtype)
+
         for _ in range(nshots):
             if circuit.density_matrix:
                 if initial_state is None:
@@ -471,14 +476,24 @@ class NumpyBackend(Backend):
             else:
                 results.append(state)
 
+            if not circuit.density_matrix:
+                probabilities += result.probabilities()
+
+        if not circuit.density_matrix:
+            probabilities /= nshots
+
         if circuit.measurements:
             final_result = CircuitResult(self, circuit, state, nshots)
             final_result._samples = self.aggregate_shots(results)
+            if not circuit.density_matrix:
+                final_result.probabilities = \
+                    lambda: self._override_probabilities_repeated_execution(probabilities)
             circuit._final_state = final_result
             return final_result
-        else:
-            circuit._final_state = CircuitResult(self, circuit, results[-1], nshots)
-            return results
+
+        circuit._final_state = CircuitResult(self, circuit, results[-1], nshots)
+            
+        return results
 
     def execute_distributed_circuit(
         self, circuit, initial_state=None, nshots=None, return_array=False
@@ -546,6 +561,14 @@ class NumpyBackend(Backend):
             else:
                 unmeasured.append(i)
         return self.np.transpose(probs, [reduced.get(i) for i in qubits])
+
+    def _override_probabilities_repeated_execution(self, probabilities):
+        """Override `.probabilties()` method from CircuitResult object
+        with the cumulative average of `nshot` probabilities from
+        repeated execution when `circuit.density_matrix=False`.
+        """
+        # TODO: add qubits argument in order to get marginal probabilities
+        return probabilities
 
     def calculate_probabilities(self, state, qubits, nqubits):
         rtype = self.np.real(state).dtype


src/qibo/backends/numpy.py
```diff
@@ -471,14 +476,27 @@ def execute_circuit_repeated(self, circuit, initial_state=None, nshots=None):
...
+
         if circuit.measurements:
             final_result = CircuitResult(self, circuit, state, nshots)
             final_result._samples = self.aggregate_shots(results)
+            if not circuit.density_matrix:
+                final_result.probabilities = (
+                    lambda: self._override_probabilities_repeated_execution(
+                        probabilities
+                    )
+                )
```
stavros11, line 493: > It isn't a pretty fix, but it does the job.

This really seems quite nasty :joy: I hope @vodovozovaliza does not take it as an example of good coding practices.

stavros11 (commented):
Thanks @renatomello for looking into this. 

Another approach could be to add an attribute to the `CircuitResult` object, for example like `self.repeated_execution_probabilities = None` and populate this with the `probabilities` you calculate during execution. Then instead of overriding the `probabilities` method you could modify `backend.circuit_result_probabilities` like:
```py
def circuit_result_probabilities(self, ...):
    if self.repeated_execution_probabilities is not None:
        return self.repeated_execution_probabilities
    # else do the calculation as it is now
```
Overally none of these approaches are optimal, so personally I am fine with what you already did.

I believe the main source of this issue is that `CircuitResult` changes depending on the kind of execution. The cleanest solution would be to return different type (not always `CircuitResult`) depending on the type of execution (eg. with shots, without shots, with repeated execution, etc.). In fact we are discussing something similar in qiboteam/qibolab#400, but this would require a larger refactoring.

src/qibo/backends/numpy.py
renatomello: hahahaha from the start I said it wasn't pretty

renatomello:
> Thanks @renatomello for looking into this.
> 
> Another approach could be to add an attribute to the `CircuitResult` object, for example like `self.repeated_execution_probabilities = None` and populate this with the `probabilities` you calculate during execution. Then instead of overriding the `probabilities` method you could modify `backend.circuit_result_probabilities` like:
> 
> ```python
> def circuit_result_probabilities(self, ...):
>     if self.repeated_execution_probabilities is not None:
>         return self.repeated_execution_probabilities
>     # else do the calculation as it is now
> ```
> 
> Overally none of these approaches are optimal, so personally I am fine with what you already did.
> 
> I believe the main source of this issue is that `CircuitResult` changes depending on the kind of execution. The cleanest solution would be to return different type (not always `CircuitResult`) depending on the type of execution (eg. with shots, without shots, with repeated execution, etc.). In fact we are discussing something similar in [qiboteam/qibolab#400](https://github.com/qiboteam/qibolab/pull/400), but this would require a larger refactoring.

Yeah I agree with the main premise of a larger refactoring, but will let that for the future. For now, I think this will suffice to correct immediate behaviour.

renatomello:
Due to popular demand, I implemented a less nasty solution. The price to pay is the creation of two extra variables in the `CircuitResult` class.

renatomello:
@wilkensJ @vodovozovaliza @stavros11 this is ready for review

renatomello:
@wilkensJ @stavros11 @vodovozovaliza just a reminder that someone needs to review this

vodovozovaliza (approved):

src/qibo/backends/numpy.py
```diff
@@ -467,18 +473,31 @@ def execute_circuit_repeated(self, circuit, initial_state=None, nshots=None):
 
             if circuit.measurements:
                 result = CircuitResult(self, circuit, state, 1)
-                results.append(result.samples()[0])
+                sample = result.samples()[0]
+                results.append(sample)
+                if not circuit.density_matrix:
+                    probabilities += result.probabilities()
+                    frequencies.append(sample[0])
```
wilkensJ, line 480: Two remarks:
1. These are not frequencies yet, they are samples, right?
2. I think the indexing `sample[0]` is not working for 2 qubits and more. See my comment.

tests/test_hamiltonians_trotter.py
```diff
@@ -126,9 +134,14 @@ def test_trotter_hamiltonian_three_qubit_term(backend):
     initial_state = random_statevector(2**4, backend=backend)
     circuit = ham.circuit(dt=dt)
     final_state = backend.execute_circuit(circuit, np.copy(initial_state))
+    if backend.__class__.__name__ == "CupyBackend":
+        mm1 = numpy_backend.cast(mm1.get(), dtype=mm1.dtype)
+        mm2 = numpy_backend.cast(mm2.get(), dtype=mm2.dtype)
+        mm3 = numpy_backend.cast(mm3.get(), dtype=mm3.dtype)
```
stavros11, line 140: I believe in the backends we have a `backend.to_numpy()` method to avoid checking for the backend through the name. Would
```suggestion
    mm1 = backend.to_numpy(mm1.get())
    mm2 = backend.to_numpy(mm2.get())
    mm3 = backend.to_numpy(mm3.get())
```
work in this case?

stavros11 (commented):
Thank you for the updates. I think this solution is better than before. The solution for probabilities and frequencies looks fine to me, except the issue @wilkensJ reported, which I believe is not fixed yet but I have not tested.

Other than that, I see that you are modifying some files (Hamiltonians and derivative) that are not related to the PR. Is there a backend/test problem?

renatomello:
> Other than that, I see that you are modifying some files (Hamiltonians and derivative) that are not related to the PR. Is there a backend/test problem?

https://github.com/qiboteam/qibo/issues/881

src/qibo/backends/numpy.py
```diff
@@ -467,18 +473,31 @@ def execute_circuit_repeated(self, circuit, initial_state=None, nshots=None):
...
+            probabilities /= nshots
+
+            frequencies = self.calculate_frequencies(frequencies)
+
         if circuit.measurements:
             final_result = CircuitResult(self, circuit, state, nshots)
             final_result._samples = self.aggregate_shots(results)
+            if not circuit.density_matrix:
+                final_result._repeated_execution_probabilities = probabilities
+                final_result._repeated_execution_frequencies = frequencies
```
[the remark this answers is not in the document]
renatomello: Both suggestions implemented. Thanks.

renatomello:
I found two more bugs on this, but since they're not related to the main point of this PR, I think their solutions should be a separate PR.

tests/test_hamiltonians_trotter.py
renatomello: This will fail because only the `cupy.ndarray`s have the `.get()` method

stavros11: Sorry, I meant
```suggestion
    mm1 = backend.to_numpy(mm1)
    mm2 = backend.to_numpy(mm2)
    mm3 = backend.to_numpy(mm3)
```
This converts tensors of any backend to numpy arrays, so it will do the `.get()` internally in the `CupyBackend` only, without having to check explicitly with `__class__.__name__`.

tests/test_derivative.py
```diff
@@ -8,7 +8,7 @@
 
 
 # defining an observable
-def hamiltonian(nqubits, backend):
+def hamiltonian(nqubits):
     return hamiltonians.hamiltonians.SymbolicHamiltonian(
         np.prod([Z(i) for i in range(nqubits)])
     )
```
stavros11, line 14: Doing this will ignore the `backend` parameter in the test and will use the default `GlobalBackend()` for all `test_derivative`. The proper solution would be
```py
def hamiltonian(nqubits, backend):
    return hamiltonians.hamiltonians.SymbolicHamiltonian(
        np.prod([Z(i) for i in range(nqubits)]), backend=backend
    )
```
so that all backends are tested.

tests/test_derivative.py
```diff
@@ -31,6 +31,8 @@ def circuit(nqubits=1):
     [(1, [-8.51104358e-02, -5.20075970e-01, 0]), (0.5, [-0.02405061, -0.13560379, 0])],
 )
 def test_derivative(backend, nshots, atol, scale_factor, grads):
+    grads = backend.cast(grads, dtype=float)
```
stavros11, line 34: Also this is not needed with the above fix (at least on main that I tested).

tests/test_models_circuit_features.py
```diff
@@ -314,3 +318,34 @@ def test_repeated_execute_with_noise(backend):
     target_state = np.stack(target_state)
     final_state = [backend.to_numpy(x) for x in final_state]
     backend.assert_allclose(final_state, target_state)
+
+
+def test_repeated_execute_probs_and_freqs(backend):
+    circuit = Circuit(1)
```
stavros11, line 324: Maybe would be good to extend this test to multiqubit circuit, to catch the issue reported by @wilkensJ.

stavros11 (commented):
> @stavros11 yes there are issues with some tests in the `cupy` backend (see #881)

Thanks, indeed these tests are failing on main. I opened #893, we can add a workflow that tests cupy to avoid such issues in the future.

> I found two more bugs on this, but since they're not related to the main point of this PR, I think their solutions should be a separate PR. Agreed?

Yes, if they are not related it is always better to open different PRs.

tests/test_derivative.py
renatomello: Implemented

renatomello:
@stavros11 @wilkensJ after @vodovozovaliza approval and implementing Stavros' suggestions, we are only missing a second approval on this in order to merge it

stavros11 (approved):
Looks good to me, thanks.

PR #883, opened 2023-05-04
Expressibility of parameterized quantum circuits

AlejandroSopena:
This PR implements the calculation of the expressibility [1905.10876](https://arxiv.org/abs/1905.10876) defined as $||A||\_{HS}$ where 
$A=\int_{\text{Haar}}\left(|\psi\rangle\langle\psi|\right)^{\otimes t}d\psi - \int_{\Theta}\left(|\psi_{\theta}\rangle\langle\psi_{\theta}|\right)^{\otimes t}d\psi$,
 and the entangling capability defined as an average for different parameters of the Meyer-Wallach entanglement $Q$,
$Q=1-\frac{1}{N}\sum_{k}\text{Tr}\left(\rho_k^2(\theta)\right)$.

These quantities allow to determine whether a parametrized quantum circuit is sufficiently expressive as well as its capacity to produce entangled states. In general, high expressibility implies high entangling capability, but the converse is not true.

These quantities are also related to the appearance of barren plateaus during optimization. [2101.02138](https://arxiv.org/abs/2101.02138)

Checklist:
- [x] Tests are passing.


src/qibo/quantum_info/metrics.py
```diff
@@ -109,6 +109,35 @@ def entropy(state, base: float = 2, validate: bool = False, backend=None):
     return ent
 
 
+def meyer_wallach(circuit, backend=None):
```
renatomello, line 112: Can you insert this by the end of the file?

renatomello:
@AlejandroSopena I think a new `expressibility` file is not needed. It seems like the two integrals belong in `utils.py`, while `entangling_capacity` belongs to `metrics`. What do you think?

AlejandroSopena:
I was just thinking whether to do that or create a new file. Both options seem good to me, so I have changed it as you suggested.

src/qibo/quantum_info/utils.py
```diff
@@ -227,3 +228,68 @@ def hellinger_fidelity(prob_dist_p, prob_dist_q, validate: bool = False, backend
...
+
+    circuit.density_matrix = True
+    dim = 2**circuit.nqubits
+    randunit_density = backend.cast(np.zeros((dim**t, dim**t), dtype=complex))
+    nparams = 0
+    for gate in circuit.queue:
+        if hasattr(gate, "trainable") and gate.trainable:
+            nparams += len(gate.parameters)
+    for _ in range(samples):
+        params = np.random.uniform(-np.pi, np.pi, nparams)
```
renatomello, line 291: Correct me if I'm wrong, but doesn't sampling all three phases from an uniform distribution create a concentration in the poles of the single-qubit Bloch-sphere?

src/qibo/quantum_info/utils.py
```diff
@@ -227,3 +228,68 @@ def hellinger_fidelity(prob_dist_p, prob_dist_q, validate: bool = False, backend
...
+        circuit.set_parameters(params)
+        rho = backend.execute_circuit(circuit).state()
+        randunit_density += ft.reduce(np.kron, [rho] * t)
+    return randunit_density / samples
```
renatomello, line 295: it's good for readability to have a blank like before the `return` (in all functions)

src/qibo/quantum_info/metrics.py
```diff
@@ -385,3 +385,92 @@ def gate_error(channel, target=None, backend=None):
     error = 1 - average_gate_fidelity(channel, target, backend=backend)
 
     return error
+
+
+def meyer_wallach(circuit, backend=None):
```
renatomello, line 390: I think `meyer_wallach_entanglement` is a better (even though longer) name, because it's clearer

src/qibo/quantum_info/metrics.py
```diff
@@ -385,3 +385,92 @@ def gate_error(channel, target=None, backend=None):
...
+
+    Args:
+        circuit (qibo.models.Circuit): Parametrized circuit.
+        backend (qibo.backends.abstract.Backend): Calculation engine.
+
+    Returns:
+        (int) : Meyer-Wallach entanglement.
+    """
+
+    if backend == None:  # pragma: no cover
```
renatomello, line 404: `is` instead of `==` (and elsewhere)

src/qibo/quantum_info/metrics.py
```diff
@@ -385,3 +385,92 @@ def gate_error(channel, target=None, backend=None):
...
+    res = []
+    nparams = 0
+    for gate in circuit.queue:
+        if hasattr(gate, "trainable") and gate.trainable:
+            nparams += len(gate.parameters)
+    for i in range(samples):
+        params = np.random.uniform(-np.pi, np.pi, nparams)
+        circuit.set_parameters(params)
+        res.append(meyer_wallach(circuit, backend))
+    return 2 * np.sum(backend.cast(res)).real / samples
```
renatomello, line 449: Here, better use `np.real` for backend-related reasons mentioned below

src/qibo/quantum_info/metrics.py
```diff
@@ -385,3 +385,92 @@ def gate_error(channel, target=None, backend=None):
...
+
+    .. math::
+        Q = 1-\\frac{1}{N}\\sum_{k}\\text{Tr}\\left(\\rho_k^2(\\theta_i)\\right) \\,
+
+    Args:
+        circuit (qibo.models.Circuit): Parametrized circuit.
+        backend (qibo.backends.abstract.Backend): Calculation engine.
+
+    Returns:
+        (int) : Meyer-Wallach entanglement.
```
renatomello, line 401: following the rest of the module, return type doesn't need to be inside parenthesis (all functions)

src/qibo/quantum_info/metrics.py
```diff
@@ -385,3 +385,92 @@ def gate_error(channel, target=None, backend=None):
...
+    for j in range(nqubits):
+        trace_q = list(range(nqubits))
+        trace_q.pop(j)
+        rho_r = backend.partial_trace_density_matrix(rho, trace_q, nqubits)
+        trace = np.trace(rho_r @ rho_r)
+        entropy += trace
+    return 1 - entropy / nqubits
+
+
+def entangling_capability(circuit, samples, backend=None):
```
renatomello, line 420: Also following the rest of the module, preferably use type hints when types are basic types (e.g. int, float, str, bool, etc)

renatomello (changes requested):

src/qibo/quantum_info/utils.py
AlejandroSopena: Exactly, that's what happens. The expressivity measures the deviation from a t-design. The distribution of single-qubit states obtained by sampling the 3 phases randomly is not a t-design, but it is closer to being one than a parameterized state that is not able to explore the entire Hilbert space.

AlejandroSopena:
I have already implemented all the suggestions. @renatomello, thank you for the detailed review.

src/qibo/quantum_info/metrics.py
```diff
@@ -385,3 +385,95 @@ def gate_error(channel, target=None, backend=None):
...
+    """Computes the expressibility of the `circuit`, :math:`||A||_{HS}` where
+
+    .. math::
+        A = \\int_{\\text{Haar}}\\left(|\\psi\\rangle\\right.\\left.\\langle\\psi|\\right)^{\\otimes t}d\\psi - \\int_{\\Theta}\\left(|\\psi_{\\theta}\\rangle\\right.\\left.\\langle\\psi_{\\theta}|\\right)^{\\otimes t}d\\psi \\,
+
+    Args:
+        circuit (qibo.models.Circuit): Parametrized circuit.
+        t (int): index t to define the t-design.
+        samples (int): number of samples to estimate the integral.
+        backend (qibo.backends.abstract.Backend): Calculation engine.
```
renatomello, line 464:
```suggestion
        backend (:class:`qibo.backends.abstract.Backend`, optional): backend to be used
            in the execution. If ``None``, it uses ``GlobalBackend()``.
            Defaults to ``None``.
```

src/qibo/quantum_info/utils.py
```diff
@@ -227,3 +228,74 @@ def hellinger_fidelity(prob_dist_p, prob_dist_q, validate: bool = False, backend
...
+    """Computes the integral over pure states over the Haar measure.
+
+    .. math::
+        \\int_{\\text{Haar}}\\left(|\\psi\\rangle\\right.\\left.\\langle\\psi|\\right)^{\\otimes t}d\\psi \\,
+
+    Args:
+        nqubits (int): Number of qubits.
+        t (int): index t to define the t-design.
+        samples (int): number of samples to estimate the integral.
+        backend (qibo.backends.abstract.Backend): Calculation engine.
```
renatomello, line 243:
```suggestion
        backend (:class:`qibo.backends.abstract.Backend`, optional): backend to be used
            in the execution. If ``None``, it uses ``GlobalBackend()``.
            Defaults to ``None``.
```

src/qibo/quantum_info/utils.py
```diff
@@ -227,3 +228,74 @@ def hellinger_fidelity(prob_dist_p, prob_dist_q, validate: bool = False, backend
...
+        parameter space described by the parameterized `circuit`.
+
+    .. math::
+        \\int_{\\Theta}\\left(|\\psi_{\\theta}\\rangle\\right.\\left.\\langle\\psi_{\\theta}|\\right)^{\\otimes t}d\\psi \\,
+
+    Args:
+        circuit (qibo.models.Circuit): Parametrized circuit.
+        t (int): index t to define the t-design.
+        samples (int): number of samples to estimate the integral.
+        backend (qibo.backends.abstract.Backend): Calculation engine.
```
renatomello, line 278:
```suggestion
        backend (:class:`qibo.backends.abstract.Backend`, optional): backend to be used
            in the execution. If ``None``, it uses ``GlobalBackend()``.
            Defaults to ``None``.
```

renatomello (changes requested):

src/qibo/quantum_info/metrics.py
```diff
@@ -385,3 +385,101 @@ def gate_error(channel, target=None, backend=None):
...
+        backend = GlobalBackend()
+
+    res = []
+    nparams = 0
+    for gate in circuit.queue:
+        if hasattr(gate, "trainable") and gate.trainable:
+            nparams += len(gate.parameters)
+    for i in range(samples):
+        params = np.random.uniform(-np.pi, np.pi, nparams)
+        circuit.set_parameters(params)
```
vodovozovaliza, line 452: `nparams` value is already stored in a circuit as `circuit.trainable_gates.nparams`, so I the there is no need to go over the whole `circuit.queue`. Also, the Identity gate is considered `trainable` in Qibo so, if I understand correctly, it is better to avoid changing its parameters. The solution I would propose is:
```python
params = {
    gate: np.random.uniform(-np.pi, np.pi, gate.nparams)
    for gate in circuit.trainable_gates
    if isinstance(gate, gates.I) is False
}
```

AlejandroSopena: Thanks @vodovozovaliza, this is much better. I will make the change when PR #895 is merged (also thank you for that).

src/qibo/quantum_info/metrics.py
```diff
@@ -385,3 +385,101 @@ def gate_error(channel, target=None, backend=None):
...
+
+    circuit.density_matrix = True
+    nqubits = circuit.nqubits
+    rho = backend.execute_circuit(circuit).state()
+    entropy = 0
+    for j in range(nqubits):
+        trace_q = list(range(nqubits))
+        trace_q.pop(j)
+        rho_r = backend.partial_trace_density_matrix(rho, trace_q, nqubits)
+        trace = np.trace(rho_r @ rho_r)
```
renatomello, line 417:
```suggestion
        trace = purity(rho_r)
```

renatomello:
@AlejandroSopena the test for `entangling_capability` on `numba` is 20x slower than on `tensorflow` and `cupy`; the `expressibility` test is 10x slower. Can you check why?

renatomello:
> @AlejandroSopena the test for `entangling_capability` on `numba` is 20x slower than on `tensorflow` and `cupy`; the `expressibility` test is 10x slower. And `numba` is using all cores at 100%. Can you check why?

renatomello (approved):

tests/test_quantum_info_metrics.py
```diff
@@ -238,3 +240,48 @@ def test_process_fidelity(backend):
...
+    c2 = Circuit(nqubits)
+    c2.add(gates.H(0))
+    c2.add(gates.CNOT(0, 1))
+    c2.add(gates.RX(0, 0, trainable=True))
+    ent_mw2 = entangling_capability(c2, samples, backend)
+
+    c3 = Circuit(nqubits)
+    ent_mw3 = entangling_capability(c3, samples, backend)
+
+    assert ent_mw3 < ent_mw1 < ent_mw2
```
renatomello, line 264:
```suggestion
    backend.assert_allclose(ent_mw3 < ent_mw1 < ent_mw2, True)
```

tests/test_quantum_info_metrics.py
```diff
@@ -238,3 +240,48 @@ def test_process_fidelity(backend):
...
+    c2 = Circuit(nqubits)
+    c2.add(gates.H(0))
+    c2.add(gates.CNOT(0, 1))
+    c2.add(gates.RX(0, 0, trainable=True))
+    expr_2 = expressibility(c2, t, samples, backend)
+
+    c3 = Circuit(nqubits)
+    expr_3 = expressibility(c3, t, samples, backend)
+
+    assert expr_1 < expr_2 < expr_3
```
renatomello, line 287:
```suggestion
    backend.assert_allclose(expr_1 < expr_2 < expr_3, True)
```

src/qibo/quantum_info/metrics.py
```diff
@@ -385,3 +385,101 @@ def gate_error(channel, target=None, backend=None):
...
+        samples (int): number of samples to estimate the integral.
+        backend (:class:`qibo.backends.abstract.Backend`, optional): backend to be used
+            in the execution. If ``None``, it uses ``GlobalBackend()``.
+            Defaults to ``None``.
+
+    Returns:
+        float: Entangling capability.
+    """
+
+    from qibo.quantum_info import haar_integral, pqc_integral
```
renatomello, line 476:
```suggestion
    from qibo.quantum_info.utils import haar_integral, pqc_integral
```

src/qibo/quantum_info/metrics.py
```diff
@@ -385,3 +385,101 @@ def gate_error(channel, target=None, backend=None):
...
+    nparams = 0
+    for gate in circuit.queue:
+        if hasattr(gate, "trainable") and gate.trainable:
+            nparams += len(gate.parameters)
+    for i in range(samples):
+        params = np.random.uniform(-np.pi, np.pi, nparams)
+        circuit.set_parameters(params)
+        res.append(meyer_wallach_entanglement(circuit, backend))
+
+    return 2 * np.real(np.sum(backend.cast(res))) / samples
```
renatomello, line 455: it seems like you don't need the `cast` here since `res` is a list of `float`s and the return of this function is also a `float`

tests/test_quantum_info_metrics.py
```diff
@@ -238,3 +240,48 @@ def test_process_fidelity(backend):
...
     backend.assert_allclose(gate_error(channel, backend=backend), 0.0)
     backend.assert_allclose(gate_error(channel, channel, backend=backend), 0.0)
+
+
+def test_entangling_capability(backend):
+    nqubits = 2
+    samples = 500
+
+    c1 = Circuit(nqubits)
+    c1.add([gates.RX(0, 0, trainable=True) for i in range(nqubits)])
```
renatomello, line 250: Here and in the other test function you're adding `I` twice to the same qubit before and after the `CNOT`. Why?

tests/test_quantum_info_metrics.py
```diff
@@ -238,3 +240,48 @@ def test_process_fidelity(backend):
...
+
+def test_expressibility(backend):
+    nqubits = 2
+    samples = 500
+    t = 1
+
+    c1 = Circuit(nqubits)
+    c1.add([gates.RX(0, 0, trainable=True) for i in range(nqubits)])
+    c1.add(gates.CNOT(0, 1))
+    c1.add([gates.RX(0, 0, trainable=True) for i in range(nqubits)])
```
renatomello, line 275:
```suggestion
    c1.add([gates.RX(0, 0, trainable=True) for _ in range(nqubits)])
    c1.add(gates.CNOT(0, 1))
    c1.add([gates.RX(0, 0, trainable=True) for _ in range(nqubits)])
```

renatomello, line 275:
```suggestion
    c1.add([gates.RX(0, 0, trainable=True) for _ in range(nqubits)])
    c1.add(gates.CNOT(0, 1))
    c1.add([gates.RX(0, 0, trainable=True) for _ in range(nqubits)])
```

tests/test_quantum_info_metrics.py
renatomello, line 250:
```suggestion
    c1.add([gates.RX(0, 0, trainable=True) for _ in range(nqubits)])
```

renatomello, line 250:
```suggestion
    c1.add([gates.RX(0, 0, trainable=True) for _ in range(nqubits)])
```

src/qibo/quantum_info/utils.py
```diff
@@ -227,3 +228,78 @@ def hellinger_fidelity(prob_dist_p, prob_dist_q, validate: bool = False, backend
...
+
+    return rand_unit_density / samples
+
+
+def pqc_integral(circuit, t: int, samples: int, backend=None):
+    """Computes the integral over pure states generated by uniformly sampling in the
+        parameter space described by the parameterized `circuit`.
+
+    .. math::
+        \\int_{\\Theta}\\left(|\\psi_{\\theta}\\rangle\\right.\\left.\\langle\\psi_{\\theta}|\\right)^{\\otimes t}d\\psi \\,
```
renatomello, line 274:
```suggestion
        \\int_{\\Theta} d\\psi \\, \\left(|\\psi_{\\theta}\\rangle\\right.\\left.\\langle
            \\psi_{\\theta}|\\right)^{\\otimes t}
```

src/qibo/quantum_info/utils.py
```diff
@@ -227,3 +228,78 @@ def hellinger_fidelity(prob_dist_p, prob_dist_q, validate: bool = False, backend
     distance = hellinger_distance(prob_dist_p, prob_dist_q, validate, backend=backend)
 
     return (1 - distance**2) ** 2
+
+
+def haar_integral(nqubits, t: int, samples: int, backend=None):
+    """Computes the integral over pure states over the Haar measure.
+
+    .. math::
+        \\int_{\\text{Haar}}\\left(|\\psi\\rangle\\right.\\left.\\langle\\psi|\\right)^{\\otimes t}d\\psi \\,
```
renatomello, line 237:
```suggestion
        \\int_{\\text{Haar}} d\\psi \\, \\left(|\\psi\\rangle\\right.\\left.\\langle
            \\psi|\\right)^{\\otimes t}
```

renatomello (dismissed):

AlejandroSopena:
@renatomello, on my computer, `numba` is comparable to `numpy` (1.5 times slower) and faster than `cupy`, `cuquantum` and `tensorflow` for both tests. After profiling the code, I have confirmed that there is no bottleneck. 75% of the time is spent calling `backend.execute_circuit()`. I am using the current main of qibojit.

renatomello:
> @renatomello, on my computer, `numba` is comparable to `numpy` (1.5 times slower) and faster than `cupy`, `cuquantum` and `tensorflow` for both tests. After profiling the code, I have confirmed that there is no bottleneck. 75% of the time is spent calling `backend.execute_circuit()`. I am using the current main of qibojit.

Which `numba` version are you using?

AlejandroSopena:
`numba==0.56.4`

renatomello:
> `numba==0.56.4`

I upgraded `qibojit` to latest and `numba` to `0.57` and the issue went away locally for me

renatomello:
@AlejandroSopena I think everything is almost good now. Last thing I'd ask you is to create separate tests in `test_quantum_info_utils.py` for `haar_integral` and `pqc_integral`, as well as a separate test for `meyer_wallach_entanglement` in `test_quantum_info_metrics.py`. Aftet tha, it's good to go.

src/qibo/quantum_info/utils.py
```diff
@@ -227,3 +228,115 @@ def hellinger_fidelity(prob_dist_p, prob_dist_q, validate: bool = False, backend
...
+    dim = 2**circuit.nqubits
+
+    rand_unit_density = np.zeros((dim**t, dim**t), dtype=complex)
+    rand_unit_density = backend.cast(rand_unit_density, dtype=rand_unit_density.dtype)
+    for _ in range(samples):
+        params = {
+            gate: np.random.uniform(-np.pi, np.pi, gate.nparams)
+            for gate in circuit.trainable_gates
+            if isinstance(gate, I) is False
+        }
```
vodovozovaliza, line 331: Same as above:
```suggestion
        params = np.random.uniform(-np.pi, np.pi, circuit.trainable_gates.nparams)
        circuit.set_parameters(params)
```

renatomello (approved):

vodovozovaliza (approved):

Issue #881, opened 2023-05-04
Multiple tests do not work for `cupy` backend


renatomello:
List:

- `test_derivative.py`
- `test_hamiltonians_terms.py`
- `test_hamiltonians_trotter.py`

PR #894, opened 2023-05-10
Workflow for selfhosted GPU (testing)

stavros11:
Testing if it works using the workflows repository so that we can share among all projects.

Checklist:
- [ ] Tests are passing.


stavros11:
@scarrazza @AleCandido the workflow seems to be working, however there are still some issues:

1. `Permission denied` when copying the tests to the `/tmp/` folder, which is needed to get the coverage. Despite that the tests are still executed.
2. `INTERNALERROR> sqlite3.OperationalError: unable to open database file` after the tests finish. Most likely it is related to 1.

The same workflow worked without issues in qibolab (qiboteam/qibolab#409).

scarrazza:
> GPU is not used even though I used --gpus-per-node 1 and installed cupy and cuquantum. If I execute the tests from my terminal I get the GPU with the same approach.

Does this work if you send a pytest job manually applying those flags?

stavros11:
> Does this work if you send a pytest job manually applying those flags?

Yes, when I use `--gpus-per-node 1` I get 6801 tests collected, while without it I get 4042 like the workflow. Also, when using GPU I get the failures reported in #881 so it is really running on GPU.

I just reran this and strangely 1, 2 were fixed, I am not sure how.

tests-slurm
AleCandido, line 1: In principle this is not specific to `slurm` in any way (or am I wrong?), so I would remove from its name.

PR #899, opened 2023-05-16
Numerical methods for calculating gradients

MatteoRobbiati:
This PR starts the re-factor of the derivative module.
I am adding here a simple central finite difference method for calculating derivatives.

In principle I could write a single function `numerical`, with methods `parameter_shift_rule` and `finite_differences`, but I think it is useful to separate the two ways to calculate derivatives, because we have in mind to implement the general parameter shift rule in the next future. At that point, the "standard" shift rule will be a remarkable case of the general one.

TO DO:

- [x] add tests;
- [ ] discuss about the derivative (or gradients?) module shape: a gradient class with methods? a collection of functions?

Checklist:

 - [ ] Tests are passing.


src/qibo/derivative.py
```diff
@@ -158,3 +158,74 @@ def circuit(nqubits = 1):
...
+        step_size (float): step size used to evaluate the finite difference
+            (default 1e-7).
+
+    Returns:
+        (float): Value of the derivative of the expectation value of the hamiltonian
+            with respect to the target variational parameter.
+    """
+
+    if parameter_index > len(circuit.get_parameters()):
+        raise_error(ValueError, """This index is out of bounds.""")
```
andrea-pasquale, line 196: Do we really need this check?
I think that the code will fail with a similar message without the error.

src/qibo/derivative.py
```diff
@@ -158,3 +158,74 @@ def circuit(nqubits = 1):
...
+    shifted[parameter_index] += step_size
+    circuit.set_parameters(shifted)
+
+    # forward evaluation
+    forward = hamiltonian.expectation(
+        backend.execute_circuit(circuit=circuit, initial_state=initial_state).state()
+    )
+
+    # backward shift and evaluation
+    shifted[parameter_index] -= 2 * step_size
```
andrea-pasquale, line 220: Here I am guessing that you are doin ` - 2 * step` since before the same parameter in shifted was incremented by ` + step`. Perhaps it is better to define two shifted variables: `backward_shifted` and `forward_shifted`.

src/qibo/derivative.py
```diff
@@ -158,3 +158,74 @@ def circuit(nqubits = 1):
...
+    if not isinstance(hamiltonian, AbstractHamiltonian):
+        raise_error(
+            TypeError,
+            "hamiltonian must be a qibo.hamiltonians.Hamiltonian or qibo.hamiltonians.SymbolicHamiltonian object",
+        )
+
+    backend = hamiltonian.backend
+
+    # parameters copies
+    parameters = np.asarray(circuit.get_parameters()).copy()
```
andrea-pasquale, line 207: Is this copy necessary?

src/qibo/derivative.py
```diff
@@ -158,3 +158,74 @@ def circuit(nqubits = 1):
...
+    shifted[parameter_index] -= 2 * step_size
+    circuit.set_parameters(shifted)
+
+    backward = hamiltonian.expectation(
+        backend.execute_circuit(circuit=circuit, initial_state=initial_state).state()
+    )
+
+    circuit.set_parameters(parameters)
+
+    result = float((forward - backward) / (2 * step_size))
```
andrea-pasquale, line 229: If you force it to float you may lose some information on the backend that you are using. But I am not to sure. Did you run tests also on gpu?

andrea-pasquale (commented):
Thanks @MatteoRobbiati.
It looks good to me. I have a few suggestions.

src/qibo/derivative.py
MatteoRobbiati: This is useful if we want to define asymmetric step_sizes, right? I wanted to implement the standard finite difference method here, but of course we can generalize it.

src/qibo/derivative.py
MatteoRobbiati: I think yes, because we cannot get parameters from circuit after the shifts.
Otherwise, of course, we can shift them again to recover the starting configuration.

src/qibo/derivative.py
andrea-pasquale: Maybe instead of  `This index` you can use a f-string to print the wrong index directly.

andrea-pasquale (approved):

PR #901, opened 2023-05-17
Adding Iterative Quantum Amplitude Estimation method

gmlejarza:
Checklist:
- [ ] Tests are passing.


scarrazza:
@gmlejarza thanks for this. Could you please fix tests/linter before we start the review?

scarrazza:
@gmlejarza just 2 quick tips:
- you can install pre-commit in your local machine so your commit history is much more cleaner.
- when you perform a commit please use a sensible message instead of "add files via upload".

Both points are important in order to maintain a proper and understandable history of changes.

MatteoRobbiati (changes requested):
Thank you @gmlejarza for this. The code seems to me to be in a very good shape.
Still, tests are missing. 
As you can see from the codecov report, it is important to add tests which cover the code you write. 

Typically, we add specific tests [here](https://github.com/qiboteam/qibo/tree/master/tests): you can create a file like `test_iqae.py` in which you run parts of your model (or a complete execution). Then, you can run pytest locally as exaplained [here](https://docs.pytest.org/en/7.3.x/); you just need to run:

```
pytest my_test.py
```

after writing your `my_test.py` file
Tests are then performed automatically by our workflow when you push new commits and you can see the reports in your PR. 

Furthermore, the automatic codecov check of your status highlights the lines of code which need to be covered by tests.

Let me know if you need any kind of help in doing this :)

src/qibo/models/iqae.py
```diff
@@ -0,0 +1,385 @@
...
+
+        return result
+
+
+# THE CLASS WITH THE RESULTS
+class IterativeAmplitudeEstimationResult:
+    """The ``IterativeAmplitudeEstimation`` result object."""
+
+    def __init__(self):
+        super().__init__()
```
Edoardo-Pedicillo, line 375: `IterativeAmplitudeEstimationResult` is not a derived class, you can drop this line
```suggestion

```
If you want you can use [dataclasses](https://docs.python.org/3/library/dataclasses.html)

Edoardo-Pedicillo (changes requested):
Thank @gmlejarza, you will find below some comments. Could you install and use pre-commit locally (https://pre-commit.com/#install)?

src/qibo/models/iqae.py
```diff
@@ -0,0 +1,385 @@
...
+            ):
+                up = True
+                return (K, up)
+            elif (
+                theta_max >= 1 / 2
+                and theta_min >= 1 / 2
+                and int(K * theta_u) == int(K * theta_l)
+            ):
+                up = False
+                return (K, up)
```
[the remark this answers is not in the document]
gmlejarza: Thanks for the suggestion but I think there was no problem for the specific case you mentioned because the function returns `up_i` instead of `up` in that case. Anyhow, thanks to that suggestion I found a more compact way to do the same. The corrected function will read as:

    def find_next_k(self, K_i, up_i, theta_l, theta_u, r=2):
        """Find the largest integer K such that the interval K*[theta_l,theta_u]
        lies completely in [0, pi] or [pi, 2pi].

        Args:
            K_i: the current K such K=4k+2, where k is the power of the Q operator
            up_i: boolean flag of whether theta_interval lies in the
                upper half-circle [0, pi] or in the lower one [pi, 2pi]
            theta_l: the current lower limit of the confidence interval for the angle theta
            theta_u: the current upper limit of the confidence interval for the angle theta
            r: lower bound for K

        Returns:
            The next power K_i, and boolean flag for the extrapolated interval
        """
        K_max = int(1 / (2 * (theta_u - theta_l)))
        K = K_max - (K_max - 2) % 4
        while K >= r * K_i:
            theta_min = K * theta_l - int(K * theta_l)
            theta_max = K * theta_u - int(K * theta_u)
            if int(K * theta_u) == int(K * theta_l):
                if(theta_max <= 1 / 2 and theta_min <= 1 / 2):
                    up=True
                    return (K,  up)
                elif(theta_max >= 1 / 2 and theta_min >= 1 / 2):
                    up=False
                    return (K,  up)
            K -= 4
        return (K_i, up_i)

src/qibo/models/iqae.py
```diff
@@ -0,0 +1,385 @@
...
+        Returns:
+            An IterativeAmplitudeEstimation results object.
+        """
+        if backend is None:
+            from qibo.backends import GlobalBackend
+
+            backend = GlobalBackend()
+        # Initializing all parameters
+        K = [2]
+        k = [0]
```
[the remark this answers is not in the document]
gmlejarza: You have a point there. But I was following the notation from the original paper where they call the variables in that way. So for an experienced reader, it will be easier to establish a relationship between the paper and the code. Do you think `k` and `capital_k` would be better, for example?

src/qibo/models/iqae.py
```diff
@@ -0,0 +1,384 @@
...
+            #Here the circuits A and Q are defined
+            ...
+            iae = IQAE(circuit_A=A, circuit_Q=Q)
+            results=iae.execute()
+            print(results.estimation)
+    """
+
+    def __init__(
+        self,
+        circuit_A=None,
```
Edoardo-Pedicillo, line 59: > Function names should be lowercase, with words separated by underscores as necessary to improve readability.

From https://peps.python.org/pep-0008/#function-and-variable-names ...
```suggestion
        circuit_a=None,
```

src/qibo/models/iqae.py
```diff
@@ -0,0 +1,384 @@
...
+                uppercase_k[-1], up[-1], theta_l, theta_u
+            )
+            k_i = int((uppercase_k_i - 2) / 4)
+            uppercase_k.append(uppercase_k_i)
+            up.append(up_i)
+            k.append(k_i)
+            if uppercase_k_i > int(max_L / eps):
+                n_shots_i = int((max_L / eps) * self.n_shots / uppercase_k_i / 10)
+                if n_shots_i == 0:
+                    n_shots_i = 1
```
Edoardo-Pedicillo, line 298: > Maybe it is better to rename the first N_shots_i. Please, put a comment explaining these lines.

scarrazza:
@gmlejarza thanks. I have merged master, lets then check the error.

scarrazza (approved):
Thanks @gmlejarza, looks good.

PR #904, opened 2023-05-18
Add elements order to `pauli_basis`

vodovozovaliza:
Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/quantum_info/basis.py
+++ b/src/qibo/quantum_info/basis.py
@@ -21,36 +21,30 @@ def pauli_basis(
 ):
     """Creates the ``nqubits``-qubit Pauli basis.
 
-        Args:
-            nqubits (int): number of qubits.
-            normalize (bool, optional): If ``True``, normalized basis is returned.
-                Defaults to False.
-            vectorize (bool, optional): If ``False``, returns a nested array with
-                all Pauli matrices. If ``True``, retuns an array where every
-                row is a vectorized Pauli matrix. Defaults to ``False``.
-            sparse (bool, optional) If ``True``, retuns Pauli basis in a sparse
-                representation. Default is ``False``.
-            order (str, optional): If ``"row"``, vectorization of Pauli basis is
-                performed row-wise. If ``"column"``, vectorization is performed
-                column-wise. If ``"system"``, system-wise vectorization is
-                performed. If ``vectorization=False``, then ``order=None`` is
-                forced. Default is ``None``.
-    <<<<<<< HEAD
-            pauli_order (str, optional): corresponds to the order of 4 single-qubit
-                Pauli elements. Default is "IXYZ".
-            backend (``qibo.backends.abstract.Backend``, optional): backend to be
-                used in the execution. If ``None``, it uses ``GlobalBackend()``.
-                Defaults to ``None``.
-    =======
-            backend (:class:`qibo.backends.abstract.Backend`, optional): backend
-                to be used in the execution. If ``None``, it uses
-                :class:`qibo.backends.GlobalBackend`. Defaults to ``None``.
-    >>>>>>> master
+    Args:
+        nqubits (int): number of qubits.
+        normalize (bool, optional): If ``True``, normalized basis is returned.
+            Defaults to False.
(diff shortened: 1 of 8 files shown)


renatomello (approved):

src/qibo/quantum_info/random_ensembles.py
```diff
@@ -363,9 +384,11 @@ def random_density_matrix(
...
         metric (str, optional): metric to sample the density matrix from. Options:
             ``"Hilbert-Schmidt"`` and ``"Bures"``. Default is ``"Hilbert-Schmidt"``.
-        basis (str, optional): if ``"pauli"``, return random density matrix in the
-            Pauli basis. If ``None``, returns it in the computational basis.
-            Default is ``None``.
+        basis (str, optional): if ``None``, returns random density matrix in the
+            computational basis. If ``"pauli"``, returns it in the Pauli basis.
+            If "pauli-<pauli_order>", (e.g. "pauli-IZXY"), returns it in the Pauli basis
+            with the corresponding order of single-qubit Pauli elements
+            (see :func:`qibo.quantum_info.pauli_basis`). Default is ``None``.
```
stavros11, line 391: If the only possible basis are `None` (for the computational basis) and `pauli-<pauli_order>`, why don't you ask for the `pauli_order` directly, as you are doing in all other functions? And if it is `None` you can fall back to the computational.

This way you will avoid the complicated string manipulations and checks in the function's body. Unless you are planning to extend this to other basis or I am missing something.

On another note, not related to this PR, I see that type annotations are used in the function arguments. If you want to be strict about the types, if an argument can be `None`, you should use
```py
from typing import Optional

rank: Optional[int] = None
```
instead of 
```
rank: int = None
```
but as I said this is not required for this PR.

stavros11 (commented):
Thanks, looks good to me. Just a curiosity below and I will approve.

stavros11 (commented):
Thanks, looks good to me. Just a curiosity below and I will approve.

src/qibo/quantum_info/random_ensembles.py
renatomello: I liked the first suggestion and implemented it.

Also agree with the second one, but will do in a different PR.

src/qibo/quantum_info/basis.py
```diff
@@ -16,32 +16,41 @@ def pauli_basis(
...
+            pauli_order (str, optional): corresponds to the order of 4 single-qubit
+                Pauli elements. Default is "IXYZ".
+            backend (``qibo.backends.abstract.Backend``, optional): backend to be
+                used in the execution. If ``None``, it uses ``GlobalBackend()``.
+                Defaults to ``None``.
+    =======
+            backend (:class:`qibo.backends.abstract.Backend`, optional): backend
+                to be used in the execution. If ``None``, it uses
+                :class:`qibo.backends.GlobalBackend`. Defaults to ``None``.
+    >>>>>>> master
```
stavros11, line 48: I think here a merge went wrong.

src/qibo/quantum_info/basis.py
```diff
@@ -118,54 +140,63 @@ def comp_basis_to_pauli(
...
+            pauli_order (str, optional): corresponds to the order of 4 single-qubit
+                Pauli elements. Default is "IXYZ".
+            backend (``qibo.backends.abstract.Backend``, optional): backend to be
+                used in the execution. If ``None``, it uses ``GlobalBackend()``.
+                Defaults to ``None``.
+    =======
+            backend (:class:`qibo.backends.abstract.Backend`, optional): backend to be
+                used in the execution. If ``None``, it uses
+                :class:`qibo.backends.GlobalBackend`. Defaults to ``None``.
+    >>>>>>> master
```
stavros11, line 194: Here too.

src/qibo/quantum_info/basis.py
```diff
@@ -199,39 +237,54 @@ def pauli_to_comp_basis(
...
```
stavros11, line 272: Here too.

renatomello:
@stavros11 fixed the docs

stavros11 (approved):

Issue #898, opened 2023-05-15
`circuit.with_noise` bug


MatteoRobbiati:
I was following the [documentation example](https://qibo.science/qibo/stable/api-reference/qibo.html#qibo.models.circuit.Circuit.with_noise) for building a noisy circuit. 
By running the following:

```python
import numpy as np
import matplotlib.pyplot as plt

from qibo import gates, set_backend
from qibo.models import Circuit

set_backend('numpy')

nqubits = 1
noise_map = {0: (0.1, 0.0, 0.2)}

c = Circuit(nqubits)
c.add(gates.RZ(q=0, theta=0))
c.add(gates.RZ(q=0, theta=0))
c.add(gates.RY(q=0, theta=0))
c.add(gates.RY(q=0, theta=0))
c.add(gates.M(0))

noisy_c = c.with_noise(noise_map)
```

I get this:

```
File "/home/matteo/Documents/PhD/qibo/src/qibo/models/circuit.py", line 505, in <listcomp>
if q in noise_map and sum([row[1] for row in noise_map[q]]) > 0:
TypeError: 'float' object is not subscriptable
```

I think is related to the new way to use the noise channels.

renatomello:
Yes, this is correct. `latest` documentation has the new way of implementing `with_noise` if you're using `qibo` from `master`. This old way still works if you're using `qibo` from `stable` version. After next release, it will stop working.

However, there is indeed something wrong with the `latest` documentation, in which the example is changed to the new way, but not the text. I'll fix that.

MatteoRobbiati:
> Yes, this is correct. `latest` documentation has the new way of implementing `with_noise()` method if you're using `qibo` from `master`. This old way still works if you're using `qibo` from `stable` version. After next release, it will stop working.

Okj, thank you. It makes sense, I use to work woth the `master` branch and `qibo` installed in editable mode.

> However, there is indeed something wrong with the `latest` documentation, in which the example is changed to the new way, but not the text. I'll fix that.

Is there any specific issue opened? I also can have a look.

> Also, there's a more complex issue that a method called `with_noise()` is only able to implement one very specific channel, but that's a different issue.

I understand, you mean it should be better to generalize this feature to different noise models?

renatomello:
> Is there any specific issue opened? I also can have a look.

I'm already going to make other small changes in the doc, so I can just do it all in one PR.

> I understand, you mean it should be better to generalize this feature to different noise models?

Correct.

[...]

PR #936, opened 2023-06-14
Rename `Circuit.with_noise` method as `Circuit.with_pauli_noise`

renatomello:
Since the `Circuit.with_noise` method works exclusively with `PauliNoiseChannel`, the method is better named as `Circuit.with_pauli_noise` in order to not create any confusion about using the method with other noise channels.

This was discussed in #898.

Checklist:
- [x] Tests are passing.


MatteoRobbiati (approved):
Thank you @renatomello. I had the same impression when I raised the issue some time ago, good catch.

vodovozovaliza (approved):

PR #937, opened 2023-06-15
Generalize `quantum_info.random_clifford` to `n` qubits

renatomello:
Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/backends/__init__.py
+++ b/src/qibo/backends/__init__.py
@@ -99,6 +99,7 @@ class QiboMatrices:
         self.S = self.matrices.S
         self.SDG = self.matrices.SDG
         self.CNOT = self.matrices.CNOT
+        self.CZ = self.matrices.CZ
 
 
 matrices = QiboMatrices()
--- a/src/qibo/quantum_info/random_ensembles.py
+++ b/src/qibo/quantum_info/random_ensembles.py
@@ -963,3 +963,42 @@ def _clifford_unitary(phase, x, y, z):
             ],
         ]
     )
+
+
+def _sample_from_quantum_mallows_distribution(nqubits, seed=None, backend=None):
+    if (
+        seed is not None
+        and not isinstance(seed, int)
+        and not isinstance(seed, np.random.Generator)
+    ):
+        raise_error(
+            TypeError, "seed must be either type int or numpy.random.Generator."
+        )
+
+    if backend is None:  # pragma: no cover
+        backend = GlobalBackend()
+
+    local_state = (
+        np.random.default_rng(seed) if seed is None or isinstance(seed, int) else seed
+    )
+
+    mute_index = list(range(nqubits))
+
+    exponents = backend.cast(list(range(nqubits, 0, -1)), dtype=int)
+
+    r = local_state.uniform(0, 1, size=nqubits)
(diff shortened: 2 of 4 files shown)


renatomello:
@igres26 @vodovozovaliza @wilkensJ @stavros11 this is ready for review

vodovozovaliza (approved):
Thank you @renatomello, everything looks good to me

renatomello:
@igres26 @wilkensJ we noticed that we do not need a flag to fuse single-qubit gates, since the class `qibo.models.Circuit` already has a `.fuse()` method.

renatomello:
> @igres26 @wilkensJ: @vodovozovaliza and I noticed that we do not need a flag to fuse single-qubit gates, since the class `qibo.models.Circuit` already has a `.fuse()` method.

```python3
clifford = random_clifford(1).fuse()
```

already does the trick.

renatomello:
@wilkensJ @vodovozovaliza @igres26 @stavros11 there's also another peculiarity that I forgot to mention during the meeting: the algorithm generates circuits with both `CZ`s and `CNOT`s. Do we force the conversion of one into the other? If yes, which one? Or do we leave as is?

stavros11 (approved):
Thanks for this, looks good to me.

> @igres26 @wilkensJ: @vodovozovaliza and I noticed that we do not need a flag to fuse single-qubit gates, since the class `qibo.models.Circuit` already has a `.fuse()` method. So,
> 
> ```python
> clifford = random_clifford(1).fuse()
> ```
> 
> already does the trick.

Just a note, if the circuit has more than one qubit, I think `fuse()` will fuse up to two-qubit gates by default. You can still do `.fuse(max_qubits=1)` to fuse only single-qubit gates.

> @wilkensJ @vodovozovaliza @igres26 @stavros11 there's also another peculiarity that I forgot to mention during the meeting: the algorithm generates circuits with both `CZ`s and `CNOT`s. Do we force the conversion of one into the other? If yes, which one? Or do we leave as is?

Personally, I have no preference, as I am not sure what will be the application of this. If the goal is to execute these circuits on hardware, as far as I know, all TII hardware uses CZ as the native two-qubit gate. But even if you have CNOTs, the transpiler in qibolab will automatically convert them to CZ before executing. If you want to disable the transpiler and manually create circuits that only have native gates, then you may need to convert everything to CZ as you say.

renatomello:
> Just a note, if the circuit has more than one qubit, I think `fuse()` will fuse up to two-qubit gates by default. You can still do `.fuse(max_qubits=1)` to fuse only single-qubit gates.

The idea was to have a flag to merge single-qubit gates when sampling single-qubit random Cliffords anyways, so the `fuse`method as is still does the job.

> Personally, I have no preference, as I am not sure what will be the application of this. If the goal is to execute these circuits on hardware, as far as I know, all TII hardware uses CZ as the native two-qubit gate. But even if you have CNOTs, the transpiler in qibolab will automatically convert them to CZ before executing. If you want to disable the transpiler and manually create circuits that only have native gates, then you may need to convert everything to CZ as you say.

I'm gonna merge this PR, but still wanna hear about other people's opinions one this, In any case, if any changes are needed, we can put it in another PR.

PR #939, opened 2023-06-19
Adding qfiae tutorial

gmlejarza:
Checklist:
- [ ] Tests are passing.


examples/qfiae/utils_qfiae.py
```diff
@@ -0,0 +1,108 @@
+import matplotlib.pyplot as plt
+import numpy as np
+
+from qibo import gates
+from qibo.models import Circuit
+from qibo.models.iqae import IQAE
+
+
+def fourier_series(coeffs, x_val, period=2 * np.pi):
+    """Computes the Fourier series for a given set of coefficients in exponencial form."""
```
MatteoRobbiati, line 10: Recalling a discussion recently had in `Qibolab`, we should try to uniform the docstrings. 
Since we are referring to the [PEP-257 lineguides](https://peps.python.org/pep-0257/), we should write descriptions according to:

_The docstring is a phrase ending in a period. It prescribes the function or method’s effect as a command (“Do this”, “Return that”), not as a description; e.g. don’t write “Returns the pathname …”._

In a few words, please change all the third persons "Computes, ..." into imperative mood "Compute, ...".

examples/qfiae/utils_qfiae.py
```diff
@@ -0,0 +1,108 @@
...
+
+from qibo import gates
+from qibo.models import Circuit
+from qibo.models.iqae import IQAE
+
+
+def fourier_series(coeffs, x_val, period=2 * np.pi):
+    """Computes the Fourier series for a given set of coefficients in exponencial form."""
+
+    series = [complex(0, 0) for _ in range(len(x_val))]
```
MatteoRobbiati, line 12: I think here you can use something more clear, like:

```python
xval_len = len(x_val)
series = np.zeros(xval_len, dtype=np.complex128)
```

and, moreover, it seems to me you can compress the two `for` cycles into one by summing in one shot the exponential with `+n` and `-n`.

examples/qfiae/utils_qfiae.py
```diff
@@ -0,0 +1,108 @@
...
+
+    return plt, R
+
+
+def exp_fourier_to_trig(fourier_coeffs):
+    """Converts the Fourier coefficients from exponential form to trigonometric form."""
+
+    n = len(fourier_coeffs)
+    c0 = fourier_coeffs[0]
+    b_coeffs = [2 * np.real(fourier_coeffs[i]) for i in range(1, int(n / 2) + 1)]
```
MatteoRobbiati, line 63: Please do only one cycle if you can do that (like here).

examples/qfiae/utils_qfiae.py
```diff
@@ -0,0 +1,108 @@
...
+        array_list[i, 2] = trig_coeffs[i]
+    for i in range(int((len(trig_coeffs) - 1) / 2) + 1, len(trig_coeffs)):
+        # sine
+        array_list[i, 0] = 1  # 1 means Sin
+        array_list[i, 1] = i - (len(trig_coeffs) - 1) / 2
+        array_list[i, 2] = trig_coeffs[i]
+    return array_list
+
+
+class Fourier_terms:
```
MatteoRobbiati, line 102: I don't understand why you need a class here. 
Can you explain how do you use this?

examples/qfiae/vqregressor_linear_ansatz.py
```diff
@@ -0,0 +1,372 @@
...
+        self.params = np.random.randn(3 * layers + 3).astype("float64")
+        self.nparams = (layers + 1) * 3
+        self.features, self.labels, self.norm = self.prepare_training_set(ndata)
+        self.nsample = len(self.labels)
+
+    def S(self, circuit, x):
+        """Data encoding circuit block."""
+        circuit.add(gates.RZ(q=0, theta=x))
+
+    def A(self, circuit, theta0, theta1, theta2):
```
MatteoRobbiati, line 37: Please pass the parameters as an array.

examples/qfiae/vqregressor_linear_ansatz.py
```diff
@@ -0,0 +1,372 @@
...
+            loss += (label - prediction_evaluation) ** 2
+            obs_gradients = self.parameter_shift(feat)
+            for i in range(self.nparams):
+                loss_gradients[i] += (2 * prediction_evaluation * obs_gradients[i]) - (
+                    2 * label * obs_gradients[i]
+                )
+
+        return loss_gradients, (loss / len(features))
+
+    def shift_a_parameter(self, i, this_feature):
```
MatteoRobbiati, line 219: As far as I understood, you are uploading your data into the first rotational gate of the circuit. Setting the entire rotational angle to be `theta=x`. 

If this is true, the parameter shift rule implemented in the `vqregressor` example can't be used as it is. As you can see, in this implementation you divide the shift by `this_feature` if `i % 3 == 0`. This is done because we needed to calculate the derivative of the circuit **only** with respect to the variational parameters $\theta_0$, which is involved into the angle definition like $\theta = \theta_0\cdot x$. 

Here two questions:
1. do you need to calculate the derivative with respect to the first rotational angle? It is not a variational parameter in this case!
2. if yes, you should fix the rule.

As general suggestion, think about the possibility to use [this newer implementation of the PSR](https://github.com/qiboteam/qibo/blob/master/src/qibo/derivative.py) we did.

MatteoRobbiati (changes requested):
Thank you for implementing this!

I start by providing you some comments on the Python scripts.
I am going to review the notebook in a second moment. 

As a general comment, I would ask you to change the docstrings adopting [Type Hints](https://peps.python.org/pep-0484/), which can be a good solution in the future. Parts of this code are quite old, so it is fine if we exploit this moment to improve the documentation! 
Before doing this, I would ask the @qiboteam/coredevs if they agree, or if you prefer to avoid these kind of modifications in `Qibo` for now.

Apart from this, some suggestions (and discussions) follow.

AleCandido:
About Qibo itself I'm fine.

About this PR, I would not charge new contributors with this extra burden. But if you're available to help me, that's fine :)

MatteoRobbiati:
> About this PR, I would not charge new contributors with this extra burden. But if you're available to help me, that's fine :)

Sure! I will take care of this together with @gmlejarza, in case.

examples/qfiae/utils_qfiae.py
gmlejarza: Thanks for the suggestions! I've already implemented some. 
Here I don't necessarily need this Class, of course, I can just work with the array, but for me, the code seems easier to understand with the notation introduced in `Fourier_terms`.  Since the coefficient of the Fourier term will be in the `self.coefficient` attribute, the angle in  `self.angle` and the type of function in `self.function`
On the other hand, I also see your point and it is probably better to remove unnecessary functions/classes. So, let me know if you want me to remove this.

examples/qfiae/vqregressor_linear_ansatz.py
gmlejarza: You're right, I do not need to calculate the derivative with respect to the first rotational angle. I've removed that part of the PSR.

examples/qfiae/utils_qfiae.py
MatteoRobbiati: If you think it is clearer, feel free to keep it :)
Maybe you can think if it is possible to add some method to this class, maybe collecting some of the functions you write above. If it doesn't make sense, keep it as it its.

MatteoRobbiati:
Hi @gmlejarza!
Ask me again for review one you are ready!

MatteoRobbiati (commented):
Thank you @gmlejarza for this.
Some comments follow.

#### On the notebook

1. there is some thing we can fix in the text (minor stuff), maybe once you have finished I can fix some typos and push an updated version of the notebook;
2. I think some variables are extremely explicit (e.g. `number_of_layers`, which can simply be `nlayers`), but this is a choice, do as you prefer!
3. are you sure you need 10 layers for this fit? It seems to me the VQC has too much expressivity;
4. can you compress the two errors in the final estimations into a single one? Or define them explicitly in the notebook

examples/qfiae/vqregressor_linear_ansatz.py
```diff
@@ -0,0 +1,363 @@
...
+        Function which set the new parameters into the circuit
+        Args:
+            new_params: np.array of the new parameters; it has to be (3 * nlayers) long
+        """
+
+        self.params = new_params
+
+    def one_prediction(self, this_feature):
+        """
+        This function execute one prediction
```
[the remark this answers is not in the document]
gmlejarza: Thank you @MatteoRobbiati for your comments!
1. Perfect.
2. Done.
3. It indeed has much expressivity as you said, but there's a tradeoff between the expressivity and the accuracy we can obtain with 100 epochs. Fewer layers will mean a (slightly) lower accuracy. So I opted for using 10 layers to ensure very high accuracy.
4. I prefer to have them separately so I defined the second error explicitly on the notebook

MatteoRobbiati (commented):
All works for me, just some final comments:

1. I think it should be better to reduce some Fig size, in order to avoid some super big images and other small ones;
2. I understand the discussion about the number of layers, thanks for the explanation. I suggest to specify that this number of layers (typically not required for performing a fit) is useful in order to get a good accuracy with a reasonable number of epochs.
3. I tested the training and, with this configuration, it takes around 30'. This is for sure related to $N_{\rm data}=200$ and $N_{\rm layers}=10$. I am ok with keeping this configuration, but it can be fine to write a sentence so that the user knows it.
4. I think you still need to define the two errors in the markdown block.

gmlejarza:
> All works for me, just some final comments:
> 
> 1. I think it should be better to reduce some Fig size, in order to avoid some super big images and other small ones;
> 2. I understand the discussion about the number of layers, thanks for the explanation. I suggest to specify that this number of layers (typically not required for performing a fit) is useful in order to get a good accuracy with a reasonable number of epochs.
> 3. I tested the training and, with this configuration, it takes around 30'. This is for sure related to Ndata=200 and Nlayers=10. I am ok with keeping this configuration, but it can be fine to write a sentence so that the user knows it.
> 4. I think you still need to define the two errors in the markdown block.

1. Could you be more specific about the Figs. I should reduce the size? 
I will add the other things you mention in 2,3 and 4

MatteoRobbiati:
> 1. Could you be more specific about the Figs. I should reduce the size?

Some figures like [this](https://github.com/qiboteam/qibo/blob/qfiae/examples/qfiae/images/iqae.png) are too big! But this is a minor thing. I also can reduce the figsize once you finish your modifications.

MatteoRobbiati (approved):
Thank you @gmlejarza, I think we are ready to proceed if you are ok with my last modifications.

gmlejarza:
> Thank you @gmlejarza, I think we are ready to proceed if you are ok with my last modifications.

Yes, thank you! I just centered the images and added the changes on "doc/source/code-examples/tutorials/qfiae/"

scarrazza:
Thank you @gmlejarza @MatteoRobbiati!

PR #947, opened 2023-06-21
anomaly detection example update

Simone-Bordoni:
This is just a small update to the code, as i was revisiting it. Unfortunately i have found problems in loading data from different folders, i will open an issue about this, as it regards also other examples.


Simone-Bordoni:
I have added the relative path to data folders in anonmaly detection and variational classifier example

AleCandido (approved):

scarrazza (approved):
Thanks @Simone-Bordoni.

PR #949, opened 2023-06-23
Add Diamond norm to `quantum_info.metrics`

### LINKED ISSUE #909: Add `diamond_norm` to `quantum_info.metrics`

renatomello:
Closes #909 

Checklist:
- [x] Tests are passing.


AlejandroSopena (approved):

renatomello:
@stavros11 I'd like your review on this PR because it introduces a new dependency that is not mandatory, so I'd like your opinion to see if everything is ok

src/qibo/quantum_info/metrics.py
```diff
@@ -852,6 +853,133 @@ def gate_error(channel, target=None, check_unitary: bool = False, backend=None):
...
+    .. note::
+        This function requires the optional CVXPY package to be installed.
+
+    """
+    try:  # pragma: no cover
+        import cvxpy  # pylint: disable=C0415
+    except ImportError:  # pragma: no cover
+        raise_error(
+            ModuleNotFoundError, "cvxpy module was not found. Please install it."
+        )
```
stavros11, line 898: I think this error is not needed since Python will raise a very similar and readable one on its own if the package is missing. In such cases it is simpler to rely on the built-in tools so that our code is shorted and we have to maintain less. This way you also save all the coverage/pylint exceptions that are done here.

src/qibo/quantum_info/metrics.py
```diff
@@ -852,6 +853,133 @@ def gate_error(channel, target=None, check_unitary: bool = False, backend=None):
...
+
+    from scipy.sparse import eye as sp_eye  # pylint: disable=C0415
```
stavros11, line 900: What is the pylint issue here? 

By the way it is also fine to change this to something like
```py
from scipy import sparse
```
and move it as a global import, as it is good to avoid local imports as much as possible. Scipy is a dependency anyway so it won't fail.

pyproject.toml
```diff
@@ -65,6 +66,12 @@ docs = "make -C doc html"
 docs-clean = "make -C doc clean"
 test-docs = "make -C doc doctest"
 
+
+
+
+
+
+
```
stavros11, line 74: Are these empty lines needed?

pyproject.toml
```diff
@@ -73,6 +80,12 @@ cupy-cuda11x = "^12.0.0"
 cuquantum-python-cu11 = "^23.3.0"
 qibojit = { git = "https://github.com/qiboteam/qibojit.git" }
 
+
+
+
+
+
+
```
stavros11, line 88: Same here.

stavros11 (commented):
Thanks @renatomello. Looks good to me, other than the following small comments.

Regarding the new dependency, personally I don't have any issue, and, as far as I know, there are no rules about adding new ones. I generally avoid libraries are not supported well, but cvxpy looks very active on GitHub so it should be fine. Did you discuss this in the meeting yesterday to hear the opinion of others?

Also, looking at pyproject.toml, you only added the new dependency in the tests group. That means that new users who just install qibo (without tests) won't get it and will have to install it manually if they want to use the new function. So it is not really a dependency. That is fine, we do the same for cupy and cuquantum in qibojit.

renatomello:
@stavros11 suggestions implemented. Thanks.

stavros11 (approved):
Thanks!

Issue #852, opened 2023-04-11
Combining `expectation_from_samples` with measurements in the Pauli X/Y basis


chmwzc:
I've been trying to use the `expectation_from_samples` method under `SymbolicHamiltonian` together with measurements in the Pauli X/Y basis (#799).
However, `expectation_from_samples` currently only accepts Hamiltonians with only Z's ([link](https://github.com/qiboteam/qibo/blob/master/src/qibo/hamiltonians/hamiltonians.py#L564)) and raises an error otherwise.
Is this check still necessary, and can it be removed?

Sample code:
```
from qibo import models, gates

from qibo.symbols import X, Y
from qibo.hamiltonians import SymbolicHamiltonian

# Hamiltonian term
ham_term = X(0)*Y(1)
sym_ham = SymbolicHamiltonian(ham_term)

# Get the basis rotation gates and target qubits from the Hamiltonian term
qubits = [factor.target_qubit for factor in sym_ham.terms[0].factors]
basis = [type(factor.gate) for factor in sym_ham.terms[0].factors]

# Run the circuit to get the output frequencies
circuit = models.Circuit(2)
circuit.add(gates.M(*qubits, basis=basis))
result = circuit(nshots=100)
freq = result.frequencies(binary=True)

# Only works for Z terms, raises an error if ham_term has X/Y terms
sym_ham.expectation_from_samples(freq, qubit_map=qubits) 
```
***

Other than that, a minor documentation [issue](https://qibo.science/qibo/stable/api-reference/qibo.html#qibo.hamiltonians.abstract.AbstractHamiltonian.expectation_from_samples) with `expectation_from_samples`: the lines under [freq](https://github.com/qiboteam/qibo/blob/4ca9ba961b4e7121d3e2d647a2030c61059d67ca/src/qibo/hamiltonians/abstract.py#L91) are missing 4 spaces😀

scarrazza:
@AlejandroSopena are you looking into this and #719?

AlejandroSopena:
I haven't looked at it in detail, but I can take care of it

PR #951, opened 2023-06-24
Fix `SymbolicHamiltonian.expectation` for different nqubits

### LINKED ISSUE #935: Wrong results using `SymbolicHamiltonian`
I have recently found a bug in the `SymbolicHamiltonian` class, whereby expectation values of certain observables are not calculated correctly when using the `qibojit` backend (version `0.0.10`).

In particular, here is an example where this behavior shows up. It is a 2-qubit circuit where I apply an `RXX(theta)` gate and then measure the `Z(0)` observable. The expected value is easy to compute, it should be `cos(theta)`. However, `qibo` (version `0.1.14`) returns a constant value of 1 for all values of `theta`.

The code to reproduce this issue is:

```python
# CREATE CIRCUIT
circ = Circuit(2)
circ.add(gates.RXX(0, 1, theta=0))

# DEFINE SYMBOLIC HAMILTONIAN
operator = SymbolicHamiltonian(Z(0))

# COLLECT RESULTS
results = []
for t in np.linspace(0, 2*np.pi, 1000):
    circ.set_parameters([t])
    state = circ().state()
    results.append(operator.expectation(state))
    
# PLOT RESULTS
plt.plot(np.linspace(0, 2*np.pi, 1000), results)
plt.plot(np.linspace(0, 2*np.pi, 1000), np.cos(np.linspace(0, 2*np.pi, 1000)))
plt.show()
```

If we now change `operator = SymbolicHamiltonian(Z(0))` by `operator = SymbolicHamiltonian(Z(0)*I(1))`, the code works as expected. Moreover, when using the `numpy` backend the snippet above doesn't work unless the `I(1)` is included.

This shows that the problem is in the `Z(0)`, not in the circuit. The code also works as expected if we decompose the `RXX` gate in terms of `H`, `RZ` and `CNOT` gates.

Any clue about what is going on? I think this is an important issue, as people may be finding wrong results in their research projects.

Also, this issue went through the tests, so I guess those should be updated as well.

stavros11:
Fixes #935. Specifically:
* `nqubits` attribute was added to `SymbolicHamiltonian` as suggested in https://github.com/qiboteam/qibo/issues/935#issuecomment-1592312035. Note that the effect of this is still equivalent to padding, for example `SymbolicHamiltonian(Z(0) * Z(1), nqubits=4)` is equivalent to `SymbolicHamiltonian(Z(0) * Z(1) * I(2) * I(3))`.
* A `ValueError` is raised if there is a mismatch between the number of qubits in the state and the Hamiltonian in expectation value calculations.
* A documentation example about `expectation_from_samples` was added (@AlejandroSopena let me know if that is okay).

Checklist:
- [x] Tests are passing.

### DIFF
--- a/doc/source/code-examples/advancedexamples.rst
+++ b/doc/source/code-examples/advancedexamples.rst
@@ -1953,7 +1953,7 @@ to allow calculation of expectation values directly from such samples:
 
     hamiltonian = XXZ(4)
 
-    result = circuit()
+    result = circuit(nshots=1024)
     expectation_value = hamiltonian.expectation_from_samples(result.frequencies())
 
 
--- a/doc/source/code-examples/advancedexamples.rst
+++ b/doc/source/code-examples/advancedexamples.rst
@@ -1897,3 +1897,74 @@ constructing each symbol:
 
     form = Z(0, commutative=True) * Z(1, commutative=True) + Z(1, commutative=True) * Z(2, commutative=True)
     ham = hamiltonians.SymbolicHamiltonian(form)
+
+
+.. _hamexpectation-example:
+
+How to calculate expectation values using samples?
+--------------------------------------------------
+
+It is possible to calculate the expectation value of a :class:`qibo.hamiltonians.Hamiltonian`
+on a given state using the :meth:`qibo.hamiltonians.Hamiltonian.expectation` method,
+which can be called on a state or density matrix. For example
+
+
+.. testcode::
+
+    import numpy as np
+    from qibo.models import Circuit
+    from qibo.hamiltonians import XXZ
+
+    circuit = Circuit(4)
+    circuit.add(gates.H(i) for i in range(4))
+    c.add(gates.CNOT(0, 1))
+    c.add(gates.CNOT(1, 2))
+    c.add(gates.CNOT(2, 3))
+
+    hamiltonian = XXZ(4)
+
+    result = circuit()
+    expectation_value = hamiltonian.expectation(result.state())
+
+In this example, the circuit will be simulated to obtain the final state vector
+and the corresponding expectation value will be calculated through exact matrix
+multiplication with the Hamiltonian matrix.
+If a :class:`qibo.hamiltonians.SymbolicHamiltonian` is used instead, the expectation
+value will be calculated as a sum of expectation values of local terms, allowing
+calculations of more qubits with lower memory consumption. The calculation of each
+local term still requires the state vector.
+
+When executing a circuit on real hardware, usually only measurements of the state are
+available, not the state vector. Qibo provides :meth:`qibo.hamiltonians.Hamiltonian.expectation_from_samples`
+to allow calculation of expectation values directly from such samples:
+
+
+.. testcode::
+
+    import numpy as np
+    from qibo.models import Circuit
+    from qibo.hamiltonians import XXZ
+
+    circuit = Circuit(4)
+    circuit.add(gates.H(i) for i in range(4))
+    c.add(gates.CNOT(0, 1))
+    c.add(gates.CNOT(1, 2))
+    c.add(gates.CNOT(2, 3))
+
+    hamiltonian = XXZ(4)
+
+    result = circuit()
+    expectation_value = hamiltonian.expectation_from_samples(result.frequencies())
+
+
+This example simulates the circuit similarly to the previous one but calculates
+the expectation value using the frequencies of shots, instead of the exact state vector.
+This can also be invoked directly from the ``result`` object:
+
+.. testcode::
+
+    expectation_value = result.expectation_from_samples(hamiltonian)
+
+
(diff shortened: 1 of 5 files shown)


doc/source/code-examples/advancedexamples.rst
```diff
@@ -1897,3 +1897,74 @@ constructing each symbol:
...
+
+    circuit = Circuit(4)
+    circuit.add(gates.H(i) for i in range(4))
+    c.add(gates.CNOT(0, 1))
+    c.add(gates.CNOT(1, 2))
+    c.add(gates.CNOT(2, 3))
+
+    hamiltonian = XXZ(4)
+
+    result = circuit()
```
igres26, line 1956:
```suggestion
    result = circuit(nshots=1024)
```
I think that for the expectations from samples we should make it clear that we are doing a certain number of shots, even if there is a default value.

stavros11: Thanks, I missed this. This code will throw an error anyway (`result.frequencies()` won't be available unless you specify `nshots`), so is strange why the doctest didn't fail...

igres26 (approved):

doc/source/code-examples/advancedexamples.rst
```diff
@@ -1897,3 +1897,74 @@ constructing each symbol:
...
+    circuit = Circuit(4)
+    circuit.add(gates.H(i) for i in range(4))
+    c.add(gates.CNOT(0, 1))
+    c.add(gates.CNOT(1, 2))
+    c.add(gates.CNOT(2, 3))
+
+    hamiltonian = XXZ(4)
+
+    result = circuit(nshots=1024)
+    expectation_value = hamiltonian.expectation_from_samples(result.frequencies())
```
AlejandroSopena, line 1957:
```python
from qibo import gates
from qibo.hamiltonians import Z

circuit = Circuit(4)
circuit.add(gates.H(i) for i in range(4))
circuit.add(gates.CNOT(0, 1))
circuit.add(gates.CNOT(1, 2))
circuit.add(gates.CNOT(2, 3))
circuit.add(gates.M(*range(4)))

hamiltonian = Z(4)

result = circuit(nshots=1024)
expectation_value = hamiltonian.expectation_from_samples(result.frequencies())
```

AlejandroSopena (commented):
Thanks @stavros11 for adding the example. The tests are not passing because `expectation_from_samples()` only works when the observable is diagonal in the z basis as you mentioned, and `XXZ()` is not. I have pending to generalize this function #852 and I will do it at some point in August.

doc/source/code-examples/advancedexamples.rst
stavros11: Thanks @AlejandroSopena. Indeed I even wrote that only diagonal Hamiltonians are supported for now but didn't follow it in the example.

doc/source/code-examples/advancedexamples.rst
```diff
@@ -1897,3 +1897,74 @@ constructing each symbol:
...
+
+    from qibo import gates
+    from qibo.models import Circuit
+    from qibo.hamiltonians import Z
+
+    circuit = Circuit(4)
+    circuit.add(gates.H(i) for i in range(4))
+    c.add(gates.CNOT(0, 1))
+    c.add(gates.CNOT(1, 2))
+    c.add(gates.CNOT(2, 3))
```
AlejandroSopena, line 1952:
```python
circuit.add(gates.CNOT(0, 1))
circuit.add(gates.CNOT(1, 2))
circuit.add(gates.CNOT(2, 3))
circuit.add(gates.M(*range(4)))
```

doc/source/code-examples/advancedexamples.rst
```diff
@@ -1897,3 +1897,74 @@ constructing each symbol:
...
+
+    from qibo import gates
+    from qibo.models import Circuit
+    from qibo.hamiltonians import XXZ
+
+    circuit = Circuit(4)
+    circuit.add(gates.H(i) for i in range(4))
+    c.add(gates.CNOT(0, 1))
+    c.add(gates.CNOT(1, 2))
+    c.add(gates.CNOT(2, 3))
```
AlejandroSopena, line 1922:
```python
  circuit.add(gates.CNOT(0, 1))
  circuit.add(gates.CNOT(1, 2))
  circuit.add(gates.CNOT(2, 3))
```

AlejandroSopena (commented):
Thanks. Just a few more comments. When adding the three CNOTs, you have to replace `c` --> `circuit` To use `expectation_from_samples()` we have to add measurement gates because the expected value is calculated from frequencies.

stavros11:
Thanks for the fixes @AlejandroSopena. I had them locally in my computer but forgot to push. I hope it is okay now.

AlejandroSopena (approved):

PR #952, opened 2023-06-26
Add `clifford` label to gates

renatomello:
This is useful for transpilers / compilers, as well as a future Clifford circuit simulator.

Checklist:
- [x] Tests are passing.


src/qibo/gates/abstract.py
```diff
@@ -290,6 +291,11 @@ def matrix(self):
         backend = GlobalBackend()
         return self.asmatrix(backend)
 
+    @property
+    def clifford(self) -> bool:
+        """Returns ``True`` if gate is a Clifford gate, else ``False``."""
+        return self._clifford
```
stavros11, line 297: Since this is very simple it could also be just an attribute. And then you directly do `self.clifford = True` in the related gates.

renatomello: Done

stavros11 (approved):
Looks good to me, thanks.

vodovozovaliza (approved):

PR #954, opened 2023-06-26
Add `Givens` gate

renatomello:
This PR adds a new gate to the `gates` module, the Givens gate, which is important to quantum chemistry applications.

Checklist:
- [x] Tests are passing.


andrea-pasquale (approved):
Thanks @renatomello for implementing this.
Everything looks good from my side.

src/qibo/gates/gates.py
```diff
@@ -337,7 +369,7 @@ class _Rn_(ParametrizedGate):
     Args:
         q (int): the qubit id number.
         theta (float): the rotation angle.
-        trainable (bool): whether gate parameters can be updated using
+        trainable (bool, optional): whether gate parameters can be updated using
```
stavros11, line 372: Here I would prefer to leave it as it is (`bool` only), because `optional` usually means that it can be `None` (but I may be wrong), which is not the case here. `trainable` can only be `True` or `False`.

src/qibo/gates/gates.py
```diff
@@ -1540,7 +1611,7 @@ def on_qubits(self, qubit_map):
         return gate
 
     def _dagger(self):
-        import numpy as np
+        import numpy as np  # pylint: disable=C0415
```
stavros11, line 1614: Not very related to this PR, but this can also be a global import (to fix pylint instead of disabling).

stavros11 (approved):

src/qibo/gates/gates.py
renatomello: Done

PR #957, opened 2023-07-03
Clifford circuit with `to_clifford` method

MatteoRobbiati:
I think it can be useful to have a method for decomposing a circuit into a one compatible with the Clifford Data Regression mitigation technique. 

To do:
- [ ] tests.

Checklist:
- [ ] Tests are passing.


src/qibo/gates/abstract.py
```diff
@@ -276,6 +276,17 @@ def generator_eigenvalue(self):
             f"Generator eigenvalue is not implemented for {self.__class__.__name__}",
         )
 
+    def is_clifford(self) -> bool:
+        """Check if the gate is a Clifford gate."""
+        return False
```
stavros11, line 281: This was implemented (as `self.clifford`) in #952. If you don't have it in your branch, try to merge the latest master. If you prefer `self.is_clifford()` instead of `self.clifford` please discuss with @renatomello. I am fine with either approach, but we should only have one to avoid redudancy.

src/qibo/models/circuit.py
```diff
@@ -438,6 +438,31 @@ def decompose(self, *free: int):
...
+            Circuit object containing only Clifford gates.
+        """
+
+        clifford_queue = []
+        # cycle on gates replacing non-clifford with clifford
+        for gate in self.queue:
+            if gate.is_clifford():
+                clifford_queue.append(gate)
+            else:
+                clifford_queue.extend(gate.decompose_into_clifford())
```
stavros11, line 457: Why don't you create the `clifford_circuit` here and add the gates directly, instead of creating a list and then adding it to the circuit?
```suggestion
        clifford_circuit = self.__class__(**self.init_kwargs)
        # cycle on gates replacing non-clifford with clifford
        for gate in self.queue:
            if gate.clifford:
                clifford_circuit.add(gate)
            else:
                clifford_circuit.add(gate.decompose_into_clifford())
```
Note that `circuit.add()` supports lists or iterables of gates (it will automatically loop over the iterable and add all gates).

stavros11 (commented):
Thanks, I think that is a useful feature also for the transpilers.

renatomello:
Just to understand the goal here: you want to transpile circuits into strictly Clifford circuits?

MatteoRobbiati:
> Just to understand the goal here: you want to compile circuits into strictly Clifford circuits? If so, isn't that a task for a compiler instead of a circuit method?

No, I want to build a circuit which satisfies [thess](https://qibo.readthedocs.io/en/stable/api-reference/qibo.html#clifford-data-regression-cdr) conditions.

renatomello:
> > Just to understand the goal here: you want to compile circuits into strictly Clifford circuits? If so, isn't that a task for a compiler instead of a circuit method?
> 
> No, I want to build a circuit which satisfies [these](https://qibo.readthedocs.io/en/stable/api-reference/qibo.html#clifford-data-regression-cdr) conditions. This simplify the procedure of using the CDR method, I think.

I see, you want to replace non-Clifford gates with Clifford gates from a fixed gate set. In that case, I would be against naming this method `to_clifford`, because I think this should be saved to be used to convert circuits composed of only Clifford gates into a phase-space representation in a future Clifford simulator. That was the initial goal of #952.

In any case, my opinion is that this is very niche to be a general `Circuit` method. My advice would be to create a function for that inside `error_mitigation.py` or maybe `quantum_info.utis.py`

MatteoRobbiati:
> I would be against naming this method `to_clifford`, because I think this should be saved to be used to convert circuits composed of only Clifford gates into a phase-space representation in a future Clifford simulator.

I agree, that notations is wrong and misunderstandable.

> In any case, my opinion is that this is very niche to be a general `Circuit` method. My advice would be to create a function for that inside `error_mitigation.py` or maybe `quantum_info.utis.py`. But other people should opine too, @stavros11 @AlejandroSopena.

Yes, since it will be useful only when dealing with CDR mitigation, it makes sense to modify that model.
Thank you for your suggestions @renatomello! I also wait the others opinion before proceed.

AlejandroSopena:
I think this is more related with the transpiler. Currently, the transpiler creates a circuit with gates `I`, `Z`, `RZ`, `U3`, and `CZ`, and then the compiler translates these gates into pulses.

So far, to use error mitigation on hardware, I have to write the `U3` gates in terms of `RX(pi/2)` and `RZ`, and then add a custom rule to the compiler to directly compile `RX(pi/2)` as a pulse.

One option would be to modify the transpiler to directly compile the circuit into `I`, `Z`, `RZ`, `U3`, and `CZ` gates, and then compile that circuit. This way, we could directly access the circuit in native gates and use it for algorithms like error mitigation.

Let me know your opinion.

MatteoRobbiati:
I was thinking it is not worth to overcomplicate the code by adding several decomposition to the gates: it is fine to leave the CDR method as it is now. I think we have two options for this PR:

1. close it;
2. use it as starting point for a more general `to_clifford` decomposition of any kind of circuit. 

I don't know if this can be useful now or in future, what do you think about @renatomello?

renatomello:
> I was thinking it is not worth to overcomplicate the code by adding several decomposition to the gates: it is fine to leave the CDR method as it is now. I think we have two options for this PR:
> 
>     1. close it;
> 
>     2. use it as starting point for a more general `to_clifford` decomposition of any kind of circuit.
> 
> 
> I don't know if this can be useful now or in future, what do you think about @renatomello?

I think the `to_clifford` method is still necessary, but the route I'd take is to have a Clifford compiler separately for more general tasks, and then the `to_clifford` method calls that compiler.

PR #965, opened 2023-07-12
`Parameter` class

WanderingMike:
Checklist:
- [ ] Tests are passing.


src/qibo/derivative.py
```diff
@@ -156,5 +233,281 @@ def circuit(nqubits = 1):
...
+    circuit's parameter is obtained by combination of a variational
+    parameter and an external object, such as a training variable in a Quantum
+    Machine Learning problem. For example, performing a re-uploading strategy
+    to embed some data into a circuit, we apply to the quantum state rotations
+    whose angles are in the form: theta' = theta * x, where theta is a variational
+    parameter and x an input variable. The PSR allows to calculate the derivative
+    with respect of theta' but, if we want to optimize a system with respect its
+    variational parameters we need to "free" this procedure from the x depencency.
+    If the `scale_factor` is not provided, it is set equal to one and doesn't
+    affect the calculation.
```
MatteoRobbiati, line 273: It seems to me you are not using the scale factor here. 
Consider to drop this part of the docstring or to pass the factor as argument.

src/qibo/derivative.py
```diff
@@ -156,5 +233,281 @@ def circuit(nqubits = 1):
...
+        circuit (:class:`qibo.models.circuit.Circuit`): custom quantum circuit.
+        hamiltonian (:class:`qibo.hamiltonians.Hamiltonian`): target observable.
+            if you want to execute on hardware, a symbolic hamiltonian must be
+            provided as follows (example with Pauli Z and ``nqubits=1``):
+            ``SymbolicHamiltonian(np.prod([ Z(i) for i in range(1) ]))``.
+        parameter_index (int): the index which identifies the target parameter
+            in the ``circuit.get_parameters()`` list.
+        initial_state (ndarray, optional): initial state on which the circuit
+            acts. Default is ``None``.
+        scale_factor (float, optional): parameter scale factor. Default is ``1``.
```
MatteoRobbiati, line 288: Same here: I think the docs are not updated with the current code structure.

src/qibo/models/circuit.py
```diff
@@ -438,6 +486,26 @@ def decompose(self, *free: int):
...
+        clifford_circuit = self.__class__(**self.init_kwargs)
+        # cycle on gates replacing non-clifford with clifford
+        for gate in self.queue:
+            if gate.is_clifford():
+                clifford_circuit.add(gate)
+            else:
+                clifford_circuit.add(gate.decompose_into_clifford())
+
+        return clifford_circuit
+
```
MatteoRobbiati, line 508:
```suggestion
```

src/qibo/models/circuit.py
```diff
@@ -512,7 +580,7 @@ def with_pauli_noise(self, noise_map: NoiseMapType):
                 noisy_circuit.add(noise_gate)
         return noisy_circuit
 
-    def add(self, gate):
+    def add(self, gate, position=-1):
```
MatteoRobbiati, line 583: I think this feature is interesting. Is it better to introduce a new `insert` method like you did in `_Queue` on top of `list` or we want to change the `add` method like it is suggested here @qiboteam/coredevs?

src/qibo/models/circuit.py
```diff
@@ -512,7 +580,7 @@ def with_pauli_noise(self, noise_map: NoiseMapType):
...
         """Add a gate to a given queue.
 
         Args:
```
MatteoRobbiati, line 586: In case we decide to keep this, the docstrings must be updated by adding the new `position` argument.

src/qibo/models/parameter.py
```diff
@@ -0,0 +1,42 @@
...
+        params = []
+        if self._featurep is not None:
+            params.append(self._featurep)
+        if fixed_params:
+            params.extend(fixed_params)
+        else:
+            params.extend(self._variational_parameters)
+        return self.lambdaf(*params)
+
+    def _update_params(self, trainablep=None, feature=None):
```
MatteoRobbiati, line 21: I think you don't need this method in case params and features are None, so I would drop the default values.

src/qibo/models/parameter.py
```diff
@@ -0,0 +1,42 @@
...
+        self._featurep = features
+        self.nparams = len(variational_parameters)
+        self.lambdaf = func
+
+    def _apply_func(self, fixed_params=None):
+        params = []
+        if self._featurep is not None:
+            params.append(self._featurep)
+        if fixed_params:
+            params.extend(fixed_params)
```
MatteoRobbiati, line 16: Can you please explain what you mean with `fixed_params`?

MatteoRobbiati (changes requested):
Thanks a lot @WanderingMike for this.

I like the idea to have a `Parameter` class, but I think it should be documented a bit more and we should add some more features (e.g. a way to compute the final angle, in order to provide a circuit with the float value to be inserted, or a method to compute the derivative of the `Parameter` with respect one of its components, because this will be very useful when using the parameter shift rule).
Some more comments follow.

src/qibo/models/parameter.py
WanderingMike: If you have a gate parameter $\theta = 10\theta_i + 20$, the fixed_params returns 20. It is basically the invariant part of the lambda function, with regards to the parameter we are considering (here $\theta_i$)

src/qibo/models/parameter.py
WanderingMike: There are cases where we update only the trainable parameters or the feature. If could be for instance that the feature is not embedded in the Parameter lambda function, and I need to prepare for that case as well.

src/qibo/models/parameter.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def get_indices(self, start_index):
+        return [start_index + i for i in range(self.nparams)]
+
+    def get_fixed_part(self, trainablep_idx):
+        params = [0] * self.nparams
+        params[trainablep_idx] = self._variational_parameters[trainablep_idx]
+        return self._apply_func(fixed_params=params)
+
+    def get_scaling_factor(self, trainablep_idx):
```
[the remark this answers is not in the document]
WanderingMike: Yes indeed. Currently, you could only have polynomials of degree 1 w.r.t a specific parameter $\theta$. For any other function, we would need to calculate the full derivative using a different method. This becomes very complicated if the user does not supply the derivative. There might be a possibility to give the lambda function as a sympy object which does automatic gradient calculation

src/qibo/gates/gates.py
```diff
@@ -2,11 +2,89 @@
...
 
 import numpy as np
+import sympy as sp
 
 from qibo.config import PRECISION_TOL, raise_error
 from qibo.gates.abstract import Gate, ParametrizedGate
 
 
+class Parameter:
+    """Object which allows complex gate parameters. Several trainable parameter
```
stavros11, line 12: I guess that here by "complex" you mean that there is some classical pre-processing in the parameter calculation. The word may be a bit confusing as one can think of complex (with real and imaginary part) numbers, but this may just be me.

src/qibo/gates/gates.py
```diff
@@ -2,11 +2,89 @@
...
+    and possibly features are linked through a lambda function which returns the
+    final gate parameter"""
```
stavros11, line 14: Here we should add the `Args:` section in the docstring and document the arguments of `__init__`.

src/qibo/gates/gates.py
```diff
@@ -2,11 +2,89 @@
...
+        return float(function(*params))
+
+    def _update_params(self, trainablep=None, feature=None):
+        """Update gate trainable parameter and feature values"""
+        if trainablep is not None:
+            self._trainablep = trainablep
+        if feature and self._featurep:
+            self._featurep = feature
+
+    def get_params(self, trainablep=None, feature=None):
```
stavros11, line 49: Same here, I would prefer
```suggestion
    def get_parameters(self, trainablep=None, feature=None):
```
particularly because this is exposed to the user.

src/qibo/gates/gates.py
```diff
@@ -2,11 +2,89 @@
...
+        """Update values with trainable parameter and calculate current gate parameter"""
+        self._update_params(trainablep=trainablep, feature=feature)
```
stavros11, line 51: From a user perspective, it is a bit counter-intuitive that you can set the parameters using the `get_parameters` function. Maybe having a separate `set_parameters` or `update_parameters` would be cleaner?

src/qibo/gates/gates.py
```diff
@@ -2,11 +2,89 @@
...
+            else:
+                params.append(self._featurep)
+        if fixed_params:
+            params.extend(fixed_params)
+        else:
+            params.extend(self._trainablep)
+        return float(function(*params))
+
+    def _update_params(self, trainablep=None, feature=None):
+        """Update gate trainable parameter and feature values"""
```
stavros11, line 43: Here we could check that the given parameters and features have the proper types/shapes, otherwise they may be incompatible with the derivatives already calculated in `__init__` and essentially break the object.

WanderingMike:
> Thanks for impementing this @WanderingMike.
> 
> Regarding integration with the rest of qibo, I see that now you only included this in the rotation gates (RX, RY, RZ). Are you planning to extend this to other parametrized gates?
> 
> Before doing so, I wanted to point out that now you are only assigning the `Parameter` when initializing the gate. If someone uses `circuit.set_parameters` it will overwrite the parameters calculated using the `Parameter` object during initialization. Is this the behavior we want? I think we needed to expose `Parameter` to the optimizers. In this case we probably need to refactor a bit how the parameter setting in circuits works.

@stavros11 Indeed we would have to implement the Parameter input for all ParametrizedGate. If you give me the greenlight for this, I can make all the changes. I have since implemented your comments. I have also created a new VariationalCircuit class, inheriting the Circuit class. This VariationalCircuit will handle setting and getting parameters (either the trainable parameters or the gate parameters obtained by the application of the lambda function). This now works with my Optimizer class as well, which will be pushed on a different PR #962 once this one is merged

stavros11 (commented):
@WanderingMike thank you for the updates. 

Before proceeding with the rest of the gates, in order to avoid double work from your side, I would like to ask the opinion of @AleCandido and @scarrazza about the new layout.

Here we are adding the `VariationalCircuit` that is compatible with the new `Parameter` class, while the original `Circuit` still supports the `set_parameters` functionality. Is there an advantage in keeping both approaches or shall we try merge them? I think I would try to merge them because, if I understand correctly, the new `Parameter` class provides more functionality than the current `set_parameters` so it should be the default. Also, from a usability perspective, having only `circuit.set_parameters` instead of a both a `circuit.set_parameters` and a `circuit.set_variational_parameters` is easier to follow.

If the worry is about not breaking the current interface, maybe we could still maintain it with proper checks. For example when initializing a trainable gate:
```py
gates.RX(0, theta=0.5, trainable=True)
```
the constructor checks if `theta` is a `float` and `if trainable` it converts it to a `Parameter`. This way all `trainable` gates have `Parameter` parameters and are easier to control in the code.

Regardless of the approach, and once this PR is finalized, we should add a how-to-use the new `Parameter` object example in the documentation (advanced examples). This would be useful for many users. It can be done here or in another PR.

src/qibo/gates/gates.py
```diff
@@ -2,11 +2,102 @@
...
 from qibo.gates.abstract import Gate, ParametrizedGate
 
 
+class Parameter:
+    """Object which allows for variational gate parameters. Several trainable parameter
+    and possibly features are linked through a lambda function which returns the
+    final gate parameter
+
+    Args:
+        func (function): lambda function describing the gate parameter
```
MatteoRobbiati, line 17: I think we should define a conventional way to write the lambdas. E.g., in your examples you always write functions like: 

```pyhton
lambda x1, x2, th1, th2 : x1 * th2 - x2 * th1
```

And this is fine with me, but I think it should be clear what is a feature and what is a parameter. For example, we can specify that the lambdas have to be provided writing features (x's) followed by parameters (th's). This can also help when calling the `update_parameters` function. 

I suggest to write these instructions in the docstrings:

```
    func (callable): lambda function which builds the gate parameter. If both features and trainable parameters 
        compose the function, it must be passed by first providing the features and then the parameters, as
        described in the code example above. 
```

src/qibo/gates/gates.py
```diff
@@ -2,11 +2,102 @@
...
+        the optimizer's trainable parameter list"""
+        return [start_index + i for i in range(self.nparams)]
+
+    def get_fixed_part(self, trainable_idx):
+        """Retrieve parameter constant unaffected by a specific trainable parameter"""
+        params = self._trainable.copy()
+        params[trainable_idx] = 0.0
+        return self._apply_func(self.lambdaf, fixed_params=params)
+
+    def calculate_derivatives(self):
```
MatteoRobbiati, line 78: Did you to adopt this approach in order not to re-calculate every time the derivatives? 

I think it can be useful to have a simpler function in which the derivative w.r.t whatever in the lambda can be calculated individually. One can be interested, e.g., in derivating the expected value w.r.t only one of the features.

src/qibo/gates/gates.py
```diff
@@ -2,11 +2,102 @@
...
+        expr = sp.sympify(self.lambdaf(*vars))
+
+        derivatives = []
+        for i in range(self.nfeat, len(vars)):
+            derivative_expr = sp.diff(expr, vars[i])
+            derivatives.append(sp.lambdify(vars, derivative_expr))
+
+        return derivatives
+
+    def get_scaling_factor(self, trainable_idx):
```
MatteoRobbiati, line 95: I would rename this function (and the `scale_factor` in `derivative.parameter_shift`) in some clearer way: `get_partial_derivative` maybe? Or more in general `get_derivative` and we ask for the index to be specified? 

Can this function be used for both trainable parameters and features? In principle, in QML the two concepts are quite symmetric.

src/qibo/models/circuit.py
```diff
@@ -1423,3 +1423,71 @@ def chunkstring(string, length):
             output += table
 
         return output.rstrip("\n")
+
+
+class VariationalCircuit(Circuit):
```
MatteoRobbiati, line 1428: Maybe it should be better to define this in a separate file.

src/qibo/gates/gates.py
```diff
@@ -2,11 +2,102 @@
...
+
+        if not isinstance(trainable, (list, np.ndarray)):
+            raise_error(ValueError, "Features must be given as list or numpy array")
+
+        if trainable is not None:
+            self._trainable = trainable
+        if feature and self._feature:
+            self._feature = feature
+
+    def get_gate_parameters(self):
```
MatteoRobbiati, line 63: I would suggest to implement this same method as a `__call__` function. What do you think about this @stavros11?

MatteoRobbiati (commented):
Thanks @WanderingMike for the modifications! Some more comments by my side follow.

src/qibo/gates/gates.py
WanderingMike: When you initialise the Parameter, it automatically calculates all possible internal derivatives. Then, based on the parameter you want to the know the partial derivative of, you just call Parameter.get_partial_derivative() (previously called get_scaling_factor). This function therefore already exists :)

src/qibo/models/circuit.py
WanderingMike: Yes, let's see what @AleCandido and @scarrazza think

AleCandido: In general, don't be afraid of using more files. It is fine to have even a file with 10 lines, if there is a good reason to keep it separate.

However, on the other side, it's not even convenient to have 1000 files with 10 lines each (even because most of them would be `import` statements).
So, your files should be approximately uniform: it's not a strict rule, they might be very small, or a bit larger than average. But never too large (as a rule of thumb, if it hits 1000 lines you might start considering which is the best way of splitting it).

In this specific case, this file is large because Python does not allow to define a class over multiple files (but other clean solutions are possible, such as splitting the most of the functions' bodies, and have some functions that operate on these objects, without being methods).
But since the addition of this PR is not part of the `Circuit` class, it's not a bad idea to put that one in a separate module.

src/qibo/gates/gates.py
MatteoRobbiati: This is exactly what I was thinking about. When we do QML stuff, we update the parameters many times! Think to the scenario in which one use a CMA optimizer in a QML algo: if we do like you suggest, the derivatives are calculated every time a parameter is updated and never used! I don't really know heavy is the differentiation, but I still believe a safest solution is to calculate when is necessary.

WanderingMike: So I think it depends a bit if the actual Parameter lambda changes a lot. If it's the same function, say theta1 * x + theta2, then the derivatives are calculated once when initialising the Parameter (you get derivatives [theta1, x, 1] for all three possible parameters) and when you update either a feature or parameter, those functions are already in memory and you simply plugin the new values into the derivative formulas. So as long as the lambda function has the same "shape", we don't need to recalculate the derivatives

MatteoRobbiati: Oh, I see. These are callable functions, so we simply have the analytical form of the derivative. That souns good, thanks for the explanation.

AleCandido (commented):
I still have to pass the `Parameter` class, but while doing it atomically, I took one step back and tried to understand the `VariationalCircuit` class first, since smaller (in order to get hints on the overall workflow).

The `VariationalCircuit` is already not fully straightforward, for the two main reasons:
1. it jumps a bit back and forth, since there are many functions that operate mutating the state of the single undeclared attribute `initparams` (not present even in the base class)
2. because it's essentially working in two modes in parallel, so you always should branch in one of the two situations (array or parameters)

So, 1. is simple to alleviate: declare `self.initparams` in `__init__`, possibly with an associated type hint and docstring (you could also include the attribute docstring in an [*Attributes*](https://www.sphinx-doc.org/en/master/usage/extensions/example_numpy.html) section in the class docstring).
To properly solve, maybe it's a bit trickier, because it would be better if the functions/methods were used through composition, rather than mutating a state (it would make the workflow more explicit). But this I'd face in a later review.

Concerning the second problem, the best solution would be to avoid one of the two modes, or to convert one into the other.
In general, it's better if one element performs one task (so one function does one action, one class represents one entity - not two or more). So, I'd rather leave the case of the array outside the `VariationalCircuit`, and only address the `Parameter` (as currently declared [in the docstring](https://github.com/qiboteam/qibo/blob/7f42798b7cc4ec7f971414606a675b37ed5b7da2/src/qibo/models/variational.py#L669-L670), dealing with the array just with the original `Circuit` class (or making a dedicated class, in case I'm missing something).

In general, if you could help me to understand the `VariationalCircuit` workflow (possibly simplifying it a bit), then I could make a better review for `Parameter` itself.

AleCandido:
Btw, in the `VariationalCircuit` the following terms are used: `params`, `initparams`, `variational_params`, and `train_params`. If there is a clear separation, it would be useful to document it (for future developers, not for users, they are mostly internal). If there is not, maybe it would be better to use a more standard vocabulary.

src/qibo/models/variational.py
```diff
@@ -659,3 +663,100 @@ def _loss(params, falqon, hamiltonian):
...
+
+        return params
+
+    def _get_train_params(self):
+        # for array
+        if isinstance(self.initparams, np.ndarray):
+            return self.initparams
+
+        # for Parameter objects
+        else:
```
AleCandido, line 710: If you `return`ed before, you don't need an `else` branch.

https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/no-else-return.html

src/qibo/models/variational.py
```diff
@@ -659,3 +663,100 @@ def _loss(params, falqon, hamiltonian):
...
+                In this case all gates in the iterable will be added in the
+                circuit.
+
+        Returns:
+            If the circuit contains measurement gates with ``collapse=True``
+            a ``sympy.Symbol`` that parametrizes the corresponding outcome.
+        """
+        res = super().add(gate)
+        if isinstance(gate, gates.ParametrizedGate):
+            self.initparams = self._get_initparams()
```
AleCandido, line 761: This is one of the things I don't fully get. This is essentially an update of `self.initparams`, but you're overwriting it every time you `add()` a gate.

Can't you just compute `self.initparams` once before execution?
Is there a time in which you need `self.initparams` before `add`ing one more gate?

MatteoRobbiati:
> In general, if you could help me to understand the `VariationalCircuit` workflow (possibly simplifying it a bit), then I could make a better review for `Parameter` itself.

The `VariationalCircuit` is indeed not necessary, especially as it is. The plan is to improve it, adding extra features which can help the user with a _TensorFlow functional API_-like  approach to the QML problem.

Anyway, your question is valid a priori, in the sense that it is useful to understand the workflow of the optimization process in which a variational circuit is the model (or part of the model). 
The main reason for which the `Parameter` is needed is that we use parametric gates (above all rotations) as nodes in a neural network, and we can be interested in uploading combinations of trainable parameters and features into the gate parameters (GP). With this in mind, the `Parameter` class becomes an useful tool, which allows the user to clearly declare what has to be optimised, in particular when we deal with gradient based optimization processes. 
In fact, if the old shape of the parameters was perfectly fine when dealing with a borrowed optimizer (e.g. the `scipy`'s), the process becomes more difficult if we implement a gradient descent using the Parameter Shift Rules. In that context, is extremely useful to be able to compute the shifts only on $\beta_{\rm trainable}$ and get access to automatic calculated derivatives of the GP with respect to the trainable ones: $\partial_{\beta_{\rm trainable}} \theta(\beta)$.

AleCandido:
At this broader level I already got it, manly because you explained me some time ago which were the critical issues in the current approach.

Instead, I tried to understand at an operational level, i.e. something like:
1. you call `.add()` to add a gate (and for non-parameterized gates that's it)
2. that reassigns `.initparams` *somehow* through `._get_initparams()`
3. and, while optimizing, you read the parameters from `.initparams` update them through `.set_variational_parameters()`

But in parallel, if `.initparams` is in a certain state, because the `add`ed gates were of certain kind, all operations are trivial (which situation you intend to cover in this way? are there incompatibility among gates that belong to one or the other scenario? are you testing that they are not causing issues, or do you need to be somehow consistent as a `Circuit` user?).

Moreover, `._get_train_params()` is new (not overwriting the base class), internal (led by `_`), and not use anywhere. It is not even documented. Which is its purpose?

src/qibo/gates/gates.py
```diff
@@ -495,12 +496,12 @@ def __init__(self, q, theta, trainable=True):
         self.name = None
         self._controlled_gate = None
         self.target_qubits = (q,)
-        self.unitary = True
```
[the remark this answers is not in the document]
ACSE-mct22: I have put the line back, this seems to be from an old review

src/qibo/gates/gates.py
```diff
@@ -495,12 +496,12 @@ def __init__(self, q, theta, trainable=True):
...
-
-        if isinstance(theta, (float, int)) and (theta % (np.pi / 2)).is_integer():
-            self.clifford = True
```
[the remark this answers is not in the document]
ACSE-mct22: I have put this lines in again. This is based on Matteo's PR #957

src/qibo/parameter.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def __init__(self, func, trainable, feature=None):
+        self._trainable = trainable
+        self._feature = feature
+        self.nparams = len(trainable)
+
+        if isinstance(feature, list):
+            self.nfeat = len(feature)
+        else:
+            self.nfeat = 0
```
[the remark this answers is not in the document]
ACSE-mct22: Done

src/qibo/parameter.py
```diff
@@ -0,0 +1,140 @@
...
+        compose the function, it must be passed by first providing the features and then the parameters, as
+        described in the code example above.
+        trainable (list or np.ndarray): array with initial trainable parameters theta
+        feature (list or np.ndarray): array containing possible input features x
+    """
+
+    def __init__(self, func, trainable, feature=None):
+        self._trainable = trainable
+        self._feature = feature
+        self.nparams = len(trainable)
```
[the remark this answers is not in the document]
ACSE-mct22: Done

src/qibo/models/variational.py
```diff
@@ -659,3 +663,100 @@ def _loss(params, falqon, hamiltonian):
...
+        return res
```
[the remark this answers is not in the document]
ACSE-mct22: In this case, we have to first add the gate to the circuit, then the self._get_initparams() retrieves the gate parameters, and only then can we return the gate. The operations must happen in that sequence

src/qibo/models/variational.py
ACSE-mct22: Corrected

src/qibo/models/variational.py
ACSE-mct22: Maybe you know of a solution to this. Basically, I need to get the initparams once the circuit is fully built. However, there is no way of knowing when that is the case, the user could keep adding gates. Since I can't make assumptions about when the circuit is finished being built, I need to call self._get_initparams after each gate is added to the circuit

ACSE-mct22:
Thanks @AleCandido and @MatteoRobbiati. I will try splitting both the Parameter and VariationalCircuit into separate PR. For context @AleCandido, the main reason I built both classes is to use them in quantum optimisation problems. I found that it was much more intuitive when the optimiser (or any use-case relying on the Parameter object to describe gate parameters) did not have to handle the mapping from trainable parameters embedded in a Parameter object to final gate parameters. The goal of the VariationalCircuit is to handle this transformation. Basically, you input the trainable parameters and get out the gate parameters, and the VariationalCircuit acts like a black box for an outsider user.

WanderingMike:
@AleCandido  Please let me know if you think we can merge the Parameter class, which has already been reviewed by @MatteoRobbiati  and @stavros11

MatteoRobbiati:
> @AleCandido Please let me know if you think we can merge the Parameter class, which has already been reviewed by @MatteoRobbiati and @stavros11

I am also going to give a last pass.

src/qibo/parameter.py
```diff
@@ -0,0 +1,158 @@
...
+            raise_error(
+                ValueError,
+                f"{len(trainable)} trainable parameters given, need {self.nparams}",
+            )
+
+        if not isinstance(feature, (list, np.ndarray)) and self._feature != feature:
+            raise_error(ValueError, "Features must be given as list or numpy array")
+
+        if self._feature is not None and self.nfeat != len(feature):
+            raise_error(ValueError, f"{len(feature)} features given, need {self.nfeat}")
```
AleCandido, line 137: Even in this case, I'd avoid checks, they take more than the function itself.

If we needed to be strict on the inputs, we'd better used a typed language (or typed Python itself). Otherwise, adopt duck typing, and keep going until it fail :)
(assuming to save potential debugging burden to the user, this is causing maintenance burden on the developers)

src/qibo/parameter.py
```diff
@@ -0,0 +1,158 @@
...
+
+        if trainable is not None:
+            self._trainable = trainable
+        if feature and self._feature:
+            self._feature = feature
+
+    def get_indices(self, start_index):
+        """Return list of respective indices of trainable parameters within
+        a larger trainable parameter list"""
+        return [start_index + i for i in range(self.nparams)]
```
AleCandido, line 147: This whole function is equivalent to:
```py
(np.arange(self.nparams) + start_index).tolist()
```
in case you need a list. If you could return an array, it's even simpler, essentially boiling down to a call to `np.arange()`, that is much more well-known.

In general, I'm a huge fan of small reusable functions, but when it comes to calling just another function it might be unnecessarily nested (thus slightly complicating for someone else to figure out what is happening at call site, because of the further step).
In this case, the trade-off is set by the usage: if you plan to use this combination many times, maybe it's even worth keeping (I'd just drop the `get_`), but if it's only in a couple of places, maybe it'd make sense to inline it.

src/qibo/parameter.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def get_fixed_part(self, trainable_idx):
+        """Retrieve parameter constant unaffected by a specific trainable parameter"""
```
AleCandido, line 150: The docstring is not incredibly straightforward.

src/qibo/parameter.py
```diff
@@ -0,0 +1,158 @@
...
+
+        if not isinstance(feature, (list, np.ndarray)) and self._feature != feature:
+            raise_error(ValueError, "Features must be given as list or numpy array")
+
+        if self._feature is not None and self.nfeat != len(feature):
+            raise_error(ValueError, f"{len(feature)} features given, need {self.nfeat}")
+
+        if trainable is not None:
+            self._trainable = trainable
+        if feature and self._feature:
```
AleCandido, line 141: Don't test for falsiness, but for their actual value, i.e.
```suggestion
        if feature is not None and self._feature is not None:
```

src/qibo/parameter.py
AleCandido: In any case, if you want to define setters with additional checks, the more idiomatic way would be to use properties wrapping the inner attributes, i.e.

```py
@property
def trainable(self):
    return self._trainable

@trainable.setter
def trainable(self, value):
    my_check(value)
    self._trainable = value
```
https://docs.python.org/3/library/functions.html#property

AleCandido (commented):
Sorry, I had a review pending, since I stopped on Sympy usage.

I'm not so keen in Sympy integration in other programs, but I will keep thinking a bit about it.
Most likely there isn't a proper way out, other than using a proper autodiff framework.

WanderingMike:
@AleCandido @MatteoRobbiati All checks now pass, please let me know if the PR is ready to merge tomorrow during the Qibo meeting.

src/qibo/parameter.py
```diff
@@ -0,0 +1,109 @@
...
+        self._feature = feature
+
+        # lambda function
+        self.lambdaf = func
+
+        self.derivatives = self._calculate_derivatives()
+
+    def __call__(self):
+        """Update values with trainable parameter and calculate current gate parameter"""
+        return self._apply_func(self.lambdaf)
```
AleCandido, line 47: Just for simplicity, I would avoid the alias, and just use `__call__` instead of `_apply_func`, i.e.:
```py
    def __call__(self, function=None, fixed_params=None):
        """Applies lambda function and returns final gate parameter"""
        params = []
        if self._feature is not None:
            params.extend(self._feature)
        if fixed_params:
            params.extend(fixed_params)
        else:
            params.extend(self._trainable)

        # run function
		f = function if function is not None else self.lambdaf
        return float(f(*params))
```

src/qibo/parameter.py
```diff
@@ -0,0 +1,109 @@
...
+
+    @property
+    def nparams(self):
+        return len(self._trainable)
+
+    @property
+    def nfeat(self):
+        return len(self._feature) if isinstance(self._feature, list) else 0
```
AleCandido, line 55: An even more Pythonic way, if not misleading, would be to just try:
```suggestion
        try:
            return len(self._feature)
        except TypeError:
            return 0
```

In this way you don't care if it's a list, a tuple, or something similar. If it has a `len()`, that's enough for this method :)

src/qibo/parameter.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def _apply_func(self, function, fixed_params=None):
+        """Applies lambda function and returns final gate parameter"""
+        params = []
+        if self._feature is not None:
+            params.extend(self._feature)
+        if fixed_params:
```
AleCandido, line 62:
```suggestion
        if fixed_params is not None:
```
Better to avoid assumptions on falsiness (e.g. even `[]` is falsy, but maybe you'd like to use it anyhow).

src/qibo/parameter.py
```diff
@@ -0,0 +1,109 @@
...
+
+        return derivatives
+
+    def update_parameters(self, trainable=None, feature=None):
+        """Update gate trainable parameter and feature values"""
+        if trainable is not None:
+            self._trainable = trainable
+
+        if feature is not None and self._feature is not None:
+            self._feature = feature
```
AleCandido, line 93: I have the impression to have already written it, but: why not using just property setters?

https://docs.python.org/3/library/functions.html#property

src/qibo/parameter.py
```diff
@@ -0,0 +1,109 @@
...
+    def get_fixed_part(self, trainable_idx):
+        """Retrieve constant term of lambda function with regard to a specific trainable parameter"""
+        params = self._trainable.copy()
+        params[trainable_idx] = 0.0
+        return self._apply_func(self.lambdaf, fixed_params=params)
+
+    def get_partial_derivative(self, trainable_idx):
+        """Get derivative w.r.t a trainable parameter"""
+        deriv = self.derivatives[trainable_idx]
+        return self._apply_func(deriv)
```
AleCandido, line 109: Just a stylistic advice: I'd drop `get_` in front of all the function names, it adds nothing.

src/qibo/parameter.py
AleCandido: Even better: since in this case the setters would be trivial, why not just exposing them as `self.trainable` and `self.feature`?

If they could be directly modified by the user, no need to hide them (and no need for this method).

AleCandido (commented):
Here there are few small suggestions for improvements. Nothing is crucial, it's just a matter of simplification.

AleCandido (approved):
Apart from the comments above, I'd be happy to merge whenever you wish. Since the comments are not crucial, they could also be delayed (even though they are simple enough that I'd apply them immediately).

src/qibo/parameter.py
```diff
@@ -0,0 +1,109 @@
+import numpy as np
+import sympy as sp
+
+from qibo.config import raise_error
+
+
+class Parameter:
+    """Object which allows for variational gate parameters. Several trainable parameter
```
MatteoRobbiati, line 8:
```suggestion
    """Object which allows for variational gate parameters. Several trainable parameters
```

src/qibo/parameter.py
```diff
@@ -0,0 +1,109 @@
...
+
+        partial_derivative = param.get_partial_derivative(3)
+
+        param.update_parameters(trainable=[15.0, 10.0, 7.0], feature=[5.0])
+        gate_value = param()
+
+
+    Args:
+        func (function): lambda function which builds the gate parameter. If both features and trainable parameters
+        compose the function, it must be passed by first providing the features and then the parameters, as
```
MatteoRobbiati, line 30: 1. I think it is still a bit ambiguous, because we are asking to write the `lambda` with features followed by parameters, but then we call the `Parameter(function, trainable, feature)`. I think is maybe better to keep the order `features` $\to$ `trainable` also in the initialisation. 
2. At this point I would call the features vector with the plural form `self.features`, because we can need to inject more than one feature.
3. I was thinking it can maybe be useful to set some default initialization of the parameters, e.g. to initialize them to 0 or to a random value if none is provided by the user.

src/qibo/parameter.py
MatteoRobbiati: Yes, maybe to remove the method is the cleaner way. One can modify the features by directly calling `self.features = new_features`.

src/qibo/parameter.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def update_parameters(self, trainable=None, feature=None):
+        """Update gate trainable parameter and feature values"""
+        if trainable is not None:
+            self._trainable = trainable
+
+        if feature is not None and self._feature is not None:
+            self._feature = feature
+
+    def get_indices(self, start_index):
```
MatteoRobbiati, line 95: I agree with @AleCandido in removing `get`.  Also, I would call this in some explicit way, e.g. `corresponding_circuit_indices`, or `indices_into_circuit`, in order to make it clear this function is returning the trainable parameters indices into a larger list of parameters (if I understood well).

src/qibo/parameter.py
```diff
@@ -0,0 +1,109 @@
...
+        """Return list of respective indices of trainable parameters within
+        a larger trainable parameter list"""
+        return (np.arange(self.nparams) + start_index).tolist()
+
+    def get_fixed_part(self, trainable_idx):
```
MatteoRobbiati, line 100: I would call this `unaffected_by`, or `lightened_of`, or something like this. If the idea here is to return the `Parameter` value if the target `trainable` is set equal to zero. On this way it will appear as:

```python
reduced_parameter = parameter.uneffected_by(parameter_index)

or

reduced_parameter = parameter.lightened_of(parameter_index)
```

I would also ask the opinion of @AleCandido here.

MatteoRobbiati (commented):
Thank you very much for your modifications. 
I have still some little comment below.

src/qibo/parameter.py
AleCandido: `unaffected_by` sounds nice.

But other than being amused by the clear sentence appearing in your code snippet (well, it is exactly in the spirit of Python, it should read like English), I don't have a strong personal opinion :)

src/qibo/parameter.py
WanderingMike: I can implement 1. and 2. without any problems. However, for 3. since Ale preferred not to do any checks on the number of features and the number of parameters, I can't initialise the parameters, since I do not have any information on the size of the parameter list

src/qibo/parameter.py
WanderingMike: I am using _apply_func in other functions, such as get_partial_derivative. I think in this case it makes sense to separate the function that applies the lambda.

src/qibo/parameter.py
MatteoRobbiati: Ok, but on this way the initialization is accepted only if the user provides the values too. Is this what you mean @WanderingMike, @AleCandido?

src/qibo/parameter.py
AleCandido: You're right it makes sense, but you could use as well `self(...)` in those places, instead of `self._apply_func(...)`. It is just a different "name".

src/qibo/parameter.py
AleCandido: Sorry @WanderingMike, I didn't mean to be so strict.
I always try to provide the rationale for my suggestions, but I might have failed. In any case, you do not need to do anything just because I said something :)

In any case, about checking, I would still avoid. It will fail anyhow, when it tries to use it, with less code on our side. But nothing prevents you to collect the same information for the suggested initialization. This is something that Python won't do for us, if we do not instruct it to do it.

About the initialization itself, I'm not sure whether it is worth it to do it silently. Maybe even requiring it from the user is not a completely bad idea.

src/qibo/parameter.py
AleCandido: > Overall, I think it is not very clean to change object variables directly, as the user might not know what format they take. Usually it is encouraged to interact with objects through functions instead.

This is exactly the Java perspective. But Python has `@property`s exactly for that, while providing a more ergonomic syntax.
But if you are not doing anything in your function, other than setting the parameter, there is no added benefit from the function itself (other than the syntax and call overhead...).

However, if you want to accept `None` and ignore it (rather than literally setting to `None`), then the setter is non-trivial, you're right. Thus, I would propose the `@property` approach, and just split `trainable` and `feature` in two separate properties.

src/qibo/parameter.py
MatteoRobbiati: > About the initialization itself, I'm not sure whether it is worth it to do it silently. Maybe even requiring it from the user is not a completely bad idea.

In case we don't allow to do it silently, I would suggest to raise a custom error, asking explicitly to provide values for both trainable parameters and features. 

Despite this, I still believe the sense of `Parameter` is more in its being flexible (a lambda) than in the value it returns. On top of this idea, I think it should be possible to initialize an "empy" `Parameter`, which e.g. symbolically returns zero, but still exists. I am thinking to a QML scenario in which I need many parameters but I don't really care about the initial values.

src/qibo/parameter.py
```diff
@@ -0,0 +1,123 @@
...
+        self._trainable = value
+
+    @property
+    def features(self):
+        """I'm the features property."""
+        return self._features
+
+    @features.setter
+    def features(self, value):
+        self._features = value if not self._nofeatures else None
```
AleCandido, line 107: I take it back, `.features` is a separate case from `.trainable`. What is `._nofeatures`?

WanderingMike: nofeatures is an optional flag that the user can set if they don't want to embed features in the lambda function. It is a way to explicitly implement the `if trainable is not None:` . That way, even if the optimiser updates the feature value of a Parameter object with no feature embedding, the whole operation is safe. I will add this to the docstring.

src/qibo/parameter.py
WanderingMike: I implemented an initialisation with an array full of 0.0 float values

src/qibo/parameter.py
```diff
@@ -0,0 +1,130 @@
...
+    Example::
+
+        from qibo.parameter import Parameter
+        param = Parameter(
+                lambda x, th1, th2, th3: x**2 * th1 + th2 * th3**2,
+                [1.5, 2.0, 3.0],
+                feature=[7.0],
+            )
+
+        partial_derivative = param.get_partial_derivative(3)
```
MatteoRobbiati, line 22:
```suggestion
        param = Parameter(
                lambda x, th1, th2, th3: x**2 * th1 + th2 * th3**2,
                features=[7.0],
                 trainable=[1.5, 2.0, 3.0],
            )

        partial_derivative = param.get_partial_derivative(3)
```

src/qibo/parameter.py
```diff
@@ -0,0 +1,130 @@
...
+
+        param.update_parameters(trainable=[15.0, 10.0, 7.0], feature=[5.0])
+        gate_value = param()
```
MatteoRobbiati, line 25:
```suggestion
        param_value = param()
```

src/qibo/parameter.py
MatteoRobbiati: Ok, I think I see the problem. This is needed in order to allow the default initialization. 
I think `self._nofeatures` should be dropped, since we already have that information from `self.trainable` and `self.features`, thus I see two options:

1. we jump back to the old version, removing the automatic initialization;
2. we exploit Sympy's symbols: if only `'th{ }'` are provided, then `self.features=None`.

MatteoRobbiati (commented):
Just some last comments follow. 
I think we also need to cover two more lines (#L62 and #L101).

src/qibo/parameter.py
WanderingMike: I think the automatic initialisation makes sense, it will make building circuits much cleaner. Whereas for proposal 2., Ale believed we should not constrain the user to specific lambda variable names, such as 'th{ }'. That's why I thought the nofeatures flag could be useful. The nofeatures flag is initialised with False, therefore it remains a completely optional feature that can facilitate the user's coding. However, there are no consequences for the user not knowing what the flag is, since with a nofeatures flag set to False the Parameter class behaves as expected

src/qibo/models/parameter.py
```diff
@@ -0,0 +1,42 @@
+"""Model for combining trainable parameters and possible features into circuit parameters."""
+
+
+class Parameter:
```
AleCandido, line 4: @MatteoRobbiati why do we have two `Parameter` classes?

MatteoRobbiati: Oh, right, I forgot to remove it from there. Thanks.

src/qibo/parameter.py
```diff
@@ -0,0 +1,141 @@
...
+            vars.append(sp.Symbol(f"p{i}"))
+
+        expr = sp.sympify(self.lambdaf(*vars))
+
+        derivatives = []
+        for i in range(len(vars)):
+            derivative_expr = sp.diff(expr, vars[i])
+            derivatives.append(sp.lambdify(vars, derivative_expr))
+
+        return derivatives
```
AleCandido, line 107: Since the only thing you're using of `self` is `self.lambdaf`, in this case I would also move this function out of the class, and just accept a single argument (i.e. `lambdaf`, for which you could also use a more standard name like `func`).

src/qibo/parameter.py
```diff
@@ -0,0 +1,141 @@
...
+        """Returns the number of features"""
+        return len(self._features) if isinstance(self._features, list) else 0
+
+    @property
+    def ncomponents(self):
+        """Returns the number of elements which compose the Parameter"""
+        return self.nparams + self.nfeat
+
+    def _apply_func(self, function, fixed_params=None):
+        """Applies lambda function and returns final gate parameter"""
```
AleCandido, line 76: I'm not sure I understand why you need a function to call a function (i.e. one already exists, it is the application and in Python is called `__call__`).

Maybe you could just make a function to construct the parameters, but it would be better if it did not depend on `function`, like just the `else` branch of the first `if`.
Maybe it is feasible if you just define `self._trainable` to be an empty list instead of `None` (because I guess in that case you want the derivatives wrt the features, otherwise if there are no features nor trainables I do not understand why someone would use the `Parameter` class).

src/qibo/parameter.py
```diff
@@ -0,0 +1,141 @@
...
+        self._trainable = value
+
+    @property
+    def features(self):
+        """Features property."""
+        return self._features
+
+    @features.setter
+    def features(self, value):
+        self._features = value
```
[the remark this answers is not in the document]
MatteoRobbiati: I was thinking so. We can set them as public variables and directly use `self.features = [new_value]`.

src/qibo/parameter.py
AleCandido: Exactly!

src/qibo/parameter.py
```diff
@@ -0,0 +1,134 @@
...
+        """Returns the number of trainable parameters"""
+        try:
+            return len(self.trainable)
+        except TypeError:
+            return 0
+
+    @property
+    def nfeat(self):
+        """Returns the number of features"""
+        return len(self.features) if isinstance(self.features, list) else 0
```
AleCandido, line 108: In these cases, `try... except...` is usually the way to go (more robust).
```suggestion
        try:
            return len(self.features)
        except TypeError:
            return 0
```

src/qibo/parameter.py
```diff
@@ -0,0 +1,134 @@
...
+    def trainable_parameter_indices(self, start_index):
+        """Return list of respective indices of trainable parameters within
+        the larger trainable parameter list of a circuit for example"""
+        return (np.arange(self.nparams) + start_index).tolist()
+
+    def unaffected_by(self, trainable_idx):
+        """Retrieve constant term of lambda function with regard to a specific trainable parameter"""
+        params = self.trainable.copy()
+        params[trainable_idx] = 0.0
+        return self.__call__(trainable=params)
```
AleCandido, line 124:
```suggestion
        return self(trainable=params)
```

They are aliases of each other. If you can use the public, prefer it over the private (in particular over magic methods).

In case you could not use `self(...)`, even better to use:
```suggestion
        import operator
        return operator.call(self, trainable=params)
```

src/qibo/parameter.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def __init__(self, func, trainable=[], features=[]):
+        self.trainable = trainable
+        self.features = features
+
+        if self.nfeat + self.nparams != func.__code__.co_argcount:
+            raise_error(
+                TypeError,
+                f"{self.nfeat + self.nparams} parameters are provided, but {func.__code__.co_argcount} are required, please initialize features and trainable according to the defined function.",
+            )
```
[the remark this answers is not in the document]
MatteoRobbiati: I only wanted to be more explicit in the error raising: I would like to underline the importance of setting both trainable and features.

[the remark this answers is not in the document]
AleCandido: I would put the emphasis in the docs, and catching the error where it will be generating (if you want to raise a more informative one, but the one are actually getting could be even already informative enough).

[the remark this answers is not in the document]
MatteoRobbiati: I think we have some ambiguities if we don't do like this. Let's suppose we define:

```python
param = Parameter(
           lambda x1, x2, th1, th2, th3: x1**2 * th1 + x2 * th2 * th3,
           trainable=[1.5, 2.0, 3.0],
       )
```

This is still be initialised as it is. But then, once one try to call `param()`, the raised error is:

```sh
TypeError: <lambda>() missing 2 required positional arguments: 'th2' and 'th3'
```

because features are appended before trainables into the components list.
I feel this ambiguous in the sense that `th2` and `th3` are here initialized.
My purposed error is more general and less confusing (my feeling).

PR #975, opened 2023-07-20
Add delay pulse option to `Align` gate

SERVCUBED:
This adds a delay option to the Align gate, also implementing the Align gate in qibocal's compiler.

Currently, the only way to add delays in circuit operations is via constructing a pulse sequence.

The delay option causes the circuit execution to pause for a specified period of time. This is currently done by sending a pulse of zero frequency and zero amplitude so that the control hardware performs the timing.

Please see the corresponding PR in qiboteam/qibolab#541.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/gates/gates.py
+++ b/src/qibo/gates/gates.py
@@ -462,7 +462,7 @@ class Align(Gate):
         delay (int): (optional) The time (in ns) for which to delay circuit execution on the specified qubits.
     """
 
-    def __init__(self, *q, delay=0):
+    def __init__(self, *q, delay: int = 0):
         super().__init__()
         self.name = "align"
         self.delay = delay
--- a/src/qibo/gates/gates.py
+++ b/src/qibo/gates/gates.py
@@ -464,11 +464,12 @@ class Align(Gate):
     """
 
     def __init__(self, *q, delay: int = 0):
-        if delay < 0.0:
+        if not isinstance(delay, int):
             raise_error(
-                ValueError,
-                "Delay must not be negative.",
+                TypeError, f"delay must be type int, but it is type {type(delay)}."
             )
+        if delay < 0.0:
+            raise_error(ValueError, "Delay must not be negative.")
 
         super().__init__()
         self.name = "align"
--- a/src/qibo/gates/gates.py
+++ b/src/qibo/gates/gates.py
@@ -455,12 +455,21 @@ class I(Gate):
 
 
 class Align(Gate):
-    def __init__(self, *q):
+    """Aligns proceeding qubit operations and (optionally) waits ``delay`` amount of time.
+
+    Args:
+        *q (int): The qubit ID numbers.
+        delay (int): (optional) The time (in ns) for which to delay circuit execution on the specified qubits.
+    """
+    def __init__(self, *q, delay=0):
         super().__init__()
         self.name = "align"
         self.draw_label = "A"
+        if delay != 0:
+            self.draw_label += f"({delay})"
+        self.init_kwargs = {"delay": delay}
         self.target_qubits = tuple(q)
-        self.init_args = q
+        self.init_args = (*q, delay)
 
 
 class _Rn_(ParametrizedGate):
--- a/src/qibo/gates/gates.py
+++ b/src/qibo/gates/gates.py
@@ -465,12 +465,10 @@ class Align(Gate):
     def __init__(self, *q, delay=0):
         super().__init__()
         self.name = "align"
-        self.draw_label = "A"
-        if delay != 0:
-            self.draw_label += f"({delay})"
+        self.delay = delay
+        self.draw_label = f"A({delay})"
         self.init_kwargs = {"delay": delay}
         self.target_qubits = tuple(q)
-        self.init_args = (*q, delay)
 
 
 class _Rn_(ParametrizedGate):
--- a/src/qibo/gates/gates.py
+++ b/src/qibo/gates/gates.py
@@ -464,6 +464,12 @@ class Align(Gate):
     """
 
     def __init__(self, *q, delay: int = 0):
+        if delay < 0.0:
+            raise_error(
+                ValueError,
+                "Delay must not be negative.",
+            )
+
         super().__init__()
         self.name = "align"
         self.delay = delay
--- a/src/qibo/gates/gates.py
+++ b/src/qibo/gates/gates.py
@@ -460,7 +460,7 @@ class Align(Gate):
     Args:
         *q (int): The qubit ID numbers.
         delay (int, optional): The time (in ns) for which to delay circuit execution on the specified qubits.
-            Defaults to ``0`` (zero).        
+            Defaults to ``0`` (zero).
     """
 
     def __init__(self, *q, delay: int = 0):
--- a/src/qibo/gates/gates.py
+++ b/src/qibo/gates/gates.py
@@ -461,6 +461,7 @@ class Align(Gate):
         *q (int): The qubit ID numbers.
         delay (int): (optional) The time (in ns) for which to delay circuit execution on the specified qubits.
     """
+
     def __init__(self, *q, delay=0):
         super().__init__()
         self.name = "align"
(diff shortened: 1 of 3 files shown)


tests/test_gates_gates.py
```diff
@@ -199,19 +199,20 @@ def test_identity(backend):
 
 
 def test_align(backend):
```
renatomello, line 201:
```suggestion
@pytest.mark.parameters("delay", [0, 1])
def test_align(backend, delay):
```

tests/test_gates_gates.py
```diff
@@ -199,19 +199,20 @@ def test_identity(backend):
...
-    gate_matrix = gate.asmatrix(backend)
-    backend.assert_allclose(gate_matrix, np.eye(4))
+    for delay in [0, 1]:
+        gate = gates.Align(0, 1, delay=delay)
+        gatelist = [gates.H(0), gates.H(1), gate]
+        final_state = apply_gates(backend, gatelist, nqubits=2)
+        target_state = np.ones_like(final_state) / 2.0
+        backend.assert_allclose(final_state, target_state)
+        gate_matrix = gate.asmatrix(backend)
+        backend.assert_allclose(gate_matrix, np.eye(4))
```
renatomello, line 209:
```suggestion
    nqubits = 2

    gate = gates.Align(0, 1, delay=delay)
    gatelist = [gates.H(0), gates.H(1), gate]
    
    final_state = apply_gates(backend, gatelist, nqubits=nqubits)
    target_state = backend.plus_state(nqubits)
    backend.assert_allclose(final_state, target_state)

    gate_matrix = gate.asmatrix(backend)
    identity = backend.identity_density_matrix(nqubits, normalize=False)
    backend.assert_allclose(gate_matrix, identity)
```

tests/test_gates_gates.py
```diff
@@ -199,19 +199,20 @@ def test_identity(backend):
...
 
-    with pytest.raises(NotImplementedError):
-        gate.qasm_label
+        with pytest.raises(NotImplementedError):
+            gate.qasm_label
```
renatomello, line 212:
```suggestion
    with pytest.raises(NotImplementedError):
        gate.qasm_label
```

tests/test_gates_gates.py
```diff
@@ -199,19 +199,20 @@ def test_identity(backend):
...
 
-    assert not gates.Align(0, 1).clifford
-    assert not gates.Align(0, 1).unitary
+        assert not gates.Align(0, 1, delay=delay).clifford
+        assert not gates.Align(0, 1, delay=delay).unitary
```
renatomello, line 215:
```suggestion
    assert not gates.Align(0, 1, delay=delay).clifford
    assert not gates.Align(0, 1, delay=delay).unitary
```

renatomello (changes requested):

src/qibo/gates/gates.py
```diff
@@ -455,12 +455,22 @@ def qasm_label(self):
...
 class Align(Gate):
-    def __init__(self, *q):
+    """Aligns proceeding qubit operations and (optionally) waits ``delay`` amount of time.
+
+    Args:
+        *q (int): The qubit ID numbers.
+        delay (int): (optional) The time (in ns) for which to delay circuit execution on the specified qubits.
+    """
+
+    def __init__(self, *q, delay=0):
```
stavros11, line 465: You can also add a new attribute `self.delay = delay` for this gate. This would make qiboteam/qibolab#541 simpler as you could directly do `gate.delay` to get the delay, instead of some weird indexing of `init_args`/`init_kwargs` (which are probably not a very good mechanism in the first place).

stavros11 (commented):
Thanks for implementing this. Just a couple of comments, otherwise looks good.

src/qibo/gates/gates.py
```diff
@@ -455,12 +455,22 @@ def qasm_label(self):
...
         super().__init__()
         self.name = "align"
         self.draw_label = "A"
+        if delay != 0:
+            self.draw_label += f"({delay})"
+        self.init_kwargs = {"delay": delay}
         self.target_qubits = tuple(q)
-        self.init_args = q
+        self.init_args = (*q, delay)
```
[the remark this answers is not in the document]
SERVCUBED: Thank you. I have removed this line.

src/qibo/gates/gates.py
```diff
@@ -455,12 +455,22 @@ def qasm_label(self):
...
+        *q (int): The qubit ID numbers.
+        delay (int): (optional) The time (in ns) for which to delay circuit execution on the specified qubits.
+    """
+
+    def __init__(self, *q, delay=0):
         super().__init__()
         self.name = "align"
         self.draw_label = "A"
+        if delay != 0:
+            self.draw_label += f"({delay})"
```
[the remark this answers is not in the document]
SERVCUBED: Changed. Thank you. There doesn't seem to be any length restrictions for the labels when drawing circuits.

src/qibo/gates/gates.py
SERVCUBED: Implemented, and updated qiboteam/qibolab#541 accordingly. Thank you.

src/qibo/gates/gates.py
```diff
@@ -455,12 +455,20 @@ def qasm_label(self):
...
 class Align(Gate):
-    def __init__(self, *q):
+    """Aligns proceeding qubit operations and (optionally) waits ``delay`` amount of time.
+
+    Args:
+        *q (int): The qubit ID numbers.
+        delay (int): (optional) The time (in ns) for which to delay circuit execution on the specified qubits.
+    """
+
+    def __init__(self, *q, delay=0):
```
renatomello, line 465:
```suggestion
    def __init__(self, *q, delay: int = 0):
```

src/qibo/gates/gates.py
```diff
@@ -455,12 +455,20 @@ def qasm_label(self):
 
 
 class Align(Gate):
-    def __init__(self, *q):
+    """Aligns proceeding qubit operations and (optionally) waits ``delay`` amount of time.
+
+    Args:
+        *q (int): The qubit ID numbers.
+        delay (int): (optional) The time (in ns) for which to delay circuit execution on the specified qubits.
```
renatomello, line 462:
```suggestion
        delay (int, optional): The time (in ns) for which to delay circuit execution on the specified qubits.
```

renatomello (commented):
`delay` cannot be negative, right? So a check should be added to guarantee that (see gates like `MS` for example).

Also, is there any physical / hardware reason to put a ceiling on how big `delay` can be? Moreover, the `draw_label` can become weird if the number is too big, I think.

SERVCUBED:
> Also, is there any physical / hardware reason to put a ceiling on how big `delay` can be? Moreover, the `draw_label` can become weird if the number is too big, I think.

I think a maximum `delay` value should be set on a hardware/platform level. Should this be done in the compiler? - I'm not yet familiar with how this works.

renatomello:
> > Also, is there any physical / hardware reason to put a ceiling on how big `delay` can be? Moreover, the `draw_label` can become weird if the number is too big, I think.
> 
> I think a maximum `delay` value should be set on a hardware/platform level. Should this be done in the compiler? - I'm not yet familiar with how this works.

I'm not familiar either, but it makes sense to leave it to the hardware connection

renatomello (approved):

stavros11 (approved):
Thanks for the updates, looks good to me. Would you like to merge this now, or wait to first make sure that qiboteam/qibolab#541 works as expected and no further changes are required here? In any case, we should first merge in qibo, so that we can update the dependency in qibolab tests.

> Also, is there any physical / hardware reason to put a ceiling on how big `delay` can be?

Delays should be implemented as "wait" instructions in the sequencer program, which shouldn't be resource or memory intensive, but I am not sure if there is any limitation in practice. If there is, it will most likely be different for each instrument, so I think it should be handled at the driver level, not here or in the compiler.

PR #981, opened 2023-07-27
Docs makeup

MatteoRobbiati:
1. Images in SVG format;
2. Ubuntu font as default (to check if this works for different OS) and remover purple links.

If you are not ok with 2., I can come back to the previous configuration (if yes, I will do the same in the other repos).

Checklist:
- [ ] Tests are passing.


scarrazza:
Thanks @MatteoRobbiati. However the logo on the left side bar has disappeared.
Frankly, I would not change the font or color, could you please revert those changes and just keep the svg?

MatteoRobbiati:
> Frankly, I would not change the font or color, could you please revert those changes and just keep the svg?

Yes sure.

scarrazza:
Great @MatteoRobbiati, I understand why all these pictures are so big, are you sure this is the best compression we can achieve? If I try to open the page, I see that the loading of the image takes few seconds...

scarrazza:
Never mind, it was just temporary issue of my browser.

PR #985, opened 2023-08-03
Change `.matrix` from property to method and remove `asmatrix` method

renatomello:
Checklist:
- [x] Tests are passing.


vodovozovaliza:
Thank you @renatomello, I think `backend` should also be optional in these functions: 
https://github.com/qiboteam/qibo/blob/5e7357f3779f753fb9011a9f2100ef01df175fc1/src/qibo/gates/abstract.py#L318-L321
https://github.com/qiboteam/qibo/blob/5e7357f3779f753fb9011a9f2100ef01df175fc1/src/qibo/gates/measurements.py#L158-L162
https://github.com/qiboteam/qibo/blob/5e7357f3779f753fb9011a9f2100ef01df175fc1/src/qibo/gates/special.py#L98-L99

renatomello:
@vodovozovaliza thanks. Changed in the `FusedGate` class. I don't think the other two need to be changed, since they are only raising errors anyways.

vodovozovaliza (approved):

stavros11 (commented):
The initial idea was to use the `.matrix` property defined in 

https://github.com/qiboteam/qibo/blob/0539b0d54e811d6fc1c889b5aae177b7d2713f2d/src/qibo/gates/abstract.py#L310

if you want to use the default `GlobalBackend` and `asmatrix` only if you need to use a specific backend, that's why it didn't have a default. But either way is fine with me. If you keep the default in `asmatrix`, you can reduce minor code repetition by changing the property to just
```py
@property
def matrix(self):
    return self.asmatrix()
```

> @vodovozovaliza thanks. Changed in the `FusedGate` class. I don't think the other two need to be changed, since they are only raising errors anyways.

I think it matters for the type of error. Now if someone calls `asmatrix` without argument in these gates they will get a Python argument missing error, instead of our `NotImplementedError`. If you decide to have a default it is better to put it everywhere so that the interface is common.

renatomello:
> The initial idea was to use the `.matrix` property defined in
> 
> https://github.com/qiboteam/qibo/blob/0539b0d54e811d6fc1c889b5aae177b7d2713f2d/src/qibo/gates/abstract.py#L310
> 
> if you want to use the default `GlobalBackend` and `asmatrix` only if you need to use a specific backend, that's why it didn't have a default. But either way is fine with me. If you keep the default in `asmatrix`, you can reduce minor code repetition by changing the property to just
> 
> ```python
> @property
> def matrix(self):
>     return self.asmatrix()
> ```

Right now, we have a redundancy in having a `gate.matrix` property and `gate.asmatrix` method. Then, I'd like to suggest either keeping the former or the latter. Either one is fine by me. Right now, it seems that the best option is to make `gate.matrix` a method and remove `gate.asmatrix` entirely. What do you think?

stavros11:
> Right now, we have a redundancy in having a `gate.matrix` property and `gate.asmatrix` method. Then, I'd like to suggest either keeping the former or the latter. Either one is fine by me. Right now, it seems that the best option is to make `gate.matrix` a method and remove `gate.asmatrix` entirely. What do you think?

Yes, you are right about the redudancy. We have this because `asmatrix` was needed for executing circuits on different backends when we made `Gate` objects backend agnostic and the `matrix` property existed before, so I didn't want to break the interface. Changing `matrix` to a function (with a `backend=None` default argument) and removing `asmatrix` is okay in terms of the backends, however it may break code of external users, ie. they would have to change `.matrix` to `.matrix()`. The main issue is that I think this particular property is used by several people.

stavros11:
> We could do something like we did with the noise channel initialisations and keep the deprecated functionality still working for one or two months while raising warnings letting them know the functionality is deprecated.

This is fine with me. If it is not urgent I would just wait to discuss it in some qibo meeting, or at least ask if @scarrazza agrees. It is not really blocking this PR anyway.

If we do the warning, note that in this case we need to raise it in the property, without changing it to a function. If we change to a function, then people who are doing `gate.matrix` will get a reference to the function (instead of the matrix array) and will not see any warning.

scarrazza:
@renatomello @stavros11 if raising an warning is complex from a dev. point of view, I would suggest going directly to the"breaking change" given that we are releasing 0.2.0 soon.

scarrazza:
@renatomello could you please perform the change, remove the old interface and add a note in the sphinx documentation, explaining that from 0.2.0 the user should use the attribute instead of function.

renatomello:
@scarrazza this is ready for review. Tests depend on https://github.com/qiboteam/qibojit/pull/137/ and vice-versa.

stavros11 (approved):
Thanks @renatomello, looks good to me.

scarrazza (approved):

PR #986, opened 2023-08-03
Remove redundancy between `gate.asmatrix` and `gate.matrix`

renatomello:
Checklist:
- [ ] Tests are passing.


renatomello:
Not necessary anymore since this will be implemented in #985

PR #989, opened 2023-08-04
Fix double callbacks in evolution models

### LINKED ISSUE #973: Adiabatic evolution class state callback returns double the expected states
When adding state callbacks to the adiabatic evolution function the resulting state object has repeated states for each state.

To reproduce one can follow the code depicted in the images which also shows the issue.

Screenshot 2023-07-19 at 10 35 22
Screenshot 2023-07-19 at 10 37 07

Im using the latest qibo package (v 0.1.14) . I would expect both lengths to match or in this case to differ by 1 if counting the initial state when t=0 which it does.

MatteoRobbiati:
Fixes #973.

I am still puzzled about the `nqubit` setter in the `EntanglementEntropy` function. I think it works well if the callback is used in "circuit mode", but it doesn't if we track che entropy during an evolution. It seems to me is missing a step in which we calculate the number of qubits from the partition or from the hamiltonian. What do you think about it @stavros11?

I am testing the callbacks with the following script, if can help:

```python
import numpy as np
from qibo import models, callbacks, hamiltonians, gates, set_backend

set_backend("numpy")

nqubits = 2
T = 1
dt = 0.25

h0 = hamiltonians.X(2)
h1 = hamiltonians.TFIM(2)

c = models.Circuit(2)
c.add(gates.H(0))
c.add(gates.M(0,1))

callbacks_list = []

#callbacks_list.append(callbacks.EntanglementEntropy([0]))
callbacks_list.append(callbacks.State())
callbacks_list.append(callbacks.Norm())
callbacks_list.append(callbacks.Energy(h1))
callbacks_list.append(callbacks.Overlap(c.execute().state()))
callbacks_list.append(callbacks.Gap())

evolution = models.AdiabaticEvolution(h0=h0, h1=h1, s=lambda t: t, dt=dt, callbacks=callbacks_list)
_ = evolution(T)

for call in callbacks_list:
    print(f"CALLBACK: {call}\nVALUES:\n{call.results}")
```

which stops working if `EntanglementEntropy` is tracked.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/callbacks.py
+++ b/src/qibo/callbacks.py
@@ -206,8 +206,10 @@ class Overlap(Callback):
         self.append(overlap)
         return overlap
 
-    def apply_density_matrix(self, backend, state):
-        return backend.calculate_overlap_density_matrix(self.state, state)
+    def apply_density_matrix(self, backend, state):  # pragma: no cover
+        overlap = backend.calculate_overlap_density_matrix(self.state, state)
+        self.append(overlap)
+        return overlap
 
 
 class Energy(Callback):
--- a/src/qibo/callbacks.py
+++ b/src/qibo/callbacks.py
@@ -324,7 +324,9 @@ class Gap(Callback):
         hamiltonian.eigenvectors()
         eigvals = hamiltonian.eigenvalues()
         if isinstance(self.mode, int):
-            return backend.np.real(eigvals[self.mode])
+            gap = backend.np.real(eigvals[self.mode])
+            self.append(gap)
+            return gap
 
         # case: self.mode == "gap"
         excited = 1
--- a/src/qibo/callbacks.py
+++ b/src/qibo/callbacks.py
@@ -208,7 +208,6 @@ class Overlap(Callback):
 
     def apply_density_matrix(self, backend, state):
         overlap = backend.calculate_overlap_density_matrix(self.state, state)
-        self.append(overlap)
         return overlap
 
 
--- a/src/qibo/callbacks.py
+++ b/src/qibo/callbacks.py
@@ -105,8 +105,6 @@ class EntanglementEntropy(Callback):
 
     @Callback.nqubits.setter
     def nqubits(self, n: int):
-        from qibo import gates
-
         if self._nqubits is not None and self._nqubits != n:
             raise_error(
                 RuntimeError,
@@ -115,10 +113,11 @@ class EntanglementEntropy(Callback):
         self._nqubits = n
         if self.partition is None:
             self.partition = list(range(n // 2 + n % 2))
-        if len(self.partition) <= self.nqubits // 2:
+        if len(self.partition) <= self._nqubits // 2:
             self.partition = [
-                i for i in range(self.nqubits) if i not in set(self.partition)
+                i for i in range(self._nqubits) if i not in set(self.partition)
             ]
+        return self._nqubits
 
     def apply(self, backend, state):
         rho = backend.partial_trace(state, self.partition, self.nqubits)
@@ -175,10 +174,14 @@ class Norm(Callback):
     """
 
     def apply(self, backend, state):
-        return backend.calculate_norm(state)
+        norm = backend.calculate_norm(state)
+        self.append(norm)
+        return norm
 
     def apply_density_matrix(self, backend, state):
-        return backend.calculate_norm_density_matrix(state)
+        norm = backend.calculate_norm(state)
+        self.append(norm)
+        return norm
 
 
 class Overlap(Callback):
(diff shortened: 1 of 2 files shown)


src/qibo/callbacks.py
```diff
@@ -175,10 +174,14 @@ class Norm(Callback):
...
 
     def apply(self, backend, state):
-        return backend.calculate_norm(state)
+        norm = backend.calculate_norm(state)
+        self.append(norm)
+        return norm
 
     def apply_density_matrix(self, backend, state):
-        return backend.calculate_norm_density_matrix(state)
+        norm = backend.calculate_norm(state)
```
stavros11, line 182:
```suggestion
        norm = backend.calculate_norm_density_matrix(state)
```

src/qibo/callbacks.py
```diff
@@ -330,6 +341,7 @@ def apply(self, backend, state):
                 "The Hamiltonian is degenerate. Using eigenvalue {} "
                 "to calculate gap.".format(excited)
             )
+        self.append(gap)
```
stavros11, line 344: This is correct, but if you check a few lines above this callback has additional returns before that, which may be triggered in some cases and if they do, it will not append. I think the safest in this case would be to append before every `return`. This should also fix the failing tests.

src/qibo/callbacks.py
```diff
@@ -115,10 +113,11 @@ def nqubits(self, n: int):
         self._nqubits = n
         if self.partition is None:
             self.partition = list(range(n // 2 + n % 2))
-        if len(self.partition) <= self.nqubits // 2:
+        if len(self.partition) <= self._nqubits // 2:
             self.partition = [
-                i for i in range(self.nqubits) if i not in set(self.partition)
+                i for i in range(self._nqubits) if i not in set(self.partition)
             ]
+        return self._nqubits
```
stavros11, line 120:
```suggestion
```
This is a setter so it doesn't need to return anything.

stavros11 (changes requested):
Thanks @MatteoRobbiati for fixing this. The solution looks good, but there are still some fixes required (see comments below), which are also probably the reason tests are failing.

> I am still puzzled about the nqubit setter in the EntanglementEntropy function. I think it works well if the callback is used in "circuit mode", but it doesn't if we track che entropy during an evolution. It seems to me is missing a step in which we calculate the number of qubits from the partition or from the hamiltonian. What do you think about it @stavros11?

Yes, you are right that it doesn't work well with evolution but the problem is not in the callback setter, but it is because the setter is never called. When you use it in a circuit, the `CallbackGate` is calling the setter in
https://github.com/qiboteam/qibo/blob/964828d24f9ca963a4ee8488af05bc8c5dcce415/src/qibo/gates/special.py#L22

We need to do something similar for evolution. For example if you add
```py
for callback in self.callbacks:
    callback.nqubits = self.nqubits
```
in the beggining of https://github.com/qiboteam/qibo/blob/964828d24f9ca963a4ee8488af05bc8c5dcce415/src/qibo/models/evolution.py#L89
the script you posted should work for `EntanglementEntropy`. 

The changes in the setter are not necessary. The setter is only setting a default bi-partition which splits the system in half for the entropy calculation, in case the user did not specify the partition when defining the callback.

MatteoRobbiati:
Thank you @stavros11 for your suggestions! I think the code is now fixed. Indeed, the right thing to do was to set the number of qubits while looping over the callbacks.

MatteoRobbiati:
> Thank you @stavros11 for your suggestions! I think the code is now fixed. Indeed, the right thing to do was to set the number of qubits while looping over the callbacks.



> Thank you @stavros11 for your suggestions! I think the code is now fixed. Indeed, the right thing to do was to set the number of qubits while looping over the callbacks.

I still need to fix tests.

scarrazza:
@MatteoRobbiati this is ready for another review/merge, correct?

MatteoRobbiati:
> @MatteoRobbiati this is ready for another review/merge, correct?

Yes.

scarrazza:
@MatteoRobbiati thanks, @stavros11 could you please have a second look?

MatteoRobbiati:
> Thank you for the updates @MatteoRobbiati, I think now it is correct. Just one curiosity, why did you remove the `append` from the `Overlap` for density matrix in https://github.com/qiboteam/qibo/pull/989/commits/f1b9c08aa99da342f85a12ae7b2aed10d0d68f23? I think this is needed for the callback to behave the same way in all cases.

The coverage decreased because, since it is not implemented, the line was never checked (an error was raised before). I thought we can add it once the method will be implemented. Am I wrong?

stavros11 (approved):
> The coverage decreased because, since it is not implemented, the line was never checked (an error was raised before). I thought we can add it once the method will be implemented. Am I wrong?

I see, the backend method is not implemented. Then I am a bit confused about how we can get 100% coverage on main since this is not covered and there is also no coverage exception.

Anyway, maybe a safer solution would be:
```py
def apply_density_matrix(self, backend, state):  # pragma: no cover
    overlap = backend.calculate_overlap_density_matrix(self.state, state)
    self.append(overlap)
    return overlap
```
so that we don't forget the append in case we implement the backend method in the future. `# pragma: no cover` will exclude this function from the coverage.

EDIT: Okay, now I see, maybe we have a test that captures the `NotImplementedError`, so it works for main, but not if you add more lines to append because these are never reached. Anyway, it is also fine as it is.

MatteoRobbiati:
Thanks! I implemented your suggestion @stavros11.

PR #997, opened 2023-08-14
Drop cached property replacement

AleCandido:
While I was looking through the code, I found this relic. I guess there is no reason to keep it there, since py3.7 is officially dead, and we do not support it since a while. 

This is just a minor simplification (just maintenance), and should change literally nothing in runtime operations (apart from parsing that code, importing `sys`, and make a trivial check... but they are all irrelevant operations...).

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/backends/npmatrices.py
+++ b/src/qibo/backends/npmatrices.py
@@ -1,19 +1,6 @@
-import sys
+from functools import cached_property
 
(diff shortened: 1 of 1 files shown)


scarrazza (approved):
Good catch, thanks!

PR #1015, opened 2023-09-21
Save gates to JSON

Jacfomg:
Naive patch to save gates in RB implemented changes in https://github.com/qiboteam/qibocal/pull/412.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/gates/abstract.py
+++ b/src/qibo/gates/abstract.py
@@ -7,6 +7,9 @@ import sympy
 from qibo.backends import GlobalBackend
 from qibo.config import raise_error
 
+REQUIRED_FIELDS = ["name", "init_kwargs", "_target_qubits", "_control_qubits"]
+REQUIRED_FIELDS_INIT_KWARGS = ["theta", "phi", "lam"]
+
 
 class Gate:
     """The base class for gate implementation.
@@ -54,16 +57,14 @@ class Gate:
     def to_json(self):
         encoded = self.__dict__
 
-        required_fields = ["name", "init_kwargs", "_target_qubits", "_control_qubits"]
         encoded_simple = {
-            key: value for key, value in encoded.items() if key in required_fields
+            key: value for key, value in encoded.items() if key in REQUIRED_FIELDS
         }
 
-        required_fields_init_kwargs = ["theta", "phi", "lam"]
         encoded_simple["init_kwargs"] = {
             key: value
             for key, value in encoded_simple["init_kwargs"].items()
-            if key in required_fields_init_kwargs
+            if key in REQUIRED_FIELDS_INIT_KWARGS
         }
 
         for value in encoded_simple:
--- a/src/qibo/gates/abstract.py
+++ b/src/qibo/gates/abstract.py
@@ -51,7 +51,7 @@ class Gate:
         self.device_gates = set()
         self.original_gate = None
 
-    def toJSON(self):
+    def to_json(self):
         encoded = self.__dict__
 
         required_fields = ["name", "init_kwargs", "_target_qubits", "_control_qubits"]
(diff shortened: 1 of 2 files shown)


AleCandido (commented):
The present one is just an addition, and it is not going to destroy anything, so for me is good to go.

I still believe we should improve the serialization, but most likely it will involve reworking a bit the `Gate` class itself.

Jacfomg:
> The present one is just an addition, and it is not going to destroy anything, so for me is good to go.
> 
> I still believe we should improve the serialization, but most likely it will involve reworking a bit the `Gate` class itself.

If you have anything that could be simple to add I can try to implement it as well.

AleCandido:
> If you have anything that could be simple to add I can try to implement it as well.

That's the main problem: it might not be too simple.
But it's not too clear even to myself. So, as soon as you provide a suitable test we can merge.

src/qibo/gates/abstract.py
```diff
@@ -50,6 +51,26 @@ def __init__(self):
         self.device_gates = set()
         self.original_gate = None
 
+    def to_json(self):
+        encoded = self.__dict__
+
+        required_fields = ["name", "init_kwargs", "_target_qubits", "_control_qubits"]
```
AleCandido, line 57: This is very suitable for a class-level constant

src/qibo/gates/abstract.py
```diff
@@ -50,6 +51,26 @@ def __init__(self):
...
+        encoded_simple = {
+            key: value for key, value in encoded.items() if key in required_fields
+        }
+
+        required_fields_init_kwargs = ["theta", "phi", "lam"]
```
AleCandido, line 62: and this one as well

src/qibo/gates/abstract.py
```diff
@@ -50,6 +51,26 @@ def __init__(self):
...
+        encoded_simple["init_kwargs"] = {
+            key: value
+            for key, value in encoded_simple["init_kwargs"].items()
+            if key in required_fields_init_kwargs
+        }
+
+        for value in encoded_simple:
+            if isinstance(encoded[value], set):
```
AleCandido, line 70: In which case is this a `set`?

Jacfomg: The `_control_qubits` were the set

AleCandido (approved):

src/qibo/gates/abstract.py
```diff
@@ -1,11 +1,15 @@
...
+import json
 from typing import List, Sequence, Tuple
 
 import sympy
 
 from qibo.backends import GlobalBackend
 from qibo.config import raise_error
 
+REQUIRED_FIELDS = ["name", "init_kwargs", "_target_qubits", "_control_qubits"]
+REQUIRED_FIELDS_INIT_KWARGS = ["theta", "phi", "lam"]
```
stavros11, line 11: You could include the `trainable` here, right? Might be relevant for some applications, but I am not sure if those would use the serialization anyway.

stavros11 (approved):
Thanks for implementing this. It is a bit minimal and I am not sure if it works for all gates (maybe it will fail for `Unitary`) but definitely better than nothing.

Jacfomg:
> Thanks for implementing this. It is a bit minimal and I am not sure if it works for all gates (maybe it will fail for `Unitary`) but definitely better than nothing.

It probably won't wont work for that. For the RB which are the only ones that asked to save gates I think will work.

src/qibo/gates/abstract.py
Jacfomg: Yeah, I don't know either, for what I needed it wasn't necessary. This can extended as the application that requires it appears. I would try to save the least for now.

AleCandido: Agreed, let's stick to something simple.

I want to rework the gates anyhow, and (if possible) write a simpler and more robust serialization. Maybe we will need to improve this in any case, but maybe not.

PR #1022, opened 2023-09-28
Fix `apply_channel()` for statevector simulation

### LINKED ISSUE #1018: `DepolarizingChannel` gives wrong probabilities when used with state vectors
**Describe the bug**

The `DepolarizingChannel` with `lam=1` on a single qubit should give 0.5 probability for both 0 and 1 but gives 0.6 - 0.4.
The problem does not appear when using density matrices.

**To Reproduce**
```py

nqubits = 1
qc = Circuit(nqubits)
lam = 1

qc.add(gates.DepolarizingChannel([0,], lam))
qc.add(gates.M(0))

# Execute circuit
result = qc.execute(nshots=10000)
counts = dict(result.frequencies(binary=True))
print(counts)
```

**Expected behavior**
Gives `{'0': 6316, '1': 3684}`. 

When using `qc = Circuit(nqubits, density_matrix=True)` it gives `{'0': 5035, '1': 4965}` which is more right.

AlejandroSopena:
This PR fixes #1018 . There was an error in the way we were sampling the gates to apply when using the  `apply_channel()` method. This also affects the `PauliNoiseChannel` and the `UnitaryChannel`.
We did not notice earlier because the tests are also not implemented correctly. I will fix them as soon as possible.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/backends/numpy.py
+++ b/src/qibo/backends/numpy.py
@@ -256,9 +256,13 @@ class NumpyBackend(Backend):
         return np.reshape(state, 2 * (2**nqubits,))
 
     def apply_channel(self, channel, state, nqubits):
-        for coeff, gate in zip(channel.coefficients, channel.gates):
-            if self.np.random.random() < coeff:
-                state = self.apply_gate(gate, state, nqubits)
+        index = np.random.choice(
+            range(len(channel.gates) + 1),
+            p=channel.coefficients + (1 - np.sum(channel.coefficients),),
+        )
+        if index != len(channel.gates):
+            gate = channel.gates[index]
+            state = self.apply_gate(gate, state, nqubits)
         return state
 
     def apply_channel_density_matrix(self, channel, state, nqubits):
--- a/src/qibo/backends/numpy.py
+++ b/src/qibo/backends/numpy.py
@@ -256,10 +256,8 @@ class NumpyBackend(Backend):
         return np.reshape(state, 2 * (2**nqubits,))
 
     def apply_channel(self, channel, state, nqubits):
-        index = np.random.choice(
-            range(len(channel.gates) + 1),
-            p=channel.coefficients + (1 - np.sum(channel.coefficients),),
-        )
+        probabilities = channel.coefficients + (1 - np.sum(channel.coefficients),)
+        index = self.sample_shots(probabilities, 1)[0]
         if index != len(channel.gates):
             gate = channel.gates[index]
             state = self.apply_gate(gate, state, nqubits)
--- a/tests/test_measurements_probabilistic.py
+++ b/tests/test_measurements_probabilistic.py
@@ -83,14 +83,15 @@ def test_measurements_with_probabilistic_noise(backend):
 
     backend.set_seed(123)
     target_samples = []
+    channel_gates = [gates.Y, gates.Z]
+    probs = [0.2, 0.4, 0.4]
     for _ in range(20):
         noiseless_c = models.Circuit(5)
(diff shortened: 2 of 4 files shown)


stavros11 (approved):
Thank you very much for the very quick fix @AlejandroSopena. Indeed, the previous implementation of `apply_channel` was not right.

Another test that we could do to find this bug earlier is to compare the shots or final probability distribution found with statevector simulation with the ones found using density matrix simulation, since the latter is deterministic (does not use random sampling) even when using channels. I did the following check to verify this PR:
```py
from qibo.quantum_info.random_ensembles import random_unitary

matrix1 = random_unitary(2)
matrix2 = random_unitary(4)

c1 = Circuit(2, density_matrix=True)
c1.add(gates.UnitaryChannel([(0,), (0, 1)], [(0.1, matrix1), (0.2, matrix2)]))
c1.add(gates.M(0))

c2 = Circuit(2)
c2.add(gates.UnitaryChannel([(0,), (0, 1)], [(0.1, matrix1), (0.2, matrix2)]))
c2.add(gates.M(0))

result = c1(nshots=10000)
print(result.frequencies(binary=True))
result = c2(nshots=10000)
print(result.frequencies(binary=True))
```
The final frequencies differ slightly so we would need a tolerance if we convert this to a test.

Anyway, this PR is fine for me as it is.

renatomello:
> Thank you very much for the very quick fix @AlejandroSopena. Indeed, the previous implementation of `apply_channel` was not right.
> 
> Another test that we could do to find this bug earlier is to compare the shots or final probability distribution found with statevector simulation with the ones found using density matrix simulation, since the latter is deterministic (does not use random sampling) even when using channels. I did the following check to verify this PR:
> 
> ```python
> from qibo import Circuit, gates
> from qibo.quantum_info.random_ensembles import random_unitary
> 
> matrix1 = random_unitary(2)
> matrix2 = random_unitary(4)
> 
> c1 = Circuit(2, density_matrix=True)
> c1.add(gates.UnitaryChannel([(0,), (0, 1)], [(0.1, matrix1), (0.2, matrix2)]))
> c1.add(gates.M(0, 1))
> 
> c2 = Circuit(2)
> c2.add(gates.UnitaryChannel([(0,), (0, 1)], [(0.1, matrix1), (0.2, matrix2)]))
> c2.add(gates.M(0, 1))
> 
> result = c1(nshots=10000)
> print(result.frequencies(binary=True))
> result = c2(nshots=10000)
> print(result.frequencies(binary=True))
> ```
> 
> The final frequencies differ slightly so we would need a tolerance if we convert this to a test.
> 
> Anyway, this PR is fine for me as it is.

This is the most appropriate way of testing this. So I'm  gonna add this test.

scarrazza:
@renatomello thanks, are you planning to do this here? I am just asking because we would like to include this bug fix in the next release.

renatomello:
> @renatomello thanks, are you planning to do this here? I am just asking because we would like to include this bug fix in the next release.

already working on it, I'll be done soon

AlejandroSopena:
Thanks @stavros11, @renatomello. I also think this the best way to test this.

renatomello (approved):

stavros11 (approved):
Thanks for adding the test @renatomello. The issue with switching `circuit.density_matrix` after adding gates should be fixed in #1024. Other than that, the test looks good to me.

The only problem I see is that it is a bit slow, but this seems to be related to qibojit, so I'd prefer to open an issue instead of blocking this PR.

PR #1025, opened 2023-10-02
Repeated execution and density matrices

stavros11:
While doing some tests for #1022 I realized that `circuit.repeated_execution` was not updated properly when the `circuit.density_matrix` flag is switched after the gates are added. For example:
```py
circuit = Circuit(2)
circuit.add(gates.UnitaryChannel(...)) # circuit.repeated_execution is now switched to True because a channel is added
circuit.density_matrix = True # does not disable repeated execution
```
so in the end the circuit will be executed repeatedly even though we are using density matrices.

Here I changed `circuit.repeated_execution` to a property that is evaluated right before the circuit is executed and is `True` when the circuit contains collapse measurements or channels and density matrices are disabled.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/models/circuit.py
+++ b/src/qibo/models/circuit.py
@@ -158,8 +158,9 @@ class Circuit:
 
         self._final_state = None
         self.compiled = None
-        self.repeated_execution = False
 
+        self.has_collapse = False
+        self.has_unitary_channel = False
         self.density_matrix = density_matrix
 
         # for distributed circuits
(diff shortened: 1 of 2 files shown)


renatomello:
Some noise channels (e.g. `ThermalRelaxation`, `AmplitudeDamping`, `PhaseDamping`, `ReadoutError`) do not inherit `UnitaryChannel`, so I think this current implementation would not work for them, right?

stavros11:
> Some noise channels (e.g. `ThermalRelaxation`, `AmplitudeDamping`, `PhaseDamping`, `ReadoutError`) do not inherit `UnitaryChannel`, so I think this current implementation would not work for them, right?

I think so, but this should also be the case with master. Here I did not change for which channels we are doing repeated execution, I just made sure it is disabled when `circuit.density_matrix` is set to `True`.

If we can extend the probabilistic execution to more channels, which I believe we can, then we can give a try, but maybe in a different PR. I am not sure if our current implementation of `backend.apply_channel` would work as it is though, I think we need some modifications.

Issue #1019, opened 2023-09-28
CVXPY transitive dependency not working on Mac M1


AleCandido:
We depend on CVXPY only in the tests:
https://github.com/qiboteam/qibo/blob/21a4f2023c307e5a7cd23fa6152256a251a611ae/pyproject.toml#L49-L52
and only because of a single test:
https://github.com/qiboteam/qibo/blob/21a4f2023c307e5a7cd23fa6152256a251a611ae/src/qibo/quantum_info/metrics.py#L901-L981

However, because of its transitive dep `qdldl`, it is required to have CMake on your machine to compile from source the package (since wheels are only available for MacOS x86).

https://github.com/qiboteam/qibo/blob/21a4f2023c307e5a7cd23fa6152256a251a611ae/poetry.lock#L889
https://github.com/qiboteam/qibo/blob/21a4f2023c307e5a7cd23fa6152256a251a611ae/poetry.lock#L2561
https://github.com/qiboteam/qibo/blob/21a4f2023c307e5a7cd23fa6152256a251a611ae/poetry.lock#L3516-L3518

I'm pointing out this isolated dependence, because it increases the number of upstreams project. However, there is no priority for this in the short term.

AleCandido:
A reasonable alternative could be:

https://docs.scipy.org/doc/scipy/tutorial/optimize.html#constrained-minimization-of-multivariate-scalar-functions-minimize

PR #1030, opened 2023-10-05
Circuit serialization

### LINKED ISSUE #809: Circuit serialization
For `Circuit` class, we could apply what described in https://github.com/qiboteam/qibo/pull/804#issuecomment-1443186870.

For the evaluated circuits (i.e. arrays), we just have to make sure that NumPy file formats, i.e. `.npy`/`.npz`, are portable. In principle, they should be, for as long as NumPy itself is available on the platform (the size of floats and integers, their endianness, the shape, and other metadata should all be saved in the file itself).
Most likely, we want some benchmarks for this, to continuously ensure that we are able to load these arrays.

AleCandido:
Closes #809 

Checklist:
- [ ] Tests are passing.


AleCandido:
For the time being is fully sketchy (unfortunately, I'm still looking for a suitable computer to develop on, since I can't use my laptop because of #1019...)

AleCandido:
Sorry for the changes in the docstring, the most of it is due to an aggressive formatter (and since it already changed some, I updated some other manually).

I'm in trouble in testing, as said above. Thus, it is also annoying to provide tests (since I can't run them). I could hack the deps and restore before merging, but I will just wait for a moment.
In case anyone wanted to keep going, the test is just serializing, and checking the presence of trivial arguments, then deserialize and check identity (for simple enough objects, serialization + deserialization should really yield identity).

src/qibo/models/circuit.py
```diff
@@ -1012,6 +1022,27 @@ def __call__(self, initial_state=None, nshots=None):
         """Equivalent to ``circuit.execute``."""
         return self.execute(initial_state=initial_state, nshots=nshots)
 
+    @property
+    def raw(self) -> dict:
+        """Serialize to dictionary.
+
+        This is a thin wrapper over :meth:`Gate.raw`.
+        """
+        return {"queue": [gate.raw for gate in self.queue], "nqubits": self.nqubits}
```
Jacfomg, line 1031: As a note, if I understood correctly what @ingoroth explained we should be saving not only the `circuit.queue` but also the `circuit.queue.moments` as the latter should be a more precise representation that could differ from the rebuild circuit by just adding the same gates back.

AleCandido: According to @stavros11 the moments are built from what you have done with `add`, so there should not be any information loss.

Could you (or @ingoroth) provide an example of a circuit in which the `circuit.queue.moments` are not determined only by the addition sequence?

stavros11: Indeed, I think a given queue is mapped to a specific moment structure, because we don't provide options on how the moments are generated and by default each gate will be applied to the earliest possible moment. Therefore, if you save the queue and then you reconstruct the circuit you should end up with the same moments.

Jacfomg: If for now we don't have any options to play with the gate placement then they should be the same. One example would be instead of playing 2 gates in parallel in 2 qubits to play one gate and hte when it finishes play the other, I think. But, as we don't even are able to generate that circuit it should no be a problem for the serialiaze but I guess some people may want this flexibility when building circuits.

AleCandido: Ok, then my current proposal would be to roll out serialization as is (just give me the time of writing tests).

The moment we will add the feature proposed (to support moments customization) we will upgrade serialization to support it as well.

AleCandido:
In f5fa275 I commented out a couple of dependencies, for the reason expressed in the commit message.

I will undo it before merging.

AleCandido:
A lockfile has to be regenerated (but I wanted to avoid downloading TensorFlow myself...)

AleCandido:
Now the serialization tests are available and passing, so a review could be considered.

However, take into account that the current serialization is quite poor, because of the existing constraints:
- we are relying on `init_args` and `init_kwargs`
	- meaning that any `Gate` implementor has to properly define these arguments; they are storing redundant information, but the proper way would be to record the data in exactly a single place, and tell the class which is this place - but this is exactly the *fields* mechanics of dataclasses
	- any gate can decide to have further fields, but we won't record them if they are not set at constructor level; even though an improvement could be to serialize the value of some attributes, whatever they are (even if they are modified after creation), as it will automatically with fields (i.e. tracking the attributes and not *init* values), but I'd not go beyond that (i.e. I would not allow the gate to serialize arbitrary information) 
- we don't have a mechanism to *register* the gate types, since it's the user that appends a gate object (already instantiated) to the circuit
	- thus, for deserialization, I'm assuming the gate is in the module `qibo.gates.gates`
	- this is quite a poor choice, but we need the class to be available to deserialize it (unless we pickle it, in which case the class could be at an arbitrary location, but it has still to be available, and even to be the same)

The message is: gates have to be refactored, sooner or later.
And the "plugin system" to give the flexibility to extend the framework (that for Qibolab would span drivers, and for Qibocal routines) for Qibo should both include backends and gates.

AleCandido:
Actually, I'm still missing a test for the circuit!

scarrazza:
@AleCandido thank you, could you please also include a `to_json` and load from `json` at the level of `Circuit`?

scarrazza:
I think this implementation does not stores the Qibo version, `density_matrix` and noise (i.e. `has_collapse`). 
Could you include those features so we have a closer map to the circuit results objects?

AleCandido:
> I think this implementation does not stores the Qibo version, `density_matrix` and noise (i.e. `has_collapse`).
> Could you include those features so we have a closer map to the circuit results objects?

Qibo version for sure, I'm doing it.

About the other two, I could store them, of course, but (if not now) we need to discuss which is the purpose of the serialization, and which arguments are in and which are out.

In particular, `has_collapse` should not be set manually (from what I see in the `Circuit` class: it is not an `__init__` parameter, and it should be set by `add`). Since the gates are `add`ed back, are you sure that I should set it separately?

stavros11:
> In particular, `has_collapse` should not be set manually (from what I see in the `Circuit` class: it is not an `__init__` parameter, and it should be set by `add`). Since the gates are `add`ed back during deserialization, are you sure that I should set it separately?

Indeed `has_collapse` does not need to be serialized as it will be automatically set by `.add` once you add a measurement that has `collapse=True`. That assumes that the `collapse` flag is properly saved and loaded when dumping the measurement though. `density_matrix` needs to be serialized because it is passed during `Circuit.__init__`.

scarrazza:
I think we should introduce a load/dump method for circuits in order to match the results approach.

AleCandido:
I'm serializing the Qibo version, but I'm not consuming it during deserialization, since I don't want to enforce unwanted behaviors (a sensible one would be to ask a compatible version, same major, same minor or higher, but we're not too consistent with major/minor meaning yet).

AleCandido:
> I think we should introduce a load/dump method for circuits in order to match the results approach.

I would argue the same for the results
https://github.com/qiboteam/qibo/blob/d8f5fd71f42fd9a4007f9f774488361fb8a51250/src/qibo/gates/measurements.py#L195-L210

Loading and dumping to dictionary is much more flexible, because you simply make a more atomic function which result is directly available.
If I prefer to serialize in a binary format and I only have `to_json` I need first to serialize to a JSON string, deserialize to a dictionary, and finally serialize again. Dictionaries are the common language (the IR), while JSON is already an opinionated choice (and I believe we have no particular reason to endorse it).

The remote connection can take the dictionary and apply JSON on its own. It's a single call to a standard library function, after all. I see no reason to hardcode it in Qibo.
https://github.com/qiboteam/qibo/blob/d8f5fd71f42fd9a4007f9f774488361fb8a51250/src/qibo/gates/measurements.py#L200
https://github.com/qiboteam/qibo/blob/d8f5fd71f42fd9a4007f9f774488361fb8a51250/src/qibo/gates/measurements.py#L205

stavros11:
We were discussing with @scarrazza and maybe it is useful to still dump/load from file directly to simplify the usage from a user point of view. One potential solution is what is implemented in #1039 for the circuit output:
https://github.com/qiboteam/qibo/blob/4dfe721a1bc61cebcf526a6cfa9365823006e2aa/src/qibo/result.py#L92

where we have `dumps` to dictionary, `dump` to json file and similarly `load` and `loads`.

AleCandido:
As you wish, it takes nothing to add a method (it takes more to remove, because it will break compatibility).

> where we have `dumps` to dictionary, `dump` to json file and similarly `load` and `loads`.

Consider that the naming choice is pretty unfortunate: in standard library `dump` means to file, but `dumps` means to string (the `s` at the end).

src/qibo/models/circuit.py
```diff
@@ -1012,6 +1024,41 @@ def __call__(self, initial_state=None, nshots=None):
...
+        Essentially the counter-part of :meth:`raw`.
+        """
+        circ = cls(raw["nqubits"], density_matrix=raw["density_matrix"])
+
+        for gate in raw["queue"]:
+            circ.add(Gate.load(gate))
+
+        return circ
+
+    def dump_json(self, stream):
```
scarrazza, line 1053: I would simply call `dump` and pass the `filename` instead of `stream`.

src/qibo/models/circuit.py
```diff
@@ -1012,6 +1024,41 @@ def __call__(self, initial_state=None, nshots=None):
...
+        """Dump json representation to the given stream."""
+        json.dump(self.raw, stream)
+
+    @classmethod
+    def load_json(cls, stream):
```
scarrazza, line 1058: This should be replaced with `load(filename)`.

src/qibo/models/circuit.py
```diff
@@ -1012,6 +1024,41 @@ def __call__(self, initial_state=None, nshots=None):
         """Equivalent to ``circuit.execute``."""
         return self.execute(initial_state=initial_state, nshots=nshots)
 
+    @property
+    def raw(self) -> dict:
```
scarrazza, line 1028: raw -> to_dict

src/qibo/gates/abstract.py
```diff
@@ -54,7 +59,15 @@ def __init__(self):
         self.device_gates = set()
         self.original_gate = None
 
-    def to_json(self):
+    @property
+    def raw(self) -> dict:
```
scarrazza, line 63: May I suggest to rename this to `to_dict` as done for other python packages?

src/qibo/gates/abstract.py
```diff
@@ -70,7 +83,35 @@ def to_json(self):
...
             if isinstance(encoded[value], set):
                 encoded_simple[value] = list(encoded_simple[value])
-        return json.dumps(encoded_simple)
+
+        encoded_simple["_class"] = type(self).__name__
+
+        return encoded_simple
+
+    @staticmethod
+    def load(raw: dict):
```
scarrazza, line 92: Maybe `from_dict` instead?

src/qibo/models/circuit.py
AleCandido: The reason why we started calling `raw` and not `to_dict` is because it is not a method but a property.

Functions (including methods) should have names of actions, but attributes (including properties) should have name of objects (what it *is* and not what it *does*).

scarrazza: Ok, fine by me then.

src/qibo/models/circuit.py
AleCandido: I put `_json` on purpose, to be explicit about the action performed.

`json.dump` itself is dumping on a file descriptor or a stream, so `filename` would be plainly wrong.
This has to be used as:
```py
with open("my_file.json", "w") as fd:
    circuit.dump_json(fd)
```

Using filenames is uselessly less general, because it prevents you to use in-memory streams to emulate files (as those available in `io`).

src/qibo/models/circuit.py
AleCandido: See above

src/qibo/gates/abstract.py
AleCandido: See above

src/qibo/gates/abstract.py
AleCandido: This is fine, but `load` is just simpler than `from_dict`. Since there was no `to_dict` I preferred the general name.

AleCandido:
All the names have been carefully chosen, motivations are given above.

Notice that, as opposed to what you say, the other packages already transitioned (at least partially) to this terminology, rather than `to_dict`.

https://github.com/qiboteam/qibolab/blob/c6d5b14f4746bf44e73feb5e49b7117ecae7cdbb/src/qibolab/native.py#L75-L85
https://github.com/qiboteam/qibolab/commit/04a29a44520186c9d0ac7bc3879fb30030102d5f
https://github.com/qiboteam/qibocal/pull/299

AleCandido:
I express the motivations for the choices about the `_json` functions, but if you believe it's better to switch to the simpler names that's fine.

I'm just waiting to implement it to get a reply about the `filename` vs `stream` issue, at least I have to touch them only once.

src/qibo/models/circuit.py
```diff
@@ -1012,6 +1024,41 @@ def __call__(self, initial_state=None, nshots=None):
...
+        return circ
+
+    def dump_json(self, stream):
+        """Dump json representation to the given stream."""
+        json.dump(self.raw, stream)
+
+    @classmethod
+    def load_json(cls, stream):
+        """Dump json representation to the given stream."""
+        return cls.load(json.load(stream))
```
scarrazza, line 1060: This `cls.load` does seem to exists in the code.

scarrazza (commented):
@AleCandido, I just had a chat with @stavros11. Please remove the `dump_json` and `load_json` because these methods are useless with stream. Put in this way, if I am a developer, I don't need these methods, only high level users would appreciate a dump/load method which writes/reads to/from disk.

src/qibo/models/circuit.py
AleCandido: Yes, it was a leftover. I received the Pylint notification.

scarrazza:
Please remove, let's keep simple.

scarrazza:
Another thing, this example:
```python
c = Circuit(1)
c.add(gates.M(0))
raw = c.raw
c2 = Circuit.from_dict(raw)
```
returns:
```
AttributeError: module 'qibo.gates.gates' has no attribute 'M'
```

AleCandido:
Unfortunately, this is part of what I was mentioning here:

> - we don't have a mechanism to _register_ the gate types, since it's the user that appends a gate object (already instantiated) to the circuit
>   * thus, for deserialization, I'm assuming the gate is in the module `qibo.gates.gates`

It is because the `M` gate is defined inside this file:
https://github.com/qiboteam/qibo/blob/master/src/qibo/gates/measurements.py
instead of together with all the others:
https://github.com/qiboteam/qibo/blob/master/src/qibo/gates/gates.py

Then, for measurement in particular, I will fix, of course.

Should I include even `channels.py` and `special.py`?
Moreover, I hope there is no name clash, but they belong to different namespaces, so it could be.
In case, I will use as priority `(gates, measurement, special, channels)`.

If I had to allow for the different namespaces, I should serialize with the full class reference, but this means that I'd not be able to deserialize if we move things around. That's why I will avoid for the time being.

scarrazza:
Let's start with `(gates, measurements)`.

scarrazza (approved):
Thanks @AleCandido. Could you please fix coverage (the error raise)? I would like to merge this on Monday to unlock the remote connection project.

AleCandido:
@scarrazza: sorry not to have noticed before, I was not running the full suite of tests for brevity, and I received the Codecov failure together with you.

I should have fixed now, but for the same reason let's wait for the Codecov report.

Remember that this PR is not pointing to `master`, but to `nix`, because I added some more files to have a reproducible development shell myself.
If it's fine for you, I can make a minimal PR even for the `nix` branch (it's just adding three decoupled files). Otherwise I could rebase this branch on `master`, and avoid the content of `nix`.

scarrazza:
Thanks, I am fine adding nix, but we should have this in master asap.

AleCandido:
I opened #1067 and I'm done with both that and this one. Feel free to merge whenever you wish.

stavros11 (approved):
Thanks @AleCandido. It works, just pointing out that due to the try-except used in `Gate.from_dict` sometimes the logger shows errors that are not relevant. For example
```

c = Circuit(5)
c.add(gates.H(i) for i in range(5))
c.add(gates.CNOT(0, 1))
c.add(gates.CNOT(2, 3))
c.add(gates.M(0, 2, 4))
c.add(gates.M(1, 3))

c2 = Circuit.from_dict(c.raw)
```
gives
```
[Qibo 0.2.2|ERROR|2023-10-30 16:20:17]: Cannot use `controlled_by` method on gate <qibo.gates.gates.CNOT object at 0x7f7ea82e2810> because it is already controlled by (0,).
[Qibo 0.2.2|ERROR|2023-10-30 16:20:17]: Cannot use `controlled_by` method on gate <qibo.gates.gates.CNOT object at 0x7f7ec83aa010> because it is already controlled by (2,).
[Qibo 0.2.2|ERROR|2023-10-30 16:20:17]: Measurement gates cannot be controlled.
[Qibo 0.2.2|ERROR|2023-10-30 16:20:17]: Measurement gates cannot be controlled.
```
even though the final circuit `c2` is correct. This is obviously not blocking for the remote connection project, but may be confusing for users.

A potential way to solve would be to use the existing `Gate.is_controlled_by` attribute (you would need to include this to the serialization), but if we are planning to refactor gates anyway we could even postpone.

scarrazza:
Thanks @AleCandido, let's merge this now, but please open an issue or another PR following up the comments by @stavros11 .

PR #1031, opened 2023-10-05
Optimizers interface

MatteoRobbiati:
### General comment
This PR is built on top of #962.
Several changes are purposed here:
- optimizers are divided into three types: meta-heuristic algorithms, minimizers (I think this name should be changed) and gradient based. To each type corresponds one file e.g. `qibo.optimizers.heuristics` or `qibo.optimizers.minimizers`;
- in `qibo.optimizers.minimizers` is introduced a generic `ScipyMinimizer` class, which can be used to call any minimization method provided by Scipy. For this reason, the specific classes `BFGS` and `Powell` have been removed, since they can be executed by setting the desired method inside the `ScipyMinimizer` class;
- the Basin-Hopping algorithm is moved to the heuristic collection;
- We defined an abstract optimizer class in `qibo.optimizers.abstract.py`, on top of which the other optimizers are built.
- the "verbosity" is totally removed, in order to rely on the single options provided by the specific packages we rely on;

### Usage 
As already purposed by @WanderingMike, the  idea is that we have an abstract `Optimizer` class, which takes four main arguments: `loss, args, initial_parameters` and `options`. The first three are straightforward:
- `loss` is the black box loss function we want to optimize;
- `args` are the arguments of the loss we don't want to optimize;
- `initial_parameters` is the initial set of trainable parameters.

The last one, `options`, can be used to pass specific arguments to the functions we use in the optimization process. 

In some specific cases, some extra argument can be used, depending on the optimization function we use. For example, the `qibo.optimizers.minimizers.ParallelBFGS` function can also accept a `parallel` argument.

Every optimizer has now a `fit()` method which relies on the specific optimization function used.

### Example

```python
import numpy as np
from qibo import models, hamiltonians, gates, optimizers
from qibo.optimizers.heuristics import CMAES, BasinHopping
from qibo.optimizers.minimizers import ScipyMinimizer, ParallelBFGS

# define a variational model
nqubits = 2
nlayers = 3

c = models.Circuit(nqubits)
for l in range(nlayers):
    for q in range(nqubits):
        c.add(gates.RY(q=q, theta=0))
        c.add(gates.RY(q=q, theta=0))
    for q in range(nqubits-1):
        c.add(gates.CNOT(q0=q, q1=q+1))

# define a loss function
h = hamiltonians.Z(nqubits)

def loss(parameters, circuit, hamiltonian):
    circuit.set_parameters(parameters)
    return hamiltonian.expectation(circuit().state())

# initialize parameters
params = np.random.randn(2 * nqubits * nlayers)
print(f"Initial loss function value: {loss(params, c, h)}")

# -------------------------------- EXAMPLE 1
# -------------------------------- optimize using Powell Scipy's minimizer
options = {
    "method": "Powell",
    "tol" : 1e-7,
}

opt = ScipyMinimizer(initial_parameters=params, loss=loss, args=(c,h), options=options)
opt.fit()

# -------------------------------- EXAMPLE 2
# -------------------------------- optimize using CMA-ES
options = {"sigma0": 0.10}

opt = CMAES(initial_parameters=params, loss=loss, args=(c,h), options=options)
opt.fit()

# and so on with the others
```

### What is still missing

Something is still missing here:
- [x] the old pure TensorFlow `sgd` method, which is still needed for optimizations which rely on the back propagation procedure;
- [ ] a custom differentiation module to compute SGD in an hardware compatible way. I think this can be a discussion for a second time.

Please give me feedback on this @qiboteam/coredevs, @WanderingMike. 
I think a first step can be to review this and once it is ready, to merge it into #962.

Checklist:
- [ ] Tests are passing.


BrunoLiegiBastonLiegi:
Thanks @MatteoRobbiati looks good, the only thing I would change is giving the ability to the user to pass directly the backend.
Could it be worth it to have a general `Loss` object to implement some standard losses and avoid defining the loss every time?

MatteoRobbiati:
> Thanks @MatteoRobbiati looks good, the only thing I would change is giving the ability to the user to pass directly the backend.

I was thinking about this, but in principle one can set the backend at the beginning of the code, without needing to pass it to the optimizer. Is this maybe the best solution? 

> Could it be worth it to have a general `Loss` object to implement some standard losses and avoid defining the loss every time?

I think this is a very useful tool. For now, the most flexible way to approach the optimization is to define a black box loss function, which is the objective of the minimization. But I think it can super nice to have a set of pre-computed losses! 
I think we can open an issue for a future improvement.

AleCandido:
> I was thinking about this, but in principle one can set the backend at the beginning of the code, without needing to pass it to the optimizer. Is this maybe the best solution?

How the backend should be used is a bit controversial.

Most of the examples are suggesting to set a global one. If it's working with that, I'd say it's sufficient for the time being.

(In general, I'd rather avoid using global objects at all, but this is for later)

WanderingMike:
@MatteoRobbiati . Looks good to me. I like the separation into different files, although we'll have to create some sort of lookup table or list on qibo.science for users to easily find the optimizer they want to use. 

One issue that might arise is related to a discussion we've had a few times this summer. To optimize the circuit for SGD on-chip, we need access to the circuit itself. Your current implementation doesn't mention how we would do this. I think @AleCandido's loss object could help here. When running SGD on-chip, we could limit the scope/nature of the loss function within this loss object, and the optimizer could access the circuit by pointing to the circuit stored in the loss object.

WanderingMike:
For the TensorFlowSGD, would it make sense to stop the iterations as soon as the mean norm of the gradients dips below a certain threshold like 1e-6? Instead of continuing to run an optimisation schedule which has reached a local or global minimum

MatteoRobbiati:
> although we'll have to create some sort of lookup table or list on qibo.science for users to easily find the optimizer they want to use.

I suggest to see how it appears in the Qibo documentation with the `sphinx` rendering of the docstrings. Also, I think we can do an Xmind diagram at some point, to show the optimization pipeline and all the available optimizers. This will be 
in line with the rest of the documentation. 

> One issue that might arise is related to a discussion we've had a few times this summer. To optimize the circuit for SGD on-chip, we need access to the circuit itself. Your current implementation doesn't mention how we would do this. I think @AleCandido's loss object could help here. When running SGD on-chip, we could limit the scope/nature of the loss function within this loss object, and the optimizer could access the circuit by pointing to the circuit stored in the loss object.

Yes, I think once this first set of optimizers is concluded, we can have a meeting and decide how to proceed with the `on-chip` SGD. Btw, I think this structure is flexible enough: we only need to add some more argument to the `HardwareCompatibleAdam` or whatever.

> For the TensorFlowSGD, would it make sense to stop the iterations as soon as the mean norm of the gradients dips below a certain threshold like 1e-6? Instead of continuing to run an optimisation schedule which has reached a local or global minimum.

Of course, we can add the possibility of setting an early-stopping condition.

src/qibo/optimizers/gradient_based.py
```diff
@@ -1 +1,83 @@
 """Gradient descent strategies to optimize quantum models."""
+
+import tensorflow as tf
+
+from qibo.backends import TensorflowBackend
+from qibo.config import log, raise_error
+from qibo.optimizers.abstract import Optimizer, check_options
+
+
+class TensorFlowSGD(Optimizer):
```
AleCandido, line 9: Stupid remark: use at least a consistent case for `Tensorflow` (I know the official one is `TensorFlow`, but if there is already one around, better to sync the two of them)

MatteoRobbiati:
Now the Tensorflow notation and optimizers tests are fixed. I think, after your reviews, we can merge this into @WanderingMike's PR, and use that one to fix all qibo tests and details.

src/qibo/optimizers/gradient_based.py
```diff
@@ -0,0 +1,135 @@
...
+from qibo.config import log, raise_error
+from qibo.optimizers.abstract import Optimizer, check_options
+
+
+class TensorflowSGD(Optimizer):
+    def __init__(
+        self,
+        initial_parameters,
+        args=(),
+        loss=None,
```
WanderingMike, line 14:
```suggestion
        loss=None,
        args=(),
```

Same order as the docstring

src/qibo/optimizers/heuristics.py
```diff
@@ -0,0 +1,107 @@
...
+
+import cma
+from scipy.optimize import basinhopping
+
+from qibo.config import log
+from qibo.optimizers.abstract import Optimizer, check_options
+
+
+class CMAES(Optimizer):
+    def __init__(self, initial_parameters, args=(), loss=None, options={"sigma0": 0.5}):
```
WanderingMike, line 11:
```suggestion
    def __init__(self, initial_parameters, loss=None, args=(), options={"sigma0": 0.5}):
```

Same order as docstring

src/qibo/optimizers/heuristics.py
```diff
@@ -0,0 +1,107 @@
...
+
+        return r[1].result.fbest, r[1].result.xbest, r
+
+
+class BasinHopping(Optimizer):
+    def __init__(
+        self,
+        initial_parameters,
+        args=(),
+        loss=None,
```
WanderingMike, line 59:
```suggestion
        loss=None,
        args=(),
```

src/qibo/optimizers/minimizers.py
```diff
@@ -0,0 +1,198 @@
...
+from qibo.config import log
+from qibo.optimizers.abstract import Optimizer, check_options
+
+
+class ScipyMinimizer(Optimizer):
+    def __init__(
+        self,
+        initial_parameters,
+        args=(),
+        loss=None,
```
WanderingMike, line 15:
```suggestion
        loss=None,
        args=(),
```

src/qibo/optimizers/minimizers.py
```diff
@@ -0,0 +1,198 @@
...
+        minimizer_kwargs={},
+    ):
+        """
+        Computes the L-BFGS-B using parallel evaluation using multiprocessing.
+        This implementation here is based on https://doi.org/10.32614/RJ-2019-030.
+
+        Args:
+            initial_parameters (np.ndarray or list): array with initial values
+                for gate parameters.
+            loss (callable): loss function to train on.
```
WanderingMike, line 83:
```suggestion
            loss (callable): loss function to train on.
            processes (int): number of parallel processes used to evaluate functions
```

WanderingMike (approved):

WanderingMike:
Fyi the tests fail because there are dependencies on the old optimizers.optimize function. Check out:
- models/variational.py
- tests/test_models_variational.py

MatteoRobbiati:
> Fyi the tests fail because there are dependencies on the old optimizers.optimize function. Check out:

Yes, thanks! I was wondering if fixing the here or on your branch.

MatteoRobbiati:
> Yes, thanks! I was wondering if fixing them here or in your branch.

I had a look into `model.variational.py` and I feel we need to discuss how to re-organize the `minimize` methods there. Each variational model has it's own minimize function, which is typically based on the old `qibo.optimizers.minimize`. Sometimes some customization is necessary, but in general it seems to me we can define an `Optimizer` and pass (or set) it to the `VariationalModel`. 

I personally think this PR should be kept like it is. We can use the #1007 pointing to the #962 to fix both tests and the Variational models compilation strategy. 

What you think about this @WanderingMike, @AleCandido, @BrunoLiegiBastonLiegi, @scarrazza?

WanderingMike:
@MatteoRobbiati What is the current state on this PR? Are we only waiting for a second review?

src/qibo/optimizers/gradient_based.py
```diff
@@ -1 +1,135 @@
...
+        for epoch in range(epochs):
+            # early stopping if loss_treshold has been set
+            if (
+                loss_treshold is not None
+                and (epoch != 0)
+                and (loss_history[-1] <= loss_treshold)
+            ):
+                continue
+
+            loss = sgd_step()
```
BrunoLiegiBastonLiegi, line 129:
```suggestion
            loss = sgd_step().numpy()
```

src/qibo/optimizers/gradient_based.py
```diff
@@ -1 +1,135 @@
...
+            loss_history.append(loss.numpy())
```
BrunoLiegiBastonLiegi, line 130:
```suggestion
            loss_history.append(loss)
```

src/qibo/optimizers/gradient_based.py
```diff
@@ -1 +1,135 @@
...
+
+            if epoch % nmessage == 0:
+                log.info("ite %d : loss %f", epoch, loss_history[-1])
```
BrunoLiegiBastonLiegi, line 133:
```suggestion
                log.info("ite %d : loss %f", epoch, loss)
```

src/qibo/optimizers/gradient_based.py
```diff
@@ -1 +1,135 @@
...
+            (np.ndarray): best parameter values
+            (list): loss function history
+        """
+
+        vparams = self.backend.tf.Variable(
+            self.params, dtype=self.backend.tf.complex128
+        )
+        loss_history = []
+
+        def sgd_step():
```
BrunoLiegiBastonLiegi, line 105: Is there a particular reason why we don't want to expose the sgd step to the user? I think in some cases it could be useful to be able to define the training loop your self for customization.

BrunoLiegiBastonLiegi (commented):
Thanks @MatteoRobbiati, I have just some minor comments.

src/qibo/optimizers/gradient_based.py
MatteoRobbiati: I was only keeping the notation we had in the old optimizer file. 
I am open to move it outside.

PR #1035, opened 2023-10-10
Add custom `decompose` method to `gates.U3` and `gates.CZ`

renatomello:
Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/gates/gates.py
+++ b/src/qibo/gates/gates.py
@@ -255,7 +255,9 @@ class SX(Gate):
         return "sx"
 
     def decompose(self):
-        """A global phase difference exists between the definitions of
+        """Decomposition of :math:`\\sqrt{X}` up to global phase.
+
+        A global phase difference exists between the definitions of
         :math:`\\sqrt{X}` and :math:`\\text{RX}(\\pi / 2)`, with :math:`\\text{RX}`
         being the :class:`qibo.gates.RX` gate. More precisely,
         :math:`\\sqrt{X} = e^{i \\pi / 4} \\, \\text{RX}(\\pi / 2)`.
@@ -296,7 +298,9 @@ class SXDG(Gate):
         return "sxdg"
 
     def decompose(self):
-        """A global phase difference exists between the definitions of
+        """Decomposition of :math:`(\\sqrt{X})^{\\dagger}` up to global phase.
+
+        A global phase difference exists between the definitions of
         :math:`\\sqrt{X}` and :math:`\\text{RX}(\\pi / 2)`, with :math:`\\text{RX}`
         being the :class:`qibo.gates.RX` gate. More precisely,
         :math:`(\\sqrt{X})^{\\dagger} = e^{-i \\pi / 4} \\, \\text{RX}(-\\pi / 2)`.
@@ -865,7 +869,9 @@ class U3(_Un_):
         return self.__class__(self.target_qubits[0], theta, phi, lam)
 
     def decompose(self) -> List[Gate]:
-        """A global phase difference exists between the definitions of
+        """Decomposition of :math:`U_{3}` up to global phase.
+
+        A global phase difference exists between the definitions of
         :math:`U3` and this decomposition. More precisely,
 
         .. math::
@@ -956,7 +962,9 @@ class CZ(Gate):
         return "cz"
 
     def decompose(self) -> List[Gate]:
-        """Decompose :math:`\\text{CZ}` gate into :class:`qibo.gates.H` in the target qubit,
+        """Decomposition of :math:`\\text{CZ}` gate.
+
+        Decompose :math:`\\text{CZ}` gate into :class:`qibo.gates.H` in the target qubit,
         followed by :class:`qibo.gates.CNOT`, followed by another :class:`qibo.gates.H`
         in the target qubit"""
         q0, q1 = self.init_args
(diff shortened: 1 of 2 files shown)


src/qibo/gates/gates.py
```diff
@@ -864,6 +864,27 @@ def _dagger(self) -> "Gate":
         theta, lam, phi = tuple(-x for x in self.parameters)  # pylint: disable=E1130
         return self.__class__(self.target_qubits[0], theta, phi, lam)
 
+    def decompose(self) -> List[Gate]:
+        """A global phase difference exists between the definitions of
+        :math:`U3` and this decomposition. More precisely,
```
AleCandido, line 869: Just a small note on the docstring style: the first line should not wrap, and just be a very brief summary of the function (e.g. `pydocstyle` would check for the first line ending with a `.`).

Everything else can be as long as you wish.

tests/test_gates_gates.py
```diff
@@ -414,6 +421,15 @@ def test_u3(backend):
 
     backend.assert_allclose(final_state, target_state)
 
+    # testing random expectation value due to global phase difference
+    observable = random_hermitian(2**nqubits, backend=backend)
```
AleCandido, line 425: Better adding a seed, otherwise the test would lack reproducibility.

AleCandido (approved):
Just a couple of comments about test and docs, but the feature on its own seems perfectly fine.

tests/test_gates_gates.py
renatomello: The point of not adding a seed is that this test should be observable-independent. It has to work for all observables, a fixed one might introduce bias.

AleCandido: > The point of not adding a seed is that this test should be observable-independent. It has to work for all observables, a fixed one might introduce bias. One option would be testing with multiple seeds, but that still doesn't eliminate the bias.

I understood the point, that's the usual one that motivates random numbers in the unit tests.
But this partially defeats the point of unit tests, since in this way it will become not reproducible.

If you want to test multiple observables, you should pick a single seed, and just extract multiple times. In this way, you will get N random observables, but they have to be always the same for all the runs.

The rationale is that if at some points something goes wrong, you want to jump back and forth and check where it stopped working. But if it's only failing sometimes for some of them, it becomes quite difficult to debug.

renatomello: Seed(s) introduced.

AleCandido:
@renatomello thanks to have applied the suggested corrections!

In general, it would be slightly better to pass a pRNG state, rather than using multiple seeds. It's because the properties of pRNG are optimized for sequential extractions, not necessarily for adjacent seeds.
However, this is certainly beyond the scope of this PR, since we should allow the various functions to get a state (i.e. a [`Generator`](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.Generator) object) instead of a seed.
And you should not notice any issue with just 10 extractions (these problems usually arise at the level of order of the generators' period extractions - or slightly less).

src/qibo/gates/gates.py
```diff
@@ -864,6 +868,29 @@ def _dagger(self) -> "Gate":
...
+        :class:`qibo.gates.RZ` and :class`qibo.gates.SX`.
+        """
+        q = self.init_args[0]
+        return [
+            RZ(q, self.init_kwargs["lam"]),
+            SX(q),
+            RZ(q, self.init_kwargs["theta"] + math.pi),
+            SX(q),
+            RZ(q, self.init_kwargs["phi"] + math.pi),
+        ]
```
AleCandido, line 892: Just one small personal doubt: why do you use the two `SX` gates?

I thought that the basic decomposition would have been just Euler angles, i.e. `[RZ(a), RX(b), RZ(c)]`.
Is there any further reason to use 5 gates?

stavros11 (approved):
Thanks @renatomello for adding this. Looks good to me.

Just noting that soon(ish) we will move the transpilers from qibolab to qibo and it would be great if these decompose methods are translated to transpilers to have a unified approach of gate transformations.

src/qibo/gates/gates.py
renatomello: That's how it is decomposed for hardware purposes. The two $\sqrt{X}$ are applied as a $\\textup{RX}(\\pi / 2)$ pulse each, and the $\\textup{RZ}$ "applied" via virtual phases.

tests/test_gates_gates.py
```diff
@@ -414,6 +423,15 @@ def test_u3(backend):
...
     backend.assert_allclose(final_state, target_state)
 
+    # testing random expectation value due to global phase difference
+    observable = random_hermitian(2**nqubits, seed=seed_observable, backend=backend)
+    backend.assert_allclose(
+        np.transpose(np.conj(final_state_decompose))
+        @ observable
+        @ final_state_decompose,
+        np.transpose(np.conj(target_state)) @ observable @ target_state,
+    )
```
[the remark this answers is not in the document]
renatomello: That's a good observation. But since this is used in multiple other tests, I prefer to replace them all in a separate PR.

src/qibo/gates/gates.py
stavros11: Indeed, that's what the qibolab compiler currently does to U3, but at a pulse level.

AleCandido: Ok, thanks for the answer :)

Then I wonder how much this `decompose` feature is overlapping with the compiler, since the choice of the decomposition is also hardware-driven.
However, we can rediscuss the moment the compiler will be moved to Qibo.

AlejandroSopena (approved):

Issue #1020, opened 2023-09-28
Output of circuit execution


stavros11:
When executing a circuit that contains channels (noise) or collapse measurements using state vectors, qibo will automatically repeat execution multiple times (specified by the given `nshots`) and each time the channel will act non-deterministically. Then the circuit can return different things:

* if no measurements were used in the end: a list of `len(nshots)` state arrays, each one corresponding to a single execution,
* if measurements were used in the end: a `CircuitResult` object for which the samples, frequencies and probabilities are constructed from the repeated executions, however the representation and state correspond to the last execution, so they are random.

Currently `CircuitResult` attempts to accommodate all the different "modes" of execution which makes it a bit hard to use. I may be useful to collect the different "modes" of execution and possibly return different custom objects in each case. 

A proposal would be:

1. no noise, `nshots is None`, no measurements -> `QuantumState`
2. no noise, `nshots is None`, measurements -> `QuantumState` (measurements are ignored if `nshots` is not given)
3. no noise, `nshots is not None`, no measurements -> raise error because `nshots` was given but there are no measurements or noise
4. no noise, `nshots is not None`, measurements -> `Measurements` (this still has a `.state()` property to obtain the state vector since it is always available in simulation - on hardware `.state()` raises an error)
5. noise, `nshots is None`, no measurements -> raise error as we cannot simulate noise without density matrices or shots
6. noise, `nshots is None`, measurements -> raise error as we cannot simulate noise without density matrices or shots
7. noise, `nshots is not None`, no measurements -> `NoisyQuantumState` (contains a list of `len(nshots)` states obtained from each execution where noise was applied randomly)
8. noise, `nshots is not None`, measurements -> `Measurements` (can be the same object as 4 but it does not provide `.state()` - similar to when executing on hardware)

For `density_matrix=True` we can follow the same pattern, except that for 5, 6 we can still return a `QuantumState` containing the density matrix. Or we could even have separate `Statevector` and `DensityMatrix` objects to be more explicit.

This is a quick sketch, any suggestions are welcome.

BrunoLiegiBastonLiegi:
After a discussion with @stavros11 we decided that, as `nshots` is usually tightly connected to whether we perform a measurement or not, to set a default value `nshots=1000`. This simplify a bit the cases we have:

1. no noise, no measurements --> `QuantumState`
2. no noise, measurements --> `QuantumState`, `MeasurementOutcomes`
3. noise, no measurements --> raise error if `density_matrix=False`
[...]

PR #1039, opened 2023-10-10
Circuit Execution Outputs

BrunoLiegiBastonLiegi:
As requested in #1020, three objects are implemented that substitute the old `CircuitResult` object. 
Depending on the different conditions, now the circuit execution returns:

1. no noise, no measurements --> `QuantumState`
2. no noise, measurements --> `CircuitResult`
3. noise, no measurements --> raise error if `density_matrix=False` else `QuantumState`
4. noise, measurements --> `CircuitResult` if `density_matrix=True` else `MeasurementOutcomes`

where the new `CircuitResult(QuantumState, MeasurementOutcomes)` object contains both the `QuantumState` and the `MeasurementOutcomes`.
`nshots` is set by default to `1000` now.
The probabilities and the state can be recovered as `QuantumState.probabilities()` and `QuantumState.state()`, the frequencies and the samples with `MeasurementOutcomes.frequencies()` and `MeasurementOutcomes.samples()`. `CircuitResult` gives access to all of them in the same way.

Currently the only case that has been left out is measurements with `collapse=True`. 
One possibility to handle them could be:
- only collapsing measurements present
      --> `QuantumState` if density_matrix=True
      --> raise Error if density_matrix=False
- both collapsing and standard measurements present
     --> `CircuitResult` if density_matrix=True
     --> `MeasurementOutcomes` if density_matrix=False

Furthermore, handling the `Circuit._final_state` attribute now is a bit problematic, as depending on the conditions different different things are obtained after execution.

Checklist:
- [ ] Tests are passing.


BrunoLiegiBastonLiegi:
Okay this should be ready for review now. I am also working on adding a table to the docs to summarize the circuit execution outcomes depending on the settings.

tests/test_measurements.py
```diff
@@ -472,3 +474,14 @@ def test_measurement_basis_list_error(backend):
     c = models.Circuit(4)
     with pytest.raises(ValueError):
         c.add(gates.M(0, 1, 2, 3, basis=[gates.X, gates.Z, gates.X]))
+
+
+def test_CircuitResult_error(backend):
```
stavros11, line 479:
```suggestion
def test_circuit_result_error(backend):
```
to follow the usual conventions. It's true that we are testing the class but capitals do not look good in function names.

tests/test_measurements_collapse.py
```diff
@@ -6,6 +6,7 @@
 from qibo.quantum_info import random_density_matrix, random_statevector
 
 
+# @pytest.mark.skip(reason="no way of currently testing this")
```
stavros11, line 9:
```suggestion
```
dead code

tests/test_measurements_collapse.py
```diff
@@ -14,18 +15,21 @@ def test_measurement_collapse(backend, nqubits, targets):
...
+    # slicer = nqubits * [slice(None)]
+    # for t, r in zip(targets, samples):
+    #    slicer[t] = int(r)
+    # slicer = tuple(slicer)
+    # initial_state = initial_state.reshape(nqubits * (2,))
+    # target_state = np.zeros_like(initial_state)
+    # target_state[slicer] = initial_state[slicer]
+    # norm = (np.abs(target_state) ** 2).sum()
+    # target_state = target_state.ravel() / np.sqrt(norm)
+    # backend.assert_allclose(final_state, target_state)
```
stavros11, line 32:
```suggestion
```
same here assuming it is no longer valid.

tests/test_models_circuit_execution.py
```diff
@@ -7,7 +7,7 @@
 def test_eager_execute(backend, accelerators):
     c = Circuit(4, accelerators)
     c.add(gates.H(i) for i in range(4))
-    final_state = backend.execute_circuit(c)
+    final_state = backend.execute_circuit(c)._state
```
stavros11, line 10: As in previous comments, prefer
```suggestion
    final_state = backend.execute_circuit(c).state()
```
outside the class.

stavros11 (approved):
Thank you for working on this @BrunoLiegiBastonLiegi. The new objects look good and the results are much cleaner now. I wrote some comments below, note that some of them are hidden because they are many. I still need to have a look in the documentation updates and a more detailed look in the test updates.

Other than that, it may be worth updating this with main and fixing the conflicts every now and then because conflicts will block the merge even if we approve. In principle you can fix once when we are ready to merge but is usually better to maintain this up-to-date with main.

scarrazza:
@BrunoLiegiBastonLiegi thanks for this. Could you please also consider the possibility to store and load the numpy object that are required in order to reallocate the results classes? This is necessary for the future remote connection mechanism based on http requests.

src/qibo/states.py
```diff
@@ -1,242 +1,44 @@
...
-            # individual register samples are registered here
-            self.circuit.final_state.samples()
-        if binary:
-            return np.array(self._samples, dtype="int32")
-        else:
-            qubits = self.measurement_gate.target_qubits
-            return self.backend.samples_to_decimal(self._samples, len(qubits))
+        self.density_matrix = len(state.shape) == 2
+        self.nqubits = int(np.log2(state.shape[0]))
+        self.qubits = tuple(range(self.nqubits))
```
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: `qubits` is actually used in the backend method `circuit_result_probabilities`:

```
    def circuit_result_probabilities(self, state, qubits=None):
        if qubits is None:
            qubits = state.qubits

        state_tensor = state._state
        if state.density_matrix:
            return self.calculate_probabilities_density_matrix(
                state_tensor, qubits, state.nqubits
            )
        else:
            return self.calculate_probabilities(state_tensor, qubits, state.nqubits)
```

maybe by setting  the default value for `calculate_probabilities` methods we could avoid this

src/qibo/states.py
```diff
@@ -1,242 +1,44 @@
...
+        self._repeated_execution_probabilities = None
```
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: `density_matrix` is needed for the `symbolic` method in `QuantumState`
`nqubits` and `qubits` are needed for the `circuit_result_probabilties` above
`_repeated_execution_probabilities` for the `_final_state` attribute of circuit and thus for when the circuit is `compile` -ed

src/qibo/backends/numpy.py
```diff
@@ -420,11 +422,25 @@ def execute_circuit(
...
 
-            if return_array:
-                return state
+            # set the circuit _final_state??
+
+            if circuit.has_unitary_channel:
+                # here we necessarily have `density_matrix=True`, otherwise
+                # execute_circuit_repeated would have been called
+                if circuit.measurements:
+                    circuit._final_state = CircuitResult(state, circuit, self, nshots)
```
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: In the end I decided to keep it, as I think the circuit compilation breaks otherwise. Still, it would probably be better to change the name to `_final_result` as now even a stateless object could be returned in principle.

tests/test_callbacks.py
```diff
@@ -172,7 +172,7 @@ def test_entropy_in_distributed_circuit(
     """Check that various entropy configurations work in distributed circuit."""
     target_c = Circuit(4)
     target_c.add([gates.H(0), gates.CNOT(0, 1)])
-    target_state = backend.execute_circuit(target_c)
+    target_state = backend.execute_circuit(target_c)._state
```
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: yeah sorry about that, the thing is that I modified the `assert_allclose` only after correcting some of the tests manually, indeed I'll remove the `.state()` calls where unnecessary

src/qibo/states.py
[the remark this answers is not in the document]
stavros11: I see, thanks. Then a potential simplification is to remove `circuit_result_probabilities` from the backends and copy its content in the `QuantumState.probabilities`. I think this is the only place this is used anyway. Then the backends will be completely free from `circuit_result_*` methods and you can probably also drop `self.qubits` by doing
```py
if qubits is not None:
    # use all qubits
    qubits = tuple(range(self.nqubits))
```
inside this method.

src/qibo/backends/numpy.py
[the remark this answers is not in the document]
stavros11: Given that we may drop this completely in the future, I wouldn't worry much about the name. However we could open an issue to remember checking if this really needed and if yes maybe update the name in the future.

src/qibo/states.py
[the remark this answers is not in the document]
stavros11: Thanks, I think I meant only the `_repeated_execution_probabilities`. Are you sure it is related to the `_final_state`/`compile`? I don't think these are releated to repeated execution.

src/qibo/backends/numpy.py
```diff
@@ -479,55 +507,64 @@ def execute_circuit_repeated(self, circuit, initial_state=None, nshots=None):
...
+            final_state = np.asarray(final_states).mean(0)
+            if circuit.measurements:
+                qubits = [q for m in circuit.measurements for q in m.target_qubits]
+                final_result = CircuitResult(
+                    final_state, circuit.measurements, self, nshots
                 )
+                probabilities = final_result.probabilities(qubits)
+                final_result._repeated_execution_probabilities = probabilities
+                circuit._final_state = final_result
+                return final_result
```
stavros11, line 531: You can set the `_final_state` and return only once at the end of this function instead of every if case. This will save some repeated lines of code.

src/qibo/backends/numpy.py
```diff
@@ -479,55 +507,64 @@ def execute_circuit_repeated(self, circuit, initial_state=None, nshots=None):
...
+            else:
+                final_result = QuantumState(final_state, self)
+                probabilities = final_result.probabilities()
+                final_result._repeated_execution_probabilities = probabilities
```
stavros11, line 535: Regarding the `QuantumState._repeated_execution_probabilities` (comment above), I think this is the only place they are used. However if I follow the code properly I think it is just doing a trivial caching. You are calculating the probabilities from the `final_result` and then you are setting them back to the same object. If you call `final_result` again after that you will get the values cached in `_repeated_execution_probabilities` but these will be the same with just redoing the calculation. If that's correct, it should be safe to drop this variable completely.

If you need caching, you can handle it internally in the `QuantumState` object, or use [standard Python methods.](https://docs.python.org/3/library/functools.html#functools.cache). However I don't think that's needed in our case.

src/qibo/backends/numpy.py
```diff
@@ -479,55 +507,64 @@ def execute_circuit_repeated(self, circuit, initial_state=None, nshots=None):
...
+                circuit._final_state = final_result
+                return final_result
+        else:
+            probabilities = probabilities / nshots
+            final_result = MeasurementOutcomes(
+                circuit.measurements, probabilities, self, nshots
+            )
+            final_result._samples = self.aggregate_shots(results)
+            final_result._repeated_execution_probabilities = probabilities
```
stavros11, line 544: Also here you are assigning the `_repeated_execution_probabilities` attribute to a `MeasurementOutcomes` object, but this does not exist in the original class definition. Python allows you to set attributes to instantiated objects, even if they do not exist in their definition but it's not a very good practice because then it's harder to track what attributes each object has. Pylint usually complains for this, I am not sure why it is passing now.

If the `_repeated_execution_probabilities` is not used we can remove, otherwise we should define it as `None` in the `MeasurementOutcomes.__init__`.

tests/test_cirq.py
```diff
@@ -59,9 +59,9 @@ def assert_gates_equivalent(
     assert c.depth == target_depth
     if accelerators and not backend.supports_multigpu:
         with pytest.raises(NotImplementedError):
-            final_state = backend.execute_circuit(c, np.copy(initial_state))
+            final_state = backend.execute_circuit(c, np.copy(initial_state))._state
```
stavros11, line 62: This should be replaced with
```suggestion
            final_state = backend.execute_circuit(c, np.copy(initial_state)).state()
```
to avoid using the "private" attribute. Same for three other instances in this file.

tests/test_gates_density_matrix.py
```diff
@@ -165,11 +165,11 @@ def test_controlled_by_no_effect(backend):
     c = Circuit(4, density_matrix=True)
     c.add(gates.X(0))
     c.add(gates.SWAP(1, 3).controlled_by(0, 2))
-    final_rho = backend.execute_circuit(c, np.copy(initial_rho))
+    final_rho = backend.execute_circuit(c, np.copy(initial_rho))._state
```
stavros11, line 168: Same as above in this file.

tests/test_hamiltonians_terms.py
```diff
@@ -50,7 +50,7 @@ def test_hamiltonian_term_gates(backend):
     final_state = term(backend, np.copy(initial_state), nqubits)
     c = models.Circuit(nqubits)
     c.add(gates.Unitary(matrix, 1, 2))
-    target_state = backend.execute_circuit(c, np.copy(initial_state))
+    target_state = backend.execute_circuit(c, np.copy(initial_state))._state
```
stavros11, line 53: Same here.

tests/test_hamiltonians_trotter.py
```diff
@@ -133,7 +133,7 @@ def test_trotter_hamiltonian_three_qubit_term(backend):
     dt = 1e-2
     initial_state = random_statevector(2**4, backend=backend)
     circuit = ham.circuit(dt=dt)
-    final_state = backend.execute_circuit(circuit, np.copy(initial_state))
+    final_state = backend.execute_circuit(circuit, np.copy(initial_state))._state
```
stavros11, line 136: Same here.

tests/test_measurements.py
```diff
@@ -262,9 +262,9 @@ def test_measurement_compiled_circuit(backend):
         {"10": 100},
     )
 
-    target_state = np.zeros_like(c.final_state)
+    target_state = np.zeros_like(c.final_state._state)
```
stavros11, line 265: Same here.

tests/test_measurements_collapse.py
```diff
@@ -68,55 +61,63 @@ def test_measurement_collapse_bitflip_noise(backend):
...
     if int(m.symbols[0].outcome()):
-        c = models.Circuit(4)
+        c = models.Circuit(4, density_matrix=True)
         c.add(gates.RY(0, theta=np.pi / 5))
         c.add(gates.RX(2, theta=np.pi / 4))
         target_state = backend.execute_circuit(c, initial_state=target_state)
     backend.assert_allclose(final_state, target_state)
 
 
+@pytest.mark.skip(reason="no way of currently testing this")
```
stavros11, line 115: For this and other tests that are skipped, we should decide whether:
* we keep as skipped but write a more explicit reason, for example: skipped because we no longer support repeated execution for density matrices (or something like that), or
* remove completely.

Either solution is fine with me.

tests/test_measurements_collapse.py
```diff
@@ -125,81 +126,94 @@ def test_measurement_result_parameters_repeated_execution(backend, use_loop):
...
     backend.set_seed(123)
-    c = models.Circuit(4)
+    c = models.Circuit(4, density_matrix=True)
     r = c.add(gates.M(1, collapse=True))
     c.add(gates.RY(0, theta=np.pi * r.symbols[0] / 3))
     c.add(gates.RY(2, theta=np.pi * r.symbols[0] / 4))
     c.add(gates.M(0, 1, 2, 3))
     result = backend.execute_circuit(c, initial_state=np.copy(initial_state), nshots=30)
     final_samples = result.samples(binary=False)
+    print(final_samples)
```
stavros11, line 165:
```suggestion
```

tests/test_models_circuit_noise.py
```diff
@@ -15,7 +20,7 @@ def test_pauli_noise_channel(backend):
     c.add(gates.H(1))
     c.add(gates.PauliNoiseChannel(0, list(zip(["X", "Z"], [0.5, 0.3]))))
     c.add(gates.PauliNoiseChannel(1, list(zip(["Y", "Z"], [0.1, 0.3]))))
-    final_rho = backend.execute_circuit(c)
+    final_rho = backend.execute_circuit(c)._state
```
stavros11, line 23: Same as above
```suggestion
    final_rho = backend.execute_circuit(c)
```

I think in the cases where you use `backend.assert_allclose`, you don't need to call `.state()` explicitly because this is done automatically in the backend method.

tests/test_models_circuit_parametrized.py
```diff
@@ -205,9 +205,9 @@ def test_set_parameters_with_light_cone(backend, trainable):
         c.set_parameters(np.random.random(4))
     else:
         c.set_parameters(np.random.random(2))
-    target_state = backend.execute_circuit(c)
+    target_state = backend.execute_circuit(c)._state
```
stavros11, line 208: Same as above.

tests/test_models_circuit_qasm_cirq.py
```diff
@@ -28,15 +28,15 @@ def test_from_qasm_simple(backend, accelerators):
         assert gate.__class__.__name__ == "H"
         assert gate.qubits == (i,)
     target_state = np.ones(32) / np.sqrt(32)
-    final_state = backend.execute_circuit(c)
+    final_state = backend.execute_circuit(c)._state
```
stavros11, line 31: Same as above.

src/qibo/result.py
```diff
@@ -0,0 +1,396 @@
...
+        return cls(payload.get("state"), backend)
+
+    @classmethod
+    def load(cls, filename):
+        payload = np.load(filename, allow_pickle=True).item()
+        return cls.loads(payload)
+
+
+class MeasurementOutcomes:
+    """Object to store the outcomes of measurements after circuit execution."""
```
stavros11, line 114: Looking also at the qibolab PR, it could be useful to have a `probabilities` method in the `MeasurementOutcomes` too. This would return `self._probs` if not `None`, otherwise it would calculate the probabilities as frequencies / nshots.

Then in `CircuitResult.probabilities` you just need to make sure that you use the `QuantumState.probabilities` instead of the `MeasurementOutcome.probabilities`.

stavros11 (commented):
Thanks for the updates @BrunoLiegiBastonLiegi, some additional comments. Also, I see an empty file called `test_result.py~`.

src/qibo/result.py
```diff
@@ -0,0 +1,397 @@
...
+            )
+        else:
+            return self.backend.calculate_probabilities(
+                self._state, qubits, self.nqubits
+            )
+
+    def __str__(self):
+        return self.symbolic()
+
+    def dumps(self):
```
scarrazza, line 92: Please rename this to `to_dict`.

scarrazza (approved):

stavros11 (approved):
I had a quick look at the docs as I have not done this before and it looks good to me. Feel free to merge this and open an issue about the probabilities or anything else we could potentially address in the future.

I will test the qibolab part later and if there is any update required here we can fix in a new PR.

PR #1042, opened 2023-10-11
Rename `gates.RXY` as `gates.RXXYY` and fix `gates.SYC` docstring

renatomello:
Checklist:
- [x] Tests are passing.

### DIFF
--- a/doc/source/api-reference/qibo.rst
+++ b/doc/source/api-reference/qibo.rst
@@ -654,10 +654,10 @@ Parametric ZX interaction (RZX)
     :members:
     :member-order: bysource
 
-Parametric XX-YY interaction (RXY)
+Parametric XX-YY interaction (RXXYY)
 """"""""""""""""""""""""""""""""""
 
-.. autoclass:: qibo.gates.RXY
+.. autoclass:: qibo.gates.RXXYY
     :members:
     :member-order: bysource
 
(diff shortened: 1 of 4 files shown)


AleCandido (commented):
This PR is pretty plain, and there is no major effect on the rest of the Qibo code.

However, in principle, is a breaking change wrt the Qibo users' code.

So either you have a strong motivation, or you should know that this gate is not much used by other people.

renatomello:
> This PR is pretty plain, and there is no major effect on the rest of the Qibo code.
> 
> However, in principle, is a breaking change wrt the Qibo users' code.
> 
> So either you have a strong motivation, or you should know that this gate is not much used by other people.

1. This gate is not much used, even though it is important to have it.
2. The previous name (that I created myself) is misleading.

AleCandido:
Sometimes poor choices have to be kept to avoid breaking compatibility (or at least some changes have to be postponed, to support stability).

However, I'm pretty sure there are not that many long-standing pieces of code relying on Qibo (yet). Other than the other packages, we develop ourselves.
That's why, if you know it's not that much used, it is fine to change it.

(never argued it was not useful, on that side it is fine to have it even if it were not crucial)

MatteoRobbiati (approved):

PR #1043, opened 2023-10-12
Add `order` parameter to `backend.calculate_norm`

renatomello:
Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/backends/abstract.py
+++ b/src/qibo/backends/abstract.py
@@ -341,7 +341,7 @@ class Backend(abc.ABC):
     def calculate_norm_density_matrix(self, state, order="nuc"):  # pragma: no cover
         """Calculate norm (trace) of a density matrix.
 
-        If ``order="nuc"``, it returns the nuclear norm os ``state``,
+        If ``order="nuc"``, it returns the nuclear norm of ``state``,
         assuming ``state`` is Hermitian. For specifications on the other
         possible values of the parameter ``order`` for the ``tensorflow`` backend,
         please refer to
--- a/src/qibo/backends/abstract.py
+++ b/src/qibo/backends/abstract.py
@@ -342,8 +342,8 @@ class Backend(abc.ABC):
         """Calculate norm of a density matrix. Default is the ``nuclear`` norm.
 
         If ``order="nuc"``, it returns the nuclear norm of ``state``,
-        assuming ``state`` is Hermitian (also known as trace norm). 
-        For specifications on the other  possible values of the 
+        assuming ``state`` is Hermitian (also known as trace norm).
+        For specifications on the other  possible values of the
         parameter ``order`` for the ``tensorflow`` backend, please refer to
         `tensorflow.norm <https://www.tensorflow.org/api_docs/python/tf/norm>`_.
         For all other backends, please refer to
--- a/src/qibo/backends/abstract.py
+++ b/src/qibo/backends/abstract.py
@@ -327,7 +327,7 @@ class Backend(abc.ABC):
 
     @abc.abstractmethod
     def calculate_norm(self, state, order=2):  # pragma: no cover
-        """Calculate norm of a state vector.
+        """Calculate norm of a state vector. Default is :math:`2`-norm.
 
         For specifications on possible values of the parameter ``order``
         for the ``tensorflow`` backend, please refer to
@@ -339,12 +339,12 @@ class Backend(abc.ABC):
 
     @abc.abstractmethod
     def calculate_norm_density_matrix(self, state, order="nuc"):  # pragma: no cover
-        """Calculate norm (trace) of a density matrix.
+        """Calculate norm of a density matrix. Default is the ``nuclear`` norm.
 
         If ``order="nuc"``, it returns the nuclear norm of ``state``,
-        assuming ``state`` is Hermitian. For specifications on the other
-        possible values of the parameter ``order`` for the ``tensorflow`` backend,
-        please refer to
+        assuming ``state`` is Hermitian (also known as trace norm). 
+        For specifications on the other  possible values of the 
+        parameter ``order`` for the ``tensorflow`` backend, please refer to
         `tensorflow.norm <https://www.tensorflow.org/api_docs/python/tf/norm>`_.
         For all other backends, please refer to
         `numpy.linalg.norm <https://numpy.org/doc/stable/reference/generated/numpy.linalg.norm.html>`_.
--- a/src/qibo/backends/abstract.py
+++ b/src/qibo/backends/abstract.py
@@ -326,13 +326,21 @@ class Backend(abc.ABC):
         raise_error(NotImplementedError)
 
     @abc.abstractmethod
-    def calculate_norm(self, state):  # pragma: no cover
-        """Calculate norm of a state vector."""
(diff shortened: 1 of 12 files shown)


src/qibo/backends/tensorflow.py
```diff
@@ -306,6 +306,20 @@ def entanglement_entropy(self, rho):
         entropy = self.np.sum(masked_eigvals * spectrum) / self.np.log(2.0)
         return entropy, spectrum
 
+    def calculate_norm(self, state, order=None):
+        state = self.cast(state)
```
MatteoRobbiati, line 310: Maybe here you can directly use [`tf.norm`](https://www.tensorflow.org/api_docs/python/tf/norm), which seems to be quite similar to `np.linalg.norm`.

src/qibo/backends/numpy.py
```diff
@@ -673,13 +673,13 @@ def entanglement_entropy(self, rho):
...
-    def calculate_norm(self, state):
+    def calculate_norm(self, state, order=None):
         state = self.cast(state)
-        return self.np.sqrt(self.np.sum(self.np.abs(state) ** 2))
+        return self.np.linalg.norm(state, ord=order)
 
-    def calculate_norm_density_matrix(self, state):
+    def calculate_norm_density_matrix(self, state, order="nuc"):
         state = self.cast(state)
-        return self.np.trace(state)
```
MatteoRobbiati, line 682: Is one of the order options corresponding to the trace?

src/qibo/backends/tensorflow.py
renatomello: I tried that too and implemented it but then moved back to `numpy` because `tf.norm` not only uses orders with different names, but also doesn't implement all the orders that `numpy` has. So, I kept using the `numpy` norm in order to unify the `order` parameter with all the other backends.

src/qibo/backends/numpy.py
renatomello: Nuclear norm (also known as trace norm)

src/qibo/backends/tensorflow.py
MatteoRobbiati: > I tried that too and implemented it but then moved back to `numpy` because `tf.norm` not only uses orders with different names, but also doesn't implement all the orders that `numpy` has. So, I kept using the `numpy` norm in order to unify the `order` parameter with all the other backends.

Okay, I understand. The only problem I see is that this is not compatible with TensorFlow gradient descent in simulation. I Think it will lose track of the state using an unknown  (by TensorFlow) numpy function.

renatomello: > > I tried that too and implemented it but then moved back to `numpy` because `tf.norm` not only uses orders with different names, but also doesn't implement all the orders that `numpy` has. So, I kept using the `numpy` norm in order to unify the `order` parameter with all the other backends.
> 
> Okay, I understand. The only problem I see is that this is not compatible with TensorFlow gradient descent in simulation. I Think it will lose track of the state using an unknown (by TensorFlow) numpy function.

you mean if `backend.calculate_norm` is used as a callback?

MatteoRobbiati: Yes, in general if you want to calculate gradients of a process which involves non-tensorflow objects, it can lose the track of the flow and return `None`.

renatomello: Do we have any test covering that right now? If so, shouldn't it be broken? If not, we need to have one then.

MatteoRobbiati: I don't think we have tests for every scenario right now (and I don't think it is possible).

My only concern is that if we built the TensorFlow backend to do automatic differentiation, it makes sense to keep the total compatibility with TF, because I suppose if one decide to use `tensorflow` as backend is because some gradient descent has to be computed. 

With this in mind, maybe I would like to customize the method if `tensorflow` is set, in order to remove the `"order"` option and manually compute only one norm if TF is used.

Maybe @stavros11 has clearer picture of this.

renatomello: We could also just use `tf.norm` and let the user know that not all orders in the other backends are compatible with the `tensorflow` backend. It is non-uniform but it works.

MatteoRobbiati: > We could also just use `tf.norm` and let the user know that not all orders in the other backends are compatible with the `tensorflow` backend. It is non-uniform but it works.

Yes, it is also a good solution.

renatomello: > Yes, it is also a good solution.

Done

src/qibo/backends/abstract.py
```diff
@@ -326,13 +326,29 @@ def entanglement_entropy(self, rho):  # pragma: no cover
...
+        """
         raise_error(NotImplementedError)
 
     @abc.abstractmethod
-    def calculate_norm_density_matrix(self, state):  # pragma: no cover
-        """Calculate norm (trace) of a density matrix."""
+    def calculate_norm_density_matrix(self, state, order="nuc"):  # pragma: no cover
+        """Calculate norm (trace) of a density matrix.
+
+        If ``order="nuc"``, it returns the nuclear norm os ``state``,
```
MatteoRobbiati, line 344:
```suggestion
        If ``order="nuc"``, it returns the nuclear norm of ``state``,
```

src/qibo/backends/tensorflow.py
```diff
@@ -306,6 +306,16 @@ def entanglement_entropy(self, rho):
...
         return entropy, spectrum
 
+    def calculate_norm(self, state, order=2):
+        state = self.cast(state)
+        return self.tf.norm(state, ord=order)
+
+    def calculate_norm_density_matrix(self, state, order="nuc"):
+        state = self.cast(state)
+        if order == "nuc":
+            return self.np.trace(state)
```
stavros11, line 316: Regarding the discussion on using tensorflow primitives, indeed, if someone uses gradient descent over a numpy operation it will break the gradient calculation. However, note that in this case `self.np` is not `numpy` but
```py
import tensorflow.experimental.numpy as tnp
```
(see the `__init__`). Back at the time we were writing this backend this was an easy way to port numpy operations to tensorflow, but I am not sure what is the status now. I am also not sure how it will behave in terms of the gradient.

src/qibo/backends/abstract.py
```diff
@@ -326,13 +326,29 @@ def entanglement_entropy(self, rho):  # pragma: no cover
         raise_error(NotImplementedError)
 
     @abc.abstractmethod
-    def calculate_norm(self, state):  # pragma: no cover
-        """Calculate norm of a state vector."""
+    def calculate_norm(self, state, order=2):  # pragma: no cover
+        """Calculate norm of a state vector.
```
MatteoRobbiati, line 330:
```suggestion
        """Calculate norm of a state vector. The 2-norm is computed as default option. 
```

MatteoRobbiati (approved):
Thanks! It looks good to me. Just two little suggestions follow.

src/qibo/backends/tensorflow.py
MatteoRobbiati: Indeed! 
This works well.

PR #1045, opened 2023-10-16
New implementation of `EntanglementEntropy` callback

renatomello:
Checklist:
- [x] Tests are passing.


src/qibo/callbacks.py
```diff
@@ -119,20 +125,23 @@ def nqubits(self, n: int):
             ]
 
     def apply(self, backend, state):
-        rho = backend.partial_trace(state, self.partition, self.nqubits)
-        entropy, spectrum = backend.entanglement_entropy(rho)
+        from qibo.quantum_info.metrics import entanglement_entropy
```
stavros11, line 128: This can also be top level.

tests/test_callbacks.py
```diff
@@ -101,44 +79,11 @@ def test_entropy_switch_partition(backend):
...
-            5e-13,
-            1e-13,
-            1e-12,
-            1e-11,
-            1e-10,
-            1e-9,
-            1e-7,
-            1,
-            4,
-            10,
```
stavros11, line 128: I remember in the past we added this test for weird values because we were getting some failures. Maybe it is still useful to test if the new calculation works with them (if you are not already doing it)?

stavros11 (approved):
Looks good to me, thanks.

src/qibo/callbacks.py
renatomello: at top level it creates a circular import error

tests/test_callbacks.py
renatomello: The `quantum_info.entropy` function already takes care of these cases internally by explicitly setting to 0.0 the values where `log` isn't defined.

MatteoRobbiati (approved):
Thanks.

PR #1046, opened 2023-10-16
Execute multiple circuits in parallel

### LINKED ISSUE #1021: Execution of multiple circuits in parallel
When we implement sequence unrolling in qibolab, we will have the option to deploy multiple circuits in parallel to speed up execution. We will need a way to expose these to users via qibo. The current approach of
```py
result = circuit(nshots=1000)
```
does not work because it is focusing on one circuit at a time.

In contrast
```py
backend.execute_circuit(circuit, nshots=1000)
```
could be extended to `backend.execute_circuits`.

I am not sure if there is a way to extend the first approach too.

stavros11:
Fixes #1021. I implemented a new parallel method for executing multiple circuits and exposed to the backends via `backend.execute_circuits`. I have not benchmarked this, as I am mostly interested in providing the interface for hardware, in order to use sequence unrolling (qiboteam/qibolab#618) with circuits, rather than the simulation method itself. It may be useful to have though.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/parallel.py
+++ b/src/qibo/parallel.py
@@ -154,6 +154,10 @@ def parallel_parametrized_execution(
         parameters (list): list of parameters for the circuit evaluation.
         initial_state (np.array): initial state for the circuit evaluation.
         processes (int): number of processes for parallel evaluation.
+            This corresponds to the number of threads, if a single thread is used
+            for each circuit evaluation. If more threads are used for each circuit
+            evaluation then some tuning may be required to obtain optimal performance.
+            Default is ``None`` which corresponds to a single thread.
 
     Returns:
         Circuit evaluation for input parameters.
--- a/src/qibo/parallel.py
+++ b/src/qibo/parallel.py
@@ -1,6 +1,12 @@
 """
 Resources for parallel circuit evaluation.
 """
+from typing import Iterable
+
+from joblib import Parallel, delayed
+
+from qibo.backends import GlobalBackend
+from qibo.config import raise_error
 
 
 def parallel_execution(circuit, states, processes=None, backend=None):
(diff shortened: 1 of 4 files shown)


src/qibo/parallel.py
```diff
@@ -35,27 +39,84 @@ def parallel_execution(circuit, states, processes=None, backend=None):
...
+            qibo.set_backend('qibojit')
+            from qibo import models, set_threads
+            from qibo.parallel import parallel_execution
+            import numpy as np
+            # create different circuits
+            circuits = [models.QFT(n) for n in range(5, 16)]
+            # set threads to 1 per process (optional, requires tuning)
+            set_threads(1)
+            # execute in parallel
+            results = parallel_circuits_execution(circuit, processes=2)
```
MatteoRobbiati, line 75:
```suggestion
            from qibo.parallel import parallel_circuits_execution
            import numpy as np
            # create different circuits
            circuits = [models.QFT(n) for n in range(5, 16)]
            # set threads to 1 per process (optional, requires tuning)
            set_threads(1)
            # execute in parallel
            results = parallel_circuits_execution(circuits, processes=2)
```

MatteoRobbiati (approved):
Thanks @stavros11! It looks good to me.
Just a comment follows.

src/qibo/parallel.py
```diff
@@ -35,27 +39,84 @@ def parallel_execution(circuit, states, processes=None, backend=None):
...
+            results = parallel_circuits_execution(circuits, processes=2)
+
+    Args:
+        circuits (list): list of circuits to execute.
+        states (optional, list): list of states to use as initial for each circuit.
+            Must have the same length as ``circuits``.
+            If one state is given it is used on all circuits.
+            If not given the default initial state on all circuits.
+        nshots (int): Number of shots when performing measurements, same for all circuits.
+        processes (int): number of processes for parallel evaluation.
```
scarrazza, line 84: Please explain what `processes=None` does and which preferred mode is used threads vs ...

scarrazza (approved):
Thanks @stavros11. My only concern is about documentation explaining how `set_threads` and `processes` work, in particular because the new keyword `processes` sounds misleading, given that we are not using processes by rather deciding how many circuit replicas the user would like to run in parallel, despite the availability of cores for simulation purposes, see my comment.

stavros11:
Thanks @scarrazza. I expanded the docstring a bit explaining this point. I also tried the following in my machine
```py
import qibo
from qibo import models, set_threads
from qibo.parallel import parallel_circuits_execution
import numpy as np

qibo.set_backend("numpy")

circuits = 30 * [models.QFT(20)]

results = parallel_circuits_execution(circuits)
```
and observed what's happening with htop. For the numpy backend, `processes` is equal to the number of threads used and `processes=None` corresponds to single thread. However for qibojit things are a bit more complicated. Even if I set `qibo.set_threads(1)` to force single-thread execution for each circuit, when `processes=1` or `None` I get single thread as expected, however when `processes >= 2` all threads are being used.

I assume the same thing is true for all `parallel_*` methods, as here I just copied the template from the existing ones.

scarrazza:
Ok, what about setting processes=None, does qibojit work as expected when changing the number of threads?

stavros11:
> Ok, what about setting processes=None, does qibojit work as expected when changing the number of threads?

Yes, if `processes` is None then qibojit is using as many threads as specified in `qibo.set_threads`. I guess in that case `parallel_*` is not doing anything and the circuits are executed sequentially.

PR #1055, opened 2023-10-24
Move transpiler from `Qibolab` to `Qibo`

Simone-Bordoni:
Checklist:
- [x] Tests are passing.


Simone-Bordoni:
With the new modifications Sabre router can also accept measurements gates

src/qibo/transpiler/unitary_decompositions.py
```diff
@@ -0,0 +1,254 @@
...
+    lambdas = -np.angle(ud_diag)
+    hx = (lambdas[0] + lambdas[2]) / 2.0
+    hy = (lambdas[1] + lambdas[2]) / 2.0
+    hz = (lambdas[0] + lambdas[1]) / 2.0
+    return hx, hy, hz
+
+
+def cnot_decomposition(q0, q1, hx, hy, hz):
+    """Performs decomposition (6) from arXiv:quant-ph/0307177."""
+    u3 = -1j * (matrices.X + matrices.Z) / np.sqrt(2)
```
renatomello, line 168:
```suggestion
    u3 = -1j * matrices.H
```

tests/test_transpiler_unitary_decompositions.py
```diff
@@ -0,0 +1,206 @@
...
+    shape = 2**nqubits
+    psi = np.random.random(shape) + 1j * np.random.random(shape)
+    return psi / np.sqrt(np.sum(np.abs(psi) ** 2))
+
+
+def random_unitary(nqubits):
+    """Generates a random unitary matrix acting on nqubits."""
+    shape = 2 * (2**nqubits,)
+    m = np.random.random(shape) + 1j * np.random.random(shape)
+    return expm(1j * (m + np.conj(m.T)))
```
renatomello, line 35: `qibo` already has a built-in function for this in `quantum_info.random_ensembles`

tests/test_transpiler_unitary_decompositions.py
```diff
@@ -0,0 +1,206 @@
...
+)
+
+NREPS = 3  # number of repetitions to execute random tests
+ATOL = 1e-12
+
+
+def random_state(nqubits):
+    shape = 2**nqubits
+    psi = np.random.random(shape) + 1j * np.random.random(shape)
+    return psi / np.sqrt(np.sum(np.abs(psi) ** 2))
```
renatomello, line 28: `qibo` already has a built-in function for this in `quantum_info.random_ensembles`.

Moreover, states sampled like this are not uniformly distributed.

tests/test_transpiler_unitary_decompositions.py
```diff
@@ -0,0 +1,206 @@
...
+def bell_unitary(hx, hy, hz):
+    ham = (
+        hx * np.kron(matrices.X, matrices.X)
+        + hy * np.kron(matrices.Y, matrices.Y)
+        + hz * np.kron(matrices.Z, matrices.Z)
+    )
+    return expm(-1j * ham)
+
+
+def purity(state):
```
renatomello, line 47: `purity` is also built-in in `quantum_info.metrics`

src/qibo/transpiler/unitary_decompositions.py
```diff
@@ -0,0 +1,254 @@
...
+    # use corrected version from PRA paper (not arXiv)
+    u2 = -u3 @ expm(-1j * (hx - np.pi / 4) * matrices.X)
+    # add an extra exp(-i pi / 4) global phase to get exact match
+    v2 = expm(-1j * hz * matrices.Z) * np.exp(-1j * np.pi / 4)
+    v3 = expm(1j * hy * matrices.Z)
```
renatomello, line 173: These can be defined, respectively, as $R_{X}(2 h_{x} - \\pi / 2)$, $e^{-i\\pi/4} \\, R_{Z}(2 h_{z})$, and $R_{Z}(-h_{y})$

src/qibo/transpiler/unitary_decompositions.py
```diff
@@ -0,0 +1,254 @@
...
+        gates.Unitary(w, q0),
+        gates.Unitary(np.conj(w).T @ H, q1),
+    ]
+
+
+def cnot_decomposition_light(q0, q1, hx, hy):
+    """Performs decomposition (24) from arXiv:quant-ph/0307177."""
+    w = (matrices.I - 1j * matrices.X) / np.sqrt(2)
+    u2 = expm(-1j * hx * matrices.X)
+    v2 = expm(1j * hy * matrices.Z)
```
renatomello, line 194: Same as the comment above

renatomello, line 194: Same as the comment above

src/qibo/transpiler/unroller.py
```diff
@@ -0,0 +1,460 @@
...
+            decomposition = decomposition(gate)
+        return [
+            g.on_qubits({i: q for i, q in enumerate(gate.qubits)})
+            for g in decomposition
+        ]
+
+
+onequbit_dec = GateDecompositions()
+onequbit_dec.add(gates.H, [gates.U3(0, 7 * np.pi / 2, np.pi, 0)])
+onequbit_dec.add(gates.X, [gates.U3(0, np.pi, 0, np.pi)])
```
renatomello, line 188: The $X$ gate should be done as one pulse like $R_{X}(\\pi)$. By decomposing it as a $U_{3}$, it is going to be done as two pulses like $R_{X}(\\pi/2)$. That introduces a lot of unnecessary noise.

src/qibo/transpiler/unroller.py
```diff
@@ -0,0 +1,460 @@
...
+    gates.FusedGate,
+    lambda gate: [gates.U3(0, *u3_decomposition(gate.matrix(backend)))],
+)
+
+# register the iSWAP decompositions
+iswap_dec = GateDecompositions()
+iswap_dec.add(
+    gates.CNOT,
+    [
+        gates.U3(0, 7 * np.pi / 2, np.pi, 0),
```
renatomello, line 230: Isn't $7\\pi / 2$ outside of the domain of $\\theta \\in [0, 2\\pi]$? In any case it's periodic, then maybe this could be either $3\\pi/2$.

tests/test_transpiler_unitary_decompositions.py
Simone-Bordoni: I have tried using the built in function but in that case i get an error.

src/qibo/transpiler/unroller.py
Simone-Bordoni: I already know this issue but when this unroller was first coded the compiler only accepted U3 and RZ as single qubit native gates. @stavros11 has this issue been solved?

renatomello:
Is the idea of the transpiler to be backend-agnostic? Because several parts of the code end up calling the `NumpyBackend` for one reason or the other.

src/qibo/transpiler/star_connectivity.py
```diff
@@ -0,0 +1,127 @@
...
+                    new.add(gates.SWAP(middle_qubit, new_middle))
+                # update gate qubits according to the new swap
+                qubits = tuple(hardware_qubits.index(q) for q in gate.qubits)
+
+            # add gate to the hardware circuit
+            if isinstance(gate, gates.Unitary):
+                # gates.Unitary requires matrix as first argument
+                backend = NumpyBackend()
+                matrix = gate.matrix(backend)
+                new.add(gate.__class__(matrix, *qubits, **gate.init_kwargs))
```
renatomello, line 121: you can get the underlying matrix from `gate.init_args` and then there's no need to rely on the `NumpyBackend` call

src/qibo/transpiler/star_connectivity.py
```diff
@@ -0,0 +1,127 @@
...
+                if add_swap:
+                    new.add(gates.SWAP(middle_qubit, new_middle))
+                # update gate qubits according to the new swap
+                qubits = tuple(hardware_qubits.index(q) for q in gate.qubits)
+
+            # add gate to the hardware circuit
+            if isinstance(gate, gates.Unitary):
+                # gates.Unitary requires matrix as first argument
+                backend = NumpyBackend()
+                matrix = gate.matrix(backend)
```
renatomello, line 120: This makes this part of the code backend-independent and also fix the tests for GPU backends

```suggestion
                matrix = gate.init_args[0]
```

stavros11:
> Is the idea of the transpiler to be backend-agnostic? Because several parts of the code end up calling the `NumpyBackend` for one reason or the other.

From a hardware point of view, using the `NumpyBackend` or even just numpy should be sufficient because the transpiler is only dealing with gates (2x2, 4x4 matrices) and numpy is the most efficient for small array sizes. 

If there is a simulation application that would benefit from involving backends in the transpiler we could consider it, but if there is no strong motivation I would avoid it to keep the code simpler.

renatomello:
> > Is the idea of the transpiler to be backend-agnostic? Because several parts of the code end up calling the `NumpyBackend` for one reason or the other.
> 
> From a hardware point of view, using the `NumpyBackend` or even just numpy should be sufficient because the transpiler is only dealing with gates (2x2, 4x4 matrices) and numpy is the most efficient for small array sizes.
> 
> If there is a simulation application that would benefit from involving backends in the transpiler we could consider it, but if there is no strong motivation I would avoid it to keep the code simpler.

That generates one issue though, which is a few tests fail on the GPU backends

tests/test_transpiler_unitary_decompositions.py
renatomello: Resolved

renatomello:
PR #1074 is needed to fix tests on GPU backends

Simone-Bordoni:
In the latest commit I have moved the transpiler tutorial from qibolab to qibo. Tell me if the location of the document is correct, in the next days I plan to make a more complete tutorial.
moreover the tutorial was both for transpiler and compiler but i think we should keep the compiler tutorial on qibolab.

Simone-Bordoni:
@renatomello with the latest commit many tests are not passing anymore. Are you taking care of making them pass or should i do it?

stavros11:
> That generates one issue though, which is a few tests fail on the GPU backends, e.g. if one creates circuits to be transpiled that involve `gates.Unitary` with matrices in one of the other backends

I would say the easiest solution to this would be to not allow creating gates with matrices that are not numpy. Cupy will cast all matrices on GPU during execution anyway and numpy should be faster for most gate operations (unless your unitary is acting on many qubits).

The only serious issue I see with this constraint is that it will break backpropagation for the Tensorflow backend. However we could relax the constraint and also support `tf.Tensor`s to keep this feature. I believe these play well with numpy (unlike cupy).

src/qibo/transpiler/unroller.py
stavros11: Indeed, right now the [compiler only has rules](https://github.com/qiboteam/qibolab/blob/main/src/qibolab/compilers/default.py) for `Z`, `RZ`, `GPI2` and `U3` so it will fail if we don't transpile the `X`. 
This is easy to fix, however I would prefer to have a more concrete solution, preferably in a different PR, where we decide the set of single-qubit native gates and we sync both the transpiler and compiler to respect that.

Simone-Bordoni:
> Thanks @Simone-Bordoni for transfering the transpiler and @renatomello, @BrunoLiegiBastonLiegi for your comments. I think it would be better to merge this as it is and open issues for all the points you don't like, because it is hard to review new changes pushed to this PR as all files appear as new. It will be easier if we open other PRs for new fixes.

I agree that it would be better to merge this as a first version and then fix all the issues separately in future PR

stavros11:
@renatomello @BrunoLiegiBastonLiegi let me know if you agree with merging this. It is blocking the corresponding PR in qibolab and a few other issues there that need to be moved here.

renatomello:
> @renatomello @BrunoLiegiBastonLiegi let me know if you agree with merging this. It is blocking the corresponding PR in qibolab and a few other issues there that need to be moved here.

I improved the docstring and other parts of the code. But there are still a few major things that need to be resolved. Mainly, the transpiler needs to be included in the API reference and also the handling of RNG seeds (at least in the router module). If the merger of this PR is urgent, I could approve it and merge it, and leave those things to be fixed in separate PRs.

src/qibo/transpiler/pipeline.py
```diff
@@ -0,0 +1,221 @@
...
+            raise_error(
+                TranspilerPipelineError,
+                "Define the hardware chip connectivity to use default transpiler",
+            )
+        default_passes = []
+        # preprocessing
+        default_passes.append(Preprocessing(connectivity=connectivity))
+        # default placer pass
+        default_passes.append(Trivial(connectivity=connectivity))
+        # default_passes.append(ReverseTraversal(connectivity=connectivity, routing_algorithm=ShortestPaths(connectivity), depth=5))
```
BrunoLiegiBastonLiegi, line 163:
```suggestion
```
remove comment

src/qibo/transpiler/pipeline.py
```diff
@@ -0,0 +1,221 @@
...
+        # default router pass
+        default_passes.append(StarConnectivity())
```
BrunoLiegiBastonLiegi, line 165: why is `StarConnectivity` used by default regardless of the `connectivity` passed?

src/qibo/transpiler/pipeline.py
```diff
@@ -0,0 +1,221 @@
...
+        # default_passes.append(ShortestPaths(connectivity=connectivity))
```
BrunoLiegiBastonLiegi, line 166:
```suggestion
```
remove comment

src/qibo/transpiler/pipeline.py
```diff
@@ -0,0 +1,221 @@
...
+        native_gates: NativeType = NativeType.CZ,
+    ):
+        self.native_gates = native_gates
+        if passes is None:
+            self.passes = self.default(connectivity)
+        else:
+            self.passes = passes
+        self.connectivity = connectivity
+
+    def default(self, connectivity: nx.Graph):
```
BrunoLiegiBastonLiegi, line 151: I would probably use as default here the best pipeline we have. Therefore, `ReverseTraversal` + `Sabre` I believe.

src/qibo/transpiler/placer.py
```diff
@@ -0,0 +1,341 @@
...
+            if gate not in graph.edges():
+                return len(gates_qubits_pairs) - allowed - 1
+        return 0
+
+
+class ReverseTraversal(Placer):
+    """
+    Places qubits based on the algorithm proposed in Reference [1].
+
+    Works with ``ShortestPaths`` routing.
```
BrunoLiegiBastonLiegi, line 251: It works with any `Router`, right? 
```suggestion
    Compatible with all the available ``Router``s.
```

src/qibo/transpiler/router.py
```diff
@@ -0,0 +1,642 @@
...
+        return tot_distance
+
+    def _swap_candidates(self):
+        """Return a list of possible candidate SWAPs (to be applied on logical qubits directly).
+        The possible candidates are the ones sharing at least one qubit with a block in the front layer.
+        """
+        candidates = []
+        for block in self._front_layer:
+            qubits = self.circuit.get_physical_qubits(block)
+            for qubit in qubits:
```
BrunoLiegiBastonLiegi, line 560:
```suggestion
            for qubit in self.circuit.get_physical_qubits(block)
```

src/qibo/transpiler/router.py
```diff
@@ -0,0 +1,642 @@
...
+        """Check if some gatesblocks in the front layer can be executed in the current configuration.
+
+        Returns:
+            (list): executable blocks if there are, ``None`` otherwise.
+        """
+        executable_blocks = []
+        for block in self._front_layer:
+            qubits = self.circuit.get_physical_qubits(block)
+            if (
+                qubits in self.connectivity.edges
```
BrunoLiegiBastonLiegi, line 584:
```suggestion
            if (
                self.circuit.get_physical_qubits(block) in self.connectivity.edges
```

stavros11:
> I improved the docstring and other parts of the code. But there are still a few major things that need to be resolved. Mainly, the transpiler needs to be included in the API reference and also the handling of RNG seeds (at least in the router module). If the merger of this PR is urgent, I could approve it and merge it, and leave those things to be fixed in separate PRs.

It is not urgent as in we will use it immediately (the Sabre part doesn't work anyway) but there are several issues in qibolab (qiboteam/qibolab#606, qiboteam/qibolab#643), including by external people, that I would like to move here and start addressing. Also, updates pushed directly to this PR are practically not reviewed because in the file change everything appears as new. Usually it is better to separate moving (copy-pasting) big chunks of code from new developments for this reason.

Regarding the above points, including in the API reference sounds easy to do. For the RNG seeds if it is complicated (as in changing many lines of code) I would rather do in a separate PR or just open an issue and assign to someone. Keep in mind that there are other issues related to this code (some in qiboteam/qibolab#606) that may be more urgent, from a usability point of view.

src/qibo/transpiler/pipeline.py
Simone-Bordoni: Because now sabre is not able to handle measurements and star connectivity is stable. when we will fix the issue with measurements we will mode to sabre as the default.

src/qibo/transpiler/pipeline.py
Simone-Bordoni: Same as before, sabre needs the measurement fix.

renatomello (approved):

PR #1066, opened 2023-10-30
Add probabilities to MeasurementOutcomes

BrunoLiegiBastonLiegi:
Tiny patch to #1039 that implements the calculation of the probabilities starting from the frequencies in `MeasurementOutcomes` objects.

Checklist:
- [ ] Tests are passing.


scarrazza:
@BrunoLiegiBastonLiegi did you test this on hardware?

BrunoLiegiBastonLiegi:
Not yet, I am going to test it asap.

tests/test_result.py
```diff
@@ -5,6 +5,18 @@
...
 
+def test_measurementoutcomes_probabilties(backend):
+    c = models.Circuit(3)
+    c.add(gates.H(0))
+    c.add(gates.M(0, 2))
+    probabilities = c().probabilities(qubits=[0, 2])
+    c.has_collapse = True
+    repeated_probabilities = c().probabilities()
+    result = MeasurementOutcomes(c.measurements, backend, probabilities=probabilities)
+    backend.assert_allclose(probabilities, repeated_probabilities, atol=1e-1)
```
stavros11, line 16: In order to increase the tolerance here, you can either use more shots so that the sampled probabilities converge more to the exact ones, or use a circuit where the outcomes are fixed (non-probabilistic), for example using X gates instead of H.

stavros11 (approved):
Thanks @BrunoLiegiBastonLiegi, looks good to me. The comments are not really blocking, although I would give a try to lower the tolerance in the test.

I also have not tested on hardware (will test the qibolab PR soon), but I don't the current PR is breaking anything since you are just calculating the probabilities from frequencies, and it shouldn't make a difference if the frequencies are simulated or obtained from hardware.

BrunoLiegiBastonLiegi:
Ok, I made a quick test on the hardware and the calculation of the probabilities from the frequencies works. 
I'll do another one after I added the `qubits` parameter to probabilities here as well.

BrunoLiegiBastonLiegi:
Ok I have added the qubits parameter and tested on hardware. It seems to work fine up to the calibration of the qubits.

tests/test_result.py
```diff
@@ -5,6 +5,31 @@
 from qibo.result import CircuitResult, MeasurementOutcomes, load_result
 
 
+@pytest.mark.parametrize("qubits", [[0, 2], [0], [1, 2]])
+def test_measurementoutcomes_probabilties(backend, qubits):
+    c = models.Circuit(3)
+    c.add(gates.X(0))
+    c.add(gates.M(0, 2))
+    global_probs = c().probabilities(qubits=[0, 2])
+    probabilities = c().probabilities(qubits=qubits)
```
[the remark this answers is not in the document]
renatomello: By running circuits in the tests without specifying the backend, the circuit execution uses the `GlobalBackend`, which will lead to using a different backend than the one it's supposed to use. That still leads to tests passing on `numpy`, `tensorflow`, and `numba`, but it breaks tests for the GPU backends

src/qibo/result.py
```diff
@@ -237,6 +246,37 @@ def frequencies(self, binary=True, registers=False):
...
+        """
+        nqubits = len(self.measurement_gate.qubits)
+        if qubits is None:
+            qubits = range(nqubits)
+        else:
+            if not set(qubits).issubset(self.measurement_gate.qubits):
+                raise RuntimeError(
+                    "Asking probabilities for qubits that were not measured."
+                )
+            else:
```
renatomello, line 263:
```suggestion
                raise_error(RuntimeError, f"Probabilities were asked for qubits {qubits}, but they were not measured.")
```

tests/test_result.py
```diff
@@ -5,6 +5,31 @@
...
+    c.has_collapse = True
+    if 1 in qubits:
+        with pytest.raises(Exception) as exc_info:
+            repeated_probabilities = c().probabilities(qubits=qubits)
+            assert (
+                str(exc_info.value)
+                == "Asking probabilities for qubits that were not measured."
+            )
```
renatomello, line 22:
```suggestion
        with pytest.raises(RuntimeError):
            repeated_probabilities = backend.execute_circuit(c).probabilities(qubits=qubits)
```

src/qibo/result.py
```diff
@@ -237,6 +246,37 @@ def frequencies(self, binary=True, registers=False):
...
+
+        if self._probs is not None:
+            return self.backend.calculate_probabilities(
+                self._probs, qubits, len(self.measurement_gate.qubits)
+            )
+
+        frequencies = self.frequencies(binary=False)
+        probs = self.backend.np.zeros(2 ** self._samples.shape[1])
+        for state, freq in frequencies.items():
+            probs[state] = freq / self.nshots
```
renatomello, line 274: This breaks in `tensorflow` because `tensorflow` does not allow item assignment on tensors

src/qibo/result.py
renatomello: It's best to delegating the raising of errors to the `qibo`'s `raise_error` function in order to keep the code unified throughout. Then, this should be changed everywhere else.

tests/test_result.py
renatomello: Like my comment about `raise_error`, this should be unified too.

renatomello (dismissed):

tests/test_result.py
renatomello: Another point about standardization, everywhere else `self` is used but here `cls` is being used.

tests/test_result.py
```diff
@@ -5,6 +5,31 @@
 from qibo.result import CircuitResult, MeasurementOutcomes, load_result
 
 
+@pytest.mark.parametrize("qubits", [[0, 2], [0], [1, 2]])
```
renatomello, line 8:
```suggestion
@pytest.mark.parametrize("qubits", [None, [0, 2], [0], [1, 2]])
```

renatomello: This fixes the coverage issues

BrunoLiegiBastonLiegi:
Ok I added a `calculate_probabilities_from_frequencies` method to the numpy and tensorflow backends to solve the problem of tensor assignement in tensorflow.
However three tests fail, the ones where you call `probabilities` with `qubits=None`. The reason is that when you calculate the probabilities from the state you have `qubits` equal to all the qubits of the circuit, measured and not measured, whereas when you do it from the frequencies only the measured qubits are availble. I could fix the tests by 'hand' but maybe there is a solution to solve this asymmetry.

stavros11 (commented):
> Ok I added a `calculate_probabilities_from_frequencies` method to the numpy and tensorflow backends to solve the problem of tensor assignement in tensorflow.

That's generally fine, however in principle we could also do this particular calculation in numpy only. The main motivation in using tensorflow is for gradient calculation using backpropagation but it is unlikely that this will go through probabilities calculated from frequencies. Maybe you did this for some tests to pass, but we could consider simplifying,  unless I am missing something.

> However three tests were failing, the ones where you call `probabilities` with `qubits=None`.  The reason is that when you calculate the probabilities from the state you have `qubits` equal to all the qubits of the circuit, measured and not measured, whereas when you do it from the frequencies only the measured qubits are availble. I fixed the tests by 'hand' but maybe there is a solution to solve this asymmetry.

Would
```py
if qubits is None:
    qubits = self.measurement_gate.qubits
```
solve this issue? I understand that probabilities from frequencies only exists for `MeasurementOutcomes` where you always have some measured qubits. This set of measured qubits is the maximum you can calculate probabilities for so I think this choice makes sense.

renatomello:
> Ok I added a `calculate_probabilities_from_frequencies` method to the numpy and tensorflow backends to solve the problem of tensor assignement in tensorflow. However three tests were failing, the ones where you call `probabilities` with `qubits=None`. The reason is that when you calculate the probabilities from the state you have `qubits` equal to all the qubits of the circuit, measured and not measured, whereas when you do it from the frequencies only the measured qubits are availble. I fixed the tests by 'hand' but maybe there is a solution to solve this asymmetry.

I think an easier solution to the item assignment issue would be to create `probs` as a `list`, use `probs.append`, and after the loop use `probs = self.backend.cast(probs)`. That is by default backend-agnostic.

src/qibo/result.py
```diff
@@ -237,6 +247,36 @@ def frequencies(self, binary=True, registers=False):
...
+                )
+
+        if self._probs is not None:
+            return self.backend.calculate_probabilities(
+                self._probs, qubits, len(self.measurement_gate.qubits)
+            )
+
+        frequencies = self.frequencies(binary=False)
+        probs = self.backend.calculate_probabilities_from_frequencies(
+            frequencies, self._samples.shape[1], self.nshots
```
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: > Would
if qubits is None:
    qubits = self.measurement_gate.qubits
solve this issue? I understand that probabilities from frequencies only exists for MeasurementOutcomes where you always have some measured qubits. This set of measured qubits is the maximum you can calculate probabilities for so I think this choice makes sense.

The problem is that you don't have access to the measurements, and thus to which qubits are measured, in `QuantumState` though. Therefore, if you don't specify which qubits to trace out to the `QuantumState.probabilities()` you will get the probabilities over all the qubits, whereas for `MeasurementOutcomes.probabilities()` you get the probabilities over the measured qubits only. I am not sure this is a problem, but it is an asymmetry.

> I have not tried this, but I suspect this will fail in the case you only have frequencies and not samples. For example if the user asks for frequencies and then for probabilities, without asking for samples in between or before.

If I am not mistaken, the calculation of the frequencies always involves the generation of the samples first, unless you have the probabilities available because you constructed the `MeasurementOutcomes` object starting from those. At that point, though, you will not call the `backend.calculate_probabilities_from_frequencies`.

BrunoLiegiBastonLiegi:
> I think an easier solution to the item assignment issue would be
> 
> ```python
> probs = [freq / self.nshots  for _, freq in frequencies.items()]
> probs = self.backend.cast(probs)
> ```
> 
> Maybe you'd only need an extra `dtype=float` inside the cast, but I think this should work for all backends

Ah yeah thanks, that's much easier indeed.

renatomello:
> > I think an easier solution to the item assignment issue would be
> > ```python
> > probs = [freq / self.nshots  for _, freq in frequencies.items()]
> > probs = self.backend.cast(probs)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Maybe you'd only need an extra `dtype=float` inside the cast, but I think this should work for all backends
> 
> Mmmh, but I think this fails if you have zero frequency for one state. I believe that frequencies right now provide only the frequency for the nonzero states. I mean, that you don't have {'state': 0}.

The point is about that anything that you do, you can do with a list and then cast to the right backend instead of starting from an array of that backend.

BrunoLiegiBastonLiegi:
> The point is about that anything that you do, you can do with a list and then cast to the right backend instead of starting from an array of that backend.

ah yeah, you're right that's far easier, you do the assignment with the list and then cast.

BrunoLiegiBastonLiegi:
All the requested changes should have been addressed now, thanks @stavros11 and @renatomello.

stavros11 (commented):
Thanks for the updates @BrunoLiegiBastonLiegi. I tried the following script with qibolab (you can run it locally in your computer, just need to install qibolab):
```py
from qibolab.backends import QibolabBackend
from qibolab.transpilers import NativeGates, Passes

c = Circuit(3)
c.add(gates.X(0))
c.add(gates.M(0, 1, 2))

backend = QibolabBackend("dummy")

result = backend.execute_circuit(c, nshots=100)

probs = result.probabilities()
freqs = result.frequencies(binary=False)

print(sum(probs))
print(sum(freqs.values()))
```
and the sum of frequencies is 100 as expected, however the sum of probabilities is not 1. Shouldn't the probabilities add to 1?

BrunoLiegiBastonLiegi:
>Shouldn't the probabilities add to 1?

Yep sorry, thanks for spotting this, there was a `sqrt()` missing. Unfortunately in the tests I used a deterministic state (`|100>`) in order to avoid using a large number of shots and small tolerance, but that masked the bug as `sqrt(1) = 1`. Now, i am testing on `|+00>` and using `nshots=10000` and `atol=1e-2`.

src/qibo/result.py
```diff
@@ -1,49 +1,53 @@
...
     return globals()[payload.pop("dtype")].from_dict(payload)
 
 
 class QuantumState:
-    """Data structure to represent the final state after circuit execution."""
+    """Data structure to represent the final state after circuit execution.
+
+    Args:
+
+    """
```
renatomello, line 31: This docstring is missing the ``args``

src/qibo/result.py
```diff
@@ -55,31 +59,33 @@ def symbolic(self, decimals=5, cutoff=1e-10, max_terms=20):
...
+                Defaults to ``False``.
 
         Returns:
             The state in the computational basis.
         """
         if numpy:
-            return np.asarray(self._state)
-        else:
-            return self._state
+            return np.array(self._state)
```
renatomello, line 74: I haven't tested it, but this looks like it will fail if the original backend is one of the GPU backends.

src/qibo/result.py
```diff
@@ -55,31 +59,33 @@ def symbolic(self, decimals=5, cutoff=1e-10, max_terms=20):
...
         When noisy model is applied to a circuit and `circuit.density_matrix=False`,
         this method returns the average probability resulting from
         repeated execution. This probability distribution approximates the
         exact probability distribution obtained when `circuit.density_matrix=True`.
 
         Args:
-            qubits (list, set): Set of qubits that are measured.
+            qubits (list or set, optional): Set of qubits that are measured.
+                If ``None``, ``qubits`` equates the total number of qubits.
+                Defauts to ``None``.
```
renatomello, line 88: Docstring missing ``Returns``

src/qibo/result.py
```diff
@@ -115,47 +119,47 @@ def dump(self, filename):
...
         payload = np.load(filename, allow_pickle=True).item()
         return cls.from_dict(payload)
 
 
 class MeasurementOutcomes:
-    """Object to store the outcomes of measurements after circuit execution."""
+    """Object to store the outcomes of measurements after circuit execution.
+
+    Args:
+    """
```
renatomello, line 153: Docstring missing ``Args``

src/qibo/result.py
```diff
@@ -237,30 +247,55 @@ def frequencies(self, binary=True, registers=False):
 
         return self._frequencies
 
+    def probabilities(self, qubits: Optional[Union[list, set]] = None):
+        """Calculate the probabilities as frequencies / nshots
+
+        Returns:
+            The array containing the probabilities of the measured qubits.
```
renatomello, line 254:
```suggestion
            (ndarray): The array containing the probabilities of the measured qubits.
```

src/qibo/result.py
```diff
@@ -237,30 +247,55 @@ def frequencies(self, binary=True, registers=False):
...
+            )
+
+        probs = [0 for _ in range(2**nqubits)]
+        for state, freq in self.frequencies(binary=False).items():
+            probs[state] = freq / self.nshots
+        probs = self.backend.cast(probs)
+        self._probs = probs
+        return self.backend.calculate_probabilities(np.sqrt(probs), qubits, nqubits)
+
     def has_samples(self):
```
renatomello, line 278: missing docstring

src/qibo/result.py
```diff
@@ -330,7 +365,7 @@ def measurement_gate(self):
 
         return self._measurement_gate
 
-    def apply_bitflips(self, p0, p1=None):
+    def apply_bitflips(self, p0: float, p1: Optional[float] = None):
```
renatomello, line 368: Missing docstring

src/qibo/result.py
```diff
@@ -330,18 +365,18 @@ def measurement_gate(self):
...
         return apply_bitflips(self, p0, p1)
 
     def expectation_from_samples(self, observable):
         """Computes the real expectation value of a diagonal observable from frequencies.
 
         Args:
             observable (Hamiltonian/SymbolicHamiltonian): diagonal observable in the
-            computational basis.
+                computational basis.
```
renatomello, line 376: typing here is missing ``:class:`` references

src/qibo/result.py
```diff
@@ -437,14 +475,14 @@ def to_dict(self):
         return args
 
     @classmethod
-    def from_dict(cls, payload):
+    def from_dict(cls, payload: dict):
         """Builds a ``CircuitResult`` object starting from a dictionary.
```
renatomello, line 479: Missing ``:class:`` reference here and other places. Same thing for some mentions of ``MeasurementOutcomes`` class.

tests/test_measurements_collapse.py
```diff
@@ -207,24 +205,24 @@ def test_measurement_result_parameters_multiple_qubits(backend):
     backend.assert_allclose(final_state, target_state)
 
 
-@pytest.mark.skip(reason="no way of currently testing this")
+@pytest.mark.skip(reason="this has to be updated for density matrices")
```
renatomello, line 208: So this is not testing all scenarios?

src/qibo/result.py
BrunoLiegiBastonLiegi: The tests with gpus were passing already I believe, but if we can get rid of an `if/else` fine by me.

tests/test_measurements_collapse.py
BrunoLiegiBastonLiegi: This is an old test that was requiring a configuration that we do not support anymore, it might be kept by updating it to make use of the density matrix, but that's not straightforward to me

src/qibo/result.py
```diff
@@ -26,10 +26,15 @@ class QuantumState:
...
     Args:
-
+        state (np.ndarray): Input quantum state as np.ndarray.
+        backend (qibo.backends.AbstractBackend): Backend used for the calculations. If not provided the :class:`qibo.backends.GlobalBackend` is going to be used.
     """
 
-    def __init__(self, state, backend):
+    def __init__(self, state, backend=None):
+        if backend is None:
+            from qibo.backends import GlobalBackend
```
renatomello, line 35: This could go to the top if it doesn't create a circular import error

src/qibo/result.py
```diff
@@ -150,12 +157,17 @@ class MeasurementOutcomes:
     """Object to store the outcomes of measurements after circuit execution.
 
     Args:
+        measurements (qibo.gates.M): The measurement gates containing the measurements.
```
renatomello, line 160:
```suggestion
        measurements (:class:`qibo.gates.M`): Measurement gates.
```

src/qibo/result.py
```diff
@@ -150,12 +157,17 @@ class MeasurementOutcomes:
...
+        backend (qibo.backends.AbstractBackend): Backend used for the calculations. If not provided the :class:`qibo.backends.GlobalBackend` is going to be used.
```
renatomello, line 161:
```suggestion
        backend (:class:`qibo.backends.AbstractBackend`): Backend used for the calculations. 
            If ``None``, then :class:`qibo.backends.GlobalBackend` is used. Defaults to ``None``.
```

src/qibo/result.py
```diff
@@ -150,12 +157,17 @@ class MeasurementOutcomes:
...
+        probabilities (np.ndarray): Use these probabilities to generate samples and frequencies.
+        samples (np.darray): Use these samples to generate probabilities and frequencies.
+        nshots (int): Number of shots used for samples, probabilities and frequencies generation.
     """
 
     def __init__(
         self,
         measurements,
-        backend,
+        backend=None,
```
renatomello, line 170: As best practice in the other modules, the arg ``backend`` is usually the last one.

src/qibo/result.py
```diff
@@ -366,6 +383,7 @@ def measurement_gate(self):
         return self._measurement_gate
 
     def apply_bitflips(self, p0: float, p1: Optional[float] = None):
+        """Apply bitflips to the measurements with probabilities `p0` and `p1`"""
```
renatomello, line 386: Docstring missing ``args``.

src/qibo/result.py
```diff
@@ -383,7 +401,7 @@ def expectation_from_samples(self, observable):
         return observable.expectation_from_samples(freq, qubit_map)
 
     def to_dict(self):
-        """Returns a dictonary containinig all the information needed to rebuild the ``MeasurementOutcomes``."""
+        """Returns a dictonary containinig all the information needed to rebuild the :class:``qibo.result.MeasurementOutcomes``."""
```
renatomello, line 404:
```suggestion
        """Returns a dictonary containinig all the information needed to rebuild the :class:`qibo.result.MeasurementOutcomes`."""
```

src/qibo/result.py
```diff
@@ -395,7 +413,7 @@ def to_dict(self):
         return args
 
     def dump(self, filename: str):
-        """Writes to file the ``MeasurementOutcomes`` for future reloading.
+        """Writes to file the :class:``qibo.result.MeasurementOutcomes`` for future reloading.
```
renatomello, line 416:
```suggestion
        """Writes to file the :class:`qibo.result.MeasurementOutcomes` for future reloading.
```

src/qibo/result.py
```diff
@@ -405,10 +423,10 @@ def dump(self, filename: str):
 
     @classmethod
     def from_dict(cls, payload: dict):
-        """Builds a ``MeasurementOutcomes`` object starting from a dictionary.
+        """Builds a :class:``qibo.result.MeasurementOutcomes`` object starting from a dictionary.
```
renatomello, line 426:
```suggestion
        """Builds a :class:`qibo.result.MeasurementOutcomes` object starting from a dictionary.
```

src/qibo/result.py
```diff
@@ -405,10 +423,10 @@ def dump(self, filename: str):
...
 
         Args:
-            payload (dict): Dictionary containing all the information to load the ``MeasurementOutcomes`` object.
+            payload (dict): Dictionary containing all the information to load the :class:``qibo.result.MeasurementOutcomes`` object.
```
renatomello, line 429:
```suggestion
            payload (dict): Dictionary containing all the information to load the :class:`qibo.result.MeasurementOutcomes` object.
```

src/qibo/result.py
```diff
@@ -432,10 +450,10 @@ def from_dict(cls, payload: dict):
 
     @classmethod
     def load(cls, filename: str):
-        """Builds the ``MeasurementOutcomes`` object stored in a file.
+        """Builds the :class:``qibo.result.MeasurementOutcomes`` object stored in a file.
```
renatomello, line 453:
```suggestion
        """Builds the :class:`qibo.result.MeasurementOutcomes` object stored in a file.
```

src/qibo/result.py
```diff
@@ -432,10 +450,10 @@ def from_dict(cls, payload: dict):
...
 
         Args:
-            filename (str): Path to the file containing the ``MeasurementOutcomes``.
+            filename (str): Path to the file containing the :class:``qibo.result.MeasurementOutcomes``.
```
renatomello, line 456:
```suggestion
            filename (str): Path to the file containing the :class:`qibo.result.MeasurementOutcomes`.
```

src/qibo/result.py
BrunoLiegiBastonLiegi: it does

src/qibo/result.py
BrunoLiegiBastonLiegi: I see, but then we will probably need to open two small PR in qibojit and qibolab to update them accordingly. The reason it was not defaulting to None from the beginning is that these object are not supposed to be exposed to the user. Usually, their construction is taken care of by the backend. I don't know, it looks a little bit too much effort just to change the order of a parameter that will probably never be seen by the user anyway.

renatomello (approved):

PR #1067, opened 2023-10-30
Nix

AleCandido:
Checklist:
- [ ] Tests are passing.


scarrazza (approved):

scarrazza:
@AleCandido why nix in the lock file? Could you please fix the conflict, it seems that I cannot simply do `poetry update`?

AleCandido:
Nix has its own lock file, `flake.lock`, and it's completely decoupled. The conflict comes from this branch being based on an old version of `master`, and the lock file having been modified in #809 (I added IPython and pdb++ as development tools).

I reran `poetry lock`, and everything should be fixed and consistent

scarrazza:
Thanks, we can merge as soon as tests pass.

scarrazza:
Windows seems not happy again.

AleCandido:
The problem doesn't seem to be Windows, but rather a conflict with #1039.

It actually defined `M.load` in a separate way from any other gate, and now it is conflicting with the general `Gate.load`, and failing in `test_result.py`.

scarrazza:
@BrunoLiegiBastonLiegi could you please have a look and chat with @AleCandido

PR #1070, opened 2023-11-02
Circuit Execution Output bugfix for the qibojit and qibolab backends

BrunoLiegiBastonLiegi:
Minor modifications on top of #1039 needed for the qibojit and qibolab backends.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/backends/numpy.py
+++ b/src/qibo/backends/numpy.py
@@ -526,7 +526,7 @@ class NumpyBackend(Backend):
 
         if circuit.density_matrix:  # this implies also it has_collapse
             assert circuit.has_collapse
-            final_state = self.np.mean(self.to_numpy(final_states), 0)
+            final_state = np.mean(self.to_numpy(final_states), 0)
             if circuit.measurements:
                 qubits = [q for m in circuit.measurements for q in m.target_qubits]
                 final_result = CircuitResult(
--- a/src/qibo/backends/numpy.py
+++ b/src/qibo/backends/numpy.py
@@ -521,6 +521,8 @@ class NumpyBackend(Backend):
                 results.append(sample)
                 if not circuit.density_matrix:
                     samples.append("".join([str(s) for s in sample]))
+                for gate in circuit.measurements:
+                    gate.result.reset()
 
         if circuit.density_matrix:  # this implies also it has_collapse
             assert circuit.has_collapse
--- a/src/qibo/backends/numpy.py
+++ b/src/qibo/backends/numpy.py
@@ -524,7 +524,7 @@ class NumpyBackend(Backend):
 
         if circuit.density_matrix:  # this implies also it has_collapse
             assert circuit.has_collapse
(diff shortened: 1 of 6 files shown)


src/qibo/backends/numpy.py
```diff
@@ -524,7 +524,7 @@ def execute_circuit_repeated(self, circuit, nshots, initial_state=None):
 
         if circuit.density_matrix:  # this implies also it has_collapse
             assert circuit.has_collapse
-            final_state = np.asarray(final_states).mean(0)
+            final_state = self.np.mean(self.to_numpy(final_states), 0)
```
stavros11, line 527: Since you are already casting to numpy, I guess you could do the `mean` in numpy
```suggestion
            final_state = np.mean(self.to_numpy(final_states), 0)
```

src/qibo/result.py
```diff
@@ -147,15 +147,6 @@ class MeasurementOutcomes:
...
         self, measurements, backend, probabilities=None, samples=None, nshots=1000
     ):
-        if probabilities is None and samples is None:
-            raise RuntimeError(
-                "You have to provide either the `probabilities` or the `samples` to build a `MeasurementOutcomes` object."
-            )
-        if probabilities is not None and samples is not None:
-            raise RuntimeError(
-                "Both the `probabilities` and the `samples` were provided to build the `MeasurementOutcomes` object. Don't know which one to use."
-            )
```
stavros11, line 157: In principle this error you could keep, as we don't have a case where we pass both `probabilities` and `samples` (right?). But it is fine to remove as well.

tests/test_callbacks.py
```diff
@@ -94,6 +94,7 @@ def test_entropy_in_circuit(backend, density_matrix, base):
 
     target = [0, 0, np.log(2)] / np.log(base)
     values = [backend.to_numpy(x) for x in entropy]
+    print(target, values)
```
stavros11, line 97:
```suggestion
```

src/qibo/result.py
BrunoLiegiBastonLiegi: The only case would be the user manually creating the object passing both. It's not really supposed to by called by the user though.

stavros11:
@renatomello can you please give another try to run the GPU tests using this branch and qiboteam/qibojit#153. I believe tests are now passing for me.

tests/test_result.py
```diff
@@ -76,6 +56,7 @@ def test_circuitresult_dump_load(backend, agnostic_load):
     # from samples and recover the same exact frequencies
     c.has_collapse = True
     result = backend.execute_circuit(c)
+    print(result.has_samples())
```
stavros11, line 59:
```suggestion
```

renatomello (approved):
Tests are passing locally for GPUs after one minor change

stavros11 (approved):
Thanks @renatomello, then I believe we can merge this.

PR #1071, opened 2023-11-02
Unary encoder for random Gaussian states

renatomello:
Checklist:
- [x] Tests are passing.


src/qibo/models/encodings.py
```diff
@@ -73,6 +74,86 @@ def unary_encoder(data):
...
+
+    sampler = _probability_distribution_gaussian_loader(
+        a=0, b=2 * math.pi, seed=local_state
+    )
+
+    circuit, pairs_rbs = _generate_rbs_pairs(nqubits)
+
+    phases = []
+    for depth, row in enumerate(pairs_rbs, 1):
+        phases += list(sampler.rvs(depth=depth, size=len(row)))
```
stavros11, line 150: Usually
```suggestion
        phases.extend(sampler.rvs(depth=depth, size=len(row)))
```
is better because it extends the existing list (while `+` recreates a new list every time).

src/qibo/models/encodings.py
```diff
@@ -92,3 +173,15 @@ def _generate_rbs_pairs(nqubits):
             circuit.add(gates.RBS(*pair, 0.0, trainable=True))
 
     return circuit, pairs_rbs
+
+
+class _probability_distribution_gaussian_loader(rv_continuous):
```
stavros11, line 178: Normally I would complain because you are not following the typical class naming rules (which would be `ProbabilityDistributionGaussianLoader`), but I am not sure if it is relevant in this case because it is a very specific scipy usage and `rv_continuous` does not follow it either.

Another possibility to make it private is to move this definition inside the `unary_encoder_random_gaussian` method, but if you are planning to use it in other methods in the future it is better to leave outside.

stavros11 (approved):

src/qibo/models/encodings.py
renatomello: I prefer to leave it outside the function because we will probably define other probability distributions later and it would be not ideal to define all of them inside the function. However, the name style can be changed, true.

AlejandroSopena (approved):

PR #1078, opened 2023-11-07
N-qubit Classifiers

Jacfomg:
Just in case someone is interested I updated the classifiers to support n-qubits, entanglent and n-dimensional data as in the README paper. The single qubit classifier could be redudant now, but I feel like as an example is easier to follow and should remain there.

Checklist:
- [ ] Tests are passing.


Jacfomg:
I don't think there is much use of having this as a example. I would keep the code simple as it was before.

PR #1081, opened 2023-11-07
DBF step optimization

MatteoRobbiati:
A first possible optimization of the step is implemented. 
We use `hyperopt` to optimize the step considering a single flow execution. 

An example of usage:

```python
import numpy as np

from qibo.models.double_bracket import DoubleBracketFlow, FlowGeneratorType
from qibo import set_backend
from qibo.hamiltonians.hamiltonians import Hamiltonian
from qibo.quantum_info import random_hermitian

set_backend("numpy")

NSTEPS=20
nqubits = 4
h0 = random_hermitian(2**nqubits)
DELTA=0.005

np.random.seed(42)

norms = []
# No hyperopt vs hyperopt
methods = [None, "opt"]
init_step = 0.01

for i, meth in enumerate(methods):
    print(f"Testing opt: {meth}")
    norm_history = []

    dbf = DoubleBracketFlow(Hamiltonian(nqubits=nqubits, matrix=h0))
    norm_history.append(dbf.off_diagonal_norm)
    if meth is not None:
        print("Big optimization for setting the initial step:")
        step = dbf.hyperopt_step(step_min=0.005, step_max=0.02, max_evals=1000, verbose=True)

    for i in range(NSTEPS):
        # evolution under canonical commutator
        if meth is not None:
            # here a hyper-parameter space and an hyperopt algorithm can be passed
            # default are hp.uniform and tpe
            step = dbf.hyperopt_step(step_min=step-DELTA, step_max=step+DELTA, max_evals=100)
            print(f"New flow duration s={step}")
        dbf(step=init_step, mode=FlowGeneratorType.canonical)
        norm_history.append(dbf.off_diagonal_norm)
    norms.append(norm_history)


plt.figure(figsize=(5, 5*6/8))
plt.plot(norms[0], label="Fixed step", color="royalblue", alpha=0.7, lw=1.5)
plt.plot(norms[1], label="Optimized", color="red", alpha=0.7, lw=1.5)
plt.legend()
plt.xlabel("Steps")
plt.ylabel("Cost")
plt.savefig("test_fig.png")
```

And an example of benchmark between fixed step size (blue), and optimized step size:

test_fig

Checklist:
- [ ] Tests are passing.


pyproject.toml
```diff
@@ -29,6 +29,7 @@ joblib = "^1.2.0"
 matplotlib = "^3.7.0"
 psutil = "^5.9.4"
 tabulate = "^0.9.0"
+hyperopt = "^0.2.6"
```
andrea-pasquale, line 32: I don't know if it is worth to keep `hyperopt` here. Perhaps it is better to have it as extra dep

src/qibo/models/double_bracket.py
```diff
@@ -95,3 +97,65 @@ def off_diagonal_norm(self):
...
+            look_ahead: number of flow steps to compute the loss function;
+        """
+        # copy initial hamiltonian
+        h_copy = self.h
+
+        # off_diagonal_norm's value before the steps
+        old_loss = self.off_diagonal_norm
+
+        for _ in range(look_ahead):
+            self.__call__(mode=FlowGeneratorType.canonical, step=step)
```
andrea-pasquale, line 153: Here you are assuming that the `FlowGeneratorType` is `canonical`.
In theory it could be any of the types present in the `Enum`.
Perhaps the easiest way would be to pass `FlowGeneratorType` in the constructor.

src/qibo/models/double_bracket.py
```diff
@@ -95,3 +97,65 @@ def off_diagonal_norm(self):
...
+    def loss(self, step: float, look_ahead: int = 1):
+        """
+        Compute loss function distance between `look_ahead` steps.
+
+        Args:
+            step: flow step.
+            look_ahead: number of flow steps to compute the loss function;
+        """
+        # copy initial hamiltonian
+        h_copy = self.h
```
andrea-pasquale, line 147: Here we were discussing with @Edoardo-Pedicillo that maybe we should add a `deepcopy`.
Did you check if the `self.h` remains unchanged?

andrea-pasquale: Also if we expect to run this at the beginning we could also consider having `h0` here.

tests/test_models_dbf.py
```diff
@@ -51,3 +52,34 @@ def test_double_bracket_flow_single_commutator(backend, nqubits):
...
+    delta = 0.02
+
+    step = dbf.hyperopt_step(
+        step_min=initial_step - delta, step_max=initial_step + delta, max_evals=100
+    )
+
+    assert step != initial_step
+
+    # evolve following the optimized first step
+    dbf(mode=FlowGeneratorType.canonical, step=step)
```
andrea-pasquale, line 74: Also here it should be better to test all `FlowGeneratorType`.

andrea-pasquale (changes requested):
Thanks @MatteoRobbiati for implementing this.
I have a few suggestions down below.

src/qibo/models/double_bracket.py
```diff
@@ -95,3 +97,65 @@ def off_diagonal_norm(self):
...
+        for _ in range(look_ahead):
+            self.__call__(mode=FlowGeneratorType.canonical, step=step)
+
+        # off_diagonal_norm's value after the steps
+        new_loss = self.off_diagonal_norm
+
+        # set back the initial configuration
+        self.h = h_copy
+
+        return new_loss - old_loss
```
Edoardo-Pedicillo, line 161: I suggest using directly `off_diagonal_norm` as loss. The result should be the same, since `old_loss` is a constant.

src/qibo/models/double_bracket.py
MatteoRobbiati: I agree, I think it is wise to set it while initializing the model.

Edoardo-Pedicillo (approved):

src/qibo/models/double_bracket.py
```diff
@@ -1,5 +1,8 @@
+from copy import deepcopy
 from enum import Enum, auto
+from functools import partial
 
+import hyperopt
```
andrea-pasquale, line 5: Now that we removed the `hyperopt` dep from the main deps, when you try to import this module it will raise a `ModuleNotFoundError`.
I think that you should move this import directly inside the `hyperopt_step` possibly in a try-except block where you can tell the user to install hyperopt if they want to perform the optimiziation.
Perhaps at this point it is easier just to keep it as a dependency.
What do you recommend @scarrazza? Shall we add `hyperopt` to the main deps or we keep it as an optional dep?

andrea-pasquale (commented):
Thanks for all the updates @MatteoRobbiati.
Just a small comment regarding how to deal with `hyperopt` again.

src/qibo/models/double_bracket.py
Edoardo-Pedicillo: > I think that you should move this import directly inside the hyperopt_step possibly in a try-except block where you can tell the user to install hyperopt if they want to perform the optimiziation.

I agree with this option

PR #1083, opened 2023-11-09
Qcnn upgrade

jykhoo1987:
Checklist:
- [x] Tests are passing.

3 'upgrades' implemented in qcnn class:
1) arbitrary 2-qubit circuit can be used as input during initialization which will be used as the base circuit when building the convolution layer
2) minimization is now implemented via qibo.optimizers.optimize, whereas it was only scipy.minimize previously
3) introduced a flag to make copy of initial state during initialization. By default, it will check backend during class initialization and be turned on if backend is qibojit to get around the in-place update issue. 

Credits to @rahula06 for implementing most of these changes.

We will work on the updated documentation and make a separate pull request for that.

@ihpcdingwj @jf-kong @rahula06


src/qibo/models/qcnn.py
```diff
@@ -1,6 +1,8 @@
 import numpy as np
 
-from qibo import Circuit, gates
+import qibo
+from qibo import gates, set_backend
```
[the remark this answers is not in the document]
MatteoRobbiati: I would remove this, it is not used.

src/qibo/models/qcnn.py
```diff
@@ -42,12 +45,32 @@ class QuantumCNN:
...
 
-    def __init__(self, nqubits, nlayers, nclasses=2, params=None):
+    def __init__(
+        self,
+        nqubits,
+        nlayers,
+        nclasses=2,
+        params=None,
+        twoqubitansatz=None,
+        Dividend=2,
```
MatteoRobbiati, line 55:
```suggestion
```

MatteoRobbiati: I think this line is not used.

src/qibo/models/qcnn.py
```diff
@@ -42,12 +45,32 @@ class QuantumCNN:
...
+        copy_init_state=None,
+    ):
         self.nclasses = nclasses
         self.nqubits = nqubits
         self.nlayers = nlayers
+        self.twoqubitansatz = twoqubitansatz
+
+        if copy_init_state is None:
+            if qibo.get_backend()[:7] == "qibojit":
```
MatteoRobbiati, line 64:
```suggestion
            if "qibojit" in qibo.get_backend():
```

tests/test_models_qcnn.py
```diff
@@ -3,7 +3,9 @@
 import numpy as np
 import pytest
```
MatteoRobbiati, line 4:
```suggestion
```

tests/test_models_qcnn.py
```diff
@@ -303,3 +305,47 @@ def test_qcnn_training():
     predictions.append(1)
     labels = np.array([[1], [-1], [1]])
     test_qcnn.Accuracy(labels, predictions)
+
+
+def test_two_qubit_ansatz():
```
MatteoRobbiati, line 310: I would pass `backend` as argument of the tests here. With this you will execute the tests in all the backends. Now that the  `qibojit` in-place update is fixed in the main of the `qcnn` model, all the tests should pass.

tests/test_models_qcnn.py
```diff
@@ -303,3 +305,47 @@ def test_qcnn_training():
...
+    c = Circuit(2)
+    c.add(gates.H(0))
+    c.add(gates.RX(0, 0))
+    c.add(gates.CNOT(1, 0))
+    test_qcnn = QuantumCNN(4, 2, 2, twoqubitansatz=c)
+
+
+def test_two_qubit_ansatz_training():
```
MatteoRobbiati, line 318: Same comment of before (in general, consider this comment valid for all the tests here).

src/qibo/models/qcnn.py
```diff
@@ -42,12 +45,32 @@ class QuantumCNN:
...
     """
 
-    def __init__(self, nqubits, nlayers, nclasses=2, params=None):
+    def __init__(
+        self,
+        nqubits,
+        nlayers,
+        nclasses=2,
+        params=None,
+        twoqubitansatz=None,
```
MatteoRobbiati, line 54: Can you tell me a bit more about this `twoqubitansatz` attribute?

MatteoRobbiati (changes requested):
Many thanks @jykhoo1987 for implementing this. 
Some comments follow.

src/qibo/models/qcnn.py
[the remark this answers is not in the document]
rahula06: Okay sure. If we don't import all of qibo, we need to import get_backend for later, I'll add that in.

scarrazza:
@jykhoo1987 I will wait for your change until later today, if you do not manage we postpone this feature for the next release.

jykhoo1987:
Apologies for the delay. @rahula06 will look into the rest of this PR later this week when he is back from leave.

tests/test_models_qcnn.py
rahula06: I think it suffices if we test the backends for the training, so we can keep it to the tests that involve training maybe

src/qibo/models/qcnn.py
rahula06: The qcnn has a two qubit convolutional layer which currently has a fixed architecture. The twoqubitansatz allows the user to change this convolutional layer into any two qubit circuit of their choice.

rahula06:
Hi everyone, sorry for the delay. I was on leave for the last couple of weeks. The new commits should address the comments that you have raised so far. In particular, the tests now check training on qibojit and numpy backends, and small changes to the style have been made as per suggestion.

Thanks!

tests/test_models_qcnn.py
MatteoRobbiati: It always better to test the code for all the available backends, in order not to have surprises if an external user decides to use this model with, e.g. `tensorflow` and training it with some automatic differentiation method. 

You only need to add `backend` as an argument of your tests, e.g.: 

```python
def test_two_qubit_ansatz(backend):
     # same code here
```

src/qibo/models/qcnn.py
```diff
@@ -16,14 +17,16 @@ class QuantumCNN:
         nclasses (int): number of classes to be classified. Default setting of 2 (phases).
         params: initial list of variational parameters. If not provided, all parameters
             will be initialized to zero.
+        twoqubitansatz (Qibo.models.Circuit object): A two qubit ansatz that can be input by the user to form the two qubit ansatz used in the convolutional circuit.
```
MatteoRobbiati, line 20:
```suggestion
        twoqubitansatz (:class:`qibo.models.Circuit`): a two qubit ansatz that can be input by the user to form the two qubit ansatz used in the convolutional circuit.
```

MatteoRobbiati (commented):
Many thanks for your modifications @jykhoo1987, @rahula06.
As mentioned before, I would suggest to pass `backend` to the tests, I think it is important in order to cover any possible scenario.

Once that is done, I think this will be ready to be merged.

A small suggestion follows.

rahula06:
> Many thanks for your modifications @jykhoo1987, @rahula06. As mentioned before, I would suggest to pass `backend` to the tests, I think it is important in order to cover any possible scenario.
> 
> Once that is done, I think this will be ready to be merged.
> 
> A small suggestion follows.

Hi Matteo, `backend' (chosen between qibojit and numpy) is currently passed to the tests that involve training. We could pass these to the other tests as well, but I think those are independent of the backend --- only the training was showing issues before. If you think it is necessary, we could do this anyway.

Many thanks for your comments!

MatteoRobbiati:
> If you think it is necessary, we could do this anyway.

Thanks for your answer!

Usually, in all the tests we write in Qibo, we include `backend` as an argument for all the test functions. It's not necessary for you to specify the backends to be tested, because just by adding the argument, all the backend tests are automatically triggered. While it may seem excessive in many situations, I believe it's always better to do so :)
Moreover, it can be even more general to test your training procedures using all the available backends and not only `numpy` and `qibojit`. Again, you can just pass `backend` to the function and remove the `@pytest.mark.parametrize` decorator.

rahula06:
> Usually, in all the tests we write in Qibo, we include `backend` as an argument for all the test functions. It's not necessary for you to specify the backends to be tested, because just by adding the argument, all the backend tests are automatically triggered. While it may seem excessive in many situations, I believe it's always better to do so :) Moreover, it can be even more general to test your training procedures using all the available backends and not only `numpy` and `qibojit`. Again, you can just pass `backend` to the function and remove the `@pytest.mark.parametrize` decorator.

Ah okay, thanks for the answer --- didn't know that simply adding backend to the tests would make it test all the backends automatically. This will make the tests fail locally, since it will say there is no 'backend' fixture, but I'm guessing that Qibo has   some global fixture defined? Am curious to understand how this works.

I will update this ASAP, thanks!

MatteoRobbiati (commented):
Thanks a lot for the modifications! 
Still, I think something is going wrong since we have a couple of lines which are not covered by tests right now (L64 and L283).

Do you think it is possible to reduce the tests duration? In my case is taking ~8' locally.

Regarding the `backend` keyword in the tests, I think it is something defined in the [`conftest.py`](https://github.com/qiboteam/qibo/blob/master/tests/conftest.py) file :)

rahula06:
> Thanks a lot for the modifications! Still, I think something is going wrong since we have a couple of lines which are not covered by tests right now (L64 and L283).

Hmm, strange. These are exactly the lines which should be checked if the qibojit backend is working. 

> Do you think it is possible to reduce the tests duration? In my case is taking ~8' locally.

I think the vast majority of this is in the training. We could reduce the number of shots in the tests, that might do it.

> Regarding the `backend` keyword in the tests, I think it is something defined in the [`conftest.py`](https://github.com/qiboteam/qibo/blob/master/tests/conftest.py) file :)

Thanks, will have a look :)

rahula06:
It seems that going back to the old test works, so that the qibojit backend test is indeed working. My best guess is that the codecov tool might not be able to evaluate these pytest fixtures very well. If you are okay, we could just keep the tests with the explicit parametrization to pass the checks. If anyone has better knowledge of how codecov works or could suggest an alternative, that would be great too :)

MatteoRobbiati (changes requested):
Thanks, a few comments follow. 
Then I think we are ready :)

rahula06:
> Thanks, a few comments follow. Then I think we are ready :)

fixed, and still seems to pass all the tests!

MatteoRobbiati (approved):
Thanks.

renatomello (approved):

PR #1076, opened 2023-11-06
Clifford Simulator

BrunoLiegiBastonLiegi:
Implementation of a `clifford` backend that allows to simulate clifford circuits following the tableau formalism introduced by [Aaronson and Gottesman](https://arxiv.org/abs/quant-ph/0406196#) .

Checklist:
- [x] Reviewers confirm new code works as expected.
- [x] Tests are passing.
- [x] Coverage does not decrease.
- [x] Documentation is updated.

PR #1093, opened 2023-11-15
Add decomposition for ``gates.FSWAP``

renatomello:
Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/gates/gates.py
+++ b/src/qibo/gates/gates.py
@@ -1488,6 +1488,11 @@ class FSWAP(Gate):
     def qasm_label(self):
         return "fswap"
 
+    def decompose(self, *free, use_toffolis: bool = True) -> List[Gate]:
+        """"""
+        q0, q1 = self.target_qubits
+        return [X(q1)] + GIVENS(q0, q1, np.pi / 2).decompose() + [X(q0)]
+
 
 class fSim(ParametrizedGate):
     """The fSim gate defined in `arXiv:2001.08343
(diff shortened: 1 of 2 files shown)


BrunoLiegiBastonLiegi (approved):
Thanks @renatomello, looks good.

stavros11 (approved):
Thanks @renatomello. This is fine with me, just wanted to mention that now that we have the transpiler in qibo, we should consider moving the `Gate.decompose` method to the `Unroller` interface. The main idea behind this was that we can have different `Unroller`s for different applications. For example the `FSWAP` may be decomposed to a set of `GPI2` and `CZ` for the hardware, but a different decomposition for some other application (if that makes sense and is useful).

That is of course if you and @BrunoLiegiBastonLiegi agree, but it is quite clear that the `decompose` does the same thing with the `Unroller` and I would rather focus on improving one approach than trying to maintain two competing ones. If you have suggestions to improve the transpiler these are also welcome.

renatomello:
> Thanks @renatomello. This is fine with me, just wanted to mention that now that we have the transpiler in qibo, we should consider moving the `Gate.decompose` method to the `Unroller` interface. The main idea behind this was that we can have different `Unroller`s for different applications. For example the `FSWAP` may be decomposed to a set of `GPI2` and `CZ` for the hardware, but a different decomposition for some other application (if that makes sense and is useful).
> 
> That is of course if you and @BrunoLiegiBastonLiegi agree, but it is quite clear that the `decompose` does the same thing with the `Unroller` and I would rather focus on improving one approach than trying to maintain two competing ones. If you have suggestions to improve the transpiler these are also welcome.

Yeah I agree we should concentrate decompositions on the transpiler. This one is being done so @BrunoLiegiBastonLiegi can move forward with the Clifford simulator in #1076 and also it can serve as a benchmark for any unroller.

PR #1097, opened 2023-11-16
Probabilistic Error Cancellation

mho291:
Added PEC to error_mitigation.py

Checklist:
- [ ] Tests are passing.


BrunoLiegiBastonLiegi:
@mho291 can you change the name of the PR to something a little bit more evocative, like for instance 'Probabilistic Error Cancellation'?

BrunoLiegiBastonLiegi:
Thanks @mho291 for implementing this. The code looks a bit too much as a notebook example. To incorporate it into `qibo` it should be a bit more general. The other mitigation methods we have right now are implemented as functions that take as inputs a circuit and an observable and return the mitigated expected value of the observable. I would implement this following a similar paradigm. Can you please transform this to a function of the kind:
```
def PEC(circuit: qibo.models.Circuit, additional_arguments_needed):
    # do something with the circuit and 
    # elaborate some output
    ...
    return output
``` 

such that a user can use it by doing:
```
# define the circuit
my_circuit = Circuit(3)
# add the gates
my_circuit.add(gates.H(0))
...
# use the mitigation
output = PEC(my_circuit, additional_arguments_needed)
```

Furthermore, by quickly looking into the code I noticed that you are defining by hand the gates. You don't need to do this, qibo is providing the API for simulating the execution of a quantum circuit already. As a rule of thumb, you should try to stick to the functions provided by qibo as much as you can.

mho291:
Hi @BrunoLiegiBastonLiegi, thank you for providing useful comments.  The code I wrote for PEC works as what you've described. One can call the PEC function directly, for example:

```
# Create noise model
lam = 0.1
depol = NoiseModel()
depol.add(DepolarizingError(lam))

# Create circuit
qcqc = Circuit(2)
qcqc.add(gates.H(0))
qcqc.add(gates.CNOT(0,1))
qcqc.add(gates.M(0))
qcqc.add(gates.M(1))

# Run PEC
nshots_GST = 10000 # number of shots for gate set tomography
nshots_MC = 10000 # number of Monte Carlo samples
err_mitigated_distribution, variance = PEC(qcqc, nshots_GST, nshots_MC, depol) # I've added variance into my local version. It's just Csample / sqrt(nshots_MC)

```

About the hand-defining the gates, it's actually part of the gate set tomography for probabilistic error cancellation. Here, a single qubit circuit is initialized in the { |0>, |1>, |+>, and |y+> } states and measured in the Pauli bases {I, X, Y, Z}. For two qubits, we just take tensor products, i.e. { |00>, |01>, ..., |y+>|y+> } for the states and {II, IX, IY, ..., ZZ} for the measurement bases. 

The PEC code contains these parts: (1) Gate set tomography, (2) Computing quasi probabilities, (3) Monte Carlo sampling, and (4) Post-processing of Monte Carlo samples. I have incorporated post-processing into the Monte Carlo sampling block of code but will work on separating the parts of PEC into functions that are usable for general use.

BrunoLiegiBastonLiegi:
>About the hand-defining the gates, it's actually part of the gate set tomography for probabilistic error cancellation. Here, a single qubit circuit is initialized in the { |0>, |1>, |+>, and |y+> } states and measured in the Pauli bases {I, X, Y, Z}. For two qubits, we just take tensor products, i.e. { |00>, |01>, ..., |y+>|y+> } for the states and {II, IX, IY, ..., ZZ} for the measurement bases.

You don't need to define them by hand. Each gate in qibo has a `matrix()` method that returns its matrix representation. The initial states can be constructed by applying the sequence of gates you need to the `|00...000>` state and collecting the final state after execution, if it's something that you have to do for initialization alone (otherwise it is going to be costly). If i remember correctly, measurement gates allow for the definition of the basis in which to measure with `gates.M(0,1,2, basis=['X', 'Y', 'Z'])`. 


> The PEC code contains these parts: (1) Gate set tomography, (2) Computing quasi probabilities, (3) Monte Carlo sampling, and (4) Post-processing of Monte Carlo samples. I have incorporated post-processing into the Monte Carlo sampling block of code but will work on separating the parts of PEC into functions that are usable for general use.

Yes, then it's better to separate the different pieces and then call them inside PEC,  but they probably should not be inside of `error_mitigation.py`. 

Moreover, I saw that you are creating directories inside PEC. What do you need them for?

mho291:
I have split up the functions of PEC in the update.

Gate set tomography has been added to src/qibo/validation. The functions for calculating the inverse noise, obtain quasi probabilities, and Monte Carlo have been split. I didn't know where to add them so I've placed them into error_mitigation.py for the time being. The final PEC function contains these individual functions and outputs the error-mitigated distribution.

As for the creation of directories, it's now part of the argument "save_data". If save_data is not None, then we'll create directories to store all the gate set tomography data and Monte Carlo samples.

scarrazza:
@mho291 could you please open another PR with just the tomography? As discussed during the last meeting, this simplifies review and tests.

mho291:
> @mho291 could you please open another PR with just the tomography? As discussed during the last meeting, this simplifies review and tests.

Will do!

PR #1103, opened 2023-11-22
Create `gates.CY`

### LINKED ISSUE #1099: Add CY gate
In qibo the `CY` gate is currently missing, as the clifford simulator #1076 will implement this, it could be useful to have it also for standard state simulation.

renatomello:
This closes #1099 and needs https://github.com/qiboteam/qibojit/pull/158 for `numba` tests to pass

Checklist:
- [x] Tests are passing.


stavros11 (commented):
Thanks @renatomello. Looks good to me, there is only an issue with tests. Using this and the corresponding branch for qibojit, the following test fails for me
```
tests/test_gates_density_matrix.py::test_controlled_by_one_qubit_gates[qibojit-numba-Y]
```
I have not checked what is wrong though. This is on CPU, I did not run on GPU yet.

Simone-Bordoni (commented):
Also for me it is ok, just fix the issue with the tests

renatomello:
@stavros11 @Simone-Bordoni I pushed the fix to https://github.com/qiboteam/qibojit/pull/158 and one can test locally that all tests are passing

stavros11 (approved):
Thanks @renatomello, it works for me now. For the CI, you probably need to run `poetry lock` because you changed the pyproject, however I don't think this is needed as we will revert back to qibojit-main before merging.

I don't remember what we usually do in this situation, but most likely we need to release qibo 0.2.3, bump the dependency on qibojit's pyproject and run `poetry lock` there.

renatomello:
> Thanks @renatomello, it works for me now. For the CI, you probably need to run `poetry lock` because you changed the pyproject, however I don't think this is needed as we will revert back to qibojit-main before merging.
> 
> I don't remember what we usually do in this situation, but most likely we need to release qibo 0.2.3, bump the dependency on qibojit's pyproject and run `poetry lock` there.

I reverted back the change

tests/test_gates_abstract.py
```diff
@@ -228,10 +228,13 @@ def test_one_qubit_rotations_controlled_by(gatename, params):
     assert gate.parameters == params
 
 
-def test_cnot_and_cz_init():
+def test_cnot_and_cy_and_cz_init():
     gate = gates.CNOT(0, 1)
     assert gate.target_qubits == (1,)
     assert gate.control_qubits == (0,)
+    gate = gates.CZ(4, 7)
```
BrunoLiegiBastonLiegi, line 235:
```suggestion
    gate = gates.CY(4, 7)
```

is this supposed to be a CY?

BrunoLiegiBastonLiegi (approved):

scarrazza:
@renatomello does the windows failure is expected (due to the missing qibojit release)?

renatomello:
> @renatomello does the windows failure is expected (due to the missing qibojit release)?

Yes

PR #1113, opened 2023-11-30
print backend

Edoardo-Pedicillo:
Print backend
Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/models/dbi/double_bracket.py
+++ b/src/qibo/models/dbi/double_bracket.py
@@ -116,6 +116,9 @@ class DoubleBracketIteration:
         """Get Hamiltonian's backend."""
         return self.h0.backend
 
+    def print_backend(self):
+        return self.backend
+
     def hyperopt_step(
         self,
         step_min: float = 1e-5,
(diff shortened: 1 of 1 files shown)


src/qibo/models/dbi/double_bracket.py
```diff
@@ -116,6 +116,9 @@ def backend(self):
         """Get Hamiltonian's backend."""
         return self.h0.backend
 
+    def print_backend(self):
```
MatteoRobbiati, line 119: Please add some docstrings.

src/qibo/models/dbi/double_bracket.py
```diff
@@ -116,6 +116,9 @@ def backend(self):
...
+        return self.backend
+
```
MatteoRobbiati, line 121:
```suggestion
        """Print backend."""
        return self.backend
```

MatteoRobbiati (changes requested):
Thanks, a suggestion follows.

MatteoRobbiati (approved):
This is fine, thanks!

PR #1119, opened 2023-12-04
Replace psutil with multiprocessing

AleCandido:
`psutil` is only used to get the CPU count

- https://github.com/search?q=repo%3Aqiboteam%2Fqibo+%2F%5Cbpsutil%2F&type=code
- it is counting the logical units, which is the default option
    https://github.com/qiboteam/qibo/blob/1dd73f7a637defd01a42023651f8499652115a7a/src/qibo/backends/tensorflow.py#L200
- the same could be achieved with `multiprocessing.cpu_count()` (which is standard library) - it would have not been possible for `logical = False`, but it is not what it's being used

#814

```py
In [8]: psutil.cpu_count(logical=False)
Out[8]: 64

In [9]: psutil.cpu_count(logical=True) == multiprocessing.cpu_count()
Out[9]: True
```

Checklist:
- [ ] Tests are passing.


scarrazza:
Could you please double check this works in the cluster when selecting a specific number of cpus?

AleCandido:
> Could you please double-check if this works in the cluster when selecting a specific number of cpus?

For sure. Is there a specific test that you want me to run?

In any case, the atomic change has been checked (as you can see from the IPython snippet above), and I've done it in _maryah_.
If that works, I don't imagine any way of failing during an actual run.

stavros11:
> Psutils was introduced when we realized that the cpu count/affinity was not correct using other strategies when submitting jobs via slurm. Therefore, the best test is to send jobs by specifying the number of cpu cores (less and more than the default) and check if the output is correct.

What is the correct way of limiting the number of cores in slurm in order to test this? I have tried both `srun` and `sbatch` with different combinations of `--ntasks`, `--cpus-per-task`, `--cores-per-socket` but I always get the total number of threads (128) with both psutil and multiprocessing.

scarrazza:
@stavros11, indeed these options are correct but I think the only way to set the proper cpu affinity is [this](https://github.com/qiboteam/qibojit/blob/c621acf3c8b8ffaaa67f76114b5ab9a086aea6d8/src/qibojit/backends/cpu.py#L70).

src/qibo/backends/tensorflow.py
```diff
@@ -195,9 +196,7 @@ def __init__(self):
         elif cpu_devices:
             self.device = cpu_devices[0].name
 
-        import psutil
-
-        self.nthreads = psutil.cpu_count(logical=True)
+        self.nthreads = multiprocessing.cpu_count()
```
scarrazza, line 199: Instead of guessing the number of available threads we might use `tf.config.get_inter_op_parallelism_threads` or `tf.config.get_intra_op_parallelism_threads` (please check which one is for single operator multi-threading) directly?

scarrazza:
Indeed, the cluster uses affinity, therefore you should try: `len(psutil.Process().cpu_affinity())`

src/qibo/backends/tensorflow.py
AleCandido:
```py
In [2]: tf.config.get_inter_op_parallelism_threads
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 tf.config.get_inter_op_parallelism_threads

AttributeError: module 'tensorflow._api.v2.config' has no attribute 'get_inter_op_parallelism_threads'

In [3]: tf.config.get_intra_op_parallelism_threads
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[3], line 1
----> 1 tf.config.get_intra_op_parallelism_threads

AttributeError: module 'tensorflow._api.v2.config' has no attribute 'get_intra_op_parallelism_threads'
```

scarrazza: https://www.tensorflow.org/api_docs/python/tf/config/threading/get_inter_op_parallelism_threads

AleCandido:
> Indeed, the cluster uses affinity, therefore you should try: `len(psutil.Process().cpu_affinity())`

```py
In [6]: import psutil

In [7]: len(psutil.Process().cpu_affinity())
Out[7]: 2
```

This works, but it's not what we are currently using. Here I wanted to clean some unneeded dependency, not to introduce a new feature.

In case, I would open a dedicated issue.

src/qibo/backends/tensorflow.py
AleCandido: I believe they are now scoped in `tf.config.threading`:
https://www.tensorflow.org/api_docs/python/tf/config/threading/get_intra_op_parallelism_threads

AleCandido:
```py
In [10]:  tf.config.threading.get_inter_op_parallelism_threads()
Out[10]: 0

In [11]:  tf.config.threading.get_intra_op_parallelism_threads()
Out[11]: 0
```

And we do not set by default they are `0`:
> *A value of 0 means the system picks an appropriate number.*

scarrazza:
Probably you have spotted a bug, in the sense that:
- the affinity is the only proper way to set the total number of available threads for OMP or thread pool library
- the tensorflow backend is the only backend which is not using affinity and instead randomly guessing the total number of threads instead of getting it from `tf.config.threading` directly.

src/qibo/backends/tensorflow.py
scarrazza: Exactly, so ideally you could check for 0, if True assume tf is using all threads given by psutil affinity.

AleCandido: And what should I do if not zero?

1. set to zero myself
2. ignore it (I will assume the user chose it manually)
3. raise an error
4. check if it's the same of the affinity (and then do something)
5. set to the affinity

My favorite option is 2., since it conflicts the least with user choices (definitely a case in which less is more).
But in that case I should do nothing in general (not even check it if it's zero, since I will ignore anyhow the outcome)

scarrazza: I think option 2 is reasonable, moreover `qibo.set_threads` consistently does not set threads when using the tf backend, but please double check if this value is not really needed anywhere.

AleCandido: Thanks to @stavros11 I found it this interesting piece of Qibo:

https://github.com/qiboteam/qibo/blob/c1227d81b09a5e99f8be9e70a03cb7bcb41713b9/src/qibo/backends/tensorflow.py#L207-L214

So, I believe that it `qibo.set_threads` is not consistently doing its job. Especially for the `TensorflowBackend`. But I would not try to fix it in this PR.

scarrazza (approved):
LGTM, thanks.

stavros11 (approved):

PR #1120, opened 2023-12-04
Drop matplotlib dependency

AleCandido:
`matplotlib` seems unused in `qibo`, and only used in the examples

- src: https://github.com/search?q=repo%3Aqiboteam%2Fqibo+%2F%5Cbmatplotlib%2F+path%3Asrc%2F&type=code
- examples: https://github.com/search?q=repo%3Aqiboteam%2Fqibo+%2F%5Cbmatplotlib%2F+path%3Aexamples%2F&type=code

#814

Checklist:
- [ ] Tests are passing.


scarrazza:
Dropping is fine but I agree with @stavros that we should add it as optional.

scarrazza (approved):
Fine by me in the tests group. We could also add a sentence in the examples/README.md that some example require `matplotlib`.

PR #1125, opened 2023-12-06
[WIP] refactoring `optimizers` structure

MatteoRobbiati:
Checklist:
- [ ] Tests are passing.


doc/source/code-examples/advancedexamples.rst
```diff
@@ -621,7 +623,10 @@ Here is a simple example using a custom loss function:
     data = np.random.normal(0, 1, size=2**nqubits)
 
     # perform optimization
-    best, params, extra = optimize(myloss, x0, args=(c, data), method='BFGS')
+    optimizer = ScipyMinimizer(options={'method'='BFGS'})
```
MatteoRobbiati, line 626: I would write here a more complete example in which we use both the `options` and the `fit_options` dictionary. I will take care of this.

examples/bell-variational/main.py
```diff
@@ -18,9 +18,12 @@ def main(nshots, backend):
     set_backend(backend)
     initial_parameters = np.random.uniform(0, 2 * np.pi, 2)
     circuits = set_parametrized_circuits()
-    best, params, _ = optimize(
-        cost_function, initial_parameters, args=(circuits, nshots)
+
+    opt = ScipyMinimizer(
```
MatteoRobbiati, line 22: This is still the old interface: we should move the loss function to `opt.fit(...)`.

src/qibo/models/evolution.py
```diff
@@ -263,7 +263,7 @@ def _loss(params, adiabatic_evolution, h1, opt_messages, opt_history):
             log.info(f"Params: {params}  -  <H1> = {loss}")
         return loss
 
-    def minimize(self, initial_parameters, method="BFGS", options=None, messages=False):
+    def minimize(self, opt, initial_parameters, fit_options=None, messages=False):
```
MatteoRobbiati, line 266:
```suggestion
    def minimize(self, opt, initial_parameters, fit_options={}, messages=False):
```

src/qibo/models/variational.py
```diff
@@ -27,51 +30,25 @@ class VQE:
             vqe = models.VQE(circuit, hamiltonian)
             # optimize using random initial variational parameters
             initial_parameters = np.random.uniform(0, 2, 1)
-            vqe.minimize(initial_parameters)
+            opt = CMAES()
+            best, params, _ = vqe.minimize(opt, initial_parameters, fit_options=options)
```
MatteoRobbiati, line 34: I think `options` is not defined and still passed to the `fit_options` argument.

src/qibo/models/variational.py
```diff
@@ -27,51 +30,25 @@ class VQE:
...
-        hess=None,
-        hessp=None,
-        bounds=None,
-        constraints=(),
-        tol=None,
-        callback=None,
-        options=None,
-        compile=False,
-        processes=None,
+        self, opt, initial_parameters, loss=None, fit_options=dict(), compile=False
```
MatteoRobbiati, line 44:
```suggestion
        self, opt, initial_parameters, loss=None, fit_options={}, compile=False
```

src/qibo/models/variational.py
```diff
@@ -527,64 +446,61 @@ def minimize(
...
+        if len(initial_parameters) % 2 != 0:
             raise_error(
                 ValueError,
                 "Initial guess for the parameters must "
                 "contain an even number of values but "
-                "contains {}.".format(len(initial_p)),
+                "contains {}.".format(len(initial_parameters)),
             )
 
+        optloss = loss
```
MatteoRobbiati, line 468: I would recall it, e.g. `loss_function`.

MatteoRobbiati (commented):
Thanks @WanderingMike for the last modifications. Some suggestions follow.

src/qibo/optimizers/abstract.py
```diff
@@ -0,0 +1,92 @@
...
+        raise_error(
+            TypeError,
+            "Parameters must be a list of Parameter objects or a numpy array.",
+        )
+
+
+class Optimizer:
+    """Qibo optimizer class."""
+
+    def __init__(self, options={}):  # pragma: no cover
```
andrea-pasquale, line 49: There are some issues in python if you use as default `{}`
https://web.archive.org/web/20200605073004/http://effbot.org/zone/default-values.htm
The workaround is to use `None` and check in the constructor if `options` is not `None`.

src/qibo/optimizers/abstract.py
```diff
@@ -0,0 +1,92 @@
...
+                filled according to each specific optimizer interface.
+        """
+        self.options = options
+        self._fit_function = None
+
+    def set_options(self, updates):
+        """Update self.options dictionary"""
+        # check the new options are arguments of the fitting function
+        check_options(self._fit_function, updates)
+        self.options.update(updates)
```
andrea-pasquale, line 62: This method could be replaced by a python `property` setter.

src/qibo/optimizers/abstract.py
```diff
@@ -0,0 +1,92 @@
...
+
+    @abstractmethod
+    def get_options_list(self):  # pragma: no cover
+        """Return all available optimizer options."""
+        raise_error(NotImplementedError)
```
andrea-pasquale, line 67: For this one you can use a `property`.

src/qibo/optimizers/abstract.py
```diff
@@ -0,0 +1,92 @@
...
+
+    @abstractmethod
+    def get_fit_options_list(self):  # pragma: no cover
+        """Return all available fitting options."""
+        raise_error(NotImplementedError)
```
andrea-pasquale, line 72: Again python property also here.

src/qibo/optimizers/abstract.py
```diff
@@ -0,0 +1,92 @@
...
+
+    @abstractmethod
+    def fit(
+        self,
+        initial_parameters,
+        loss,
+        args=(),
+        fit_options={},
```
andrea-pasquale, line 80: Same comment as in the constructor, use `None` and check if `args` and `fit_options` are None.

src/qibo/optimizers/gradient_based.py
```diff
@@ -0,0 +1,168 @@
...
+        # perform the training
+        res = opt.fit(loss=loss, initial_parameters=params, args=(circuit, hamiltonian), fit_options={"epochs": 100}, nmessage=1)
+    """
+
+    def __init__(
+        self,
+        optimizer="Adagrad",
+        options={"learning_rate": 0.001},
+    ):
+        self.options = {}
```
andrea-pasquale, line 66:
```suggestion
```
This one can be removed, right?

src/qibo/models/evolution.py
```diff
@@ -263,7 +263,7 @@ def _loss(params, adiabatic_evolution, h1, opt_messages, opt_history):
             log.info(f"Params: {params}  -  <H1> = {loss}")
         return loss
 
-    def minimize(self, initial_parameters, method="BFGS", options=None, messages=False):
+    def minimize(self, opt, initial_parameters, fit_options={}, messages=False):
```
andrea-pasquale, line 266: Put `None` as default value and check if it is not `None`.
https://web.archive.org/web/20200605073004/http://effbot.org/zone/default-values.htm

src/qibo/models/evolution.py
```diff
@@ -280,17 +280,19 @@ def minimize(self, initial_parameters, method="BFGS", options=None, messages=Fal
                 optimization.
         """
         self.opt_messages = messages
-        if method == "sgd":
+        if isinstance(opt, optimizers.gradient_based.TensorflowSGD) == "sgd":
```
andrea-pasquale, line 283:
```suggestion
        if isinstance(opt, optimizers.gradient_based.TensorflowSGD):
```
Right, or am I missing something?

src/qibo/optimizers/abstract.py
andrea-pasquale: For `check_options` if you know that you need to do it before setting it you could also define a decorator if you want.

src/qibo/optimizers/gradient_based.py
```diff
@@ -0,0 +1,168 @@
...
+        self.name = "tensorflow"
+        # This optimizer works only with tensorflow backend
+        self.backend = TensorflowBackend()
+        self.options = options
+
+        # options are automatically checked inside the tf.keras.optimizer
+        self.optimizer = getattr(self.backend.tf.optimizers, optimizer)(**self.options)
+        self.name += f"_{self.optimizer.name}"
```
andrea-pasquale, line 74: Instead of having a `name` attribute you can define a `__str__` or `__repr__` magic method.

andrea-pasquale (commented):
Thanks @MatteoRobbiati.
I have a few comments.

I looked quickly at `optimizers`. I need to check the other code.
As a general remark, in python you should be careful when using default value for mutable variable.
https://web.archive.org/web/20200605073004/http://effbot.org/zone/default-values.htm
The standard workaround is to define `None` and put and `if` statement afterwards.

src/qibo/optimizers/abstract.py
```diff
@@ -0,0 +1,92 @@
...
+    method arguments.
+    """
+
+    for arg in options:
+        if arg not in list(inspect.signature(function).parameters):
+            raise_error(
+                TypeError,
+                f"Given argument {arg} is not accepted by {function.__code__.co_name} function, which is the fitting function of the chosen optimizer.",
+            )
+
```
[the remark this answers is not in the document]
MatteoRobbiati: This checking function is useful if we keep the dictionaries approach, because a dictionary can be generally updated with any key, but then we should check if the keys are accepted by the optimization function or not. 
I had a discussion about dictionaries with both @alecandido and @andrea-pasquale. Here there is a conflict between the flexibility I want to guarantee and the complexity of doing this.

A solution can be to use the safer `dataclasses`, or to constrain the optimizers arguments to the ones we want to accept, but by doing this we will loose the flexibility of the interface I am proposing. 

Maybe we can discuss a little bit during the Qibo meeting :)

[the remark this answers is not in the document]
alecandido: My suggestion would have been to pass all arguments down below. If it's not accepted by optimization, the optimization itself will fail ^^

PR #1129, opened 2023-12-11
Quantum Error Mitigation: Readout, Importance Clifford Sampling

AlejandroSopena:
General improvement of the code and docstrings. 

This PR implements [Iterative Bayesian Unfolding](https://arxiv.org/abs/1910.01969) to avoid quasiprobabilities when using readout mitigation with the response matrix. It also implements the [Importance Clifford Sampling](https://arxiv.org/abs/2112.06255) method. 

Checklist:
- [x] Tests are passing.


renatomello:
@AlejandroSopena maybe this is for a different PR, but we should aim to integrate these Clifford-based methods (e.g. ICS, CDR) with the new Clifford backend from @BrunoLiegiBastonLiegi as much as it is possible / worth it.

MatteoRobbiati (commented):
Thanks a lot @AlejandroSopena for all these modifications and for the big refactoring of the mitigation module. 

My only doubt here is about the usage of the `apply_readout_mitigation`. I agree with your choice, it lighten the code a lot and it is flexible enough to apply the desired readout mit technique, but I find it ambiguous when a mitigation method (e.g. CDR) is applied without applying readout mitigation, because still relying on `apply_readout_mitigation`.

What about renaming the `apply_readout_mitigation` function into something more generic like: `apply_problem_configuration`? Alternatively, we can make the two things explicit in a longer but clearer name: `apply_problem_with_readout_conf`.

A couple of minor comments follow.

src/qibo/models/error_mitigation.py
```diff
@@ -592,8 +632,340 @@ def apply_randomized_readout_mitigation(
...
+    return mitigated_expectation
+
+
+def _circuit_conf(circuit, qubit_map, noise_model=None, nshots=10000, backend=None):
+    """
+    Helper function to execute the given circuit with the specified parameters.
+
+    Args:
+        circuit (qibo.models.Circuit): input circuit.
+        qubit_map (list): the qubit map. If ``None``, a list of range of circuit's qubits is used. Defaults to ``None``.
```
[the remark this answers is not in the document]
AlejandroSopena: I tried to adapt the "manual" transpilation process we use in [RTQEM](https://github.com/qiboteam/rtqem/tree/clean/src/rtqem) but using the qibo transpiler. Depending on what is missing in #1121, we can merge this and then open a small pull request, or wait.

src/qibo/models/error_mitigation.py
```diff
@@ -115,57 +120,59 @@ def ZNE(
...
-        readout (dict, optional): It has the structure
-            {'calibration_matrix': `numpy.ndarray`, 'ncircuits': `int`}.
-            If passed, the calibration matrix or the randomized method is
-            used to mitigate readout errors.
-        backend (:class:`qibo.backends.abstract.Backend`, optional): Calculation engine.
+            Defaults to ``"CNOT"``.
+        readout (dict, optional): a dictionary that may contain the following keys:
+
+            *    ncircuits: int, specifies the number of random circuits to use for the randomized method of readout error mitigation.
+            *    calibration_matrix: numpy.ndarray, used for applying a pre-computed calibration matrix for readout error mitigation.
```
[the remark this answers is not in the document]
AlejandroSopena: The response matrix is the matrix obtained from the noisy frequencies when preparing each state. This matrix can be used in two ways to mitigate measurement error: by calculating the calibration matrix as the inverse of the response matrix or by using iterative Bayesian unfolding. This new method does not involve calculating the inverse of the response matrix and avoids some issues with the calibration matrix method.

I have updated the code and documentation to make this clear.

AlejandroSopena:
> Thanks a lot @AlejandroSopena for all these modifications and for the big refactoring of the mitigation module.
> 
> My only doubt here is about the usage of the `apply_readout_mitigation`. I agree with your choice, it lighten the code a lot and it is flexible enough to apply the desired readout mit technique, but I find it ambiguous when a mitigation method (e.g. CDR) is applied without applying readout mitigation, because still relying on `apply_readout_mitigation`.
> 
> What about renaming the `apply_readout_mitigation` function into something more generic like: `apply_problem_configuration`? Alternatively, we can make the two things explicit in a longer but clearer name: `apply_problem_with_readout_conf`.
> 
> A couple of minor comments follow.


I agree. I have changed it to `apply_problem_with_readout_conf()`.

AlejandroSopena:
> @AlejandroSopena maybe this is for a different PR, but we should aim to integrate these Clifford-based methods (e.g. ICS, CDR) with the new Clifford backend from @BrunoLiegiBastonLiegi as much as it is possible / worth it.


Yes, I agree @renatomello . I think we can implement this in a new PR because it only involves an improvement when running the training circuits but not the final result (although a very useful improvement).

src/qibo/models/error_mitigation.py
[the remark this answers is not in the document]
MatteoRobbiati: Okay perfect, your answer is the missing piece. 
Fine with this implementation.

src/qibo/models/error_mitigation.py
```diff
@@ -592,8 +632,340 @@ def apply_randomized_readout_mitigation(
...
+    )
+
+    if backend is None:  # pragma: no cover
+        backend = GlobalBackend()
+
+    non_clifford_gates = [
+        (i, gate)
+        for i, gate in enumerate(circuit.queue)
+        if not gate.clifford and not isinstance(gate, gates.M)
+    ]
```
BrunoLiegiBastonLiegi, line 725: I don't think the gates of this tuples are used anywhere in the following
```suggestion
    non_clifford_gates_indices = [
        i
        for i, gate in enumerate(circuit.queue)
        if not gate.clifford and not isinstance(gate, gates.M)
    ]
```

src/qibo/models/error_mitigation.py
```diff
@@ -592,8 +632,340 @@ def apply_randomized_readout_mitigation(
...
+            for q in gate.qubits:
+                gate_rand = gates.Unitary(
+                    random_clifford(1, backend=backend, return_circuit=False),
+                    q,
+                )
+                gate_rand.clifford = True
+                sampled_circuit.add(gate_rand)
+            sampled_circuit.add(gate)
+        else:
+            if i in [index for index, _ in non_clifford_gates]:
```
BrunoLiegiBastonLiegi, line 743: Better not to build the same list every iteration of the loop
```suggestion
            if i in non_clifford_gates_indices:
```

src/qibo/models/error_mitigation.py
```diff
@@ -592,8 +632,340 @@ def apply_randomized_readout_mitigation(
...
+        data["noisy"].append(noisy_expectation)
+        lambda_list.append(1 - noisy_expectation / expectation)
+
+    dep_param = np.mean(lambda_list)
+    dep_param_std = np.std(lambda_list)
+
+    noisy_expectation = apply_problem_with_readout_conf(
+        circuit, observable, noise_model, nshots, readout, qubit_map, backend=backend
+    )
+
```
BrunoLiegiBastonLiegi, line 917: Maybe compute these separately to avoid multiple identical calculations
```suggestion
    1_dep_squared = (1 - dep_param) ** 2
    dep_std_squared = dep_param_std ** 2
```

BrunoLiegiBastonLiegi (commented):
Tanks @AlejandroSopena, it looks good. I made some minor comments and in general I am not a huge fan of the two names of the functions `_circuit_conf` and `apply_problem_with_readout_conf`. I would probably prefer something like `_execute_circuit` and `_get_expectation_val_with_readout_mitigation`.

src/qibo/gates/special.py
```diff
@@ -54,7 +55,7 @@ def __init__(self, *q):
     def from_gate(cls, gate):
         fgate = cls(*gate.qubits)
         fgate.append(gate)
-        if isinstance(gate, (M, SpecialGate)):
+        if isinstance(gate, (M, SpecialGate, I)):
```
renatomello, line 58: @AlejandroSopena could you comment why you need to add `I` here?

src/qibo/models/circuit.py
```diff
@@ -81,7 +81,7 @@ def from_fused(self):
                     queue.append(gate.gates[0])
                 else:
                     queue.append(gate)
-            elif isinstance(gate.gates[0], (gates.SpecialGate, gates.M)):
+            elif isinstance(gate.gates[0], (gates.SpecialGate, gates.M, gates.I)):
```
renatomello, line 84: Here too

src/qibo/models/error_mitigation.py
```diff
@@ -592,8 +632,340 @@ def apply_randomized_readout_mitigation(
...
+        circuit_result = _circuit_conf(
+            circuit, qubit_map, noise_model, nshots, backend=backend
+        )
+
+    if "response_matrix" in readout:
+        circuit_result = apply_resp_mat_readout_mitigation(
+            circuit_result,
+            readout["response_matrix"],
+            readout.get("ibu_iters", None),
+        )
```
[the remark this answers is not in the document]
AlejandroSopena: No, one or the other. I have updated the code.

src/qibo/gates/special.py
AlejandroSopena: This was necessary in an initial version of the code, but not anymore. I have removed it.

BrunoLiegiBastonLiegi (approved):

src/qibo/models/error_mitigation.py
```diff
@@ -101,10 +105,11 @@ def ZNE(
     observable,
     noise_levels,
     noise_model=None,
-    nshots=int(1e4),
+    nshots=10000,
     solve_for_gammas=False,
     insertion_gate="CNOT",
     readout: dict = {},
```
MatteoRobbiati, line 111: On top of the review I just received from @andrea-pasquale, it would be better to set `None` as default value of a dictionary instead of a mutable variable. Then you can do:

```python
if readout is None:
    readout = {}
```

Take this comment valid for all the readout arguments that follow.

src/qibo/models/error_mitigation.py
```diff
@@ -375,182 +362,229 @@ def vnCDR(
...
-        nshots (int, optional): number of shots.
-        backend (:class:`qibo.backends.abstract.Backend`, optional): calculation engine.
+        nshots (int, optional): number of shots. Defaults to :math:`10000`.
+        backend (:class:`qibo.backends.abstract.Backend`, optional): backend to be used
+            in the execution. If ``None``, it uses :class:`qibo.backends.GlobalBackend`.
+            Defaults to ``None``.
 
     Returns:
-        numpy.ndarray : The computed (`nqubits`, `nqubits`) calibration matrix for
+        numpy.ndarray : The computed (`nqubits`, `nqubits`) response matrix for
```
MatteoRobbiati, line 485:
```suggestion
        numpy.ndarray : the computed (`nqubits`, `nqubits`) response matrix for
```

src/qibo/models/error_mitigation.py
```diff
@@ -375,182 +362,229 @@ def vnCDR(
...
-        calibration_matrix (numpy.ndarray): calibration matrix for readout mitigation.
+        state (:class:`qibo.measurements.CircuitResult`): the input state to be updated. This state should contain the
+            frequencies that need to be mitigated.
+        response_matrix (numpy.ndarray): the response matrix for readout mitigation.
+        iterations (int, optional): the number of iterations to use for the Iterative Bayesian Unfolding method.
+            If ``None`` the 'inverse' method is used. Defaults to ``None``.
 
     Returns:
-        :class:`qibo.measurements.CircuitResult`: the input state with the updated frequencies.
+        :class:`qibo.measurements.CircuitResult`: The input state with the updated (mitigated) frequencies.
```
MatteoRobbiati, line 530:
```suggestion
        :class:`qibo.measurements.CircuitResult`: the input state with the updated (mitigated) frequencies.
```

src/qibo/models/error_mitigation.py
```diff
@@ -592,8 +627,339 @@ def apply_randomized_readout_mitigation(
...
+            *    ncircuits: int, specifies the number of random circuits to use for the randomized method of readout error mitigation.
+            *    response_matrix: numpy.ndarray, used for applying a pre-computed response matrix for readout error mitigation.
+            *    ibu_iters: int, specifies the number of iterations for the iterative Bayesian unfolding method of readout error mitigation. If provided, the corresponding readout error mitigation method is used. Defaults to {}.
+
+        qubit_map (list, optional): the qubit map. If None, a list of range of circuit's qubits is used. Defaults to ``None``.
+        backend (qibo.backends.abstract.Backend, optional): the backend to be used in the execution.
+            If None, it uses the global backend. Defaults to ``None``.
+
+    Returns:
+        float: The mitigated expectation value of the observable.
```
MatteoRobbiati, line 665:
```suggestion
        float: the mitigated expectation value of the observable.
```

src/qibo/models/error_mitigation.py
```diff
@@ -592,8 +627,339 @@ def apply_randomized_readout_mitigation(
...
+    """Samples a training circuit for CDR by susbtituting some of the non-Clifford gates.
+
+    Args:
+        circuit (:class:`qibo.models.Circuit`): circuit to sample from.
+        backend (:class:`qibo.backends.abstract.Backend`, optional): backend to be used
+            in the execution. If ``None``, it uses :class:`qibo.backends.GlobalBackend`.
+            Defaults to ``None``.
+
+    Returns:
+        :class:`qibo.models.Circuit`: The sampled circuit.
```
MatteoRobbiati, line 706:
```suggestion
        :class:`qibo.models.Circuit`: the sampled circuit.
```

src/qibo/models/error_mitigation.py
```diff
@@ -592,8 +627,339 @@ def apply_randomized_readout_mitigation(
...
+                    q,
+                )
+                gate_rand.clifford = True
+                sampled_circuit.add(gate_rand)
+            sampled_circuit.add(gate)
+        else:
+            if i in non_clifford_gates_indices:
+                gate = gates.Unitary(
+                    random_clifford(1, backend=backend, return_circuit=False),
+                    gate.qubits[0],
```
MatteoRobbiati, line 740: Is this true also in case of two qubit gates? Or in that case you need to consider the target qubit?
If the second, I woud use `gate.target_qubits[0]`.

MatteoRobbiati (approved):
Thanks!
A couple of comments and some minor suggestions follow.

src/qibo/models/error_mitigation.py
AlejandroSopena: Thanks for spotting this. If there is a non-clifford two qubit gate, it should be replaced with a two-qubit Clifford gate, so

``` Python
gate = gates.Unitary(random_clifford(len(gate.qubits), backend=backend, return_circuit=False), *gate.qubits,)
```

PR #1132, opened 2023-12-11
Darray-based multi-node

alecandido:
- [ ] execute operation on disk
- [ ] properly test

## Rationale

The idea of this backend should be pretty simple:

- take a NumPy array (the initial state) and turn it into a Dask one (`dask.array`)
- apply all the other operations as usual
- eventually, use the client to `compute` the result

In Dask the operations are recorded in a computational graph, and NumPy operations are also supported.
So, the idea would be to do something like:
```py
import dask.array as da

initial = da.zeros(...)
final = np.sin(initial + 3)
result = client.compute(final)
```

Checklist:
- [ ] Tests are passing.


AleCandido:
Tests are definitely missing, that's why it is nothing more than a sketch currently.

AleCandido:
I decided to duplicate the `execute_circuit`, since I really need to customize it.

AleCandido:
@scarrazza I'm not finding a way to execute the operations on a Dask array without receiving the whole result in memory.

It seems that the suggested workflow would be to keep all the operations delayed, for as long as you do not trace enough to get an object that you can hold in memory.
Instead, I would have assumed that we'd like to perform the actual computation during execution, which would be impossible if the result is going to be too large to store.

I hoped this line could have given me the desired result
https://github.com/qiboteam/qibo/blob/0cbb0d0e2f67026a6c1ae9bb612fbfe3d756c193/src/qibo/backends/darray.py#L96-L97
but it seems like `.persist()` is very similar to `.compute()` (possibly keeping the result in distributed memory, if computed in a distributed way, rather done downloading on the master).

alecandido:
This was a proof-of-concept, and it's not going to be merged anytime soon.

PR #1137, opened 2023-12-13
Creation of `quantum_info.quantum_networks` submodule and  `QuantumNetwork` class

### LINKED ISSUE #1077: Implementation of `QuantumNetwork` class and `link_product`
[Quantum networks](https://journals.aps.org/pra/abstract/10.1103/PhysRevA.80.022339), as a generalization of *quantum channels* to higher order maps, is a powerful tool in the study of quantum systems. It is important in simulating complex quantum systems, such as in the analysis of [causal structures](https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.130.240201).

To simulate a quantum network, we need to be able to represent the quantum network as a Choi operator with tensor structures. The `QuantumNetwork` class should store the information on the labeling of all input and output systems, which identifies the leads of the Choi operator tensor. The `link_product` function is also required, which takes two `QuantumNetwork` objects and return a new `QuantumNetwork` object that represents the link product of the inputs.

The overall goal of this class is to help people construct and manipulate quantum networks in a more intuitive way, where they can be done as simple as the construction and manipulation of tensor networks.
```[tasklist]
### Tasks
- [ ] https://github.com/qiboteam/qibo/pull/1116
- [ ] https://github.com/qiboteam/qibo/pull/1137
```

renatomello:
This closes #1077 

Checklist:
- [x] Tests are passing.


Canoming (changes requested):
I think the current version of `QuantumNetwork` is usable. But I found several issues in the implementation. Although I have fixed the issues, they are not detected by the `pytest`. We may need to add the following tests to the test files:

1. The `QuantumNetwork` object should not be mutated after calling `link_product` or `@`;
2. The `__add__` , `__truediv__` and `__mul__` functions should work for both pure and non-pure `QuantumNetworks`. We need to test them on both conditions.

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -431,15 +430,22 @@ def __mul__(self, number: Union[float, int]):
...
-        return QuantumNetwork(
-            number * matrix,
-            self.partition,
-            self.system_output,
-            self._pure,
-            self._backend,
-        )
+        if self.pure():
+            return QuantumNetwork(
+                np.emath.sqrt(number) * self.matrix(),
```
renatomello, line 435: @Canoming do you need `np.emath.sqrt` here to return complex numbers or does `np.sqrt` suffice?

renatomello:
@Canoming is the use of `np.emath.sqrt` necessary? i.e. can a quantum network be multiplied by a complex number?

renatomello:
>     2. The `__add__` , `__truediv__` and `__mul__` functions should work for both pure and non-pure `QuantumNetworks`. We need to test them on both conditions.

Just to confirm with you: a pure QN, after added to another pure network becomes non-pure, correct? Same thing when multiplying / dividing the pure QN by a number, right?

renatomello:
>     1. The `QuantumNetwork` object should not be mutated after calling `link_product` or `@`;

This is fixed now.

>     2. The `__add__` , `__truediv__` and `__mul__` functions should work for both pure and non-pure `QuantumNetworks`. We need to test them on both conditions.

Just to confirm with you: a pure QN, after added to another pure network becomes non-pure, correct? Same thing when multiplying / dividing the pure QN by a number, right?

Canoming:
> @Canoming is the use of `np.emath.sqrt` necessary? i.e. can a quantum network be multiplied by a complex number?

Ah, sorry. I just rethought it. According to our construction of non-pure networks from pure networks, the phase of the complex will be omitted. So here is the expected behavior of the multiplication and division:
1. If the `number` multiplied or divided is a positive number, we'll keep a pure network to be pure, so we can save the memory. The pure network will be multiplied/divided by the square root of the `number` (which is equivalent to multiply/divide the non-pure network with the target `number`)
2. If the `number` is non-positive, we'll convert everything to non-pure networks. The multiplication and division of negative number will convert a semidefinite positive operator to semidefinite negative operator, which cannot be represented by pure networks.

I've updated `__mul__` and `__truediv__` accordingly. Please check if my code behaves as what

Canoming:
I updated the test file to match the change of behavior of `__mul__` and `__truediv__`.

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -0,0 +1,644 @@
...
+        Args:
+            state (ndarray): density matrix of a ``state``.
+
+        Returns:
+            ndarray: Resulting state :math:`\\mathcal{E}(\\varrho)`.
+        """
+        matrix = np.copy(self._matrix)
+
+        if self.pure():
+            return np.einsum("jk,lm,jl -> km", matrix, np.conj(matrix), state)
```
renatomello, line 284: @Canoming let me ask you something: this method, when ``self.pure()`` is ``True``, is applying $U^{T} \\, \rho \\, U^{*}$ instead of the traditional $U \\, \rho \\, U^{\\dagger}$. Is that supposed to be the case or is it a bug?

renatomello:
@Canoming could you help me beef up the documentation for `QuantumNetworks` in general? It would be nice to have detailed documentation for other people that are going to use this class.  Maybe with some examples on how to use it too

renatomello:
> > @Canoming is the use of `np.emath.sqrt` necessary? i.e. can a quantum network be multiplied by a complex number?
> 
> Ah, sorry. I just rethought it. According to our construction of non-pure networks from pure networks, the phase of the complex will be omitted. So here is the expected behavior of the multiplication and division:
> 
>     1. If the `number` multiplied or divided is a positive number, we'll keep a pure network to be pure, so we can save the memory. The pure network will be multiplied/divided by the square root of the `number` (which is equivalent to multiply/divide the non-pure network with the target `number`)
> 
>     2. If the `number` is non-positive, we'll convert everything to non-pure networks. The multiplication and division of negative number will convert a semidefinite positive operator to semidefinite negative operator, which cannot be represented by pure networks.
> 
> 
> I've updated `__mul__` and `__truediv__` accordingly. Please check if my code behaves as what

Could you add that to the docstrings of these functions too? Thanks.

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -0,0 +1,645 @@
...
+
+        return matrix
+
+    def _check_subscript_pattern(self, subscripts: str):
+        """Checks if input subscript match any implemented pattern."""
+        braket = "[a-z]"
+        pattern_two = re.compile(braket * 2 + "," + braket * 2 + "->" + braket * 2)
+        pattern_four = re.compile(braket * 4 + "," + braket * 4 + "->" + braket * 4)
+        # pattern_two = re.compile("[a-z][a-z],[a-z][a-z]->[a-z][a-z]")
+        # pattern_four = re.compile("[a-z][a-z][a-z][a-z],[a-z][a-z]->[a-z][a-z]")
```
AlejandroSopena, line 641:
```suggestion
```

AlejandroSopena (approved):

Canoming:
> @Canoming could you help me beef up the documentation for `QuantumNetworks` in general? It would be nice to have detailed documentation for other people that are going to use this class. Maybe with some examples on how to use it too

Ah, sorry for the late reply. I can write some documentation for this class. But should we restore the `quantum_network` branch? The current version is only a prototype; we should add more functionality to this class later. Are we supposed to create a new branch for the next stage of development?

renatomello:
> Ah, sorry for the late reply. I can write some documentation for this class. But should we restore the `quantum_network` branch? The current version is only a prototype; we should add more functionality to this class later. Are we supposed to create a new branch for the next stage of development?

As long as the branch is up-to-date with `master`, it makes no difference if you restore this one or create a new one

Canoming:
Emm.. I just found an issue in the latest version of `QuantumNetwork`. The following code will crash the kernel:

```python
import numpy as np

from qibo.quantum_info.quantum_networks import QuantumNetwork
from qibo.gates import DepolarizingChannel

# Create a quantum network from numpy array is fine
test = np.ones((2,2,2,2))
test_choi = QuantumNetwork(test, (2,2,2,2), pure=True)
print(test_choi.partition)
print(test_choi.system_output)

# But create a quantum network from the output of the old quantum channel will crash the kernal
test_ch = DepolarizingChannel(0,0.5)
N = len(test_ch.target_qubits)
partition = (2**N, 2**N)
depolar_choi = QuantumNetwork(test_ch.to_choi(), partition)
print(f'A quantum channel is a quantum netowrk of the form {depolar_choi}')
```

The creation of the `QuantumNetwork`, `depolar_choi`, will prompt the error and crash the kernel:
```
python: symbol lookup error: /home/ximing/miniconda3/envs/qibo_dev/lib/python3.11/site-packages/mkl/../../../libmkl_intel_thread.so.2: undefined symbol: omp_get_num_procs
```

I used these codes as the demo use case of `QuantumNetwork`, and they worked before. Not sure what causes the problem. Seems `DepolarizingChannel` hasn't changed in the recent commits. But as the kernel crashes, the debug functions also crash. Could you help to figure out what happened to these codes?

By the way, I'm also not sure where to put the documentation of `QuantumNetwork`. I think all the examples in `qibo/examples` are aiming for very specific algorithms. But `QuantumNetwork` is a very fundamental mathematical concept, and we haven't implemented the conversion of `QuantumNetwork` with other classes in `qibo` back and forth. Do you have any suggestions of where to put the documentation?

Canoming:
Emm.. I just found an issue in the latest version of `QuantumNetwork`. The following code will crash the kernel:

```python
import numpy as np

from qibo.quantum_info.quantum_networks import QuantumNetwork
from qibo.gates import DepolarizingChannel

# Create a quantum network from numpy array is fine
test = np.ones((2,2,2,2))
test_choi = QuantumNetwork(test, (2,2,2,2), pure=True)
print(test_choi.partition)
print(test_choi.system_output)

# But create a quantum network from the output of the old quantum channel will crash the kernal
test_ch = DepolarizingChannel(0,0.5)
N = len(test_ch.target_qubits)
partition = (2**N, 2**N)
depolar_choi = QuantumNetwork(test_ch.to_choi(), partition)
print(f'A quantum channel is a quantum netowrk of the form {depolar_choi}')
```

The creation of the `QuantumNetwork`, `depolar_choi`, will prompt the error and crash the kernel:
```
python: symbol lookup error: /home/ximing/miniconda3/envs/qibo_dev/lib/python3.11/site-packages/mkl/../../../libmkl_intel_thread.so.2: undefined symbol: omp_get_num_procs
```

I used these codes as the demo use case of `QuantumNetwork`, and they worked before. Not sure what causes the problem. Seems `DepolarizingChannel` hasn't changed in the recent commits. But as the kernel crashes, the debug functions also crash. Could you help to figure out what happened to these codes?

By the way, I'm also not sure where to put the documentation of `QuantumNetwork`. I think all the examples in `qibo/examples` are aiming for very specific algorithms. But `QuantumNetwork` is a very fundamental mathematical concept, and we haven't implemented the convertions of

PR #1139, opened 2023-12-18
Add `Clifford.to_circuit` and `Clifford.copy` methods + initialization via circuits

renatomello:
Checklist:
- [x] Tests are passing.


renatomello:
@BrunoLiegiBastonLiegi @AlejandroSopena this should be ready for review (with some urgency) after passing tests

BrunoLiegiBastonLiegi (commented):
Thanks @renatomello I just finished giving this a first look and it's looking good. I have got just a couple of general comments. 
The first one regards the ability to pass a circuit directly to the `Clifford` object constructor. From my perspective this is somewhat redundant as you can already do that with `Clifford.from_circuit()` and it's not really adding any additional functionality.
Secondly I wonder whether many of the `clifford_utils` functions should be moved to the `CliffordBackend` object. They are extensively making use of the backend through the `Clifford` object already, at this point they should probably just be part of it and called at need inside of the `Clifford`. 
Finally, why do you need to store the circuit inside of the `Clifford` object? In general I would like to keep it as light as possible, keeping in mind that at some point we might want to create an abstract `Result` object from which both this `Clifford`, the `QuantumState`, `MeasurementOutcomes`, `CircuitResult` and all the possible future outcomes object inherit from.

renatomello:
> Thanks @renatomello I just finished giving this a first look and it's looking good. I have got just a couple of general comments. The first one regards the ability to pass a circuit directly to the `Clifford` object constructor. From my perspective this is somewhat redundant as you can already do that with `Clifford.from_circuit()` and it's not really adding any additional functionality. 

Redundancy is not necessarily a bad thing. For instance, we have `Clifford.from_circuit(circuit)` and `CliffordBackend.execute_circuit(circuit)`. The question is what level of redundancy is acceptable and desirable. And to this point, I insist that `Clifford(circuit)` is the right level of redundancy and adds to UX. In my opinion, `Clifford(circuit)` is the most natural way of executing a circuit with the Clifford backend from a user's perspective. Moreover, other libraries, e.g. `qiskit` support this type of initialization, so it's also natural that users will expect us to support the same thing.

> Secondly I wonder whether many of the `clifford_utils` functions should be moved to the `CliffordBackend` object. They are extensively making use of the backend through the `Clifford` object already, at this point they should probably just be part of it and called at need inside of the `Clifford`. 

I have no objection to this. However, this sounds like a different PR.

> Finally, why do you need to store the circuit inside of the `Clifford` object? In general I would like to keep it as light as possible, keeping in mind that at some point we might want to create an abstract `Result` object from which both this `Clifford`, the `QuantumState`, `MeasurementOutcomes`, `CircuitResult` and all the possible future outcomes object inherit from.

You are right that it can be memory intensive to store the `circuit.queue` in the `Clifford` object. I will roll that back.

AlejandroSopena (approved):
Thanks @renatomello, everything looks good. I agree with @BrunoLiegiBastonLiegi 's comment to move some `clifford_utils` functions to the `CliffordBackend`.

BrunoLiegiBastonLiegi (approved):

PR #1148, opened 2023-12-21
increasing doctest speed

scarrazza:
I was debugging `doctest` when I realized that `testsetup` was not fixing the backend properly. I have changed this call to `testcode` and performance has improved a lot. We can reduce the requirements for examples (e.g. nqubits, epochs, shots) however the impact is not relevant in comparison to the small change applied in this branch.

Checklist:
- [x] Tests are passing.


andrea-pasquale (approved):
Thanks @scarrazza

Issue #1128, opened 2023-12-08
Improvements on top of the Clifford Simulator


BrunoLiegiBastonLiegi:
#1076 is introducing in qibo a way for simulating clifford circuits efficiently, several improvements can still be introduced on top of that though:
- [x] Currently the ``CliffordBackend`` is backend agnostic, in the sense that any one of the backend available in qibo and qibojit right now (except for the ``TensorflowBackend``) can be used  as engines for performing the computation happening under the hood of the ``CliffordBackend``. In practice, though, the only difference is whether numpy or cupy primitives are used. Therefore, there's no difference between using, for example,  the ``NumbaBackend`` or ``NumpyBackend`` as engine right now. Numba parallelization can be implemented using the ``@njit(parallel=True)`` decorator with the clifford operations that define how gates act in the phase space. Similarly custom cuda kernels can be implemented in cupy using the ``@jit.rawkernel()`` decorator on the clifford operations.
- [x] Add support for clifford noise channels through repeated execution, as it works for collapsing measurements already. 
- [ ] Come up with a way for representing ensembles of states in the phase space which is what you get when collapsing measurement or noise channels are present in the circuit. This [https://arxiv.org/abs/2002.06181](https://arxiv.org/abs/2002.06181) might provide some ideas. 
- [ ] Currently the action of a clifford gate in the phase space is represented by a series of row or column wise operations on the symplectic matrix, like for example swapping rows or summing columns. It would be more flexible to have a matrix multiplication representation for each operation. 
- [x] As per suggestion of @AlejandroSopena and @AleCandido, it might be a good idea to integrate [stim](https://github.com/quantumlib/stim) as a separate alternative clifford backend.
- [ ] The clifford gates + the T gate form a universal basis for computation, several methods exist for simulating circuits decomposed in this basis with a cost that grows exponentially only in the number of T gates present. An additional hybrid clifford backend could be implemented following this idea.
- [ ] Python native ints appear to be faster than numpy arrays of bools with boolean algebra. Namely, the operation ``5 ^ 3`` is faster than ``np.array([True, False, True]) ^ np.array([False, True, True])``. Therefore, a substantial speed up might be obtained by rethinking the symplectic matrix in terms of the integers encoded by its rows and columns, however, this would require to rewrite also all the clifford gates in the new representation.
- [ ] (@alecandido) It would really be nice to have a syntax for gates composition in order to avoid the repetition of patterns, to make it something like:
```suggestion
    theta_switch(
        theta,
[...]

PR #1150, opened 2023-12-27
Customizing Clifford operations for `qibojit` engines

BrunoLiegiBastonLiegi:
This PR removes the `CliffordOperations` object and substitute it with standalone functions. This contains also other minor changes for compatibility with the `numba` and `cupy` customized operations that are implemented in qiboteam/qibojit#161.

Checklist:
- [ ] Tests are passing.


renatomello:
@BrunoLiegiBastonLiegi let me know if this is ready for review

BrunoLiegiBastonLiegi:
Thanks @renatomello, I believe this is ready for a first round of review now.

src/qibo/backends/clifford.py
```diff
@@ -589,6 +85,29 @@ def zero_state(self, nqubits: int):
...
+
+        Args:
+            state (ndarray): The input state.
+            nqubits (int): Number of qubits.
+
+        Returns:
+            (ndarray): The reshaped state.
+        """
+        return self.engine.clifford_post_execution_reshape(state, nqubits)
+
```
renatomello, line 110: Do these two functions need to be user-facing?

src/qibo/backends/clifford_operations.py
```diff
@@ -0,0 +1,436 @@
+import numpy as np
```
renatomello, line 1: Same here. Seems like the entire module does not need to be user-facing

src/qibo/backends/numpy.py
```diff
@@ -75,6 +77,12 @@ def to_numpy(self, x):
     def compile(self, func):
         return func
 
+    def clifford_pre_execution_reshape(self, state):
+        return state
+
+    def clifford_post_execution_reshape(self, state, nqubits):
+        return state
```
renatomello, line 84: Why are these functions not doing anything?

tests/test_backends_clifford.py
```diff
@@ -250,6 +247,7 @@ def test_bitflip_noise(backend):
 
 
 def test_set_backend(backend):
+    set_backend(backend.name)
```
renatomello, line 250: Does this work for the `qibojit` backends as well?

src/qibo/backends/clifford_operations.py
```diff
@@ -0,0 +1,450 @@
...
+    x1_neq_z1 = ~x1_eq_z1
+    x1_eq_0 = x1 == False
+    x1_eq_1 = ~x1_eq_0
+    ind2 = x1_eq_z1 & x1_eq_1
+    ind3 = x1_eq_1 & x1_neq_z1
+    ind4 = x1_eq_0 & x1_neq_z1
+    exp[ind2] = z2[ind2].astype(int) - x2[ind2].astype(int)
+    exp[ind3] = z2[ind3].astype(int) * (2 * x2[ind3].astype(int) - 1)
+    exp[ind4] = x2[ind4].astype(int) * (1 - 2 * z2[ind4].astype(int))
+    return exp
```
alecandido, line 356: Not sure why he did not simplify the expression in the paper, but you could write it in a single one, without any `if` required (it's just boolean algebra, with 4 rather than 2 variables):
```py
-2*x1*x2*z1 + 2*x1*x2*z2 + 2*x1*z1*z2 - x1*z2 - 2*x2*z1*z2 + x2*z1
```

alecandido: Btw, it's interesting to notice that is actually a determinant:
```py
[ins] In [1]: import sympy
                                                                                                        
[ins] In [2]: x1, z1, x2, z2 = sympy.symbols("x1 z1 x2 z2")
                                                                                                        
[ins] In [3]: m = sympy.Matrix([[x1, z1, 2 * x1], [x2, z2, 2 * z2], [x2, z1, 1]])
                                                                                                        
[ins] In [4]: m
Out[4]:
Matrix([
[x1, z1, 2*x1],
[x2, z2, 2*z2],
[x2, z1,    1]])
                                                                                                        
[ins] In [5]: for p in [(0, 0), (1, 1), (1, 0), (0, 1)]:
         ...:     t = (-m.det()).subs(zip((x1, z1), p)).simplify()
         ...:     print(f"{p}: {t}")
         ...:
(0, 0): 0
(1, 1): -x2 + z2
(1, 0): z2*(2*x2 - 1)
(0, 1): x2*(1 - 2*z2)
                                                                                                        
[ins] In [6]: -m.det()
Out[6]: -2*x1*x2*z1 + 2*x1*x2*z2 + 2*x1*z1*z2 - x1*z2 - 2*x2*z1*z2 + x2*z1
```

But I'm not sure it's an advantage wrt to just using the function above.

alecandido: (in principle the determinant could be more optimal, if you start from the matrix, because it could cache and reuse part of the computation - but here it's also simpler than an arbitrary determinant, so I'd rather manually collect some of the terms or defining some variables, if the speed-up is soooo relevant...)

src/qibo/backends/clifford.py
BrunoLiegiBastonLiegi: no you'e right, should I call them `_clifford_pre_execution_reshape` and `_clifford_post_execution_reshape`

src/qibo/backends/clifford_operations.py
BrunoLiegiBastonLiegi: what do you mean?

src/qibo/backends/numpy.py
BrunoLiegiBastonLiegi: because for the `numpy` backend you don't need to do anything, this is used by the `cupy` backend only for reshaping the symplectic matrix

src/qibo/backends/clifford.py
renatomello: I think so

src/qibo/backends/clifford_operations.py
renatomello: I think this entire file does not need to be user-facing. `_clifford_operations.py`

src/qibo/backends/numpy.py
renatomello: I see. So they're overwritten in `qibojit`?

tests/test_backends_clifford.py
BrunoLiegiBastonLiegi: Apparently it works, but it's actually not needed. I'll update this test.

src/qibo/backends/clifford_operations.py
BrunoLiegiBastonLiegi: Interesting, speeding up the `_exponent` function is quite important for measurements actually. I am not sure how to deal with the determinant though, as each `x` and `z` is a matrix itself. I'll probably go with the easier version for now, something like this:
```
x1_x2 = x1 & x2
x1_z2 =  x1 & z2
x2_z1 = x2 & z1
exp = (
-2 * (x1_x2 & z1).astype(int) + 
2 * (x1_x2 & z2).astype(int) + 
2 * (x1_z2 & z1).astype(int) - 
x1_z2.astype(int) - 
2 * (x2_z1 & z2).astype(int)  +
x2_z1.astype(int)
)
```

src/qibo/backends/numpy.py
BrunoLiegiBastonLiegi: yes they are

src/qibo/backends/clifford_operations.py
BrunoLiegiBastonLiegi: I cannot rename the gate functions names as they need to be called the same as the standard qibo gates. I could rename the module though.

src/qibo/backends/clifford_operations.py
alecandido: If nothing else, just for the sake of readability, I'd apply the cast immediately to the 4 variables (they are also 4 cast operations instead of 6), and just use `*` instead of `&` (while `|` is only `+ mod 2`, `&` is `*` even in the target set of natural numbers).

Then you can collect as you prefer, if you want to speed up. There is no unique way of doing it, but I'd collect at least the factor of 2, and an arbitrary choice among the equivalent ones (despite not being unique - that's why I wrote it expanded - collecting reduces the number of operations).
E.g.
```py
x1_ = x1.as_type(int)
x2_ = x2.as_type(int)
z1_ = z1.as_type(int)
z2_ = z2.as_type(int)
exp = 2 * (x1_*x2_ * (z2_ - z1_) + z1_*z2_ * (x1_ - x2_)) - x1_*z2_ + x2_*z1_
```

I did this manually (and spontaneously attempting to preserve some symmetry), but you could try to ask Sympy to collect them, and check if you could manage to get even fewer operations in `exp`.

Moreover, I'm appending an `_` to the casted variables, for the general principle of avoiding shadowing. I expect any compiler to be able to drop from memory unused variables, but you could check explicitly if shadowing the name forces the memory free (I don't believe so, but with Python, and even its frameworks, memory management is an art of guessing...).

renatomello:
Yeah just the module is enough

src/qibo/backends/clifford_operations.py
BrunoLiegiBastonLiegi: But wait, you cannot cast them before I believe, you want to do the boolean operations first and then sum all the resulting ints. Otherwise you'll end up doing boolean operations with numpy ints that should be considerably slower.

src/qibo/backends/clifford_operations.py
renatomello: Yeah just the module is enough

renatomello (approved):

src/qibo/backends/_clifford_operations.py
```diff
@@ -342,17 +342,17 @@ def _exponent(x1, z1, x2, z2):
...
+    x2_z1 = x2 & z1
+    exp = (
+        -2 * (x1_x2 & z1).astype(int)
+        + 2 * (x1_x2 & z2).astype(int)
+        + 2 * (x1_z2 & z1).astype(int)
+        - x1_z2.astype(int)
+        - 2 * (x2_z1 & z2).astype(int)
+        + x2_z1.astype(int)
+    )
     return exp
```
alecandido:
```suggestion
    return (
        -2 * (x1_x2 & z1).astype(int)
        + 2 * (x1_x2 & z2).astype(int)
        + 2 * (x1_z2 & z1).astype(int)
        - x1_z2.astype(int)
        - 2 * (x2_z1 & z2).astype(int)
        + x2_z1.astype(int)
    )
```
No need to name the return value :)

src/qibo/backends/clifford_operations.py
alecandido: As I said, the only boolean operations you're doing before are the `&`, that is the same of the `*` for the domain you're interested in.

Consider that NumPy booleans are still `u8` (they use entire bytes), and that the integer product is an assembly operation (so I'm not user how much you could get out of the `&`...).

For as long as you're not working on bitstrings, I'd suggest to just use `int8` for the type (since you could go negative, but you're bounded to be `|exp| < 10`... actually the bound should be `|exp| <= 2`, but you could get more in the intermediate operations, though it's not relevant).

If you were using bitstrings, then I would agree with you, that could be faster.

alecandido: If you were using bitstrings instead, then I would agree with you, that could be faster. But it's a trade-off with the exceeding casts.

So, eventually you should benchmark them (or print the assembly and read the instructions).

BrunoLiegiBastonLiegi: I tried running this simple code snippet
```
tottime = 0
for _ in range(1000):
     t = time()
     a * b
     tottime += time() - t
print(tottime/1000)
```
with `a` and `b`:
```
a = np.random.randint(2, size=int(1e6), dtype=int)
b = np.random.randint(2, size=int(1e6), dtype=int)
```
I get on average `8*10^(-4)` with `dtype=int` and `5*10^(-5)` with `dtype=bool`

src/qibo/backends/_clifford_operations.py
```diff
@@ -0,0 +1,449 @@
...
+        symplectic_matrix = CZ(symplectic_matrix, control_q, target_q, nqubits)
+        symplectic_matrix = X(symplectic_matrix, target_q, nqubits)
+        symplectic_matrix = CZ(symplectic_matrix, control_q, target_q, nqubits)
+        return X(symplectic_matrix, target_q, nqubits)
+    # theta = 3 * pi + 4 * n * pi
+    elif (theta / np.pi - 3) % 4 == 0:
+        symplectic_matrix = CNOT(symplectic_matrix, control_q, target_q, nqubits)
+        symplectic_matrix = X(symplectic_matrix, target_q, nqubits)
+        symplectic_matrix = CY(symplectic_matrix, control_q, target_q, nqubits)
+        return X(symplectic_matrix, target_q, nqubits)
```
alecandido, line 302: You know that if no condition is satisfied, the function will just return `None`?

I'd rather raise an error...

alecandido: Moreover, this is another repeated pattern. It would really be nice to have a syntax for gates composition, to make it something like:
```suggestion
    theta_switch(
        theta,
        I,
        X | CY | X | CNOT,
        CZ | X | CZ | X,
        CNOT | X | CY | X,
        symplectic_matrix, control_q, target_q, nqubits
    )
```
with `theta_switch` a function applying the repeated pattern (and I chose `|` at random, any operator would be fine).

If you're interested in, we could try to design something like that.

BrunoLiegiBastonLiegi: I am not really familiar with monads, but sounds like an interesting idea!

renatomello:
@BrunoLiegiBastonLiegi @alecandido may I propose something? We need these custom ops at least on `master` very soon in order to perform more complex benchmarks. So, I propose we merge this and the `qibojit` PR with a minimal working version of `numba` and `cupy`. After that we could create follow-ups with more intricate syntax optimizations and speed-ups. What do you guys think?

src/qibo/backends/_clifford_operations.py
alecandido: > I am not really familiar with monads, but sounds like an interesting idea!

That was kind of a joke: there are no monads in Python. With monads, you can manage a state in a pure functional way, without explicitly passing it in every function invocation, since it will be part of the type (it's slightly more complicate actually https://learnyouahaskell.com/a-fistful-of-monads...)

In this case, the idea is that your state is exactly the symplectic matrix, and you'd like to make operations on that.

However, we don't need such a fancy thing (though it would be kind of elegant), but here we just need a way to batch some operations and execute them.
We can just make the operations instances of a class defining a `__or__` containing the function, something like:

```py
@dataclass
class CliffordOp:
    ops: list[Callable]

    def __or__(self, other):
        ...

    def __call__(self, state):
        for f in ops:
            state = f(state)
        return state

    @classmethod
    def atom(cls, f):
        return cls([f])

def _CNOT(state):
    ...

CNOT = CliffordOp.atom(_CNOT)
```

The details could be improved, but that's the idea.

Another option is to just compose the function in `__or__`:
```py
@dataclass
class CliffordOp:
    f:  Callable

    def __or__(self, other):
        cls = type(self)
        return cls(lambda state: self(other(state)))

    def __call__(self, state):
        return self.f(state)

def _CNOT(state):
    ...

CNOT = CliffordOp(_CNOT)
```
but Python is not a functional language, so it has limited support for optimizations on calls, I'm not sure how much I'd play this game of nesting...

alecandido:
> @BrunoLiegiBastonLiegi @alecandido may I propose something? We need these custom ops at least on `master` very soon in order to perform more complex benchmarks. So, I propose we merge this and the `qibojit` PR with a minimal working version of `numba` and `cupy`. After that we could create follow-ups with more intricate syntax optimizations and speed-ups. What do you guys think?

I'd agree to merge this soon (i.e. even today) postponing syntax optimizations to a further PR.

But I'd start the further PR almost immediately, because the problem itself is not having a fancy syntax for the sake of itself, but to avoid much repetition and a bunch of barely maintainable code.

I'm not sure I agree about Numba and CuPy instead: they are only needed for performance optimization. And that's what we're doing in the review, i.e. further optimizing performances. 

Btw, the bare Clifford is already in `master`, this is already improving the internals. Is it impossible to start working on the complex benchmarks with that?

renatomello:
> Btw, the bare Clifford is already in `master`, this is already improving the internals. Is it impossible to start working on the complex benchmarks with that?

`master` now allows us to benchmark `numpy`,which is what I'm doing right now. But I'd like to benchmark `numba` and `cupy` soon, even if we know we can do better.

alecandido:
> `master` now allows us to benchmark `numpy`,which is what I'm doing right now. But I'd like to benchmark `numba` and `cupy` soon, even if we know we can do better.

Ok, then I just miss part of the motivation. But yes, if you need them in `master` we could for sure merge.

Btw, the only thing that makes `master` special, in a certain sense, is that it's the branch we're releasing.
In general (not necessarily for this case), if you don't need a feature to be part of a release, you could start working with that even before (using the corresponding branch wherever relevant).

I'm pointing this out because currently we don't have many users, but in case we had we should pay much care to what we're releasing, and since we have tight release cycles (once per month) we should keep working on new features in separate branches, even for a long time (of course there is a trade-off, since you might need to update it many times with `master`, but if it is isolated enough it's affordable).

BrunoLiegiBastonLiegi:
Ok I think I addressed all the comments except the one about the monads and updated the qibojit PR accordingly. So, how do we want to procede?

src/qibo/backends/clifford_operations.py
alecandido: I tested the full function:
- the current one:
```
143 µs ± 1.01 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
- the current one with `np.int8` instead of `int` (that should be an alias for `np.int32` or `np.int64` within NumPy)
```
59 µs ± 264 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
- my proposal
```
20.4 µs ± 27.2 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

(I've been lazy and I did it in the IPython interpreter, I attach the input log [quickbench.py.zip](https://github.com/qiboteam/qibo/files/14219209/quickbench.py.zip), you can just load with `%load` in IPython - it's a bit of a mess, but if you delete a bunch of lines, rearrange, and import `timeit` instead of using the magic it becomes a script in no time)

src/qibo/backends/numpy.py
```diff
@@ -30,6 +30,8 @@ def __init__(self):
             np.complex128,
         )
 
+        self.clifford_operations = _clifford_operations
```
alecandido, line 33: This should be in the `CliffordBackend`, NumPy doesn't have to know nothing about Clifford.

src/qibo/backends/numpy.py
```diff
@@ -75,6 +77,12 @@ def to_numpy(self, x):
     def compile(self, func):
         return func
 
+    def _clifford_pre_execution_reshape(self, state):
+        return state
+
+    def _clifford_post_execution_reshape(self, state, nqubits):
+        return state
+
```
alecandido, line 85: These as well, there is no need to keep them in `NumpyBackend`

src/qibo/gates/abstract.py
```diff
@@ -391,7 +391,7 @@ def apply_density_matrix(self, backend, state, nqubits):
         return backend.apply_gate_density_matrix(self, state, nqubits)
 
     def apply_clifford(self, backend, state, nqubits):
-        return backend.apply_gate_clifford(self, state, nqubits)
+        return backend.apply_gate_clifford(self, state[:-1], nqubits)
```
alecandido, line 394: Are you overriding this for the measurement gate?

I would have said that it should have been cut where you can distinguish measurement from everything else.

src/qibo/backends/clifford_operations.py
BrunoLiegiBastonLiegi: I've retried your script with x and z of shape (100, 1000) (which is a sensible dimension for 1000 qubits) and using `int8` in both cases. I still see my version being faster: `~0.06s` vs `~0.15s` for 1000 repetitions.
 
`quickbench.py`

```
# coding: utf-8
import numpy as np
import timeit

def _exponent(x1, z1, x2, z2):
    x1_x2 = x1 & x2
    x1_z2 = x1 & z2
    x2_z1 = x2 & z1
    return (
        -2 * (x1_x2 & z1).astype(np.int8)
        + 2 * (x1_x2 & z2).astype(np.int8)
        + 2 * (x1_z2 & z1).astype(np.int8)
        - x1_z2.astype(np.int8)
        - 2 * (x2_z1 & z2).astype(np.int8)
        + x2_z1.astype(np.int8)
    )
    
e1 = _exponent
def _exponent(x1, z1, x2, z2):
    x1_ = x1.astype(np.int8)
    x2_ = x2.astype(np.int8)
    z1_ = z1.astype(np.int8)
    z2_ = z2.astype(np.int8)
    return 2 * (x1_*x2_ * (z2_ - z1_) + z1_*z2_ * (x1_ - x2_)) - x1_*z2_ + x2_*z1_
    
e2 = _exponent

n = int(1e3)
nrows = int(1e2)
x1 = np.random.randint(2, size=(nrows,n), dtype=bool)
z1 = np.random.randint(2, size=(nrows,n), dtype=bool)
x2 = np.random.randint(2, size=(nrows,n), dtype=bool)
z2 = np.random.randint(2, size=(nrows,n), dtype=bool)
print(timeit.timeit("e1(x1, z1, x2, z2)", globals=globals(), number=1000))
print(timeit.timeit("e2(x1, z1, x2, z2)", globals=globals(), number=1000))
```

src/qibo/backends/numpy.py
BrunoLiegiBastonLiegi: I mean each backend (engine) has it's own definition of the `clifford_operations`. When you create the `CliffordBackend` you also decide which engine to use and then it's up to engine to define how to perform the clifford operations.

src/qibo/backends/numpy.py
BrunoLiegiBastonLiegi: Mmmh but then how do you diversify between the different engines (backends)? Because in the `cupy` backend these are linearizing the matrix.

src/qibo/gates/abstract.py
BrunoLiegiBastonLiegi: Yes the `M` gates overrides this and passes the complete matrix instead.

src/qibo/backends/numpy.py
alecandido: There's a problem in the design of `CliffordBackend`: I didn't notice that `CliffordBackend` is both subclassing `NumpyBackend` and setting a backend as an engine.

So, in the case of CuPy, you would have a `CliffordBackend` with an engine being `CupyBackend`, and `clifford_operations_gpu` as operations, but still subclassing `NumpyBackend` directly.

I believe it is too much tangled, but possibly error-prone.

alecandido: However, here the main issue is isolation: having Clifford related functions where Clifford is potentially not involved is anything but modular.

E.g. the call to `_clifford_pre_execution_reshape` (comment below) is 
https://github.com/qiboteam/qibo/blob/9d15645e6fe3a3629bedd0ab3d95541f71d75e9a/src/qibo/backends/clifford.py#L158
that is delegating to
https://github.com/qiboteam/qibo/blob/9d15645e6fe3a3629bedd0ab3d95541f71d75e9a/src/qibo/backends/clifford.py#L97

But in principle you could directly invoke a function from the suitable operations module, or something like.

alecandido: In this case in particular, you're setting the operations in the engine, and the engine in `CliffordBackend`.

I know the engine is useful, since you could use to cast and things like that. Though, you're getting the same through subclassing.
But in case you still want to have the engine around as you set the engine you could also set the operations, and you could just expose a variable with a tuple `(engine, operations_module)`, that allows to avoid engine modifications and still set both in a single shot.
Or instead revert the dependency, and put an `ENGINE` constant in the operations modules, intended as default engine in case those operations are chosen.

src/qibo/backends/clifford_operations.py
alecandido: I obtained similar numbers, but actually swapped wrt to what you're quoting. I changed once more the script to make the result clearer, here is the report:

```
Benchmarking '6 casts - late'
        mean: 1.2002543163718655        std: 0.015500483036567113
Benchmarking '4 casts - eager'
        mean: 0.8415164417820051        std: 0.01536581466982159

The best is '4 casts - eager'
        confidence: 16.436280886102384sigma
```
[quickbench.py.zip](https://github.com/qiboteam/qibo/files/14221920/quickbench.py.zip)

src/qibo/backends/numpy.py
BrunoLiegiBastonLiegi: Although I agree that inheriting from the numpy backend and having a different engine inside is not the cleanest solution. I am not sure I agree that having `clifford` related functions in non clifford backends is a problem. I mean in the end the `CliffordSimulator` is just an alternative to state vector simulation, and you still have all the ingredients to perform density matrix and state vector evolution in the backends. 
Anyway, in general I think that this is a bit out of the scope of this PR, whose sole aim was accommodating for the updates needed by the  PR in `qibojit`.

src/qibo/backends/numpy.py
alecandido: We're already discussing the issue in the former comment. I'm closing this to reduce noise.

src/qibo/backends/_clifford_operations.py
```diff
@@ -0,0 +1,450 @@
...
+def FSWAP(symplectic_matrix, control_q, target_q, nqubits):
+    """Decomposition --> X-CNOT-RY-CNOT-RY-CNOT-CNOT-X"""
+    symplectic_matrix = X(symplectic_matrix, target_q, nqubits)
+    symplectic_matrix = CNOT(symplectic_matrix, control_q, target_q, nqubits)
+    symplectic_matrix = RY(symplectic_matrix, control_q, nqubits, np.pi / 2)
+    symplectic_matrix = CNOT(symplectic_matrix, target_q, control_q, nqubits)
+    symplectic_matrix = RY(symplectic_matrix, control_q, nqubits, -np.pi / 2)
+    symplectic_matrix = CNOT(symplectic_matrix, target_q, control_q, nqubits)
+    symplectic_matrix = CNOT(symplectic_matrix, control_q, target_q, nqubits)
+    return X(symplectic_matrix, control_q, nqubits)
```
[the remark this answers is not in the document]
alecandido: See comment below.

src/qibo/backends/_clifford_operations.py
```diff
@@ -0,0 +1,445 @@
...
+    z_control_q = xtq ^ zcq
+    z_target_q = ztq_xor_xcq
+    symplectic_matrix[:, ind_zc] = z_control_q
+    symplectic_matrix[:, ind_zt] = z_target_q
+    return symplectic_matrix
+
+
+def S(symplectic_matrix, q, nqubits):
+    r, x, z = _get_rxz(symplectic_matrix, nqubits)
+    symplectic_matrix[:, -1] = r ^ (x[:, q] & z[:, q])
```
alecandido, line 61: Btw, in cases like this one, you already have `r` around. Consider that the `r` defined above is not a copy of a slice of the symplectic matrix, but a sliced view on it.
```suggestion
    r[:] = r ^ (x[:, q] & z[:, q])
```

(essentially, you could this for as long as you are using `__setitem__`, while it would not work if you're just redefining the variable, that would be instead overwritten)

src/qibo/backends/_clifford_operations.py
```diff
@@ -0,0 +1,445 @@
...
+    symplectic_matrix[:, nqubits + q] = z[:, q] ^ x[:, q]
```
alecandido, line 62: Same here, this should be:
```suggestion
    z[:, q] = z[:, q] ^ x[:, q]
```
And this should also work:
```suggestion
    z[:, q] ^= x[:, q]
```

alecandido (commented):
Further simplifications to the operations

src/qibo/backends/_clifford_operations.py
alecandido: @BrunoLiegiBastonLiegi this we decided to postpone anyhow, so you could open an issue and resolve the conversation.

src/qibo/backends/numpy.py
BrunoLiegiBastonLiegi: But in main it is already like that. The `CliffordBackend` inherits from the `NumpyBackend` this PR is not changing anything in that sense.

src/qibo/backends/clifford_operations.py
BrunoLiegiBastonLiegi: Weird. I get exactly the opposite on my laptop.
```
$ python quickbench.py 
Benchmarking '6 casts - late'
	mean: 0.4699267947999942	std: 0.0028983781006549974
Benchmarking '4 casts - eager'
	mean: 1.423648849600022	std: 0.012184825263011127

The best is '6 casts - late'
	confidence: 76.14670215152795sigma
```
But yeah, on the cluster I get the same as you.

BrunoLiegiBastonLiegi: I was also thinking that performance wise, maybe we should switch directly to `np.uint8` as the boolean operations are going to be much faster, at least with numpy. Do you know how cupy handles boolean?

src/qibo/backends/numpy.py
BrunoLiegiBastonLiegi: I believe that this is a similar problem to the one we encountered in the cloud backends PR. Indeed, those two inherit from the `NumpyBackend` already, but ideally they wouldn't need to. Therefore, I was thinking, that this might be a good moment to start thinking about having a more comprehensive abstract backend which will take care of the more upper level functions, like for instance: samples generation, probabilities and frequencies calculation.

renatomello: Some of those backend functions could be moved to `quantum_info` as suggested [here](https://github.com/qiboteam/qibo/issues/1168)

alecandido: > Some of those backend functions could be moved to `quantum_info` as suggested [here](https://github.com/qiboteam/qibo/issues/1168)

@renatomello, the only problem of moving functions out of the backends is that you will then have a single implementation (having multiple implementations is the raison d'être of the backends' mechanism).
However, if we have functions that are not performance-intensive, and which do not require optimized implementations, they certainly do not belong to the backends in the first place.

BrunoLiegiBastonLiegi: Then I am getting a bit lost. So you are saying that the problem is the presence of the various clifford related functions in the backends? Like for example the clifford reshape and the `clifford_operations` attribute. But then, I did not get what is your proposed solution.

alecandido: The problem is the presence of these functions (and invocations) in the `NumpyBackend`.

The solution is to make the `CliffordBackend` directly aware of the `clifford_` functions to be invoked, not passing through the `engine`.
The reason you're passing through the engine is that you have multiple implementations of these functions, and one single `CliffordBackend`. But if the Clifford backend is aware of the Clifford operations module separately from the backend, then you don't need any longer to pass through it.

```py
class CliffordBackend:
    def __init__(self, ..., engine, operations, ...):
        self.engine = Engine
        self.clifford_operations = operations  # not `engine.clifford_operations`

    def _clifford_pre_execution_reshape(self, ...):
        return self.operations._clifford_pre_execution_reshape(...)
        # not
        # ```
        # return self.engine._clifford_pre_execution_reshape(...)
        # ```
```
and if you want to avoid passing both `engine` and `operations` in the `__init__`, then your options are:
- define `ENGINE` in `operations` and use
    ```py
    self.engine = operations.ENGINE
    ```
    (currently is the opposite)
- or define suitable tuples `(engine, operations)` such that the user can just pass the pair defined somewhere in Qibo as a constant (or use the conventional name of that pair)

renatomello: By definition, it's not for the used to choose the operators. The user chooses the `engine` (aka one of the other backends) and the `engine` deals with the rest. And the whole point is to have the `CliffordBackend` to be compatible with all the other engines (aka backends) available in `qibo`. The clifford operations being an input to the user is a no-no.

BrunoLiegiBastonLiegi: I agree, the operations should not be an input the user has to specify. From my perspective the user should pass the backend they want to use as engine to the `CliffordBackend`, that's it. Therefore, the only other alternative, I guess, is to check which backend was passed as engine and import the correct operations accordingly, which I honestly don't like...
I don't know, to me, the original solutions seems to be the more natural one.

alecandido: > By definition, it's not for the used to choose the operators. The user chooses the `engine` (aka one of the other backends) and the `engine` deals with the rest. And the whole point is to have the `CliffordBackend` to be compatible with all the other engines (aka backends) available in `qibo`. The clifford operations being an input to the user is a no-no.

I'd say that *all the other engines* is not a great design choice, because:
- `CliffordBackend` is a backend itself, and it should not be in any way special (trivially, requesting that is just compatible with all the engines would make it recursive)
- there are (and will be) backends that are not sensible (or simple) for `CliffordBackend` to use as engine, e.g. the TensorFlow backend (unless there is a sensible way to differentiate it), or the tensor networks ones (since they will be tailored for Circuits, not for generic array operations)

I would treat the Clifford engines as a thing on their own. The fact that internally you will reuse backends is an implementation detail, and indeed you are not reusing the backend gates (which are reimplemented by the operations) but just some array operations.

In particular, from what I can see you're using only `.cast()`, and `.np`
https://github.com/qiboteam/qibo/blob/b3e839ef9c8433341999d6ad6bdc068aa07fbdf6/src/qibo/backends/clifford.py#L61
https://github.com/qiboteam/qibo/blob/b3e839ef9c8433341999d6ad6bdc068aa07fbdf6/src/qibo/backends/clifford.py#L260
https://github.com/qiboteam/qibo/blob/b3e839ef9c8433341999d6ad6bdc068aa07fbdf6/src/qibo/backends/clifford.py#L64
(and the second should be `self.cast()`, otherwise is worthless defining it in the first place...)
(and the third should be `self.np`, for the same reason)

EDIT: I forgot to inspect the `Clifford` in `quantum_info` as well, there is also a `self.identity_density_matrix`
https://github.com/qiboteam/qibo/blob/b3e839ef9c8433341999d6ad6bdc068aa07fbdf6/src/qibo/quantum_info/clifford.py#L411-L413

alecandido: I'm not saying you should not use an engine (at least not now), nor that you should change everything in this PR.

But if this PR arrives to the point in which it is unavoidable to modify the engine for Clifford, then I'd suggest taking the time to minimally rearrange the structure, and even just ask the user for an `engine` string (`"numpy"`, `"numba"`, `"cupy"`) irrespectively of the fact that they are also available as Qibo backends (keep this internal).

alecandido:
> I tried to refactor the `CliffordBackend` to avoid having backends as engine, but this broke `cupy` and I am not sure that `numba` is working correctly anymore. This is going to take some time....

Oh no!

Could you report the error?

src/qibo/backends/clifford.py
```diff
@@ -589,8 +105,31 @@ def zero_state(self, nqubits: int):
...
+        """Reshape the symplectic matrix to the shape needed by the engine after circuit execution.
+
+        Args:
+            state (ndarray): The input state.
+            nqubits (int): Number of qubits.
+
+        Returns:
+            (ndarray): The reshaped state.
+        """
+        return self.engine._clifford_post_execution_reshape(state, nqubits)
```
alecandido, line 129: In principle, it's just a function consisting of a single function call. You could simplify and inline the inner function call in the execution (these functions should be called in a single place, if I'm not wrong)

src/qibo/backends/numpy.py
```diff
@@ -638,7 +638,7 @@ def samples_to_binary(self, samples, nqubits):
     def samples_to_decimal(self, samples, nqubits):
         qrange = self.np.arange(nqubits - 1, -1, -1, dtype="int32")
         qrange = (2**qrange)[:, self.np.newaxis]
-        return self.np.matmul(self.to_numpy(samples), qrange)[:, 0]
+        return self.np.matmul(self.np.vstack(self.to_numpy(samples)), qrange)[:, 0]
```
alecandido, line 641: Just to understand, is there a reason for this further `vstack`?

src/qibo/backends/clifford.py
```diff
@@ -540,19 +37,38 @@ def __init__(self, engine=None):
 
         from qibo.backends import _check_backend
 
-        engine = _check_backend(engine)
+        if engine is None:
+            engine = _check_backend(engine)
+            engine = engine.name if engine.platform is None else engine.platform
```
alecandido, line 42: Perfectly fine, but you could define directly in this module the `get_engine` function, and use that.

(you could even just call `engine()`, or if it sounds too ambiguous to you, then `resolve_engine()` - `get_` adds very little, but I might be too keen on this)

src/qibo/backends/clifford.py
alecandido: Ok, the first one is actually used twice...

src/qibo/backends/_clifford_operations.py
```diff
@@ -0,0 +1,471 @@
+from functools import reduce
+
+import numpy as np
+from scipy import sparse
+
+name = "numpy"
+
+np = np
```
alecandido, line 8: This is not needed :)

In this scope, `np` is already bound to the `numpy` module, by the `import numpy as np` statement ;)

tests/test_quantum_info_clifford.py
```diff
@@ -17,32 +17,32 @@
...
-            )
+        with pytest.raises(NotImplementedError):
+            clifford_backend = CliffordBackend(backend.name)
     else:
+        backend = backend.name if backend.platform is None else backend.platform
         return CliffordBackend(backend)
 
 
+def get_engine(backend):
+    return backend.platform if backend.platform is not None else backend.name
```
alecandido, line 28: As said above, this function is sensible, and it could be part of the main package itself (it is also used there)

tests/test_quantum_info_clifford.py
```diff
@@ -17,32 +17,32 @@
...
-        with pytest.raises(NotImplementedError) as excinfo:
-            clifford_backend = CliffordBackend(backend)
-            assert (
-                str(excinfo.value)
-                == "TensorflowBackend for Clifford Simulation is not supported yet."
-            )
+        with pytest.raises(NotImplementedError):
+            clifford_backend = CliffordBackend(backend.name)
     else:
+        backend = backend.name if backend.platform is None else backend.platform
```
alecandido, line 23: This is another instance of `get_engine()` (with a compensating negation and swap)

tests/test_quantum_info_clifford.py
```diff
@@ -56,7 +56,7 @@ def test_clifford_from_circuit(backend, measurement):
         c.add(gates.M(*np.random.choice(3, size=2, replace=False)))
 
     result = clifford_backend.execute_circuit(c)
-    obj = Clifford.from_circuit(c, engine=backend)
+    obj = Clifford.from_circuit(c, engine=get_engine(backend))
```
alecandido, line 59: Not urgent, but you could also parametrize on something else than `backend` (i.e. put an `engine` argument, or `clifford_engine`, already in the test's parameters)

alecandido:
I added a bunch of tiny comments, but I reviewed the content of the PR, and, to me, it's pretty good to go.

I don't see why it should not work with CuPy, but maybe the problem is on the other side.

src/qibo/backends/clifford.py
```diff
@@ -671,13 +212,13 @@ def execute_circuit_repeated(self, circuit, initial_state=None, nshots: int = 10
...
         Returns:
-            (:class:`qibo.quantum_info.clifford.Clifford`): Object giving access to the final results.
+            :class:`qibo.quantum_info.clifford.Clifford`: Object storing to the final results.
         """
         circuit_copy = circuit.copy()
         samples = []
         for _ in range(nshots):
             res = self.execute_circuit(circuit_copy, initial_state, nshots=1)
-            [m.result.reset() for m in circuit_copy.measurements]
+            [measurement.result.reset() for measurement in circuit_copy.measurements]
```
renatomello, line 221: Is this line needed? If so, why is it not assigned to anything

alecandido: If I understand correctly, it should be a one-line way to write:
```suggestion
            for measurement in circuit_copy.measurements:
                measurement.result.reset()
```
since `.reset()` has no output, but just modifying the result state.
https://github.com/qiboteam/qibo/blob/428265540de7b5d03de8877eab1f945b9119fe59/src/qibo/measurements.py#L130-L133

However, the two-lines version is clearer, and not much more verbose. I'd suggest that.

alecandido: Btw: in this case you're also creating a list, so allocating memory for it (that will be negligible, and you hope the garbage collector to get rid of it soon). For other cases, consider using generators
```suggestion
            (measurement.result.reset() for measurement in circuit_copy.measurements)
```

But in this case it would not work, because if the generator is not looped, the internal loop does not execute (that's its purpose, you make the loop lazy to avoid memory allocation, required for a loop happening ahead of time).

BrunoLiegiBastonLiegi:
> Oh no!
> 
> Could you report the error?

The problem appears to be that the operations that are calling other operations, like for example `RX` that calls `SX`, call the original `qibo` defined operations rather than the ones overwritten by `qibojit`. With `cupy` this results in an error due to the different shape, but probably `numba` is experiencing the same but not manifesting any error.

src/qibo/backends/clifford.py
BrunoLiegiBastonLiegi: Yes, this resets the measurements for repeated execution. Otherwise, you would generate the samples the first time only and keep using them over and over instead of regenerating them. They are not assigned to anything because there's nothing to be assigned.

src/qibo/backends/numpy.py
BrunoLiegiBastonLiegi: I remember this was failing otherwise, probably with `cupy`.

alecandido:
> The problem appears to be that the operations that are calling other operations, like for example `RX` that calls `SX`, call the original `qibo` defined operations rather than the ones overwritten by `qibojit`. With `cupy` this results in an error due to the different shape, but probably `numba` is experiencing the same but not manifesting any error.

This is quite weird, since functions are handled at the Python level, so there is no compiler magic involved.
In this case, I'd suggest you to make a further module, e.g. copying Numba, and replacing all the gates (all those you're using, so even one or two to start), with things like:

```py
def H(symplectic_matrix, q, nqubits):
    print("H")
    return symplectic_matrix

def CNOT(symplectic_matrix, control_q, target_q, nqubits):
    print("CNOT")
    return symplectic_matrix

...
```
to check that the assignment and dispatching is happening correctly.

src/qibo/backends/clifford.py
renatomello: Then it's the two-line loop that Alessandro suggested is preferred. The suggestion is already committed.

BrunoLiegiBastonLiegi:
> This is quite weird, since functions are handled at the Python level, so there is no compiler magic involved. In this case, I'd suggest you to make a further module, e.g. copying Numba, and replacing all the gates (all those you're using, so even one or two to start), with things like:
> 
> ```python
> def H(symplectic_matrix, q, nqubits):
>     print("H")
>     return symplectic_matrix
> 
> def CNOT(symplectic_matrix, control_q, target_q, nqubits):
>     print("CNOT")
>     return symplectic_matrix
> 
> ...
> ```
> 
> to check that the assignment and dispatching is happening correctly.

I am checking with the debugger and the `self.engine` correctly contains the `RX` defined in `qibo/clifford_operations` and the `SX` of `qibojit/clifford_operations_gpu`. However, when I move inside the `RX` the original `SX` of `qibo` is called. I also wonder whether this has some relation with the errors that are raised by `pylint` now...

src/qibo/backends/clifford.py
```diff
@@ -39,12 +40,9 @@ def __init__(self, engine=None):
             engine = _check_backend(engine)
             engine = engine.name if engine.platform is None else engine.platform
 
-        class CliffordEngine:
-            pass
-
-        self.engine = CliffordEngine()
-        for method in dir(_clifford_operations):
-            setattr(self.engine, method, getattr(_clifford_operations, method))
+        spec = find_spec("qibo.backends._clifford_operations")
```
alecandido, line 43: Why did you decide for the existing module?

BrunoLiegiBastonLiegi: I don't know it looked simpler to me and  I don't think there is any difference with the empty module.

alecandido: The differences are that:
- it is non-empty, that is an advantage
- it could be registered in `sys.modules`, or somehow hooked to the `import` mechanism
    - if that's the case, you would be permanently modifying that module, loosing access to the original content (same of a static import)

alecandido: Anyhow, I see it as a temporary solution. If point 2. does not apply (you could modify that module, without modifying the module imported statically), then it's alright.

alecandido:
@BrunoLiegiBastonLiegi you could have a user-wide configuration to ignore some patterns in all your repositories. It is formatted exactly like a regular `.gitignore` file, and you could put it at `$XDG_CONFIG_HOME/git/ignore`.

> Patterns which a user wants Git to ignore in all situations (e.g., backup or temporary files generated by the user’s editor of choice) generally go into a file specified by `core.excludesFile` in the user’s `~/.gitconfig`. Its default value is `$XDG_CONFIG_HOME/git/ignore`. If `$XDG_CONFIG_HOME` is either not set or empty, `$HOME/.config/git/ignore` is used instead.

https://git-scm.com/docs/gitignore

alecandido (approved):

alecandido:
@BrunoLiegiBastonLiegi, please, it would be nice if you could review the conversations still open, closing those that have been addressed, and raising issues in case there is any postponed one.

alecandido:
@BrunoLiegiBastonLiegi you ignored the wrong lines in 0a599d9.

The conditions in the `if` are evaluated, but it never enters, so the block is not executed.

Actually: do we need those lines at all? Couldn't you just delete them?

src/qibo/backends/numpy.py
BrunoLiegiBastonLiegi: Ok I double checked this and, indeed, it's not necessary, I honestly don't remember when I was getting the failure...

BrunoLiegiBastonLiegi:
> @BrunoLiegiBastonLiegi you ignored the wrong lines in [0a599d9](https://github.com/qiboteam/qibo/commit/0a599d99a5f07f924694a99cb57325a97eb2d0b1).
> 
> The conditions in the `if` are evaluated, but it never enters, so the block is not executed.
> 
> Actually: do we need those lines at all? Couldn't you just delete them?

I think that all the lines inside the block are ignored. Yes, I could remove them, I just don't know whether they might be useful in the future.

BrunoLiegiBastonLiegi:
Before merging this one, we should remember to merge the corresponding `qibojit` PR, release `qibojit` and update the `pyproject.toml` here to point at `qibojit` main, I believe. Right?

scarrazza:
I think we should merge the qibojit PR to main, update this PR to point to main and then release qibo and qibojit just after.

alecandido:
Formally, it's `qibojit` that depends on `qibo`:
https://github.com/qiboteam/qibojit/blob/ce537c898e7d1a98329eaaaf359db01ccb60499d/pyproject.toml#L25
(though it's quasi-circular, since `qibo` depends on `qibojit` for the tests)

So, you should release `qibo` first, then `qibojit` (if you want to have the released versions to work with each other).

Concerning how to merge branches, it could be natural to do in the same way of releases, but I don't really care.

BrunoLiegiBastonLiegi:
Ok then @renatomello  @alecandido, I added to #1128 the discussions that are still open here and in the `qibojit` PR. If everything is ready and you agree, please approve the `qibojit` PR such that we merge that one, I update the `pyproject.toml` here and then merge this one.

Issue #586, opened 2022-05-18
Marginal expectation values from counts


renatomello:
As I discussed with @stavros11 in person, Qiskit has a [built-in function](https://qiskit.org/documentation/stubs/qiskit.ignis.verification.expectation_counts.html) that returns all marginal expectation values from commuting observables of a fixed measurement set. A function like this is useful for several applications, e.g. evaluating several terms of a Hamiltonian with just one measurement setup.

renatomello:
I'm not sure if this is relevant anymore, so I'm closing it.

PR #1153, opened 2024-01-09
Return real expectation values in `Hamiltonian.expectation_from_samples`

MatteoRobbiati:
A very small fix (as the title says).

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/hamiltonians/hamiltonians.py
+++ b/src/qibo/hamiltonians/hamiltonians.py
@@ -155,7 +155,7 @@ class Hamiltonian(AbstractHamiltonian):
             for i in qubit_map:
                 index += int(k[qubit_map.index(i)]) * 2 ** (size - 1 - i)
             expval += obs[index, index] * counts[j]
-        return expval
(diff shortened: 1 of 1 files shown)


renatomello (approved):

renatomello:
@MatteoRobbiati let me use this PR to ask you if this feature is related to this old issue https://github.com/qiboteam/qibo/issues/586 and if the issue should be closed

MatteoRobbiati:
> @MatteoRobbiati let me use this PR to ask you if this feature is related to this old issue #586 and if the issue should be closed

I think that issue was related to a specific feature request (compute more expv with one set of frequencies?).

BrunoLiegiBastonLiegi:
Thanks @MatteoRobbiati, my only concern is that discarding the imaginary part there might hide possible bugs. Maybe adding a check that the imaginary part is actually zero (or small enough) would be safer, although it will add some overhead...

BrunoLiegiBastonLiegi (approved):

PR #1154, opened 2024-01-10
use diagonal matrix with group commutator

Edoardo-Pedicillo:
Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/models/dbi/double_bracket.py
+++ b/src/qibo/models/dbi/double_bracket.py
@@ -74,7 +74,7 @@ class DoubleBracketIteration:
             )
         elif mode is DoubleBracketGeneratorType.group_commutator:
             if d is None:
-                raise_error(ValueError, f"Cannot use group_commutator with matrix {d}")
+                d = self.diagonal_h_matrix
             operator = (
                 self.h.exp(-step)
                 @ self.backend.calculate_matrix_exp(-step, d)


andrea-pasquale (approved):

MatteoRobbiati (approved):
Thanks.
The tests errors are due to the `dbf_migrate` branch, right?

andrea-pasquale:
> Thanks. The tests errors are due to the `dbf_migrate` branch, right?

Not really, I think that it is better if we fix them in this PR. @Edoardo-Pedicillo feel free to do it whenever you have time otherwise I can do it.

Sam-XiaoyueLi:
@MatteoRobbiati  Hi Matteo, I think this PR broke `test_models_dbi.py` by removing the line `raise_error(ValueError, f"Cannot use group_commutator with matrix {d}")`. I'm not sure if we should update the test to reflect the changes or should I try to fix it in my branch (or this?). Please advise. Thanks!

/cc @Edoardo-Pedicillo  @andrea-pasquale

MatteoRobbiati:
@Sam-XiaoyueLi you are right, because now we don't need to raise an error, since we have a default matrix `d`. I guess the solution is to remove the lines:

```python
        dbf(mode=DoubleBracketGeneratorType.single_commutator, step=0.01)
```

from the tests. 

What do you think @Edoardo-Pedicillo, @andrea-pasquale?

marekgluza:
> @Sam-XiaoyueLi you are right, because now we don't need to raise an error, since we have a default matrix `d`. I guess the solution is to remove the lines:
> 
> ```python
>     with pytest.raises(ValueError):
>         dbf(mode=DoubleBracketGeneratorType.single_commutator, step=0.01)
> ```
> 
> from the tests.
> 
> What do you think @Edoardo-Pedicillo, @andrea-pasquale?

@Sam-XiaoyueLi One comment, maybe others can help to understand the dependancies: I believe `d` is a numpy array, shouldn't it be `Hamiltonian`? That way other backends eg hardware won't have an issue?

For now DBI is not deployable because we don't unfold so it's a classical simulator model but I'm preparing to do the transpiler into Hamiltonian simulation and then we will have to switch. 

Possible solution: add `transpiled_double_bracket.py` which inherits and works with appropriate evolution oracles which here would derive from `Circuit`

andrea-pasquale:
> @Sam-XiaoyueLi you are right, because now we don't need to raise an error, since we have a default matrix `d`. I guess the solution is to remove the lines:
> 
> ```python
>     with pytest.raises(ValueError):
>         dbf(mode=DoubleBracketGeneratorType.single_commutator, step=0.01)
> ```
> 
> from the tests.
> 
> What do you think @Edoardo-Pedicillo, @andrea-pasquale?

Hi all, indeed if we remove the lines that @MatteoRobbiati pointed out tests should pass. 
Basically this PR makes the `d` argument optional for `DoubleBracketGeneratorType.single_commutator` and if not provided it will use the diagonal part of the current Hamiltonian.
The fix needs to be performed in your branch @Sam-XiaoyueLi, I can take care of it.

PR #1156, opened 2024-01-16
Split star transpiler into placer and router pass

Simone-Bordoni:
This update allows to uniform star transpiler with all the other transpiler steps. This should simplify the transpiling pipeline and make the code manteinance easier.

Checklist:
- [x] Tests are passing.


src/qibo/transpiler/router.py
```diff
@@ -40,6 +40,108 @@ def assert_connectivity(connectivity: nx.Graph, circuit: Circuit):
...
+    for next_gate in queue:
+        if len(next_gate.qubits) == 2:
+            possible_qubits &= {hardware_qubits.index(q) for q in next_gate.qubits}
+            if not possible_qubits:
+                # freedom of choice
+                return qubits[0]
+            elif len(possible_qubits) == 1:
+                return possible_qubits.pop()
+    # freedom of choice
+    return qubits[0]
```
renatomello, line 142: This same function here is outside of the class but in the placer is a method of the class. Any particular reason why?

Simone-Bordoni: Thank you for noticing this. I have left just the function outside the class in the router and used it for the StarConnectivityPlacer

renatomello (approved):

src/qibo/transpiler/router.py
```diff
@@ -40,6 +40,115 @@ def assert_connectivity(connectivity: nx.Graph, circuit: Circuit):
...
+    by adding SWAP gates when needed.
+
+    Args:
+        connectivity (:class:`networkx.Graph`): chip connectivity, not used for this transpiler.
+        middle_qubit (int, optional): qubit id of the qubit that is in the middle of the star.
+    """
+
+    def __init__(self, connectivity=None, middle_qubit: int = 2):
+        self.middle_qubit = middle_qubit
+        self.connectivity = connectivity
```
BrunoLiegiBastonLiegi, line 63: same as above, why do we need to keep this?

BrunoLiegiBastonLiegi (approved):

PR #1163, opened 2024-01-23
Refactoring `random_statevector`

renatomello:
Now it is fixed to only sample from the Haar measure, with a simpler and faster code

This should be merged after #1171 

Checklist:
- [x] Tests are passing.

### DIFF
--- a/tests/test_quantum_info_random.py
+++ b/tests/test_quantum_info_random.py
@@ -238,8 +238,7 @@ def test_random_statevector(backend, seed):
     # tests if random statevector is a pure state
     dims = 4
     state = random_statevector(dims, seed=seed, backend=backend)
-    backend.assert_allclose(purity(state) <= 1.0 + PRECISION_TOL, True)
-    backend.assert_allclose(purity(state) >= 1.0 - PRECISION_TOL, True)
+    backend.assert_allclose(abs(purity(state) - 1.0) < PRECISION_TOL, True)
 
 
 @pytest.mark.parametrize("normalize", [False, True])


src/qibo/quantum_info/random_ensembles.py
```diff
@@ -426,25 +419,13 @@ def random_statevector(dims: int, haar: bool = False, seed=None, backend=None):
...
-        state += 2 * local_state.random(dims) - 1
-        state /= np.linalg.norm(state)
-        state = backend.cast(state, dtype=state.dtype)
-    else:
-        # select a random column of a haar random unitary
-        k = local_state.integers(low=0, high=dims)
-        state = random_unitary(dims, measure="haar", seed=seed, backend=backend)[:, k]
+    state = local_state.standard_normal(dims).astype(complex)
+    state += 1.0j * local_state.standard_normal(dims)
+    state /= np.linalg.norm(state)
```
AlejandroSopena, line 426:
```suggestion
    state /= np.linalg.norm(state)
    state =  backend.cast(state, dtype=state.dtype)
```

src/qibo/quantum_info/random_ensembles.py
```diff
@@ -426,25 +419,13 @@ def random_statevector(dims: int, haar: bool = False, seed=None, backend=None):
...
 
-    return state
+    return backend.cast(state, dtype=state.dtype)
```
AlejandroSopena, line 428:
```suggestion
    return state
```

src/qibo/quantum_info/random_ensembles.py
AlejandroSopena: This is not important but in the rest of the functions of the module, a variable `x = backend.cast()` is defined and then `x` is returned. This way we are consistent.

AlejandroSopena (approved):

BrunoLiegiBastonLiegi (approved):

PR #1166, opened 2024-01-24
Fix test failing and remove unnecessary prints from tests

renatomello:
Checklist:
- [x] Tests are passing.

### DIFF
--- a/tests/test_backends_clifford.py
+++ b/tests/test_backends_clifford.py
@@ -136,7 +136,7 @@ def test_random_clifford_circuit(backend, prob_qubits, binary):
     clifford_bkd = construct_clifford_backend(backend)
     if not clifford_bkd:
         return
-    c = random_clifford(5, backend=backend)
+    c = random_clifford(5, seed=1, backend=backend)
     c.density_matrix = True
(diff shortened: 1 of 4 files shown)


renatomello:
@BrunoLiegiBastonLiegi this one Clifford test keeps failing randomly because it returns a number slightly above 200. I increased the limit to 250, but could you take a deeper look into what it is exactly being tested to see if there's a way to refactor this test (if necessary)?

Simone-Bordoni:
> @BrunoLiegiBastonLiegi this one Clifford test keeps failing randomly because it returns a number slightly above 200. I increased the limit to 250, but could you take a deeper look into what it is exactly being tested to see if there's a way to refactor this test (if necessary)?

I was also getting a test randomly failing (maybe setting a fixed seed for the RNG would fix the problem)

Simone-Bordoni (approved):

BrunoLiegiBastonLiegi:
Thanks for looking into this, I re-wrote the test of the frequencies in a slightly different way that should solve the instability problem.

BrunoLiegiBastonLiegi (approved):

PR #1183, opened 2024-02-01
Noise Channels with the `CliffordSimulator`

BrunoLiegiBastonLiegi:
This PR adds support for noise channels to the `CliffordSimulator`.

Checklist:
- [ ] Tests are passing.


src/qibo/gates/channels.py
```diff
@@ -39,6 +39,9 @@ def apply(self, backend, state, nqubits):  # pragma: no cover
     def apply_density_matrix(self, backend, state, nqubits):
         return backend.apply_channel_density_matrix(self, state, nqubits)
 
+    def apply_clifford(self, backend, state, nqubits):
+        return backend.apply_channel(self, state, nqubits)
```
renatomello, line 43: This will break if one calls this method in any other scenario that isn't the Clifford backend acting on a symplectic matrix, right?

Maybe we should add a docstring explicitly mentioning that and/or raising an error when `state` isn't a symplectic matrix and/or raising an error when `backend` isn't Clifford

renatomello (commented):
I think this is missing an explicit how-to in the docs of Pauli noise channels and Clifford simulation

renatomello (approved):

AlejandroSopena (approved):

Issue #1181, opened 2024-01-31
Paramether Shift Rule (PSR) test failing on CI


renatomello:
This test randomly fails in the CI (`qibojit` and `numba`)

https://github.com/qiboteam/qibo/actions/runs/7723418367/job/21053473674?pr=1172#step:11:219

@scarrazza @MatteoRobbiati

renatomello:
@scarrazza @MatteoRobbiati

renatomello:
This was clsed by #1185

PR #1185, opened 2024-02-02
Fix `numpy.random` seed when testing Parameter Shift Rule.

MatteoRobbiati:
Fixing #1181. 
Thanks @renatomello for raising the issue. Fixing the random generator seed should prevent such failed tests.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/tests/test_derivative.py
+++ b/tests/test_derivative.py
@@ -33,6 +33,7 @@ def circuit(nqubits=1):
 def test_standard_parameter_shift(backend, nshots, atol, scale_factor, grads):
     # initializing the circuit
     c = circuit(nqubits=1)
+    np.random.seed(42)
 
     # some parameters
     # we know the derivative's values with these params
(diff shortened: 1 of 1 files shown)


renatomello (dismissed):

renatomello:
Does fixing the seed using `numpy` also solves the issue for GPU backends?

tests/test_derivative.py
```diff
@@ -33,6 +33,7 @@ def circuit(nqubits=1):
 def test_standard_parameter_shift(backend, nshots, atol, scale_factor, grads):
     # initializing the circuit
     c = circuit(nqubits=1)
+    np.random.seed(42)
```
renatomello, line 36:
```suggestion
    backend.set_seed(42)
```

Maybe this is more general

tests/test_derivative.py
```diff
@@ -88,6 +89,7 @@ def test_standard_parameter_shift(backend, nshots, atol, scale_factor, grads):
 def test_finite_differences(backend, step_size):
     # initializing the circuit
     c = circuit(nqubits=1)
+    np.random.seed(42)
```
renatomello, line 92:
```suggestion
    backend.set_seed(42)
```

MatteoRobbiati:
> Does fixing the seed using `numpy` also solves the issue for GPU backends?

Oh yes, you are right! Thanks!

renatomello (approved):

BrunoLiegiBastonLiegi (approved):

PR #1188, opened 2024-02-03
Move `gates.Gate.decompose` to `transpiler.decompositions`

### LINKED ISSUE #1040: Move `Gate.decompose()` methods to transpiler
For when we move the transpilers from qibolab to qibo, it would be good to have a unified approach of gate transformations.

_Originally posted by @stavros11 in https://github.com/qiboteam/qibo/pull/1035#pullrequestreview-1670424696_

renatomello:
Closes #1040 

Checklist:
- [x] Tests are passing.


Simone-Bordoni (approved):
Thank you for moving the Gate.decompose to the transpiler decompositions. This is a good start for the next step of having a general decomposer. I plan on starting it next week.

stavros11 (commented):
Thanks @renatomello, I believe this is a step in the right direction since it is reducing code duplication (or at least puts all decompositions in a single place). I have some questions:

1. Do we still the `Gate.decompose` interface? If this is not used anywhere I would consider dropping completely. This way we avoid the ugly pylint exceptions and we don't have to maintain two interfaces for the same thing - we also force the users to use the transpiler interface directly. Personally I am not aware of any applications of `Gate.decompose`.
2. Is there something special about the `standard_decomposition`? What makes it more "standard" than the other decompositions?

Not necessarily for this PR, but in my mind it would be useful have a map or function in which we pass the available native gates and we get back a `GateDecomposition` that decomposes everything to these gates, for example
```py
get_decomposition(NativeGates.GPI2 | NativeGates.CZ) -> cz_dec
```
etc. Then there is a trickier part of whether multiple decompositions can be combined to reach the desired gates, but that we could even leave for a later step.

renatomello:
>     1. Do we still the `Gate.decompose` interface? If this is not used anywhere I would consider dropping completely. This way we avoid the ugly pylint exceptions and we don't have to maintain two interfaces for the same thing - we also force the users to use the transpiler interface directly. Personally I am not aware of any applications of `Gate.decompose`.

I would say yes. I, for one, sometimes want to get a decomposition of one specific gate without being in the context of transpiling an entire circuit. It'd be annoying to have to create a circuit and add the gate and then transpile it just to get the decomposition of one gate.

>     2. Is there something special about the `standard_decomposition`? What makes it more "standard" than the other decompositions?

No. Just like with any "standard", it is, by definition, arbitrary. In this case it just means the ones that were hardcoded before.

stavros11:
> I would say yes. I, for one, sometimes want to get a decomposition of one specific gate without being in the context of transpiling an entire circuit. It'd be annoying to have to create a circuit, add the gate, and then transpile the circuit just to get the decomposition of one gate.

You can already do that using just
```py
from qibo.transpiler.decompositions import standard_decompositions

dec_gate =  standard_decompositions(gate)
```
without having to create any circuits. This is also more flexible because you can use any decomposition in the same fashion, not only the "standard" one.

If the main issue is the names and import, we could propose better names and arrange the modules so that you can simply do
```py
from qibo.transpiler import mydecomposition
```
or
```py
from qibo.decompositions import mydecomposition
```

Simone-Bordoni:
I think that the best way is either to use only the "transpiler method" or to allow in the decompose method to decide the basis gates. However as a first implementation only to store the decompositions in a single place this should be enough. Why don't we discuss the structure of the general unroller in person at the beginning of next week?

renatomello:
> I think that the best way is either to use only the "transpiler method" or to allow in the decompose method to decide the basis gates. However as a first implementation only to store the decompositions in a single place this should be enough. Why don't we discuss the structure of the general unroller in person at the beginning of next week?

I agree with this.

stavros11 (approved):
I still don't see how `Gate.decompose` itself can be useful but okay.

PR #1190, opened 2024-02-05
Add `phase_encoder` to `models.encodings` and improve documentation of unary encoder(s)

renatomello:
Checklist:
- [x] Tests are passing.


Simone-Bordoni (approved):
Thank you for adding this feature, it may be very useful for QML applications. (just fix coverage before merging)

BrunoLiegiBastonLiegi (approved):
Thanks, this is very helpful indeed! I would just make the figures a tad smaller if possible.

Issue #1034, opened 2023-10-10
Double-bracket iterations


marekgluza:
### Context
In https://arxiv.org/abs/2206.11772 I proposed something that I will refer to below as a double-bracket iteration (DBI) quantum algorithms, in particular they can be applied to the task of diagonalization and eigenstate preparation.

My code
https://github.com/marekgluza/double_bracket_flow_as_a_diagonalization_quantum_algorithm/
exemplifies what DBIs can do, in particular it showcases the expressive power of DBI quantum circuits. In this feature request I will motivate why it is realistic to add it to QIBO.

### Short definition of DBI
Let $D_k$ be a sequence of operators diagonal in the computational basis and let $s_k$ be a sequence of numbers which will serve as DBI recursion step durations.
A variational diagonalizing DBI for some input Hamiltonian $H_0$ is a recursion for $k\ge 0$ with the recursion step given by
$$H_{k+1} = e^{s_k W_k} H_k e^{-s_k W_k}\ , $$
where
$$W_k = [D_k, H_k] = D_k H_k - H_k D_k\ .$$

It can be shown that there exist $D_k$'s and $s_k$'s such that the magnitude of the off-diagonal terms in $H_{k+1}$ are smaller than those of $H_k$ as captured by the $2$-norm.
Such a DBI is a sequence of unitary rotations so it is a diagonalization iteration.

### Feature 1: Group commutators
Define the unitary
$$V = e^{i s D_0}e^{i s H_0}e^{-i s D_0}e^{-i s H_0}$$
then if $s=s_0$ we have an approximation to the first DBI recursion step
$$\|| H_1 - V H_0 V^\dagger\|| = O(s^3)$$

Adding a module allowing to work with group commutators will allow to additionally explore compiling, cf Solovay-Kitaev algorithm.

### Feature 2: Unfolding frame-shifting covariant quantum recursions.

Let $\left(Q_k\right)_{k=0,\ldots K}$ by $K$ unitary matrices $U(d)$ of dimension $d$ and define  $U^{(\cdot)}: U(d) \rightarrow U(d)$ which maps
$$U(d) \ni \psi \mapsto U^{(\psi)} = Q_0 \psi Q_1 \psi Q_2 \ldots \psi Q_K\ .$$
This gives rise to quantum recursions which are covariant under frame shifting in the sense that for an initial state
$$|\psi_0\rangle = \psi_0 |0\rangle$$
with some reference state $|0\rangle$ 
[...]

PR #1194, opened 2024-02-06
Added circuit composition via __matmul__ to qibo.Circuit 

marekgluza:
… because circuits are representations of  unitaries so it's also natural to multiply them. 

For example one can imagine a class which has a mode of running things in numpy or more symbolically on qibo.Circuit, with the @ symbol the same code will run irrespective of what is the underlying object.

```
class something:
    def __init__(self, data):
        self.data = data
        
    def __matmul__(self, thing):
        return __class__(self.data @ thing.data)

# @ will operate on Circuit c1
s1 = something(c1)
# @ will operate on the matrix
s2 = something(c1.unitary())

print( (s1 @ s1).data.unitary() - (s2 @ s2).data)
```

This arises from needing to wrap around myself conditioned on the type so for the numpy mode use @ and for Circuit use + 

The PR includes:
- implementation of __matmul__ by wrapping __add__
- test function checking norm equality after some compositions

Some more test code
```
from test_models_circuit import *
test_circuit_matmul()
c = Circuit(2)
c.add(gates.CNOT(0,1))
c1 = c+c
c2 = c @ c
import numpy as np
np.linalg.norm( c1.unitary() - c2.unitary())

print(c1.draw())
print(c2.draw())
print(c1.fuse().draw())
print(c2.fuse().draw())
c1.add(gates.H(1))
c3 = c1 @ c2 @ c1
print(c3.draw())
```

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/models/circuit.py
+++ b/src/qibo/models/circuit.py
@@ -274,6 +274,18 @@ class Circuit:
 
         return newcircuit
 
+    def __matmul__(self, circuit):
+        """Compose circuits.
+        In qibo this is the same as add circuits
+
+        Args:
+            circuit: Circuit to be composed with the current one.
+
+        Returns:
+            The circuit resulting from the composition.
+        """
+        return self + circuit
+
     @property
     def wire_names(self):
         return self._wire_names
--- a/tests/test_models_circuit.py
+++ b/tests/test_models_circuit.py
@@ -93,6 +93,7 @@ def test_circuit_add():
     assert c.ngates == 3
     assert list(c.queue) == [g1, g2, g3]
 
+
 def test_circuit_matmul():
     c = Circuit(2)
     g1, g2, g3 = gates.H(0), gates.H(1), gates.CNOT(0, 1)
@@ -102,7 +103,9 @@ def test_circuit_matmul():
     c1 = c + c
     c2 = c @ c
     import numpy as np
-    assert np.linalg.norm( c1.unitary() - c2.unitary() ) < 1e-10
+
+    assert np.linalg.norm(c1.unitary() - c2.unitary()) < 1e-10
+
 
 def test_circuit_add_errors():
     c = Circuit(2)
--- a/tests/test_models_circuit.py
+++ b/tests/test_models_circuit.py
@@ -93,6 +93,16 @@ def test_circuit_add():
     assert c.ngates == 3
     assert list(c.queue) == [g1, g2, g3]
 
+def test_circuit_matmul():
+    c = Circuit(2)
+    g1, g2, g3 = gates.H(0), gates.H(1), gates.CNOT(0, 1)
+    c.add(g1)
+    c.add(g2)
+    c.add(g3)
+    c1 = c + c
+    c2 = c @ c
+    import numpy as np
+    assert np.linalg.norm( c1.unitary() - c2.unitary() ) < 1e-10
 
 def test_circuit_add_errors():
     c = Circuit(2)


renatomello:
Correct me if I'm wrong, but isn't this what `__add__` already does? And also, if it's to be done on the gate level, isn't this what gate fusion does?

If `@` requires to execute `Circuit.unitary()`, I can see two things happening: 

1) the `unitary` method is both memory and time inefficient (same complexity as circuit execution using density matrices). This makes it kind of unfeasible for a moderate number of qubits (over 20-25).
2) the result of the operation is a big unitary matrix. This would be equivalent of "gate fusing" the entirety of both circuits. This goes against the idea of keeping a queue and doing operations locally to be more efficient. "Executing" the resulting circuit would be just equivalent of doing a very costly matrix multiplication.

marekgluza:
> Correct me if I'm wrong, but isn't this what `__add__` already does?

Yes, I was just surprised that the functionality comes from `__add__` and not `__matmul__`

The PR proposes a wrapper on `__add__` because composition of circuits is more naturally represented as a kind of multiplication than addition:

1.  Numpy) If you cast to unitary then the operations naturally commute 
` (c1 @ c2).unitary()`
consistently uses @ like numpy
` c1.unitary() @ c2.unitary()`
so they look analogous and not confusing like now
` (c1 + c2).unitary()`
where you need to know whether the casting happens before or after

2. tensor networks) If you would run tensor networks then individual layers of a circuit are MPOs. If I have MPO then the code most like will use @ to propagate the circuit forward. If you want to write a code which is agnostic whether c1 is a Circuit or MPO then Circuit should have `__matmul__`

3. backend agnostic codes) In my applications I'm working with a type EvolutionOracle (feature 3 in #1034)  which emulates qm operations so if you evolve a state twice then you query the oracle twice and compose them. Right now I need to use `__add__`
https://github.com/qiboteam/qibo/blob/6f9745f137d03f2674838e5d4bf52c4e920b2b14/src/qibo/models/dbi/group_commutator_iteration_transpiler.py#L116
However. to combine them it is natural to be consistent with what would happen if the EvolutionOracle is numpy exponentiation, then I would use `__matmul__`. Currently if the EvolutionOracle is a qibo.Circuit then suddenly composing two oracle queries need to switch to `__add__` which makes the whole code clumsy

4. geometry) Circuit represents unitary circuits which are sparse elements of the unitary group which is multiplicative.

In my opinion `__add__` is an artefact of the underlying data structure which places gates in sequence. For a temporal sequence adding to a scheduler makes sense. For operating on circuits multiplication is the natural operator.



> If `@` requires to execute `Circuit.unitary()`, I can see two things happening:

The PR implements a wrapper on `__add__` and no casting occurs. The dummy class example I posted above is to show why being consistent can appear in practice. The actual data structure I use is this https://github.com/qiboteam/qibo/blob/6f9745f137d03f2674838e5d4bf52c4e920b2b14/src/qibo/models/dbi/double_bracket_evolution_oracles.py#L25

> I guess at least some sort of warning would have to be raised letting users know what this operation is inefficient in the general case.

There would be no inefficiency. However the warning is that I'm asking to block an operator so please evaluate (that's why I added many reviewers) whether to block another operator:
- Idealistic solution: create backwards incompatibility and compose circuits with @ and not + (I'm not suggesting it)
- Proposed solution: wrapper on + for @ (seems intuitive to me)
- Current solution: there's only + and @ is free for some possible future operation (sure, as you wish)

renatomello:
The `__add__` is not to be understood as a sum, but as a concatenation. This is consistent with other other libraries e.g. [Cirq](https://github.com/quantumlib/Cirq/blob/v1.3.0/cirq-core/cirq/circuits/circuit.py#L1918) and [Qiskit](https://github.com/Qiskit/qiskit/blob/2f95629a3e2cea5e13c64e273f578ab5a3eec645/qiskit/circuit/quantumcircuitdata.py#L116). Also consistent with other libraries is the fact that `@` is not used for any operation. 

Since the `Circuit` object is not a `dataclass` wrapping a matrix (that's the `CircuitResult` object *after* execution), it doesn't make much sense to think of commutatitvity of `+` or `@` because, again, the operation is a concatenation and not an addition.
For these same reasons, it does not make a difference if the underlying execution of circuits is done via statevector simulation, density matrix simulation, Clifford simulation, or tensor networks, because any matrix operation is only performed at the execution level, not when concatenation of two circuits.

I'll let others chime in, but to me it makes `qibo` inconsistent with other libraries while not adding any substantial advantage. Of course this is just my opinion and I might be in the minority here.

MatteoRobbiati:
I personally agree with @renatomello. I see your point @marekgluza but I think in this context of simulation is well known that the computational `+` operation correspond to the `matmul`. For this reason, since `qibo.Circuit` already supports the addition, this extra feature is not really necessary :)

PR #1197, opened 2024-02-07
Rename qubits of transpiled circuit based on connectivity

Simone-Bordoni:
As it is possible to name the wires of a circuit this functionality can be used to give the quabits of the transpiled circuit the same name of the hardware qubits when the connectivity is restricted. I have also solved some minos problems regarding an error message that was being printed during the block decomposition and the unroller not initialising the unrolled circuit with the same kwargs of the original circuit.

Checklist:
- [x] Tests are passing.


src/qibo/transpiler/blocks.py
```diff
@@ -65,8 +65,8 @@ def fuse(self, block, name: Optional[str] = None):
             (:class:`qibo.transpiler.blocks.Block`): fusion of the two input blocks.
         """
         if not self.qubits == block.qubits:
-            raise_error(
-                BlockingError, "In order to fuse two blocks their qubits must coincide."
+            raise BlockingError(
+                "In order to fuse two blocks their qubits must coincide."
```
renatomello, line 69: Should let `raise_error` deal with the exception since it's a `qibo` exception and not a regular `python` exception

src/qibo/transpiler/unroller.py
```diff
@@ -85,7 +85,7 @@ def __call__(self, circuit: Circuit):
         Returns:
             (:class:`qibo.models.circuit.Circuit`): equivalent circuit with native gates.
         """
-        translated_circuit = circuit.__class__(circuit.nqubits)
+        translated_circuit = Circuit(**circuit.init_kwargs)
```
renatomello, line 88: this might be a problem if in the future more `kwargs` are added, no?

src/qibo/transpiler/blocks.py
Simone-Bordoni: I modified this line because if I use the qibo raise_error the error message is printed several times when i call the block decomposition function:
```python
while len(initial_blocks) > 0:
        first_block = initial_blocks[0]
        remove_list = [first_block]
        if len(initial_blocks[1:]) > 0:
            for second_block in initial_blocks[1:]:
                try:
                    first_block = first_block.fuse(second_block)
                    remove_list.append(second_block)
                except BlockingError:
                    if not first_block.commute(second_block):
                        break
        blocks.append(first_block)
        _remove_gates(initial_blocks, remove_list)
``` 
The error message is printed in the try/except. 
Maybe a good solution would be to directly change the raire_error function

src/qibo/transpiler/unroller.py
Simone-Bordoni: I think this is better this way as the translated circuit should keep all the properties of the original circuit like density_matrix, qubit_names ecc...
I all the other steps of the transpiler i create a new circuit passing the init_kwargs

src/qibo/transpiler/blocks.py
renatomello: I think first we should find out why it's being raised multiple times

Simone-Bordoni:
```python
def raise_error(exception, message=None, args=None):
    """Raise exception with logging error.

    Args:
        exception (Exception): python exception.
        message (str): the error message.
    """
    log.error(message)
    if args:
        raise exception(message, args)
    else:
        raise exception(message)
``` 
i think the log.error(message) is the problem

BrunoLiegiBastonLiegi (approved):

PR #1202, opened 2024-02-08
`pytorch` backend

Simone-Bordoni:
Backend based on `pytorch` for optimised automatic differentiation.

Checklist:
- [x] Tests are passing.


Simone-Bordoni:
I need a bit of help to finish this work. 
For now i have created the pytorch backend starting from the tf backend and basically replacing all tf functions/tensors with torch functions/tensors.
Where can I add the tests of the code?
What else do i have to change to make everything work?
Where can i write the documentation of the backend?
I get a pylint error when using torch.linalg.eigvalsh, torch.linalg.eigh, torch.linalg.matrix_exp, torch.sparse.mm (not-callable error) for now i have disabled this error on pylint, however i can't find a similar error and I don't know if there is a way to solve it without disabling the pylint error.

scarrazza (commented):
Thanks @Simone-Bordoni for this. Could you please include pytorch in tests by updating [this file](https://github.com/qiboteam/qibo/blob/45e4e224e3d9183d817bab1458d5cb330f31bf3f/tests/conftest.py#L14) and check coverage?

src/qibo/backends/pytorch.py
```diff
@@ -0,0 +1,312 @@
...
+    # Redefine parametrized gate matrices for backpropagation to work
+
+    def __init__(self, dtype):
+        self.np = np
+        super().__init__(dtype)
+
+    def RX(self, theta):
+        cos = self.np.cos(theta / 2.0) + 0j
+        isin = -1j * self.np.sin(theta / 2.0)
+        return torch.tensor([[cos, isin], [isin, cos]], dtype=self.dtype)
```
renatomello, line 22: I'd like to propose something for both the `torch` and the `tensorflow` backends. Instead of completely rewriting the parametrized gates, we could do something like:
```suggestion
    def RX(self, theta):
        matrix = getattr(NumpyMatrices(self.dtype), "RX")(theta)
        return self.torch.tensor(matrix, dtype=self.dtype)
```

This way we still cast the matrices to the proper tensor to get differentiation working, but also do not have to rewrite every parametrized matrix.

Simone-Bordoni: It makes sense, for now i am implementing it on the pytorch backend

renatomello:
I converted to draft just because it's gonna take a little bit to finish it

Simone-Bordoni:
> I converted this PR to draft just because it's gonna take a little bit to finish it

I thought it would have been an easy thing but there are many parts of qibo that are not backend agnostic and struggle with pytorch.

tests/test_gates_channels.py
```diff
@@ -209,7 +209,7 @@ def test_pauli_noise_channel(backend, pauli_order):
     )
     gate = gates.X(1)
     target_state = backend.apply_gate_density_matrix(gate, np.copy(initial_state), 2)
-    target_state = 0.3 * target_state + 0.7 * initial_state
+    target_state = 0.3 * target_state + 0.7 * backend.cast(initial_state)
```
renatomello, line 212: this cast should not be necessary since `initial_state` is supposedly generated in the correct backend

Simone-Bordoni:
@renatomello I am having problems making the pytorch backend work with hamiltonians, next days if you are available i can show you the issues.

src/qibo/gates/gates.py
```diff
@@ -2329,7 +2329,12 @@ def __init__(
         }
 
         # checking unitarity without invoking any backend
+        # maybe here having the backend would be useful?
         if check_unitary:
+            import torch
```
renatomello, line 2334: this will fail if `torch` is not installed

tests/test_callbacks.py
```diff
@@ -289,6 +289,12 @@ def test_norm(backend, density_matrix, seed):
     backend.assert_allclose(final_norm, target_norm)
 
 
+def to_numpy(x):
+    if isinstance(x, np.ndarray):
+        return x
+    return x.detach().cpu().numpy()
```
renatomello, line 295: Should use the `torch` backend method

src/qibo/derivative.py
```diff
@@ -102,6 +103,12 @@ def circuit(nqubits = 1):
 
     # inheriting hamiltonian's backend
     backend = hamiltonian.backend
+    # TODO: make this work wih pytorch backend
+    if isinstance(backend, PyTorchBackend):
+        raise_error(
+            NotImplementedError,
+            "PyTorchBackend for the parameter shift rule is not supported.",
+        )
```
renatomello, line 111: This #TODO is probably one of the most important things here, so it should be done before merging

tests/test_gates_gates.py
```diff
@@ -68,6 +68,12 @@ def test_z(backend):
     assert gates.Z(0).unitary
 
 
+def to_numpy(array):
+    if isinstance(array, np.ndarray):
+        return array
+    return array.detach().cpu().numpy()
```
renatomello, line 74: Same thing here

tests/test_gates_gates.py
```diff
@@ -94,10 +100,12 @@ def test_sx(backend):
...
-        np.transpose(np.conj(final_state_decompose))
-        @ observable
-        @ final_state_decompose,
-        np.transpose(np.conj(target_state)) @ observable @ target_state,
+        np.transpose(np.conj(to_numpy(final_state_decompose)))
+        @ to_numpy(observable)
+        @ to_numpy(final_state_decompose),
+        np.transpose(np.conj(to_numpy(target_state)))
+        @ to_numpy(observable)
+        @ to_numpy(target_state),
```
renatomello, line 108: method called multiple times on same objects, could run one and save it in variables

tests/test_quantum_info_clifford.py
```diff
@@ -23,15 +23,16 @@ def construct_clifford_backend(backend):
...
-    elif isinstance(backend, PyTorchBackend):
+
+    if isinstance(backend, PyTorchBackend):
         with pytest.raises(NotImplementedError) as excinfo:
             clifford_backend = CliffordBackend(backend)
             assert (
                 str(excinfo.value)
                 == "PyTorchBackend for Clifford Simulation is not supported."
             )
-    else:
```
Simone-Bordoni, line 26: Removing this "else" breaks the tests

renatomello:
@Simone-Bordoni @scarrazza this is now compatible with the `quantum_info` module

src/qibo/backends/pytorch.py
```diff
@@ -0,0 +1,574 @@
...
+    "complex128": torch.complex128,
+}
+
+
+class TorchMatrices(NumpyMatrices):
+    # Redefine parametrized gate matrices for backpropagation to work
+
+    def __init__(self, dtype):
+        super().__init__(dtype)
+        self.torch = torch
```
renatomello, line 32: I think doing `self.np = torch` will lead to the lesser possible amount of code rewriting

src/qibo/backends/pytorch.py
```diff
@@ -0,0 +1,462 @@
...
+            dtype (Union[str, torch.dtype, np.dtype, type]): Target data type.
+                If ``None``, the default dtype of the backend is used.
+                Defaults to ``None``.
+            copy (bool, optional): If ``True``, the input tensor is copied before casting.
+                Defaults to ``False``.
+        """
+        if dtype is None:
+            dtype = self.dtype
+        elif isinstance(dtype, self.np.dtype):
+            dtype = dtype
```
renatomello, line 89: This conditional is not doing anything

src/qibo/backends/numpy.py
```diff
@@ -93,7 +93,7 @@ def identity_density_matrix(self, nqubits, normalize: bool = True):
 
     def plus_state(self, nqubits):
         state = self.np.ones(2**nqubits, dtype=self.dtype)
-        state /= self.np.sqrt(2**nqubits)
+        state /= self.np.sqrt(self.cast(2**nqubits))
```
renatomello, line 96: I'd prefer something like 

```suggestion
        state /= math.sqrt(2**nqubits)
```

src/qibo/backends/numpy.py
```diff
@@ -114,15 +114,15 @@ def matrix_parametrized(self, gate):
 
     def matrix_fused(self, fgate):
         rank = len(fgate.target_qubits)
-        matrix = np.eye(2**rank, dtype=self.dtype)
+        matrix = np.eye(2**rank, dtype=np.complex128)
```
renatomello, line 117: I did not test this out, but my impression is that something like
```python
        matrix = self.np.eye(2**rank, dtype=self.dtype)
```

would work. Or maybe:
```python
        matrix = self.cast(np.eye(2**rank, dtype=complex), dtype=self.dtype)
```

src/qibo/backends/numpy.py
```diff
@@ -114,15 +114,15 @@ def matrix_parametrized(self, gate):
...
         for gate in fgate.gates:
             # transfer gate matrix to numpy as it is more efficient for
             # small tensor calculations
             # explicit to_numpy see https://github.com/qiboteam/qibo/issues/928
             gmatrix = self.to_numpy(gate.matrix(self))
             # Kronecker product with identity is needed to make the
             # original matrix have shape (2**rank x 2**rank)
-            eye = np.eye(2 ** (rank - len(gate.qubits)), dtype=self.dtype)
+            eye = np.eye(2 ** (rank - len(gate.qubits)), dtype=np.complex128)
```
renatomello, line 125: This could probably also be a ```self.np.eye``` instead

src/qibo/backends/numpy.py
```diff
@@ -137,7 +137,7 @@ def matrix_fused(self, fgate):
             gmatrix = np.reshape(gmatrix, original_shape)
             # fuse the individual gate matrix to the total ``FusedGate`` matrix
             matrix = gmatrix @ matrix
-        return matrix
+        return self.cast(matrix)
```
renatomello, line 140: Then you wouldn't need this ```cast```

src/qibo/backends/numpy.py
```diff
@@ -114,30 +115,33 @@ def matrix_parametrized(self, gate):
...
-            gmatrix = np.reshape(gmatrix, original_shape)
+            gmatrix = self.transpose(gmatrix, transpose_indices)
+            gmatrix = self.np.reshape(gmatrix, original_shape)
             # fuse the individual gate matrix to the total ``FusedGate`` matrix
             matrix = gmatrix @ matrix
-        return self.cast(matrix)
+        return matrix
+
+    def transpose(self, matrix, transpose_indices):
+        return self.np.transpose(matrix, transpose_indices)
```
renatomello, line 144: you don't need to create this function.The `torch` equivalent for transposition in `numpy` is `torch.permute`

src/qibo/backends/pytorch.py
```diff
@@ -165,38 +146,11 @@ def samples_to_binary(self, samples, nqubits):
...
-        state = self.np.reshape(state, 2 * nqubits * (2,))
-        state = self.np.reshape(state.permute(*order), shape)
-        probs = self.np.abs(self.np.einsum("abab->a", state))
-        probs = self.np.reshape(probs, len(qubits) * (2,))
-        return self._order_probabilities(probs, qubits, nqubits).view(-1)
+    def dimensions(self, x):
+        return x.dim()
+
+    def ravel(self, x):
+        return x.reshape(-1)
```
renatomello, line 153: Also `torch` has its own `ravel` already

Simone-Bordoni:
I was fixing the errors on unitary_decomposition and I found a strange behaviour of pytorch regarding the function torch.dot.
the problem is this:
```python
    print(torch.dot(torch.tensor([1+0j, 0, 0, 0], dtype=torch.complex128), torch.tensor([1+0j, 0, 0, 0], dtype=torch.complex128)))
    print(torch.dot(torch.tensor([1, 0, 0, 0], dtype=torch.float64), torch.tensor([1, 0, 0, 0], dtype=torch.float64)))
``` 
the first result gives zero while the second correctly gives 1. it seems like torch.dot does not work for complex vectors (i have tried also using torch.complex64 and other variants).
Do you get the same problem?
in case how may I solve this?
@BrunoLiegiBastonLiegi @renatomello

renatomello:
> I was fixing the errors on unitary_decomposition and I found a strange behaviour of pytorch regarding the function torch.dot. the problem is this:
> 
> ```python
>     print(torch.dot(torch.tensor([1+0j, 0, 0, 0], dtype=torch.complex128), torch.tensor([1+0j, 0, 0, 0], dtype=torch.complex128)))
>     print(torch.dot(torch.tensor([1, 0, 0, 0], dtype=torch.float64), torch.tensor([1, 0, 0, 0], dtype=torch.float64)))
> ```
> 
> the first result gives zero while the second correctly gives 1. it seems like torch.dot does not work for complex vectors (i have tried also using torch.complex64 and other variants). Do you get the same problem? in case how may I solve this? @BrunoLiegiBastonLiegi @renatomello

idk if you already solved this, but matrix multiplication in `torch` is `torch.mm`

Simone-Bordoni:
> > I was fixing the errors on unitary_decomposition and I found a strange behaviour of pytorch regarding the function torch.dot. the problem is this:
> > ```python
> >     print(torch.dot(torch.tensor([1+0j, 0, 0, 0], dtype=torch.complex128), torch.tensor([1+0j, 0, 0, 0], dtype=torch.complex128)))
> >     print(torch.dot(torch.tensor([1, 0, 0, 0], dtype=torch.float64), torch.tensor([1, 0, 0, 0], dtype=torch.float64)))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > the first result gives zero while the second correctly gives 1. it seems like torch.dot does not work for complex vectors (i have tried also using torch.complex64 and other variants). Do you get the same problem? in case how may I solve this? @BrunoLiegiBastonLiegi @renatomello
> 
> idk if you already solved this, but matrix multiplication in `torch` is `torch.mm`

I solved the problem using torch.sum and torch.mul
I had the same problem using torch.mm

renatomello:
> ```python
> print(torch.dot(torch.tensor([1+0j, 0, 0, 0], dtype=torch.complex128), torch.tensor([1+0j, 0, 0, 0], dtype=torch.complex128)))
>     print(torch.dot(torch.tensor([1, 0, 0, 0], dtype=torch.float64), torch.tensor([1, 0, 0, 0], dtype=torch.float64)))
> ```

BTW I just ran the examples that you gave and there were no issues, they both give the correct result.

src/qibo/backends/numpy.py
```diff
@@ -114,15 +115,15 @@ def matrix_parametrized(self, gate):
 
     def matrix_fused(self, fgate):
         rank = len(fgate.target_qubits)
-        matrix = np.eye(2**rank, dtype=self.dtype)
+        matrix = np.eye(2**rank, dtype=np.complex128)
```
renatomello, line 118:
```suggestion
        matrix =self.identity_density_matrix(2**rank, normalize=False)
```

src/qibo/backends/numpy.py
```diff
@@ -114,15 +115,15 @@ def matrix_parametrized(self, gate):
...
         for gate in fgate.gates:
             # transfer gate matrix to numpy as it is more efficient for
             # small tensor calculations
             # explicit to_numpy see https://github.com/qiboteam/qibo/issues/928
             gmatrix = self.to_numpy(gate.matrix(self))
             # Kronecker product with identity is needed to make the
             # original matrix have shape (2**rank x 2**rank)
-            eye = np.eye(2 ** (rank - len(gate.qubits)), dtype=self.dtype)
+            eye = np.eye(2 ** (rank - len(gate.qubits)), dtype=np.complex128)
```
renatomello, line 126:
```suggestion
            eye = self.identity_density_matrix(2 ** (rank - len(gate.qubits)), normalize=False)
```

src/qibo/backends/numpy.py
```diff
@@ -137,7 +138,7 @@ def matrix_fused(self, fgate):
             gmatrix = np.reshape(gmatrix, original_shape)
             # fuse the individual gate matrix to the total ``FusedGate`` matrix
             matrix = gmatrix @ matrix
-        return matrix
+        return self.cast(matrix)
```
renatomello, line 141: If the suggestions above work, this `cast` is probably not needed

src/qibo/backends/numpy.py
```diff
@@ -522,7 +524,7 @@ def execute_circuit_repeated(self, circuit, nshots, initial_state=None):
                 sample = result.samples()[0]
                 results.append(sample)
                 if not circuit.density_matrix:
-                    samples.append("".join([str(s) for s in sample]))
+                    samples.append("".join([str(s) for s in self.to_numpy(sample)]))
```
renatomello, line 527:
```suggestion
                    samples.append("".join([str(int(s)) for s in sample]))
```

src/qibo/backends/numpy.py
```diff
@@ -592,8 +594,13 @@ def calculate_symbolic_density_matrix(
                 return terms
         return terms
 
+    def dimensions(self, x):
+        return x.ndim
```
renatomello, line 598: Why is this needed in the `numpy` backend?

src/qibo/backends/numpy.py
```diff
@@ -592,8 +594,13 @@ def calculate_symbolic_density_matrix(
...
+
     def _order_probabilities(self, probs, qubits, nqubits):
         """Arrange probabilities according to the given ``qubits`` ordering."""
+        if self.dimensions(probs) == 0:
+            return probs
```
renatomello, line 603: I guess this is where the new method is needed. It is better to just redefine `_order_probabilities` in the `torch` class

src/qibo/backends/numpy.py
```diff
@@ -606,10 +613,15 @@ def calculate_probabilities(self, state, qubits, nqubits):
         rtype = self.np.real(state).dtype
         unmeasured_qubits = tuple(i for i in range(nqubits) if i not in qubits)
         state = self.np.reshape(self.np.abs(state) ** 2, nqubits * (2,))
-        probs = self.np.sum(state.astype(rtype), axis=unmeasured_qubits)
+        # This is necessary to use the same function on pytorch backend
+        if len(unmeasured_qubits) == 0:
+            probs = self.cast(state, dtype=rtype)
+        else:
+            probs = self.np.sum(self.cast(state, dtype=rtype), axis=unmeasured_qubits)
```
renatomello, line 620: Then this modification wouldn't be needed anymore

src/qibo/backends/numpy.py
```diff
@@ -606,10 +613,15 @@ def calculate_probabilities(self, state, qubits, nqubits):
...
         return self._order_probabilities(probs, qubits, nqubits).ravel()
 
     def calculate_probabilities_density_matrix(self, state, qubits, nqubits):
+        state = self.cast(state)
```
renatomello, line 624: `state` is not of the correct type already?

src/qibo/backends/numpy.py
```diff
@@ -641,8 +653,7 @@ def samples_to_decimal(self, samples, nqubits):
         return self.np.matmul(self.to_numpy(samples), qrange)[:, 0]
 
     def calculate_frequencies(self, samples):
-        res, counts = self.np.unique(samples, return_counts=True)
-        res, counts = self.np.array(res), self.np.array(counts)
+        res, counts = np.unique(samples, return_counts=True)
         return collections.Counter({k: v for k, v in zip(res, counts)})
```
renatomello, line 657: I know this isn't something that you did, but
```suggestion
        return collections.Counter(zip(res, counts))
```
works.

src/qibo/backends/numpy.py
```diff
@@ -702,9 +715,9 @@ def calculate_norm_density_matrix(self, state, order="nuc"):
         return self.np.linalg.norm(state, ord=order)
 
     def calculate_overlap(self, state1, state2):
-        state1 = self.cast(state1)
-        state2 = self.cast(state2)
-        return self.np.abs(self.np.sum(self.np.conj(state1) * state2))
+        return self.np.abs(
+            self.np.sum(self.np.conj(self.cast(state1)) * self.cast(state2))
+        )
```
renatomello, line 720: Not related to this PR per se, but this method should be removed entirely since there are already functions that do the same thing in `quantum_info`

src/qibo/backends/numpy.py
```diff
@@ -702,9 +715,9 @@ def calculate_norm_density_matrix(self, state, order="nuc"):
...
 
     def calculate_overlap_density_matrix(self, state1, state2):
```
renatomello, line 722: Same here

src/qibo/backends/numpy.py
```diff
@@ -737,22 +750,22 @@ def calculate_matrix_exp(self, a, matrix, eigenvectors=None, eigenvalues=None):
...
     def calculate_expectation_state(self, hamiltonian, state, normalize):
+        state = self.cast(state)
         statec = self.np.conj(state)
-        hstate = hamiltonian @ state
+        hstate = self.cast(hamiltonian @ state)
         ev = self.np.real(self.np.sum(statec * hstate))
         if normalize:
-            norm = self.np.sum(self.np.square(self.np.abs(state)))
-            ev = ev / norm
+            ev = ev / self.np.sum(self.np.square(self.np.abs(state)))
```
renatomello, line 763:
```suggestion
            ev /= self.np.sum(self.np.square(self.np.abs(state)))
```

src/qibo/backends/numpy.py
```diff
@@ -737,22 +750,22 @@ def calculate_matrix_exp(self, a, matrix, eigenvectors=None, eigenvalues=None):
...
-            return self.np.matmul(eigenvectors, self.np.matmul(expd, ud))
+        expd = self.np.diag(self.np.exp(-1j * a * eigenvalues))
+        ud = self.np.transpose(self.np.conj(eigenvectors))
+        return self.np.matmul(eigenvectors, self.np.matmul(expd, ud))
 
     def calculate_expectation_state(self, hamiltonian, state, normalize):
+        state = self.cast(state)
         statec = self.np.conj(state)
-        hstate = hamiltonian @ state
+        hstate = self.cast(hamiltonian @ state)
```
renatomello, line 760: Here as well: why aren't `hamiltonian` and `state` already in the correct backend?

src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -242,11 +243,17 @@ def __mul__(self, o):
...
             if self.backend.np.real(o) >= 0:  # TODO: check for side effects K.qnp
                 r._eigenvalues = o * self._eigenvalues
             elif not self.backend.issparse(self.matrix):
-                r._eigenvalues = o * self._eigenvalues[::-1]
+                if self.backend.__class__.__name__ == "PyTorchBackend":
+                    import torch
+
+                    r._eigenvalues = o * torch.flip(self._eigenvalues, [0])
+                else:
+                    r._eigenvalues = o * self._eigenvalues[::-1]
```
renatomello, line 256:
```suggestion
                r_engenvalues = (
                    o * self.backend.np.flip(self._eigenvalues, [0]) 
                    if isinstance(self.backend, PyTorchBackend) 
                    else o * self._eigenvalues[::-1]
                )
```

src/qibo/models/error_mitigation.py
```diff
@@ -809,12 +813,12 @@ def error_sensitive_circuit(circuit, observable, backend=None):
     for i in range(num_qubits):
         observable_i = pauli_gates[observable_pauli[i]]
         random_init = pauli_gates["I"]
-        while np.any(abs(observable_i - pauli_gates["Z"]) > 1e-5) and np.any(
-            abs(observable_i - pauli_gates["I"]) > 1e-5
-        ):
+        while backend.np.any(
+            backend.np.abs(observable_i - pauli_gates["Z"]) > 1e-5
+        ) and backend.np.any(abs(observable_i - pauli_gates["I"]) > 1e-5):
```
renatomello, line 818: is the `np.any` even needed here or can the built-in `any` from `python` be used?

src/qibo/quantum_info/utils.py
```diff
@@ -237,10 +237,10 @@ def hellinger_distance(prob_dist_p, prob_dist_q, validate: bool = False, backend
                 ValueError,
                 "All elements of the probability array must be between 0. and 1..",
             )
-        if np.abs(np.sum(prob_dist_p) - 1.0) > PRECISION_TOL:
+        if np.abs(np.sum(backend.to_numpy(prob_dist_p)) - 1.0) > PRECISION_TOL:
```
renatomello, line 240:
```suggestion
        if backend.np.abs(backend.np.sum(prob_dist_p) - 1.0) > PRECISION_TOL:
```

src/qibo/quantum_info/utils.py
```diff
@@ -237,10 +237,10 @@ def hellinger_distance(prob_dist_p, prob_dist_q, validate: bool = False, backend
                 ValueError,
                 "All elements of the probability array must be between 0. and 1..",
             )
-        if np.abs(np.sum(prob_dist_p) - 1.0) > PRECISION_TOL:
+        if backend.np.abs(backend.np.sum(prob_dist_p) - 1.0) > PRECISION_TOL:
             raise_error(ValueError, "First probability array must sum to 1.")
 
-        if np.abs(np.sum(prob_dist_q) - 1.0) > PRECISION_TOL:
+        if np.abs(np.sum(backend.to_numpy(prob_dist_q)) - 1.0) > PRECISION_TOL:
```
renatomello, line 243:
```suggestion
        if backend.np.abs(backend.np.sum(prob_dist_q) - 1.0) > PRECISION_TOL:
```

renatomello:
This needs #1245

tests/test_measurements_probabilistic.py
```diff
@@ -146,8 +146,9 @@ def test_measurementresult_apply_bitflips(backend, i, p0, p1):
 
     c = models.Circuit(3)
     c.add(gates.M(*range(3)))
-    state = backend.np.zeros(8)
+    state = np.zeros(8)
     state[0] = 1.0
+    state = backend.cast(state)
```
renatomello, line 151:
```suggestion
    state = backend.zero_state(8)
```

renatomello:
@stavros11 could you help us with this `dill` test that is failing here? The same test is passing for me locally, so I cannot reproduce it.

src/qibo/derivative.py
renatomello: @Simone-Bordoni now that tests and coverage are passing, it is time to come back to this. This should be done before approval.

renatomello: Since this PR is already gigantic, one option could be merging like this but immediately open a new one that implements the derivatives / variational circuits part. What do you think @BrunoLiegiBastonLiegi ?

renatomello:
@BrunoLiegiBastonLiegi Since tests are passing now, could you give a first review on this? Especially regarding `torch` applicability itself.

src/qibo/gates/gates.py
```diff
@@ -2328,14 +2328,31 @@ def __init__(
...
+                transpose_function = np.transpose
+
+            product = transpose_function(conj_function(unitary), (1, 0)) @ unitary
+            diagonals = all(np.abs(1 - diag_function(product)) < PRECISION_TOL)
+            off_diagonals = bool(
+                all_function(
+                    np.abs(product - diag_function(diag_function(product)))
+                    < PRECISION_TOL
+                )
+            )
```
BrunoLiegiBastonLiegi, line 2352: I believe that here the only function causing problems is `all`, but for both `torch.tensor` and `np.array` you can call `x.all()` rather than explicitely `np/torch.all(x)`. 
Therefore if you do:
```
off_diagonals = bool(
                    (np.abs(product - np.diag(np.diag(product)))
                    < PRECISION_TOL).all()
                )
```
it should work for both

src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -242,11 +245,16 @@ def __mul__(self, o):
...
         if self._eigenvalues is not None:
             if self.backend.np.real(o) >= 0:  # TODO: check for side effects K.qnp
                 r._eigenvalues = o * self._eigenvalues
             elif not self.backend.issparse(self.matrix):
-                r._eigenvalues = o * self._eigenvalues[::-1]
+                r._eigenvalues = (
+                    o * self.backend.np.flip(self._eigenvalues, [0])
+                    if isinstance(self.backend, PyTorchBackend)
+                    else o * self._eigenvalues[::-1]
+                )
```
BrunoLiegiBastonLiegi, line 257:
```suggestion
                r._eigenvalues = o * self.backend.np.flip(self._eigenvalues, (0,))
```

src/qibo/transpiler/unitary_decompositions.py
```diff
@@ -61,22 +61,31 @@ def calculate_psi(unitary, magic_basis=magic_basis, backend=None):
...
+    )
     # construct and diagonalize UT_U
-    ut_u = np.transpose(u_magic) @ u_magic
+    ut_u = backend.np.transpose(u_magic, (1, 0)) @ u_magic
     # When the matrix given to np.linalg.eig is a diagonal matrix up to machine precision the decomposition
     # is not accurate anymore. decimals = 20 works for random 2q Clifford unitaries.
-    eigvals, psi_magic = np.linalg.eig(np.round(ut_u, decimals=20))
-    # orthogonalize eigenvectors in the case of degeneracy (Gram-Schmidt)
-    psi_magic, _ = np.linalg.qr(psi_magic)
+    if backend.__class__.__name__ == "TensorflowBackend":
```
BrunoLiegiBastonLiegi, line 76: Is this because tensorflow's numpy doesn't have `linalg.eig`? If that's the case you could always add it in the backend `__init__`: `self.np.linalg.eig = np.linalg.eig`

src/qibo/transpiler/unitary_decompositions.py
```diff
@@ -86,7 +95,11 @@ def schmidt_decompose(state):
         (ndarray, ndarray): decomposition
 
     """
-    u, d, v = np.linalg.svd(np.reshape(state, (2, 2)))
+    backend = _check_backend(backend)
+    if backend.__class__.__name__ == "TensorflowBackend":
```
BrunoLiegiBastonLiegi, line 99: Same as above `self.np.linalg.svd = np.linalg.svd`

src/qibo/transpiler/unitary_decompositions.py
```diff
@@ -106,131 +119,161 @@ def calculate_single_qubit_unitaries(psi):
...
 
     See Eq. (A1) in arXiv:quant-ph/0011050.
     Ud is diagonal in the magic and Bell basis.
     """
+    backend = _check_backend(backend)
     # normalize U_A, U_B, V_A, V_B so that detU_d = 1
     # this is required so that sum(lambdas) = 0
     # and Ud can be written as exp(-iH)
-    det = np.linalg.det(unitary) ** (1 / 16)
+    if backend.__class__.__name__ == "TensorflowBackend":
```
BrunoLiegiBastonLiegi, line 164: `self.np.linalg.det = np.linalg.det`

BrunoLiegiBastonLiegi (commented):
Thanks @Simone-Bordoni and @renatomello for the huge work, I left some comments mostly to try and make the functions as agnostic as possible to numpy and torch. I did not look into the test yet, but I will asap.

pyproject.toml
```diff
@@ -67,8 +67,9 @@ dill = "^0.3.6"
 pytest-cov = "^4.0.0"
 pylint = "^3.0.3"
 matplotlib = "^3.7.0"
-qibojit = { git = "https://github.com/qiboteam/qibojit.git" }
 tensorflow = { version = "^2.14.1,<2.16", markers = "sys_platform == 'linux'" }
+torch = "^2.1.1"
```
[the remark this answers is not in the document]
renatomello: Done

src/qibo/backends/numpy.py
```diff
@@ -93,7 +94,7 @@ def identity_density_matrix(self, nqubits, normalize: bool = True):
 
     def plus_state(self, nqubits):
         state = self.np.ones(2**nqubits, dtype=self.dtype)
-        state /= self.np.sqrt(2**nqubits)
+        state /= math.sqrt(2**nqubits)
```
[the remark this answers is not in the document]
renatomello: `torch.sqrt` only works on `torch.Tensor`s, not on `float` / `int`

src/qibo/backends/numpy.py
```diff
@@ -105,24 +106,28 @@ def matrix(self, gate):
...
         """Convert a parametrized gate to its matrix representation in the computational basis."""
         name = gate.__class__.__name__
-        return getattr(self.matrices, name)(*gate.parameters)
+        matrix = getattr(self.matrices, name)(*gate.parameters)
+        return self.cast(matrix, dtype=matrix.dtype)
 
     def matrix_fused(self, fgate):
         rank = len(fgate.target_qubits)
-        matrix = np.eye(2**rank, dtype=self.dtype)
+        matrix = np.eye(2**rank)
```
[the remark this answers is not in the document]
renatomello: Since the matrices are generally small, the entire fusion of gates is done with `numpy` arrays and then cast to the corresponding backend. That is faster --- especially for GPU backends --- than casting all the matrices to the corresponding backend during calculation.

src/qibo/backends/numpy.py
```diff
@@ -153,10 +158,13 @@ def control_matrix(self, gate):
...
                 "Cannot use ``control_unitary`` method on "
-                "gate matrix of shape {}.".format(shape),
+                + f"gate matrix of shape {shape}.",
             )
         zeros = self.np.zeros((2, 2), dtype=self.dtype)
-        part1 = self.np.concatenate([self.np.eye(2, dtype=self.dtype), zeros], axis=0)
+        zeros = self.cast(zeros, dtype=zeros.dtype)
+        identity = self.np.eye(2, dtype=self.dtype)
+        identity = self.cast(identity, dtype=identity.dtype)
+        part1 = self.np.concatenate([identity, zeros], axis=0)
```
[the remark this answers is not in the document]
renatomello: The problem here is the GPU backends. Without casting, `part1` was trying to concatenate `numpy.array` with a `cupy.array`. CI was not catching this.

src/qibo/backends/pytorch.py
```diff
@@ -0,0 +1,223 @@
...
+            if r:
+                state = self.np.cat([self.np.zeros_like(state), state], dim=q)
+            else:
+                state = self.np.cat([state, self.np.zeros_like(state)], dim=q)
+        return state
+
+    def _order_probabilities(self, probs, qubits, nqubits):
+        """Arrange probabilities according to the given ``qubits`` ordering."""
+        if probs.dim() == 0:  # pragma: no cover
+            return probs
```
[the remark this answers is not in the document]
renatomello: To be honest, I also don't know why this is needed

src/qibo/backends/pytorch.py
```diff
@@ -0,0 +1,223 @@
...
+    def samples_to_decimal(self, samples, nqubits):
+        samples = self.cast(samples, dtype="int32")
+        qrange = self.np.arange(nqubits - 1, -1, -1, dtype=torch.int32)
+        qrange = (2**qrange).unsqueeze(1)
+        return self.np.matmul(samples, qrange).squeeze(1)
+
+    def calculate_overlap_density_matrix(self, state1, state2):
+        return self.np.trace(
+            self.np.matmul(self.np.conj(self.cast(state1)).T, self.cast(state2))
+        )
```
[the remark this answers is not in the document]
renatomello: I tried to remove this method and tests started failing. My guess is that the issue is due to `self.np.transpose` in the `torch` backend being overwritten by `torch.permute` because of other methods.

src/qibo/backends/pytorch.py
```diff
@@ -0,0 +1,223 @@
...
+        ud = self.np.conj(eigenvectors).T
+        return self.np.matmul(eigenvectors, self.np.matmul(expd, ud))
+
+    def calculate_hamiltonian_matrix_product(self, matrix1, matrix2):
+        if self.issparse(matrix1) or self.issparse(matrix2):  # pragma: no cover
+            return self.np.sparse.mm(matrix1, matrix2)  # pylint: disable=E1102
+        return self.np.matmul(matrix1, matrix2)
+
+    def calculate_hamiltonian_state_product(self, matrix, state):
+        return self.np.matmul(matrix, state)
```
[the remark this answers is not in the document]
renatomello: This is beyond this PR, but this method should be removed from all backends, since it's adding anything of value.

src/qibo/gates/gates.py
renatomello: I'm not sure `tensorflow.EagerTensor` has an `all()` method

src/qibo/transpiler/unitary_decompositions.py
renatomello: > Is this because tensorflow's numpy doesn't have `linalg.eig`? 

Yes.

> If that's the case you could always add it in the backend `__init__`: `self.np.linalg.eig = np.linalg.eig`

Problem is that `tensorflow`'s experimental `numpy` has no `linalg` module at all.

src/qibo/transpiler/unitary_decompositions.py
renatomello: Same answer from above.

src/qibo/transpiler/unitary_decompositions.py
renatomello: Same answer from above.

src/qibo/hamiltonians/hamiltonians.py
renatomello: This is failing for `tensorflow`, probably because the axis are being passed as `tuple`. However, I think `torch` does not accept axis as `int`.

renatomello: Just pushed a fix for it

src/qibo/backends/numpy.py
```diff
@@ -629,21 +638,20 @@ def sample_shots(self, probabilities, nshots):
...
         qrange = self.np.arange(nqubits - 1, -1, -1, dtype="int32")
-        qrange = (2**qrange)[:, self.np.newaxis]
+        qrange = (2**qrange)[:, None]
         return self.np.matmul(self.to_numpy(samples), qrange)[:, 0]
 
     def calculate_frequencies(self, samples):
-        res, counts = self.np.unique(samples, return_counts=True)
-        res, counts = self.np.array(res), self.np.array(counts)
-        return collections.Counter({k: v for k, v in zip(res, counts)})
+        res, counts = np.unique(samples, return_counts=True)
```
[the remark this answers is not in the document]
renatomello: This created a problem for `torch`, so I'm reverting it.

Simone-Bordoni (commented):
i was just correcting the same thing. i added a comment in the code to explain why

Simone-Bordoni:
The test "test_repeated_execute_probs_and_freqs" in test_models_circuit_features.py changes every time even if both np and torch seeds are fixed...

renatomello:
It doesn’t change every time. It’s always the same dict on CI, but it’s a
different duct when the test is run locally

Simone-Bordoni:
> It doesn’t change every time. It’s always the same dict on CI, but it’s a different duct when the test is run locally

Ok, now it makes sense. sorry if i changed it again, i will not touch it anymore

src/qibo/backends/pytorch.py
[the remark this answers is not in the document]
Simone-Bordoni: I have improved the method, in future PR we will remove it

src/qibo/backends/pytorch.py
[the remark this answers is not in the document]
Simone-Bordoni: I solved this using the .T method that works for both backends

Simone-Bordoni:
I have reverted the changes to the test with seed so now everything should pass. Thanks to @BrunoLiegiBastonLiegi I also managed to remove some more unuseful methods in the pytorch backend.

src/qibo/backends/numpy.py
```diff
@@ -767,20 +767,18 @@ def calculate_expectation_density_matrix(self, hamiltonian, state, normalize):
             ev = ev / norm
```
renatomello, line 767:
```suggestion
            ev /= norm
```

src/qibo/backends/numpy.py
```diff
@@ -279,11 +279,11 @@ def apply_channel_density_matrix(self, channel, state, nqubits):
     def _append_zeros(self, state, qubits, results):
         """Helper method for collapse."""
         for q, r in zip(qubits, results):
-            state = self.np.expand_dims(state, axis=q)
+            state = self.np.expand_dims(state, q)
             if r:
```
BrunoLiegiBastonLiegi, line 283:
```suggestion
            if len(r) == 0:
```
Is this a check for an empty list? Better to make the check explicit for readability.

src/qibo/backends/numpy.py
```diff
@@ -539,7 +539,7 @@ def execute_circuit_repeated(self, circuit, nshots, initial_state=None):
             assert circuit.has_collapse
             final_state = self.cast(np.mean(self.to_numpy(final_states), 0))
             if circuit.measurements:
-                qubits = [q for m in circuit.measurements for q in m.target_qubits]
+                # qubits = [q for m in circuit.measurements for q in m.target_qubits]
```
BrunoLiegiBastonLiegi, line 542:
```suggestion
```
remove comment

tests/test_hamiltonians.py
```diff
@@ -177,6 +177,11 @@ def test_hamiltonian_matmul(backend, sparse_type):
         backend.assert_allclose((H1 @ H2).matrix, (m1 @ m2))
         backend.assert_allclose((H2 @ H1).matrix, (m2 @ m1))
 
+    try:
+        H1 @ np.zeros(3 * (2**nqubits,), dtype=m1.dtype)
+    except Exception as error:
+        print(error)
```
BrunoLiegiBastonLiegi, line 183: Do we need this? How should the exception be handled?

tests/test_backends_clifford.py
```diff
@@ -13,14 +21,37 @@
...
 
 def construct_clifford_backend(backend):
-    if backend.__class__.__name__ in ["TensorflowBackend", "CuQuantumBackend"]:
+    if (
+        isinstance(backend, (TensorflowBackend, PyTorchBackend))
+        or backend.__class__.__name__ == "CuQuantumBackend"
+    ):
         with pytest.raises(NotImplementedError):
             clifford_backend = CliffordBackend(backend.name)
         pytest.skip("Clifford backend not defined for the this engine.")
```
BrunoLiegiBastonLiegi, line 30:
```suggestion
        pytest.skip("Clifford backend not defined for this engine.")
```

tests/test_gates_gates.py
```diff
@@ -747,7 +758,7 @@ def test_fswap(backend):
             [0, 1, 0, 0],
             [0, 0, 0, -1],
         ],
-        dtype=backend.dtype,
+        dtype=np.complex128,
```
BrunoLiegiBastonLiegi, line 761: What is the reason why we are changing this everywhere? In principle it should be safer to rely on the `backend.dtype`, right?

tests/test_hamiltonians.py
```diff
@@ -169,6 +177,11 @@ def test_hamiltonian_matmul(backend, sparse_type):
...
```
BrunoLiegiBastonLiegi, line 183: Same here, better to handle the exception properly.

tests/test_hamiltonians.py
```diff
@@ -352,6 +372,7 @@ def test_hamiltonian_eigenvectors(backend, dtype, dense):
     H3 = H1 * c2
     V3 = backend.to_numpy(H3.eigenvectors())
     U3 = backend.to_numpy(H3._eigenvalues)
+    print(U3)
```
BrunoLiegiBastonLiegi, line 375:
```suggestion
```
remove print

BrunoLiegiBastonLiegi (commented):
Thanks @Simone-Bordoni, please address these last minor things and open an issue about adding support to gradients in the pytorch backend. Then we can merge this I guess.

tests/test_gates_gates.py
Simone-Bordoni: If the backend is pytorch the backend.dtype will be torch.dtype which can't be used in the function np.array

tests/test_hamiltonians.py
Simone-Bordoni: sorry, it is a leftover from previous tests

src/qibo/backends/numpy.py
BrunoLiegiBastonLiegi: But then what is `r` and what are we checking there? I would say it's probably checking whether it is `None` at this point, but better to make sure and in case write it explicitely.

Simone-Bordoni: def _append_zeros(self, state, qubits, results):
        """Helper method for collapse."""
        for q, r in zip(qubits, results):
            state = self.np.expand_dims(state, q)
>           if len(r):
E           TypeError: object of type 'numpy.intc' has no len()

src\qibo\backends\numpy.py:283: TypeError

renatomello: I tested locally. This is checking if `r` is `0` or `1`. For `r=0`, the `if r` returns `False`. For `r=1`, the `if r` returns `True`.

renatomello (approved):

BrunoLiegiBastonLiegi (approved):

PR #1206, opened 2024-02-09
Calculation of `Clifford.stabilizers` and `Clifford.destabilizers` in parallel

renatomello:
Checklist:
- [x] Tests are passing.


src/qibo/quantum_info/clifford.py
```diff
@@ -150,6 +152,8 @@ def stabilizers(self, symplectic: bool = False, return_array: bool = False):
             return_array (bool, optional): To be used when ``symplectic = False``.
                 If ``True`` returns the stabilizers as ``ndarray``.
                 If ``False``, returns stabilizers as strings. Defaults to ``False``.
+            cores (int, optional): number of processes to be used to calculate all stabilizers.
```
BrunoLiegiBastonLiegi, line 155: I think I would rather call this `ncpus` or similar, which from my experience is a more commonly used name for this kind of things.

src/qibo/quantum_info/clifford.py
```diff
@@ -422,8 +431,18 @@ def _construct_operators(self, generators: list, phases: list):
...
+
+        operators = [(identity, g) for g in operators]
+
+        if cores > 1:
+            from multiprocessing import Pool  # pylint: disable=E0401
+
+            with Pool(cores) as pool:
+                results = pool.map(_string_product, product(*operators))
+                pool.close()
+                pool.join()
```
BrunoLiegiBastonLiegi, line 444: Do we need these two?

renatomello:
I'm closing this PR because we might go in a different direction.

PR #1208, opened 2024-02-10
Addition of parametrized gates missing in `TensorflowMatrices`

renatomello:
This follows a discussion in #1202 

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/backends/tensorflow.py
+++ b/src/qibo/backends/tensorflow.py
@@ -20,87 +20,273 @@ class TensorflowMatrices(NumpyMatrices):
         self.tf = tf
         self.np = tnp
 
-    def _get_numpy_gate(self, name, params):
-        matrix = getattr(NumpyMatrices(self.dtype), name)(*params)
-        return self.tf.cast(matrix, dtype=self.dtype)
-
     def RX(self, theta):
-        return self._get_numpy_gate("RX", [theta])
+        cos = self.np.cos(theta / 2.0) + 0j
+        isin = -1j * self.np.sin(theta / 2.0)
+        return self.tf.cast([[cos, isin], [isin, cos]], dtype=self.dtype)
 
     def RY(self, theta):
-        return self._get_numpy_gate("RY", [theta])
+        cos = self.np.cos(theta / 2.0) + 0j
+        sin = self.np.sin(theta / 2.0) + 0j
+        return self.tf.cast([[cos, -sin], [sin, cos]], dtype=self.dtype)
 
     def RZ(self, theta):
-        return self._get_numpy_gate("RZ", [theta])
+        phase = self.np.exp(0.5j * theta)
+        return self.tf.cast([[self.np.conj(phase), 0], [0, phase]], dtype=self.dtype)
 
     def GPI(self, phi):
-        return self._get_numpy_gate("GPI", [phi])
+        phase = self.np.exp(1.0j * phi)
+        return self.tf.cast([[0, self.np.conj(phase)], [phase, 0]], dtype=self.dtype)
 
     def GPI2(self, phi):
-        return self._get_numpy_gate("GPI2", [phi])
+        phase = self.np.exp(1.0j * phi)
+        return self.tf.cast(
+            [[1, -1.0j * self.np.conj(phase)], [-1.0j * phase, 1]], dtype=self.dtype
+        ) / self.np.sqrt(2)
 
     def U1(self, theta):
-        return self._get_numpy_gate("U1", [theta])
+        phase = self.np.exp(1j * theta)
+        return self.tf.cast([[1, 0], [0, phase]], dtype=self.dtype)
 
     def U2(self, phi, lam):
-        return self._get_numpy_gate("U2", [phi, lam])
+        eplus = self.np.exp(1j * (phi + lam) / 2.0)
+        eminus = self.np.exp(1j * (phi - lam) / 2.0)
+        return self.tf.cast(
+            [[self.np.conj(eplus), -self.np.conj(eminus)], [eminus, eplus]],
+            dtype=self.dtype,
+        ) / self.np.sqrt(2)
 
     def U3(self, theta, phi, lam):
-        return self._get_numpy_gate("U3", [theta, phi, lam])
+        cost = self.np.cos(theta / 2)
+        sint = self.np.sin(theta / 2)
+        eplus = self.np.exp(1j * (phi + lam) / 2.0)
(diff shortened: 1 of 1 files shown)


Simone-Bordoni:
Why did you go back to the previous implementation where you define again the gates? Do I have to go back to that implementation also on the PyTorch backend?

renatomello:
> Why did you go back to the previous implementation where you define again the gates? Do I have to go back to that implementation also on the PyTorch backend?

`tensorflow` loses track of the variable if it goes through a `numpy` function (in this case, `np.array`) and then back to `tf`. I don't know if the same thing happens in `torch`

Simone-Bordoni (approved):

stavros11 (approved):
Thanks @renatomello.

Regarding what you tried in 15c707d29405e5d5d9acd7c942e5e8dffc2ede89, I think there should be a way to simplify `TensorflowMatrices` to just inherit the gates from `NumpyMatrices` so that we don't have to redefine everything, while still not breaking backprop. 

One simple idea is to modify `NumpyMatrices` as follows:
```py
class NumpyMatrices:

    def _cast(self, x):
        return self.np.array(x, dtype=self.dtype)

    def U3(self, theta, phi, lam):
        cost = self.np.cos(theta / 2)
        sint = self.np.sin(theta / 2)
        eplus = self.np.exp(1j * (phi + lam) / 2.0)
        eminus = self.np.exp(1j * (phi - lam) / 2.0)
        return self._cast(
            [
                [self.np.conj(eplus) * cost, -self.np.conj(eminus) * sint],
                [eminus * sint, eplus * cost],
            ],
        )
```
and similar for other parametrized gates, and then `TensorflowMatrices` would just be
```py
class TensorflowMatrices(NumpyMatrices):
    def __init__(self, dtype):
        super().__init__(dtype)
        import tensorflow as tf  # pylint: disable=import-error
        import tensorflow.experimental.numpy as tnp  # pylint: disable=import-error

        self.tf = tf
        self.np = tnp

    def _cast(self, x):
        return self.tf.cast(x, dtype=self.dtype)
```
without having to redefine `U3` and other gates. I have not tested if this fully works, though.

I am not sure if even having different classes for matrices in each backend is necessary, most likely there is a way to get around this with just functions but until that, maybe the proposal above is a doable simplification.

renatomello:
> Thanks @renatomello.
> 
> Regarding what you tried in [15c707d](https://github.com/qiboteam/qibo/commit/15c707d29405e5d5d9acd7c942e5e8dffc2ede89), I think there should be a way to simplify `TensorflowMatrices` to just inherit the gates from `NumpyMatrices` so that we don't have to redefine everything, while still not breaking backprop.
> 
> One simple idea is to modify `NumpyMatrices` as follows:
> 
> ```python
> class NumpyMatrices:
> 
>     def _cast(self, x):
>         return self.np.array(x, dtype=self.dtype)
> 
>     def U3(self, theta, phi, lam):
>         cost = self.np.cos(theta / 2)
>         sint = self.np.sin(theta / 2)
>         eplus = self.np.exp(1j * (phi + lam) / 2.0)
>         eminus = self.np.exp(1j * (phi - lam) / 2.0)
>         return self._cast(
>             [
>                 [self.np.conj(eplus) * cost, -self.np.conj(eminus) * sint],
>                 [eminus * sint, eplus * cost],
>             ],
>         )
> ```
> 
> and similar for other parametrized gates, and then `TensorflowMatrices` would just be
> 
> ```python
> class TensorflowMatrices(NumpyMatrices):
>     def __init__(self, dtype):
>         super().__init__(dtype)
>         import tensorflow as tf  # pylint: disable=import-error
>         import tensorflow.experimental.numpy as tnp  # pylint: disable=import-error
> 
>         self.tf = tf
>         self.np = tnp
> 
>     def _cast(self, x):
>         return self.tf.cast(x, dtype=self.dtype)
> ```
> 
> without having to redefine `U3` and other gates. I have not tested if this fully works, though.

I will try this and see if it works.

> I am not sure if even having different classes for matrices in each backend is necessary, most likely there is a way to get around this with just functions but until that, maybe the proposal above is a doable simplification.

I was thinking that maybe the parametrized gates should be defined as returning a list instead of an array and then each backend would just need to cast it, but I think your proposal above is cleaner.

renatomello:
> > Thanks @renatomello.
> > Regarding what you tried in [15c707d](https://github.com/qiboteam/qibo/commit/15c707d29405e5d5d9acd7c942e5e8dffc2ede89), I think there should be a way to simplify `TensorflowMatrices` to just inherit the gates from `NumpyMatrices` so that we don't have to redefine everything, while still not breaking backprop.
> > One simple idea is to modify `NumpyMatrices` as follows:
> > ```python
> > class NumpyMatrices:
> > 
> >     def _cast(self, x):
> >         return self.np.array(x, dtype=self.dtype)
> > 
> >     def U3(self, theta, phi, lam):
> >         cost = self.np.cos(theta / 2)
> >         sint = self.np.sin(theta / 2)
> >         eplus = self.np.exp(1j * (phi + lam) / 2.0)
> >         eminus = self.np.exp(1j * (phi - lam) / 2.0)
> >         return self._cast(
> >             [
> >                 [self.np.conj(eplus) * cost, -self.np.conj(eminus) * sint],
> >                 [eminus * sint, eplus * cost],
> >             ],
> >         )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and similar for other parametrized gates, and then `TensorflowMatrices` would just be
> > ```python
> > class TensorflowMatrices(NumpyMatrices):
> >     def __init__(self, dtype):
> >         super().__init__(dtype)
> >         import tensorflow as tf  # pylint: disable=import-error
> >         import tensorflow.experimental.numpy as tnp  # pylint: disable=import-error
> > 
> >         self.tf = tf
> >         self.np = tnp
> > 
> >     def _cast(self, x):
> >         return self.tf.cast(x, dtype=self.dtype)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > without having to redefine `U3` and other gates. I have not tested if this fully works, though.
> 
> I will try this and see if it works.
> 
> > I am not sure if even having different classes for matrices in each backend is necessary, most likely there is a way to get around this with just functions but until that, maybe the proposal above is a doable simplification.
> 
I was thinking that maybe the parametrized gates should be defined as returning a list instead of an array and then each backend would just need to cast it, but I think your proposal above is cleaner.
I will merge as is and try the proposed solution in a new PR.

PR #1217, opened 2024-02-13
Backend-agnostic matrix definitions

Simone-Bordoni:
Store a backend agnostic matrix definition tht can be casted with a specific _cast function in the proper backend.
Necessary for PR #1202 

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/backends/npmatrices.py
+++ b/src/qibo/backends/npmatrices.py
@@ -163,6 +163,7 @@ class NumpyMatrices:
 
     @cached_property
     def CSXDG(self):
+        # return self.np.conj(self.CSX)
         a = (1 - 1j) / 2
         b = self.np.conj(a)
         return self._cast(
--- a/src/qibo/backends/npmatrices.py
+++ b/src/qibo/backends/npmatrices.py
@@ -245,7 +245,13 @@ class NumpyMatrices:
     @cached_property
     def iSWAP(self):
         return self._cast(
-            [[1, 0, 0, 0], [0, 0, 1j, 0], [0, 1j, 0, 0], [0, 0, 0, 1]], dtype=self.dtype
+            [
+                [1+0j, 0j, 0j, 0j], 
+                [0j, 0j, 1j, 0j], 
+                [0j, 1j, 0j, 0j], 
+                [0j, 0j, 0j, 1+0j],
+            ], 
+            dtype=self.dtype
         )
 
     @cached_property
--- a/src/qibo/backends/npmatrices.py
+++ b/src/qibo/backends/npmatrices.py
@@ -25,7 +25,7 @@ class NumpyMatrices:
 
     @cached_property
     def Y(self):
-        return self._cast([[0, -1j], [1j, 0]], dtype=self.dtype)
+        return self._cast([[0j, -1j], [1j, 0j]], dtype=self.dtype)
 
     @cached_property
     def Z(self):
@@ -41,11 +41,11 @@ class NumpyMatrices:
 
     @cached_property
     def S(self):
-        return self._cast([[1, 0], [0, 1j]], dtype=self.dtype)
+        return self._cast([[1 + 0j, 0j], [0j, 1j]], dtype=self.dtype)
 
     @cached_property
     def SDG(self):
-        return self._cast([[1, 0], [0, -1j]], dtype=self.dtype)
+        return self._cast([[1 + 0j, 0j], [0j, -1j]], dtype=self.dtype)
 
     @cached_property
     def T(self):
@@ -132,7 +132,12 @@ class NumpyMatrices:
     @cached_property
     def CY(self):
         return self._cast(
-            [[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 0, -1j], [0, 0, 1j, 0]],
+            [
+                [1 + 0j, 0j, 0j, 0j],
+                [0j, 1 + 0j, 0j, 0j],
+                [0j, 0j, 0j, -1j],
+                [0j, 0j, 1j, 0j],
+            ],
             dtype=self.dtype,
         )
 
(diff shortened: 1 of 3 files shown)


Simone-Bordoni:
@renatomello 
I am getting an error for the CSXDG gate.
previously is was:
```python
self.np.conj(self.CSX)
``` 
Now i changed to:
```python
a = (1 - 1j) / 2
b = self.np.conj(a)
      return self._cast(
          [
              [1, 0, 0, 0],
              [0, 1, 0, 0],
              [0, 0, a, b],
              [0, 0, b, a],
          ],
          dtype=self.dtype,
      )
``` 
It should be the same thing as CSX is defined the same way but with a=(1+1j)/2

renatomello:
> @renatomello I am getting an error for the CSXDG gate. previously is was:
> 
> ```python
> self.np.conj(self.CSX)
> ```
> 
> Now i changed to:
> 
> ```python
> a = (1 - 1j) / 2
> b = self.np.conj(a)
>       return self._cast(
>           [
>               [1, 0, 0, 0],
>               [0, 1, 0, 0],
>               [0, 0, a, b],
>               [0, 0, b, a],
>           ],
>           dtype=self.dtype,
>       )
> ```
> 
> It should be the same thing as CSX is defined the same way but with a=(1+1j)/2

Which error though

Simone-Bordoni:
> > @renatomello I am getting an error for the CSXDG gate. previously is was:
> > ```python
> > self.np.conj(self.CSX)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Now i changed to:
> > ```python
> > a = (1 - 1j) / 2
> > b = self.np.conj(a)
> >       return self._cast(
> >           [
> >               [1, 0, 0, 0],
> >               [0, 1, 0, 0],
> >               [0, 0, a, b],
> >               [0, 0, b, a],
> >           ],
> >           dtype=self.dtype,
> >       )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It should be the same thing as CSX is defined the same way but with a=(1+1j)/2
> 
> Which error though

Assertion error in test_csxdg and test_dagger (the tests that failed also on github)

renatomello:
> > Which error though
> 
> Assertion error in test_csxdg and test_dagger (the tests that failed also on github)

It's only failing for `numba`. So, most likely a PR in `qibojit` is needed [here](https://github.com/qiboteam/qibojit/blob/48b51665775a70dfd4ab577864f57623f15e0073/src/qibojit/backends/matrices.py#L33) adding
```python
@cached_property
def CSXDG(self):
    return self.SXDG
```
I have not tested locally but I'd imagine this would fix it.

src/qibo/backends/npmatrices.py
```diff
@@ -17,51 +17,53 @@ def _cast(self, x, dtype):
...
+        return self._cast([[1, 0], [0, -1]], dtype=self.dtype)
 
     @cached_property
     def SX(self):
-        return self.np.array([[1 + 1j, 1 - 1j], [1 - 1j, 1 + 1j]], dtype=self.dtype) / 2
+        return self._cast([[1 + 1j, 1 - 1j], [1 - 1j, 1 + 1j]], dtype=self.dtype) / 2
 
     @cached_property
     def SXDG(self):
-        return self.np.transpose(self.np.conj(self.SX))
```
stavros11, line 40: What is the issue with the `transpose` here? I am assuming it fails in the tensorflow/pytorch inheritance?

src/qibo/backends/npmatrices.py
```diff
@@ -17,51 +17,53 @@ def _cast(self, x, dtype):
...
+        return self._cast([[1 - 1j, 1 + 1j], [1 + 1j, 1 - 1j]], dtype=self.dtype) / 2
 
     @cached_property
     def S(self):
-        return self.np.array([[1, 0], [0, 1j]], dtype=self.dtype)
+        return self._cast([[1 + 0j, 0j], [0j, 1j]], dtype=self.dtype)
 
     @cached_property
     def SDG(self):
-        return self.np.conj(self.S)
```
stavros11, line 48: Same about the `conj`. Here I am not sure if there is really an issue because `np.conj` is being used in some other matrices (ie in `CSXDG`), maybe you need to modify those as well?

stavros11:
> It's only failing for `numba`. So, most likely a PR in `qibojit` is needed [here](https://github.com/qiboteam/qibojit/blob/48b51665775a70dfd4ab577864f57623f15e0073/src/qibojit/backends/matrices.py#L33) adding
> 
> ```python
> @cached_property
> def CSXDG(self):
>     return self.SXDG
> ```
> 
> I have not tested locally but I'd imagine this would fix it.

@renatomello do you have any idea why this is the case? I understand this PR is changing the calculations of some matrices a bit, but the returned values should remain the same. I don't see how this would break qibojit.

src/qibo/backends/npmatrices.py
renatomello: Correct

src/qibo/backends/npmatrices.py
renatomello: Yes, all matrices would have to be hard-coded instead of relying in any `numpy` function

Simone-Bordoni:
> > > Which error though
> > 
> > 
> > Assertion error in test_csxdg and test_dagger (the tests that failed also on github)
> 
> It's only failing for `numba`. So, most likely a PR in `qibojit` is needed [here](https://github.com/qiboteam/qibojit/blob/48b51665775a70dfd4ab577864f57623f15e0073/src/qibojit/backends/matrices.py#L33) adding
> 
> ```python
> @cached_property
> def CSXDG(self):
>     return self.SXDG
> ```
> 
> I have not tested locally but I'd imagine this would fix it.

I have opened the PR in qibojit

renatomello:
> @renatomello @stavros11 by adding the CSXDG gate in qibojit the tests fail only on linux...

it's because only ubuntu is testing `tensorflow`

renatomello (approved):
This looks good to advance #1202, but it needs to be merged with https://github.com/qiboteam/qibojit/pull/168

Simone-Bordoni:
Now that the qibojit branch "add CSXDG" has been merged this PR is ready to be merged (after one missing review)

stavros11 (approved):

PR #1220, opened 2024-02-15
Reduce Codecov noise

alecandido:


alecandido:
@renatomello this should prevent noise from your single line issue (it doesn't solve the uncovered line, but it will avoid bothering you)

andrea-pasquale (approved):
Thanks @alecandido.

.github/codecov.yml
```diff
@@ -0,0 +1,12 @@
...
+    project:
+      default:
+        threshold: 0.5%
+
+github_checks:
+  annotations: false
+
+ignore:
+  - "setup.py"
+  - "src/qibo/tests/test_custom_functions.py"
```
stavros11, line 12: I know it existed before this PR, but I have the impression that these files do not exist anymore, so we don't need these lines.

stavros11 (commented):
What is the motivation of not failing when a single line is not covered? I agree that it is annoying but may be the only way to maintain coverage to 100% (excluding `# pragma: no cover` ofcourse).

.github/codecov.yml
alecandido: You're right, I simply ignored the issue. But it costs little, so it's worth to do it.

Solved in 47b2518

alecandido:
> What is the motivation of not failing when a single line is not covered? I agree that it is annoying but may be the only way to maintain coverage to 100% (excluding `# pragma: no cover` ofcourse).

To be fair, the main motivation is that I do not care too much about the 100% coverage.

Coverage is definitely useful, but `99% == 100%` in all respects.
(you could always cover, but not test effectively, or not test it, but with little to no impact)

To me, it's more relevant that sensible PRs are marked with the green tick, while the red cross is reserved to those failing tests, or with a sizeable drop in coverage. We could still target the 100%, but I would not do it PR by PR (e.g. if for a release we don't have 100%, and we'd want it, we could make an effort to compensate).

However, the 100% could also be achieved through human revision, paying more attention to tests and suggesting meaningful ones.

stavros11 (approved):
> To be fair, the main motivation is that I do not care too much about the 100% coverage.

Thanks, that was what I was expecting, just wanted to make sure.

> Coverage is definitely useful, but `99% == 100%` in all respects. (you could always cover, but not test effectively, or not test it, but with little to no impact)

I agree about the effectiveness of tests. The only advantage of covering 100% is that you are at least sure that every line can be executed, so there are no obvious issues (many of which would be captured by pylint anyway). For sure it's far from a guarantee that the code does what is supposed to.

Anyway, I agree with merging this PR.

PR #1224, opened 2024-02-16
Fixing missing dependencies

### LINKED ISSUE #1118: Missing dependencies
I already published an issue about dependencies, namely #814. But in that case I was focused on limiting the current dependencies, and I didn't notice we were missing a few ones.

https://github.com/qiboteam/qibo/blob/1dd73f7a637defd01a42023651f8499652115a7a/pyproject.toml#L23-L34

In particular:
- [ ] the most notable mention is NumPy: it is undeclared (even though we're getting it through SciPy, for sure)
- [ ] despite the TensorFlow backend, there is not even an optional dependency in here (it is only in the `tests` group, but this will not survive packaging)
- [ ] I didn't scan the code yet for other ones (i.e. `grep import`, and check which ones are used)

scarrazza:
Fixes #1118 .

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -4927,6 +4927,85 @@ files = [
 ]
 
 [[package]]
+name = "wrapt"
+version = "1.16.0"
+description = "Module for decorators, wrappers and monkey patching."
+optional = false
+python-versions = ">=3.6"
+files = [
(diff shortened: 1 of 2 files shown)


andrea-pasquale (approved):

alecandido (approved):
Thanks @scarrazza

PR #1231, opened 2024-02-28
Support Python 3.12

alecandido:


alecandido:
@natestemen this branch could work (in principle), and we've been kind of lucky: TensorFlow didn't support py3.12 until 5 days ago, when they published a release candidate with initial support (not yet a stable one).

I wonder whether it's this the triggered your update.

I still have to solve part of the problem with the installation, but the dependencies should now be alright.

alecandido:
@scarrazza is it acceptable to have a TensorFlow release candidate dep in `master`?

alecandido:
Assuming that tests are passing, the plan is the following:

- merge Qibo py3.12 in `master`, with Qibojit branch dep
- update Qibojit to Qibo `master`
- merge Qibojit py3.12 in `main`, with Qibo `master` dep
- update Qibo to Qibojit `main`
- release Qibo, with Qibojit `main` (it is only a test dep, so it should be reasonable)
- update Qibojit to Qibo release
- release Qibojit
- update Qibo to Qibojit release

alecandido:
Tests are failing because `hyperopt` is a relatively old package (no release since two years) and it implicitly requires `setuptools`
https://github.com/hyperopt/hyperopt/tags
https://github.com/hyperopt/hyperopt/blob/d8515263a3168ba8f111179d29e74ffb09670610/setup.py#L46-L52
https://github.com/hyperopt/hyperopt/blob/d8515263a3168ba8f111179d29e74ffb09670610/hyperopt/atpe.py#L19

I'll try to circumvent the issue, but I'd recommend replacing it, if possible.

alecandido:
Required for https://github.com/unitaryfund/mitiq/pull/2066

alecandido:
Currently stuck because of TensorFlow installation.

alecandido:
Tests are passing, even though not doctests. The reason is that if `tensorflow` is not installed, the related tests are skipped:
https://github.com/qiboteam/qibo/blob/aa0fab287fa51362e552116776c108bfecad5269/tests/conftest.py#L31-L41

stavros11:
> Tests are passing, even though not doctests. The reason is that if `tensorflow` is not installed, the related tests are skipped:

Thanks @alecandido. This is expected, I think tests (not doctests) were set up so that they only run on backends that are available, mostly for convenience when running them locally in an environment with a subset backends. However, I do not understand why tensorflow is not installed in the CI, since we still have it in the pyproject?

If tensorflow is the only issue blocking this I would consider disabling the related doctests temporarily and opening an issue until tensorflow is "officially" relased for py3.12. That also depends on what release we would like to include this. If I remember correctly is not for the one coming soon so maybe we could even keep on hold. The main motivation is that it is needed by an external project, though.

alecandido:
> However, I do not understand why tensorflow is not installed in the CI, since we still have it in the pyproject?

This is an issue with the release candidate package, most likely. That's why this PR is on hold, waiting for a more stable support for py3.12 from TensorFlow (or to make TensorFlow optional, e.g. outsourcing to `qiboml`).

alecandido:
> If tensorflow is the only issue blocking this I would consider disabling the related doctests temporarily and opening an issue until tensorflow is "officially" relased for py3.12. That also depends on what release we would like to include this. If I remember correctly is not for the one coming soon so maybe we could even keep on hold. The main motivation is that it is needed by an external project, though.

It's not too urgent, apparently, let's keep going with `qibo-core` and the reorganization first.

natestemen:
Thank you for so much effort after I pinged y'all! Just want to say it's not super urgent on our side.

alecandido:
> Thank you for so much effort after I pinged y'all! Just want to say it's not super urgent on our side.

Thanks for the update @natestemen!
Indeed, I was initially a bit worried to be blocking `mitiq`, but now we're all waiting for TF.

I also checked the various common distributions, and no major one is shipping py3.12 as default Python yet.

In any case, we'll work on reducing the default dependencies later on, to make upgrades easier in general. For the time being, as soon as TF will be unlocked, we'll complete this PR.

renatomello:
https://github.com/tensorflow/tensorflow/releases/tag/v2.16.1

alecandido:
> https://github.com/tensorflow/tensorflow/releases/tag/v2.16.1

On PyPI since Saturday
image

I'll now make the package!

alecandido:
Ok, now the situation is even more complicated, since the same problem of QiboJIT is now present for QiboTN as well.

I will move forward with this PR, and the matching ones
https://github.com/qiboteam/qibojit/pull/170
https://github.com/qiboteam/qibotn/pull/48

If tests are passing, we can merge. But we should try to push for `qibo-core` as well, otherwise this
https://github.com/qiboteam/qibo/pull/1231#issuecomment-1968362299
will get worse and worse

alecandido:
@andrea-pasquale @renatomello, there is a Clifford test failing at random:
https://github.com/qiboteam/qibo/actions/runs/8241280424/job/22538218856#step:11:433

alecandido:
And we have a problem with `dtypes` in Keras, generated by the QGAN tests
https://github.com/qiboteam/qibo/actions/runs/8241324205/job/22538355397?pr=1231#step:11:230

Funnily, the problem is on Linux, but not on Windows.

renatomello:
> @BrunoLiegiBastonLiegi @renatomello, there is a Clifford test failing at random: https://github.com/qiboteam/qibo/actions/runs/8241280424/job/22538218856#step:11:433
> 
> (as I wrote somewhere else, I expect that is because you're testing with random values, without fixing a seed - so tests are not reproducible; if you want fuzzying, consider using https://hypothesis.readthedocs.io/en/latest/ instead)

The fix is in #1202

alecandido:
I guess it is https://github.com/qiboteam/qibo/pull/1202/commits/edeff3ada35e83e7d48fc784ae58b6621dce1f75. I could cherry-pick it, but I'm not sure how many things I need from #1202.

renatomello:
> I guess it is [edeff3a](https://github.com/qiboteam/qibo/commit/edeff3ada35e83e7d48fc784ae58b6621dce1f75). I could cherry-pick it, but I'm not sure how many things I need from #1202.

#1202 is close to being merged into master. You could just merge master here when it happens, I guess.

BrunoLiegiBastonLiegi:
> Ok, the Clifford problem has been fixed, now we are stuck on the `dtype`. I will try to inspect the issue, but I don't expect to be related to the Python version (it is failing even before), but to TensorFlow.
> 
> We already encountered it before, correct @BrunoLiegiBastonLiegi? I believe I resorted to pinning the version of TensorFlow, that fixed the version of Keras (that should be the offending one). However, now we'd be stuck because of Python 3.12, because we need TensorFlow >= 2.16 (i.e. 2.16.1, there is nothing else at the moment).

Yep we encountered the same `complex128` dtype problem a couple of weeks ago that temporarily patched by limiting the tensorflow version.

alecandido:
@renatomello for the time being, I dropped the dependency on CVXPY, since it's not even possible to install on Linux without compiling the `qdldl` wheel with CMake (required by the OSQP solver).

However, I did not touch anything in the `diamond_norm` code, and I'm just skipping the related test. If you happen to have `cvxpy` installed, then `diamond_norm` will just work as before.

It seems like there are [other people actively worried about these wheels](https://github.com/cvxpy/cvxpy/issues/2269#issuecomment-1985947222), and the CVXPY team is aware. As soon as they will solve, I will reintegrate the dependency.
However, it is just a missing dependency declaration. Of course, it's not the best, but I believe it is better than requiring compilation to run qibo tests (or to install a qibo extra).

alecandido:
@scarrazza now we have a problem of mutual exclusion due to the new TensorFlow release:

- only TensorFlow 2.16 supports py3.12
- TensorFlow 2.16 is aliasing `tf.keras` to the `keras` package, now Keras 3
  - Keras does not support complex dtypes https://github.com/keras-team/keras/blob/2cb05218baa92accc151ea47f1a198f75508bb35/keras/backend/common/variables.py#L389 (I wonder if it's because of the multi-framework support)

Since the QGAN has been built with keras, and of course it requires complex dtypes, then the (current) QGAN and Python 3.12 are mutually exclusive.

That's the last unfortunate bit missing for Python 3.12 support, everything else seems working.

So, either we decide to drop the QGAN (or have someone rewrite it without the current Keras) or we can wait until complex will possibly appear in Keras.
https://github.com/keras-team/keras/pull/19309
(but it could never happen)

I will focus instead on `qibo-core` and attempt to split the TensorFlow backend to `qiboml`. At least we will unlock Python 3.12 for everything else.

scarrazza:
Thanks for this report. I believe we could re-implement the style-QGAN directly `qiboml` (translating to the new features) and drop this model from `qibo`. @MatteoRobbiati is that feasible? If @bpcarlos is available, he could also help here.

MatteoRobbiati:
> Thanks for this report. I believe we could re-implement the style-QGAN directly `qiboml` (translating to the new features) and drop this model from `qibo`. @MatteoRobbiati is that feasible? If @bpcarlos is available, he could also help here.

I was thinking the same.

alecandido:
Great, then I will coordinate with @MatteoRobbiati to move the QGAN model.

Just let me know if it's better to avoid missing this feature in any release, or whether it's more urgent to release support for Python 3.12 (I don't believe any of the two is really fundamental, but it's not clear to me which is more relevant).

bpcarlos:
Yes, I can help. Let us coordinate @MatteoRobbiati @alecandido, or simply drop me a message with anything I can help with.

alecandido:
@MatteoRobbiati I take back what I said in the meeting, I forgot one step: it is true that we would have a problem in TensorFlow, but only because of `tf.keras`, the support for direct access to complex dtype (without Keras involved) should be unchanged.

So, without the QGAN in Qibo, there should be no obstacle in merging this PR.

src/qibo/quantum_info/metrics.py
```diff
@@ -10,7 +11,7 @@
 
 
 def purity(state):
-    """Purity of a quantum state :math:`\\rho`, which is given by
+    """Purity of a quantum state :math:`\\rho`, which is given by.
```
stavros11, line 14:
```suggestion
    """Purity of a quantum state :math:`\\rho`, which is given by
```
I believe it was caused by some editor extension that is automatically adding `.` at the end of the docstrings, however I don't think this works correctly in these cases since the sentence continues to the equation below.

The same applies to most docstring changes applied to this file, but I will only write here. It may be a bit tedious to fix.

src/qibo/quantum_info/metrics.py
```diff
@@ -574,7 +574,6 @@ def diamond_norm(channel, target=None, backend=None, **kwargs):
 
     .. note::
         This function requires the optional CVXPY package to be installed.
-
     """
     import cvxpy  # pylint: disable=C0415
```
stavros11, line 578:
```suggestion
    import cvxpy
```
is the exception still needed since you added `cvxpy` as pylint ignored module in pyproject.toml? I am not sure.

src/qibo/quantum_info/metrics.py
```diff
@@ -667,8 +666,8 @@ def expressibility(
     order: Optional[Union[int, float, str]] = 2,
     backend=None,
 ):
-    """Returns the expressibility :math:`\\|A\\|` of a parametrized
-    circuit, where
+    """Returns the expressibility :math:`\\|A\\|` of a parametrized circuit,
+    where.
```
stavros11, line 670:
```suggestion
    """Returns the expressibility :math:`\\|A\\|` of a parametrized circuit,
    where
```
here it looks a bit funny

tests/test_quantum_info_metrics.py
```diff
@@ -311,6 +311,7 @@ def test_process_fidelity_and_infidelity(backend):
     )
 
 
+@pytest.mark.skip
```
stavros11, line 314: I'd open an issue linked to https://github.com/qiboteam/qibo/pull/1231#issuecomment-1997932717 to remember that this may be possible to fix at some point.

stavros11 (approved):
Thanks @alecandido. Other than the comments below, it looks good to me.

src/qibo/quantum_info/metrics.py
```diff
@@ -544,22 +546,20 @@ def diamond_norm(channel, target=None, backend=None, **kwargs):
     If a ``target`` channel :math:`\\Lambda` is specified,
     then it calculates :math:`\\| \\mathcal{E} - \\Lambda\\|_{\\diamond}`.
 
-    Example:
+    Example::
```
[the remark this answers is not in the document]
alecandido: The `::` is an alias for `code-block:: <default-language>`

https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks

src/qibo/quantum_info/metrics.py
alecandido: Yes, you're right, I didn't notice it.

I will immediately fix the docstrings.

src/qibo/quantum_info/metrics.py
alecandido: Most likely not, let me try to remove it.

tests/test_quantum_info_metrics.py
alecandido: Feel free :)

stavros11: #1298 1298

src/qibo/quantum_info/metrics.py
```diff
@@ -574,9 +576,8 @@ def diamond_norm(channel, target=None, backend=None, **kwargs):
 
     .. note::
         This function requires the optional CVXPY package to be installed.
-
     """
-    import cvxpy  # pylint: disable=C0415
+    import cvxpy
```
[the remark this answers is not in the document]
alecandido: Done in 2dee38

scarrazza (approved):

PR #1157, opened 2024-01-16
Client backends for the remote execution of circuits

BrunoLiegiBastonLiegi:
This PR add support for the  `QiboClientBackend` and a `QiskitClientBackend` implemented in [https://github.com/qiboteam/qibo-cloud-backends](https://github.com/qiboteam/qibo-cloud-backends) that allow for the remote execution of circuits. Furthermore, support for QASM `gate` command is added to the parser. 

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.

PR #1235, opened 2024-02-28
Add qibotn option

Tankya2:
Hi, this is to link up QiboTN as a backend to Qibo. Under this backend, user will be able to select either CuQuantum or Quimb for tensor network computation. Additional settings will be passed in via the Runcard input.

### DIFF
--- a/src/qibo/backends/__init__.py
+++ b/src/qibo/backends/__init__.py
@@ -36,7 +36,7 @@ def construct_backend(backend, **kwargs):
 
     elif backend == "qibotn":  # pragma: no cover
         from qibotn.backends import CuTensorNet, QuimbBackend  # pylint: disable=E0401
-        
+
         platform = kwargs.get("platform")
         if platform == "cutensornet":  # pragma: no cover
             return CuTensorNet(kwargs["runcard"])
--- a/src/qibo/backends/__init__.py
+++ b/src/qibo/backends/__init__.py
@@ -33,10 +33,10 @@ def construct_backend(backend, **kwargs):
 
     elif backend == "qibolab":  # pragma: no cover
         from qibolab.backends import QibolabBackend  # pylint: disable=E0401
-        
+
     elif backend == "qibotn":  # pragma: no cover
         from qibotn.backends import CuTensorNet, QuimbBackend  # pylint: disable=E0401
- 
+
         if platform == "cutensornet":  # pragma: no cover
             return CuTensorNet(runcard)
         elif platform == "qutensornet":  # pragma: no cover
--- a/src/qibo/backends/__init__.py
+++ b/src/qibo/backends/__init__.py
@@ -36,7 +36,8 @@ def construct_backend(backend, **kwargs):
 
     elif backend == "qibotn":  # pragma: no cover
         from qibotn.backends import CuTensorNet, QuimbBackend  # pylint: disable=E0401
-
+        
+        platform = kwargs.get("platform")
         if platform == "cutensornet":  # pragma: no cover
             return CuTensorNet(kwargs["runcard"])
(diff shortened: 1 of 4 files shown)


alecandido (dismissed):

alecandido:
I changed base to `client_backends`, since in that PR @BrunoLiegiBastonLiegi already added:
https://github.com/qiboteam/qibo/blob/91f85680a1b22c8cb67f5dac83cdc8e880583d3c/src/qibo/backends/__init__.py#L11
and it will soon land in `master`.

You could use that to receive the `runcard` argument you're missing.

src/qibo/backends/__init__.py
```diff
@@ -33,6 +33,14 @@ def construct_backend(backend, platform=None):
     elif backend == "qibolab":  # pragma: no cover
         from qibolab.backends import QibolabBackend  # pylint: disable=E0401
 
+    elif backend == "qibotn":  # pragma: no cover
+        from qibotn.backends import CuTensorNet, QuimbBackend  # pylint: disable=E0401
+
+        if platform == "cutensornet":  # pragma: no cover
+            return CuTensorNet(runcard)
```
alecandido, line 40: `runcard` is currently undefined, but after #1157 you will be able to retrieve it from arbitrary keyword arguments.
```suggestion
            return CuTensorNet(kwargs["runcard"])
```

alecandido:
I just rebased your commit on that branch, to save you the burden.

(in case, I just ran `git rebase --onto client_backends master qibotn_interface`, no more)

src/qibo/backends/__init__.py
Tankya2: Thanks for the rebase, I have made the changes

Tankya2:
> @Tankya2 could you add `qibotn` as a test dependency, and add some minimal test to check that is actually working?
> 
> You could follow the `qibojit` model (but avoid extensive testing, those should be placed in the `qibotn` repository).

Hi @alecandido, sure I have added the dependencies and a minimal test.

tests/test_backend_qibotn.py
```diff
@@ -0,0 +1,18 @@
...
+def test_backend_qibotn(platform):
+
+    qibo.set_backend(backend="qibotn", platform=platform, runcard=None)
+
+    if platform == "qutensornet":
+        qibotn_backend = QuimbBackend(runcard=None)
+    else:
+        qibotn_backend = CuTensorNet(runcard=None)
+
+    assert isinstance(GlobalBackend(), type(qibotn_backend))
```
alecandido, line 18:
```suggestion
    if platform == "qutensornet":
        qibotn_backend = QuimbBackend
    else:
        qibotn_backend = CuTensorNet

    assert isinstance(GlobalBackend(), qibotn_backend)
```

If you're just testing the type, there is no need to create a full instance :)

alecandido:
When you change the dependencies in the `pyproject.toml` file you need to regenerate the lock file (i.e. `poetry.lock`) with `poetry lock` or `poetry update` command.
That's why tests are failing.
https://github.com/qiboteam/qibo/actions/runs/8106116994/job/22155525745?pr=1235#step:8:19

I'm going to do it now.

tests/test_backend_qibotn.py
alecandido: Applied in 583d187

alecandido:
Workflows are failing, because the dependency is pointing to the master branch of `qibotn`, where there is no backend definition.

liweintu:
> Workflows are failing, because the dependency is pointing to the master branch of `qibotn`, where there is no backend definition.

Now that qiboteam/qibotn#24 has been merged, the `qibotn` backend should be available to fix the dependency.

PR #1236, opened 2024-02-29
Fix `cuquantum` test bug in `test_backends_clifford.py`

renatomello:
Checklist:
- [x] Tests are passing.

### DIFF
--- a/tests/test_backends_clifford.py
+++ b/tests/test_backends_clifford.py
@@ -4,12 +4,7 @@ import numpy as np
 import pytest
 
 from qibo import Circuit, gates, set_backend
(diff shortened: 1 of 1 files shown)


Simone-Bordoni (approved):

alecandido:
@renatomello I'd say that a better fix would be to decouple the Clifford engine from the other backends.
By now, it is not really reusing anything of the backends. So, I'd just parametrize tests on `engine`, and spell out which are the available engines.

However, if you prefer to have this immediately in `master` as a hotfix, I have nothing against it.

BrunoLiegiBastonLiegi (approved):

PR #1238, opened 2024-02-29
Fix tests for `cupy` and `cuquantum` backends

renatomello:
Checklist:
- [x] Tests are passing.


tests/test_quantum_info_clifford.py
```diff
@@ -17,11 +17,12 @@
 
 
 def construct_clifford_backend(backend):
-    if isinstance(backend, TensorflowBackend):
+    if backend.__class__.__name__ in ["TensorflowBackend", "CuQuantumBackend"]:
```
alecandido, line 20:
```suggestion
    if isinstance(backend, (TensorflowBackend, CuQuantumBackend)):
```

`isinstance` works even for multiple options (of course they are in OR)

tests/test_models_encodings.py
```diff
@@ -125,17 +125,22 @@ def test_unary_encoder(backend, nqubits, architecture, kind):
...
     data = 2 * sampler.random(nqubits) - 1
-    data = backend.cast(data, dtype=data.dtype)
+    data = data.tolist() if kind is not None else backend.cast(data, dtype=data.dtype)
 
-    if kind is not None:
-        data = kind(data)
+    print(type(data), type(data[0]))
+
+    # if kind is not None:
+    #     data = kind(data)
```
alecandido, line 133: I'm not sure I understand what is happening here, but I guess it is also work-in-progress.

alecandido (commented):
Thanks @renatomello, I see that it is not completed yet, but it already seems a simplification.

As soon as you complete, I'll be glad to approve!

tests/test_quantum_info_clifford.py
renatomello: I specifically avoided `isinstance` so not to have to import `CuQuantumBackend` from `qibojit` on a `qibo` test. Also because github is not installing `cuquantum` to test, so it would fail anyways.

tests/test_models_encodings.py
renatomello: The function needs to be tested on inputs that are either `list` or an `array`. The print is just part of the work in progress.

alecandido: Ok. I see.

Often, in this kind of situations, you can use the property of `np.array()` (and similar, I guess) being idempotent. So, if you want to ensure a list, you first ensure it is an array, i.e., `np.array(data).tolist()`.
I would discourage this in the package code, because it is much better to have clear input types, rather than converting them everywhere (reduce bloating), but for tests is alright.

(I was not commenting the `print()`, it was a marker that was work-in-progress, as the commented code, but that is perfectly fine in a draft)

renatomello: > Ok. I see.
> 
> Often, in this kind of situations, you can use the property of `np.array()` (and similar, I guess) being idempotent. So, if you want to ensure a list, you first ensure it is an array, i.e., `np.array(data).tolist()`. I would discourage this in the package code, because it is much better to have clear input types, rather than converting them everywhere (reduce bloating), but for tests is alright.
> 
> (I was not commenting the `print()`, it was a marker that was work-in-progress, as the commented code, but that is perfectly fine in a draft)

It is not part of the code, just the tests because of how the input `data` is generated.

tests/test_quantum_info_clifford.py
alecandido: I acknowledge the problem, though I still believe that is not the optimal solution in general. But it could be for this PR.

I already wrote this somewhere else, but it would be better that tests would be shipped together with the code. So, we should face `cuquantum` issues only on the `qibojit` side.
(but we will improve this, starting from `qibo-core`, I just need some time to get back to it)
In `qibojit`, we could gate tests with marks, and possibly mock if it is too entangled.

But all of this is overly complex for this PR, so maybe let's just keep your original solution.

renatomello:
Some tests in `tests/test_models_error_mitigation.py` fail on GPU only if the tests are ran in sequence. If the tests are ran individually, the tests pass. This is probably an issue with `GlobalBackend` being set up by a previous test

renatomello:
@MatteoRobbiati the majority of DBI tests are failing on GPU. It seems that it is because `self.backend` is not setup properly in the `DoubleBracketIteration` class. That would also explain the need for all those `self.backend.to_numpy` calls inside the class.

alecandido:
> Some tests in `tests/test_models_error_mitigation.py` fail on GPU only if the tests are ran in sequence. If the tests are ran individually, the tests pass. This is probably an issue with `GlobalBackend` being set up by a previous test

If you are receiving the global backend from a *fixture*, as in here:
https://github.com/qiboteam/qibo/blob/31bbbb416789a315cac0844816d5793b40802436/tests/conftest.py#L57-L59
you could add some cleanup code after the `yield`.

andrea-pasquale:
> Some tests in `tests/test_models_error_mitigation.py` fail on GPU only if the tests are ran in sequence. If the tests are ran individually, the tests pass. This is probably an issue with `GlobalBackend` being set up by a previous test

I also saw the same behavior some time ago, very weird.



> @MatteoRobbiati the majority of DBI tests are failing on GPU. It seems that it is because `self.backend` is not setup properly in the `DoubleBracketIteration` class. That would also explain the need for all those `self.backend.to_numpy` calls inside the class.

You are right, given that we don't tests on GPU with the CI sometimes we forget. I can also have a look if you want @MatteoRobbiati.

renatomello:
It's better to move forward with this PR only after https://github.com/qiboteam/qibo/pull/1202 is merged

renatomello:
All tests should be passing on GPUs now. Someone else should test locally too.

PR #1239, opened 2024-03-01
`CliffordBackend` with packed bits

BrunoLiegiBastonLiegi:
This PR introduces a further optimization for the `CliffordBackend`. Namely the bits of the symplectic matrix are now packed as `uint8` arrays by the `np.packbits` function.

Checklist:
- [ ] Tests are passing.


src/qibo/gates/abstract.py
```diff
@@ -391,7 +391,7 @@ def apply_density_matrix(self, backend, state, nqubits):
         return backend.apply_gate_density_matrix(self, state, nqubits)
 
     def apply_clifford(self, backend, state, nqubits):
-        return backend.apply_gate_clifford(self, state[:-1], nqubits)
+        return backend.apply_gate_clifford(self, state, nqubits)
```
alecandido, line 394: Why did you remove this?

BrunoLiegiBastonLiegi: Because index `-1` now is not the scratch row anymore, but rather the 8 bit integer that also include the scratch but it's not limited to it.

alecandido: Ok, it definitely makes sense. I was just confused thinking that it was cropping the phases column, not the scratch row.

src/qibo/backends/_clifford_operations.py
```diff
@@ -424,22 +424,30 @@ def _random_outcome(state, p, q, nqubits):
...
 
 
 # valid for a standard basis measurement only
 def M(state, qubits, nqubits, collapse=False):
     sample = []
-    state_copy = state if collapse else state.copy()
+    if collapse:
+        state = np.unpackbits(state, axis=0)
+    else:
+        state = state.copy()
```
renatomello, line 442:
```suggestion
    state = np.unpackbits(state, axis=0) if collapse else state.copy()
```

src/qibo/backends/_clifford_operations.py
```diff
@@ -454,11 +462,14 @@ def cast(x, dtype=None, copy=False):
     return np.array(x, dtype=dtype, copy=copy)
 
 
-def _clifford_pre_execution_reshape(state):
+def _clifford_pre_execution_reshape(state, pack=False):
```
renatomello, line 465:
```suggestion
def _clifford_pre_execution_reshape(state, pack: bool = False):
```
Also maybe add a docstring since there are more arguments in the funciton

src/qibo/backends/_clifford_operations.py
```diff
@@ -454,11 +462,14 @@ def cast(x, dtype=None, copy=False):
...
+    if pack:
+        state = np.packbits(state, axis=0)
     return state
 
 
 def _clifford_post_execution_reshape(state, nqubits):
```
renatomello, line 471:
```suggestion
def _clifford_post_execution_reshape(state, nqubits: int):
```
Same here

src/qibo/backends/clifford.py
```diff
@@ -106,7 +106,7 @@ def zero_state(self, nqubits: int):
         symplectic_matrix[nqubits:-1, nqubits : 2 * nqubits] = self.np.copy(identity)
         return symplectic_matrix
 
-    def _clifford_pre_execution_reshape(self, state):
+    def _clifford_pre_execution_reshape(self, state, pack=False):
```
renatomello, line 109:
```suggestion
    def _clifford_pre_execution_reshape(self, state, pack: bool = False):
```

src/qibo/backends/_clifford_operations.py
```diff
@@ -341,13 +341,7 @@ def _exponent(x1, z1, x2, z2):
...
         (np.array): The calculated exponents.
     """
-    x1_ = x1.astype(np.int8)
-    x2_ = x2.astype(np.int8)
-    z1_ = z1.astype(np.int8)
-    z2_ = z2.astype(np.int8)
-    return (
-        2 * (x1_ * x2_ * (z2_ - z1_) + z1_ * z2_ * (x1_ - x2_)) - x1_ * z2_ + x2_ * z1_
-    )
+    return 2 * (x1 * x2 * (z2 - z1) + z1 * z2 * (x1 - x2)) - x1 * z2 + x2 * z1
```
andrea-pasquale, line 344: A type hint for this function could be useful.

src/qibo/backends/_clifford_operations.py
```diff
@@ -390,57 +384,108 @@ def _rowsum(symplectic_matrix, h, i, nqubits, determined=False):
...
-    return state[nqubits:-1, q].nonzero()[0]
+@cache
+def _get_dim(nqubits):
+    """Returns the dimension of the symplectic matrix for a given number of qubits."""
+    return 2 * nqubits + 1
+
+
+@cache
+def _get_packed_size(n):
+    """Returns the size of an arry of `n` booleans after packing."""
```
andrea-pasquale, line 433:
```suggestion
    """Returns the size of an array of `n` booleans after packing."""
```

src/qibo/backends/_clifford_operations.py
```diff
@@ -390,57 +384,108 @@ def _rowsum(symplectic_matrix, h, i, nqubits, determined=False):
...
+    return np.ceil(n / 8).astype(int)
```
andrea-pasquale, line 434: I saw 8 appearing more than once. Could we define it as a top level constant?

src/qibo/backends/_clifford_operations.py
```diff
@@ -390,57 +384,108 @@ def _rowsum(symplectic_matrix, h, i, nqubits, determined=False):
...
+
+
+def _packbits(array, axis):
+    return np.packbits(array, axis=axis)
+
+
+def _unpackbits(array, axis):
+    return np.unpackbits(array, axis=axis)
```
andrea-pasquale, line 442: What is the purpose of renaming these numpy primitives?

pyproject.toml
```diff
@@ -69,7 +69,7 @@ pylint = "^3.0.3"
 matplotlib = "^3.7.0"
 tensorflow = { version = "^2.14.1,<2.16", markers = "sys_platform == 'linux'" }
 torch = "^2.1.1"
-qibojit = { git = "https://github.com/qiboteam/qibojit.git" }
+qibojit = { git = "https://github.com/qiboteam/qibojit.git", branch="clifford_simulator_numba" }
```
andrea-pasquale, line 72: Just a reminder to avoid merging until we merge the corresponding PR in qibojit

andrea-pasquale (commented):
Thanks @BrunoLiegiBastonLiegi. I have some minor suggestions.

src/qibo/backends/_clifford_operations.py
BrunoLiegiBastonLiegi: Yes in principle we could, but I don't know if it makes any sense. This 8 is just the size scaling between bits and bytes.

src/qibo/backends/_clifford_operations.py
BrunoLiegiBastonLiegi: This is there mainly for the gpu implementation, which will overwrite these two. The plan was actually to have the `_pack/_unpack_for_measurements` functions only here and just use them with the corresponding `_pack/unpack_bits` . This is the same mechanism that is in place for the `M`, that is defined here only, however, in the end I had to copy the `_pack/_unpack_for_measurements` in the gpu module anyway because for some reason that is unclear to me it did not work. I might want to try it again. Anyway, if in the end we keep the `_pack/_unpack_for_measurements` in the gpu module as well, then I agree we can get rid fo the `_unpackbits` function, but the `_packbits` is needed by the `M` function below, thus we have to keep it.

renatomello (approved):

BrunoLiegiBastonLiegi:
@alecandido I was following your proposed order in https://github.com/qiboteam/qibo/pull/1231#issuecomment-1968362299 for this PR. However, now that in main the qibojit test dependency points to the branch `py3.12`, I cannot do the same with this PR, right? Because I would need to point to another qibojit branch. One possible solution, I guess, would be to merge my qibojit PR into `py3.12`.
Maybe, while we wait for `qibo-core` to be ready, we could set up a `tmp` branch in each repo for handling this situations.

alecandido:
In this case, the solution would be to rebase your PR (https://github.com/qiboteam/qibojit/pull/173) on top of Qibojit's `py3.12`, and you can point this PR to that branch.

And yes, it's not great, but now PRs are rather coupled. So, to avoid broken tests, we should first merge https://github.com/qiboteam/qibojit/pull/170

src/qibo/backends/_clifford_operations.py
```diff
@@ -329,7 +329,7 @@ def ECR(symplectic_matrix, control_q, target_q, nqubits):
     return X(symplectic_matrix, control_q, nqubits)
 
 
-def _exponent(x1, z1, x2, z2):
+def _exponent(x1: np.array, z1: np.array, x2: np.array, z2: np.array) -> np.array:
```
alecandido, line 332: In case you care, `np.array` is a function, so it's not suitable as a type hint.

The correct type would be `np.ndarray` (just try `type(np.array([]))`), but an even better typing support is coming from the `numpy.typing` module:
https://numpy.org/doc/stable/reference/typing.html

src/qibo/backends/_clifford_operations.py
```diff
@@ -390,57 +384,108 @@ def _rowsum(symplectic_matrix, h, i, nqubits, determined=False):
...
+def _unpackbits(array, axis):
+    return np.unpackbits(array, axis=axis)
+
+
+def _pack_for_measurements(state, nqubits):
+    """Prepares the state for measurements by packing the rows of the X and Z sections of the symplectic matrix."""
+    r, x, z = _get_rxz(state, nqubits)
+    x = _packbits(x, axis=1)
+    z = _packbits(z, axis=1)
+    return np.hstack((x, z, r[:, None]))
```
alecandido, line 450: Why are you packing `x` and `z` separately?

src/qibo/backends/_clifford_operations.py
```diff
@@ -454,11 +499,32 @@ def cast(x, dtype=None, copy=False):
...
+    Returns:
+        (np.array) The packed and reshaped state.
+    """
+    if pack:
+        state = _packbits(state, axis=0)
     return state
 
 
-def _clifford_post_execution_reshape(state, nqubits):
+def _clifford_post_execution_reshape(state, nqubits: int):
```
alecandido, line 517: This is asymmetric with respect to the one above, i.e. there is no `pack` flag (and it is always unpacking).

What's the reason?

src/qibo/backends/_clifford_operations.py
alecandido: I guess the motivation is the `_rowsum` function above, in which you need to treat separately `x`, `z`, and `r`, but just to get a confirmation.

src/qibo/backends/_clifford_operations.py
```diff
@@ -390,57 +384,108 @@ def _rowsum(symplectic_matrix, h, i, nqubits, determined=False):
...
+def _pack_for_measurements(state, nqubits):
+    """Prepares the state for measurements by packing the rows of the X and Z sections of the symplectic matrix."""
+    r, x, z = _get_rxz(state, nqubits)
+    x = _packbits(x, axis=1)
+    z = _packbits(z, axis=1)
+    return np.hstack((x, z, r[:, None]))
+
+
+@cache
+def _get_pad_size(n):
```
alecandido, line 454: Stupid naming remark:
```suggestion
def _pad_size(n):
```
since `get_` doesn't add anything (other than 4 characters :P)

alecandido: Same for `_get_dim` and `_get_packed_size` above

alecandido (approved):
I didn't really check all the corners thoroughly (i.e. rechecking the meaning of existing functions, and that has been preserved), but I read the whole diff and I only have stylistic suggestions, nothing else

alecandido:
Of course, whenever you can, merge/rebase on `master`

src/qibo/backends/_clifford_operations.py
BrunoLiegiBastonLiegi: Yes precisely, because a single row of the matrix contains the encoding for both x, z and r, thus you cannot pack the complete row in one go but pack the different sections of the row separately.

BrunoLiegiBastonLiegi:
I believe there is a problem with the `qibotn` dependency, the `py3.12` branch needs to be updated as it still points to the `qibo/py3.12` branch that doesn't exist anymore (I presume) and this make the `poetry lock` crash.

alecandido:
> I believe there is a problem with the `qibotn` dependency, the `py3.12` branch needs to be updated as it still points to the `qibo/py3.12` branch ([qiboteam/qibotn@`3718d5c`/pyproject.toml#L24C1-L25C1](https://github.com/qiboteam/qibotn/blob/3718d5c8bca63f63bfd42234a2679adf9776951e/pyproject.toml#L24C1-L25C1)) that doesn't exist anymore (I presume) and this make the `poetry lock` crash.

I saw you updated it, thanks.

Simply, no one touched it after the merge.

alecandido:
> due to some problem with the seed most likely.

Is the seed not fixed explicitly anywhere? Are we relying on some default?

BrunoLiegiBastonLiegi:
Yes the seed is set https://github.com/qiboteam/qibo/blob/036e882adfd7878b566e90eae02ff2c72e16e296/tests/test_models_circuit_features.py#L321
but for some reason pytorch yields different results with different systems using the same seed.

I believe the problem exists in main as well, as I see the test failing in #1310 as well. Weirdly, this came out all of a sudden...

src/qibo/backends/_clifford_operations.py
BrunoLiegiBastonLiegi: I double checked and indeed the pack arg can be dropped

alecandido:
@scarrazza just provided #1317 as a (temporary) solution for the test problem. I'd suggest merging/rebasing on it (or just wait that it will be merged in `master`)

PR #1241, opened 2024-03-01
Polish Qibo cloud

### LINKED ISSUE #1240: Change qibo-cloud with qibo-client

alecandido:
Sorry @BrunoLiegiBastonLiegi, there is some aftermath after closing #1157, I didn't notice during the review.

- [x] rename backend `"qibo-client"`
- [x] drop the `qibo-cloud-backends` dependency

Closes #1240


alecandido:
The motivation for the rename is that `qibo-cloud` is the global client supporting remote execution, but on multiple backends, while the Qibo server is really `qibo-client`, so that's the backend.

(you can think as `qibo-cloud-qiskit` and `qibo-cloud-qibo-client`, with the initial `qibo-cloud-` omitted)

Sorry not to have pointed out when we discussed the naming in #1157

BrunoLiegiBastonLiegi:
> The motivation for the rename is that `qibo-cloud` is the global client supporting remote execution, but on multiple backends, while the Qibo server is really `qibo-client`, so that's the backend.
> 
> (you can think as `qibo-cloud-qiskit` and `qibo-cloud-qibo-client`, with the initial `qibo-cloud-` omitted)
> 
> Sorry not to have pointed out when we discussed the naming in #1157

No, indeed, I was planning of changing it in the `MetaBackend` PR anyway.

alecandido:
The rationale for the dependency removal is instead the same of `qibojit`: otherwise it would be circular.

Currently (before `qibo-core` will land), the dependency on `qibo` by external backends has to be explicit. Then, `qibo` is not going to depend on them.
Following the discussion in #822 most likely the explicit dependency on the external backends won't happen even afterward, but `qibo` will *automatically discover* dependencies (scoped import) when they are installed.

BrunoLiegiBastonLiegi (approved):

scarrazza:
@alecandido thanks. At the end you are not moving the cloud-backends dependency to tests, right?

scarrazza (approved):

alecandido:
> @alecandido thanks. At the end you are not moving the cloud-backends dependency to tests, right?

I scan through it, and it seems the two backends are not currently used in tests, thus there is no need for the dependency

scarrazza:
Ok, thanks, could you please fix the conflict and merge?

alecandido:
> Ok, thanks, could you please fix the conflict and merge?

Conflicts fixed, I'm just waiting for the tests to pass

PR #1243, opened 2024-03-01
 Add qibotn option

alecandido:
@Tankya2 @liweintu sorry for the mess: I was cleaning up the branches of the merged PRs, and I didn't notice that #1235 was not pointing to `master` automatically, but still pointing to the old branch (that I deleted).


Tankya2:
> @Tankya2 @liweintu sorry for the mess: I was cleaning up the branches of the merged PRs, and I didn't notice that #1235 was not pointing to `master` automatically, but still pointing to the old branch (that I deleted).

I see, no worries @alecandido

liweintu (approved):

liweintu:
> @Tankya2 @liweintu sorry for the mess: I was cleaning up the branches of the merged PRs, and I didn't notice that #1235 was not pointing to `master` automatically, but still pointing to the old branch (that I deleted).

No worries. Thanks for cleaning up the branches. Actually, I should've flagged it up 'cos I noticed it's not pointing to `main`. However, I thought it was intentional then.

Anyway, I approved it now.

tests/test_backend_qibotn.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+from qibotn.backends import CuTensorNet, QuimbBackend
```
liweintu, line 2: Importing `CuTensorNet` will fail the CI test because the GitHub runner doesn't have `cuquantum` for GPUs.

@alecandido Shall we take it out and only test `QuimbBackend` for now? We can add it back once we have self-hosted GPU machines ready as the runners.

scarrazza: Yes, I think so.

scarrazza (approved):
Looks good but we have to fix tests.

tests/test_backend_qibotn.py
liweintu: We tried a few ways to remove `cuquantum`, but even only importing `quimb` will inevitably trigger importing of `cuquantum`. So it's a bit tricky to test `quimb` without involving `cuquantum`.

scarrazza: I have remove the import from the test in this PR, however in https://github.com/qiboteam/qibotn/blob/main/src/qibotn/backends/__init__.py we are importing cuquantum. We should change that in the qibotn repo first.

liweintu: Yeah, indeed. It's doable, but after taking `cuquantum` out from `__init__.py`, there're some files that need to be modified to import `cuquantum` in a not-so-convenient way.

liweintu: As @Tankya2 pointed out, this `test_backend_qibotn.py` file works if there's a GPU runner. Does it make sense if we use a self-hosted GPU runner for this?

liweintu: To catch up with the release, could we temporarily skip the test of qibotn for now, and once the GPU runners are set up, we enable the test again?

scarrazza: As you can see, after my commits, tests are not crashing due to cuquantum. However quimb and qibojit are interfering with each other through numba. Most likely @Tankya2 and @Vinitha-balachandran have already faced this problem in the past...

scarrazza:
@alecandido could you please double check my latest fix? If you agree we merge and finally unlock the qibo release.

src/qibo/backends/__init__.py
```diff
@@ -34,6 +34,18 @@ def construct_backend(backend, **kwargs):
...
         from qibolab.backends import QibolabBackend  # pylint: disable=E0401
 
+    elif backend == "qibotn":  # pragma: no cover
+
+        platform = kwargs.get("platform")
+        if platform == "cutensornet":  # pragma: no cover
+            from qibotn.backends.cutensornet import CuTensorNet  # pylint: disable=E0401
+
+            return CuTensorNet(kwargs["runcard"])
+        elif platform == "qutensornet":  # pragma: no cover
```
scarrazza, line 44: Btw, at this point we can remove this pragma.

alecandido:
> @alecandido could you please double check my latest fix? If you agree we merge and finally unlock the qibo release.

Yes, that's exactly how I remembered it.

tests/test_backend_qibotn.py
liweintu: @Tankya2 and I didn't see this before.
@Vinitha-balachandran suggested setting the backend to numpy only for quimb test. I noticed the backends to be tested on are defined in `conftest.py` below.
```py
# backends to be tested
BACKENDS = ["numpy", "tensorflow", "qibojit-numba", "qibojit-cupy", "qibojit-cuquantum"]
```
Is it possible to only select `numpy` backend for the quimb test for now, so as to avoid the involvement with `qibojit-numba`?

scarrazza: @liweintu this is now fixed (see my last commit). We can know test quimb with numba.

liweintu: @scarrazza That's great! Thanks a lot for the help on the fix.

PR #1244, opened 2024-03-02
refactor `QuantumNetwork` to support general quantum networks

Canoming:
Checklist:
- [x] Tests are passing.

I've refactored the `QuantumNetworks` to support general quantum networks and make it more compatible with usual tensor contraction methods.

To make things clear, I made a class diagram for the future development of the module

class diagram


Canoming:
The build of the `wheel` fails after I merge the `master` into this branch. Not sure what causes the problem

renatomello:
@Canoming I'm sorry for my absence. I can't review this PR this week but I can next week.

Canoming:
Hi, @AlejandroSopena , I've updated the code according to your comments. The `Readme.md` file is a deprecated 'proof of principle' demonstration for earlier development. I accidentally committed that file to the branch, and it's deleted now. I'm working on a more practical example for the `quantum_network`, which will be added to a separate PR for documentations later.

Canoming:
Hi, @AlejandroSopena , I've updated the code according to your comments. The `Readme.md` file is a deprecated 'proof of principle' demonstration for earlier development. I accidentally committed that file to the branch. I'm working on a more practical example for the `quantum_network`, which will be added to a separate PR for documentation later.

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +565,497 @@ def _run_checks(self, partition, system_output, pure):
...
+            TypeError,
+            f"subscripts must be type str, but it is type {type(subscripts)}.",
         )
+
+    for i, operand in enumerate(operands):
+        if not isinstance(operand, QuantumNetwork):
+            raise_error(TypeError, f"The {i}th operator is not a ``QuantumNetwork``.")
+
+    if backend is None:
+        backend = operands[0]._backend
```
AlejandroSopena, line 972:
```suggestion
    if backend is None: # pragma: no cover
        backend = operands[0]._backend
```

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +565,497 @@ def _run_checks(self, partition, system_output, pure):
...
-        """Checks if input subscript match any implemented pattern."""
-        braket = "[a-z]"
-        pattern_two = re.compile(braket * 2 + "," + braket * 2 + "->" + braket * 2)
-        pattern_four = re.compile(braket * 4 + "," + braket * 2 + "->" + braket * 2)
+def link_product(
+    subscripts: str = "ij,jk -> ik",
+    *operands: QuantumNetwork,
+    backend=None,
+    surpress_warning=False,
+    casting: bool = True,
```
AlejandroSopena, line 940: It seems `casting` is not needed.

Canoming: Ah, sure. I tried to implement the functions that cast the output of the link product to quantum channels if possible. But later I found the casting may mislead the users and I need to check many complex cases to prevent bugs. This functionality is removed now.

AlejandroSopena (approved):

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +565,497 @@ def _run_checks(self, partition, system_output, pure):
...
+    Args:
+        dim (int): Dimension of the identity operator.
+        backend (:class:`qibo.backends.abstract.Backend`, optional): Backend to be used
+            to return the identity operator. If ``None``, defaults to the backend defined
+            when initializing the :class:`qibo.quant
+    """
+
+    return QuantumChannel.from_nparray(
+        np.eye(dim), [dim, dim], pure=True, backend=backend
+    )
```
renatomello, line 1046: This `identity` function does not seem to be used anywhere. Do we need it?

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +565,497 @@ def _run_checks(self, partition, system_output, pure):
...
+    Args:
+        dim (int): Dimension to be traced.
+        backend (:class:`qibo.backends.abstract.Backend`, optional): Backend to be used
+            to return the identity operator. If ``None``, defaults to the backend defined
+            when initializing the :class:`qibo.quant
+    """
+
+    return QuantumNetwork.from_nparray(
+        np.eye(dim), [dim], [True], pure=False, backend=backend
+    )
```
renatomello, line 1061: This seems like it should be a method of the `QuantumNetwork` class. Also, in principle, this should allow for multiple dimensions, no?

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +565,497 @@ def _run_checks(self, partition, system_output, pure):
...
+
+        dim_out = self.partition[-1]
+        dim_in = self.partition[-2]
+
+        trace_out = backend.cast(
+            trace(dim_out, backend=backend).full(), dtype=self._tensor.dtype
+        )
+        trace_in = backend.cast(
+            trace(dim_in, backend=backend).full(), dtype=self._tensor.dtype
+        )
```
renatomello, line 722: If `trace` accepts `backend` as argument, then the resulting object should already be returned in the correct backend and the `cast` outside should not be needed.

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +565,497 @@ def _run_checks(self, partition, system_output, pure):
...
+            return False
+        elif len(self.partition) == 2:
+            return True
+        else:
+            return QuantumComb(
+                sub_comb, self.partition[:-2], pure=False, backend=self._backend
+            ).is_causal(order, precision_tol)
+
+    @classmethod
+    def from_nparray(
```
renatomello, line 745: Since the class accepts arrays of all backends, the method should be renamed to something more agnostic like `from_array` (or `from_tensor` or another suggestion).

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +565,497 @@ def _run_checks(self, partition, system_output, pure):
...
+
+        dim_out = self.partition[1]
+        dim_in = self.partition[0]
+
+        trace_out = backend.cast(
+            trace(dim_out, backend=backend).full(), dtype=self._tensor.dtype
+        )
+        trace_in = backend.cast(
+            trace(dim_in, backend=backend).full(), dtype=self._tensor.dtype
+        )
```
renatomello, line 858: Same thing here about casting

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +565,497 @@ def _run_checks(self, partition, system_output, pure):
...
 
         if self.is_pure():
-            matrix = self._einsum("jk,lm -> kjml", matrix, np.conj(matrix))
+            return self._einsum("ij,lk,il", operator, conj(operator), state)
+
+        return self._einsum("ijkl, jl", operator, state)
+
+
+class StochQuantumNetwork:
+    pass
```
renatomello, line 932: ?

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +565,497 @@ def _run_checks(self, partition, system_output, pure):
...
+                            warning(
+                                f"Index {ind} connects two {'input' if is_input else 'output'} systems."
+                            )
+                    is_input = operands[inds[i]].system_input[index]
+                    if found > 2:
+                        warning(
+                            f"Index {ind} appears multiple times in the input subscripts {input_str}."
+                        )
+                except:
+                    continue
```
renatomello, line 1013: Which exception are we looking to bypass exactly?

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +565,497 @@ def _run_checks(self, partition, system_output, pure):
...
+    for ind in results_index:
+        for i, script in enumerate(inputs):
+            try:
+                index = script.index(ind)
+
+                partition.append(operands[inds[i]].partition[index])
+                system_input.append(operands[inds[i]].system_input[index])
+
+            except:
+                continue
```
renatomello, line 1027: Same here

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +565,497 @@ def _run_checks(self, partition, system_output, pure):
...
+
+    if not isinstance(subscripts, str):
+        raise_error(
+            TypeError,
+            f"subscripts must be type str, but it is type {type(subscripts)}.",
         )
+
+    for i, operand in enumerate(operands):
+        if not isinstance(operand, QuantumNetwork):
+            raise_error(TypeError, f"The {i}th operator is not a ``QuantumNetwork``.")
```
renatomello, line 969:
```suggestion
            raise_error(TypeError, f"The {i}-th operator is not a ``QuantumNetwork``.")
```

src/qibo/quantum_info/quantum_networks.py
renatomello, line 940: `backend` and `casting` arguments are redundant

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +565,497 @@ def _run_checks(self, partition, system_output, pure):
...
+        )
+        for operand in operands
+    ]
+
+    # keep track of the `partition` and `system_input` of the network
+    _, contracrtion_list = np.einsum_path(
+        subscripts, *tensors, optimize=False, einsum_call=True
+    )
+
+    inds, idx_rm, einsum_str, remaining, blas = contracrtion_list[0]
```
renatomello, line 988: variables `remaining` and `blas` are not used anywhere after this

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +565,497 @@ def _run_checks(self, partition, system_output, pure):
...
+    :math:`A.\\textup{link_product}(B)` means "applying :math:`B` to :math:`A`".
+    However, the ``link_product`` is associative, so we override the `@` operation
+    in order to simplify notation.
+
+    Args:
+        subscripts (str, optional): Specifies the subscript for summation using
+            the Einstein summation convention. For more details, please refer to
+            `numpy.einsum <https://numpy.org/doc/stable/reference/generated/numpy.einsum.html>`_.
+        operands (:class:`qibo.quantum_info.quantum_networks.QuantumNetwork`): Quantum
+            networks to be contracted.
```
renatomello, line 954: `supress_warnings` and `backend` are missing in the docstring

src/qibo/quantum_info/quantum_networks.py
Canoming: Both the `identity` and the `trace` are special cases of `QuantumNetwork`. They are important tools in constructing quantum networks. So I provide them here for convenience.

src/qibo/quantum_info/quantum_networks.py
Canoming: I haven't tested the constructor from other data types. I'm not sure if this function can construct `QuantumNetwork` from all data types properly. I remember I used some `numpy` function in the intermediate steps. We may need more tests if we want it to be universal.

src/qibo/quantum_info/quantum_networks.py
Canoming: Sure. I'll delete that

src/qibo/quantum_info/quantum_networks.py
Canoming: Stochastic Quantum Networks are planned but not to be implemented in this PR. Do we need to delete it here?

renatomello: Yes, it's best to delete from this PR and in the future create a separate PR for it

src/qibo/quantum_info/quantum_networks.py
Canoming: Ah, this is for checking the index of each letter. As the letter may not be in the script, in which case nothing needs to be done, I simply suppress the error for `script.index(ind)`

src/qibo/quantum_info/quantum_networks.py
renatomello: Excluding something out of the norm, everything in the `quantum_info` module (and `qibo` in general) should be backend agnostic.

src/qibo/quantum_info/quantum_networks.py
renatomello: It seems like the best way to implement them would be an `IdentityNetwork` class and a `TraceNetwork` class, both inheriting from `QuantumNetwork`

src/qibo/quantum_info/quantum_networks.py
Canoming: I've updated the code regarding most of the comments. But I found this `cast` is still necessary as the data type between the trace operator and the target operator may not be the same (the trace operator only contains real number, but the target tensor can be complex)

src/qibo/quantum_info/quantum_networks.py
Canoming: The `casting` was here for casting a one-input-one-output `QuantumNetwork` to a `QuantumChannel`. But I've deleted this feature as it's very hard to check all boundary cases. I've removed this argument here.

src/qibo/quantum_info/quantum_networks.py
renatomello: Even if the tensor contains only real numbers, its best to leave it as complex `dtype` by default in order to avoid these issues. If an user wants to change the `dtype` for whatever reason, leave it to them.

Canoming:
@renatomello Hi, the comments should be all resolved

tests/test_quantum_info_quantum_networks.py
```diff
@@ -113,14 +126,47 @@ def test_errors(backend):
     with pytest.raises(ValueError):
         QuantumNetwork(matrix, (1, 2), backend=backend)
 
+    with pytest.raises(ValueError):
+        QuantumNetwork.from_operator(matrix, (1, 2), pure=True, backend=backend)
+
+    vec = np.random.rand(4)
```
renatomello, line 132:
```suggestion
    vec = np.random.rand(4)
    vec = backend.cast(vec, dtype=vec.dtype)
```

tests/test_quantum_info_quantum_networks.py
```diff
@@ -283,11 +412,89 @@ def test_non_hermitian_and_prints(backend):
...
     assert not network.is_hermitian()
-    assert not network.is_causal()
     assert not network.is_positive_semidefinite()
-    assert not network.is_channel()
 
-    assert network.__str__() == "J[4 -> 4]"
+    assert network.__str__() == "J[┍4┑, ┕4┙]"
+
+
+def test_uility_func():
```
renatomello, line 425:
```suggestion
def test_utility_func():
```

tests/test_quantum_info_quantum_networks.py
```diff
@@ -283,11 +412,89 @@ def test_non_hermitian_and_prints(backend):
...
+    traced = link_product("ij,j", id_ch, tr_ch, backend=backend)
+
+    backend.assert_allclose(
+        tr_ch.matrix(backend=backend), traced.matrix(backend=backend), atol=1e-8
+    )
+
+
+def test_default_construction(backend):
+    vec = np.random.rand(4).reshape([4, 1])
+    mat = np.random.rand(16).reshape([2, 2, 2, 2])
```
renatomello, line 466:
```suggestion
    vec = np.random.rand(4).reshape([4, 1])
    mat = np.random.rand(16).reshape([2, 2, 2, 2])
    vec = backend.cast(vec, dtype=vec.dtype)
    mat = backend.cast(mat, dtype=mat.dtype)
```

tests/test_quantum_info_quantum_networks.py
```diff
@@ -283,11 +412,89 @@ def test_non_hermitian_and_prints(backend):
...
+    tensor = np.random.rand(16).reshape([4, 4])
```
renatomello, line 467:
```suggestion
    vec = np.random.rand(4).reshape([4, 1])
    mat = np.random.rand(16).reshape([2, 2, 2, 2])
    tensor = np.random.rand(16).reshape([4, 4])
    vec = backend.cast(vec, dtype=vec.dtype)
    mat = backend.cast(mat, dtype=mat.dtype)
    tensor = backend.cast(tensor, dtype=tensor.dtype)
```

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -1,7 +1,8 @@
 """Module defining the `QuantumNetwork` class and adjacent functions."""
 
-import re
 from functools import reduce
+from logging import warning
+from numbers import Number
```
renatomello, line 5: Do we actually need this?

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -44,145 +45,185 @@ class QuantumNetwork:
...
 
-        self._set_tensor_and_parameters()
+        self._set_parameters()
 
-    def matrix(self, backend=None):
-        """Returns the Choi operator of the quantum network in matrix form.
+        if len(self.partition) > 0:
+            self.dims = reduce(mul, self.partition)
+        else:
+            self.dims = 1
```
renatomello, line 67:
```suggestion
        self.dims = reduce(mul, self.partition) if len(self.partition) > 0 else 1
```

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -44,161 +45,204 @@ class QuantumNetwork:
...
 
-        Args:
-            backend (:class:`qibo.backends.abstract.Backend`, optional): Backend to be used
-                to return the Choi operator. If ``None``, defaults to the backend defined
-                when initializing the :class:`qibo.quantum_info.quantum_networks.QuantumNetwork`
-                object. Defaults to ``None``.
```
renatomello, line 71: This shouldn't be removed from the docstring.

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -44,161 +45,204 @@ class QuantumNetwork:
...
 
-        Returns:
-            ndarray: Choi matrix of the quantum network.
-        """
-        if backend is None:  # pragma: no cover
-            backend = self._backend
+    @staticmethod
+    def _order_operator2tensor(n: int):
+        order = list(sum(zip(list(range(0, n)), list(range(n, n * 2))), ()))
+        return order
```
renatomello, line 77: Both methods need a docstring, and `n` should be renamed to a more expressive name.

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -44,161 +45,204 @@ class QuantumNetwork:
...
-        return self._pure
+        try:
+            tensor = np.transpose(operator.reshape(list(partition) * 2), order)
+            return tensor.reshape([dim**2 for dim in partition])
+        except:
+            raise_error(
+                ValueError,
+                "``partition`` does not match the shape of the input matrix. "
+                + f"Cannot reshape matrix of size {operator.shape} to partition {partition}",
+            )
```
renatomello, line 92: This is another instance of a `try / except` that doesn't specify which `Exception` is being bypassed. They should be specified every time.

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -44,161 +45,204 @@ class QuantumNetwork:
...
+                Defaults to ``None``.
+            pure (bool, optional): ``True`` when ``arr`` is a "pure" representation (e.g. a pure
+                state, a unitary operator, etc.), ``False`` otherwise. Defaults to ``False``.
+            backend (:class:`qibo.backends.abstract.Backend`, optional): Backend to be used in
+                calculations. If ``None``, defaults to :class:`qibo.backends.GlobalBackend`.
+                Defaults to ``None``.
 
         Returns:
-            bool: Hermiticity condition.
+            :class:`qibo.quantum_info.quantum_networks.QuantumNetwork`:
```
renatomello, line 124: The `Returns` item of the docstring needs also a small sentence describing what is being returned.

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +569,500 @@ def _run_checks(self, partition, system_output, pure):
...
+    def __init__(
+        self,
+        tensor,
+        partition: Optional[Union[List[int], Tuple[int]]] = None,
+        *,
+        system_input: Optional[Union[List[bool], Tuple[bool]]] = None,
+        pure: bool = False,
+        backend=None,
+    ):
+        if partition == None:
```
renatomello, line 675:
```suggestion
        if partition is None:
```

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +569,500 @@ def _run_checks(self, partition, system_output, pure):
...
+        if float(norm) > precision_tol:
+            return False
+        elif len(self.partition) == 2:
+            return True
+        else:
+            # Unital is defined for quantum channels only.
+            # But we can extend it to quantum combs as follows:
+            return QuantumChannel(  # pragma: no cover
+                sub_comb, self.partition[2:], pure=False, backend=self._backend
+            ).is_unital(order, precision_tol)
```
renatomello, line 887:
```suggestion
        if float(norm) > precision_tol:
            return False
        
        if len(self.partition) == 2:
            return True
        
        # Unital is defined for quantum channels only.
        # But we can extend it to quantum combs as follows:
        return QuantumChannel(  # pragma: no cover
            sub_comb, self.partition[2:], pure=False, backend=self._backend
        ).is_unital(order, precision_tol)
```

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -44,161 +45,204 @@ class QuantumNetwork:
...
 
-        Args:
-            backend (:class:`qibo.backends.abstract.Backend`, optional): Backend to be used
-                to return the Choi operator. If ``None``, defaults to the backend defined
-                when initializing the :class:`qibo.quantum_info.quantum_networks.QuantumNetwork`
-                object. Defaults to ``None``.
+    @staticmethod
+    def _order_tensor2operator(n: int):
+        order = list(range(0, n * 2, 2)) + list(range(1, n * 2, 2))
+        return order
```
renatomello, line 72: This method is only used once in this class. Is it important enough to be it's own stand-alone method?

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -44,161 +45,204 @@ class QuantumNetwork:
...
-        self, order: Optional[Union[int, str]] = None, precision_tol: float = 1e-8
+    @classmethod
+    def from_operator(
+        cls,
+        arr: np.ndarray,
+        partition: Optional[Union[List[int], Tuple[int]]] = None,
+        system_input: Optional[Union[List[bool], Tuple[bool]]] = None,
+        pure: bool = False,
+        backend=None,
     ):
```
renatomello, line 102: This method receives a variable named `arr` being typed as `numpy.ndarray`. Later, this method is overwritten by other classes, with this same variable being renamed as `tensor`. We should stick to one name throughout. Moreover, as we discussed before, this can be more than just a `numpy.ndarray`, so I'd just leave it untyped for simplicity.

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -44,161 +45,204 @@ class QuantumNetwork:
...
+        )
 
-        For specifications on the other  possible values of the
-        parameter ``order`` for the ``tensorflow`` backend, please refer to
-        `tensorflow.norm <https://www.tensorflow.org/api_docs/python/tf/norm>`_.
-        For all other backends, please refer to
-        `numpy.linalg.norm <https://numpy.org/doc/stable/reference/generated/numpy.linalg.norm.html>`_.
+    def operator(self, backend=None, full=False):
+        """Returns the Choi operator of the quantum network.
+        The shape of the returned operator is :math:`(*self.partition, *self.partition)`.
```
renatomello, line 169:
```suggestion
        The shape of the returned operator is :math:`(*partition, *partition)`.
```

renatomello, line 169:
```suggestion
    def operator(self, full=False, backend=None):
        """Returns the Choi operator of the quantum network.
        
        The shape of the returned operator is :math:`(*partition, *partition)`.
```

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -44,161 +45,204 @@ class QuantumNetwork:
...
-            precision_tol (float, optional): threshold :math:`\\epsilon` that defines
-                if Choi operator of the network is :math:`\\epsilon`-close to unitality
-                in the norm given by ``order``. Defaults to :math:`10^{-8}`.
+            backend (:class:`qibo.backends.abstract.Backend`, optional): Backend to be used
+                to return the Choi operator. If ``None``, defaults to the backend defined
+                when initializing the :class:`qibo.quantum_info.quantum_networks.QuantumNetwork`
+                object. Defaults to ``None``.
+            full (bool, optional): If this is ``False``, and the network is pure, the method
+                will only return the eigenvector (unique when the network is pure).
+                If ``True``, returns the full tensor of the quantum network. Defaults to ``False``.
```
renatomello, line 178:
```suggestion
            full (bool, optional): If this is ``False``, and the network is pure, the method
                will only return the eigenvector (unique when the network is pure).
                If ``True``, returns the full tensor of the quantum network. Defaults to ``False``.
            backend (:class:`qibo.backends.abstract.Backend`, optional): Backend to be used
                to return the Choi operator. If ``None``, defaults to the backend defined
                when initializing the :class:`qibo.quantum_info.quantum_networks.QuantumNetwork`
                object. Defaults to ``None``.
```

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +569,500 @@ def _run_checks(self, partition, system_output, pure):
...
-            ] * len(self.partition)
-            for k in range(len(self.partition) // 2):
-                self.system_output[k * 2] = False
-            self.system_output = tuple(self.system_output)
+    def full(self, backend=None, update=False):
+        """Convert the internal representation to the full tensor of the network.
+
+        Returns:
+            ndarray: The full reprentation of the quantum network.
+        """
```
renatomello, line 616: Docstring missing args

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +569,500 @@ def _run_checks(self, partition, system_output, pure):
...
+        backend (:class:`qibo.backends.abstract.Backend`, optional): Backend to be used in
+            calculations. If ``None``, defaults to :class:`qibo.backends.GlobalBackend`.
+            Defaults to ``None``.
+    """
+
+    def __init__(
+        self,
+        tensor,
+        partition: Optional[Union[List[int], Tuple[int]]] = None,
+        *,
```
renatomello, line 671: It's best to handle this with either `*args` or `**kwargs` if it is really needed.

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +567,500 @@ def _run_checks(self, partition, system_output, pure):
...
+
+        if partition is not None:
+            if len(partition) > 2:
+                raise_error(
+                    ValueError,
+                    "A quantum channel should only contain one input system and one output system. "
+                    + "For general quantum networks, one should use the ``QuantumNetwork`` class.",
+                )
+            if len(partition) == 1:
+                if system_input == None:  # Assume the input is a quantum state
```
renatomello, line 806:
```suggestion
                if system_input is None:  # Assume the input is a quantum state
```

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +567,500 @@ def _run_checks(self, partition, system_output, pure):
...
+
+        return self._einsum("ijkl, jl", operator, state)
+
 
-        return matrix
+def link_product(
+    subscripts: str = "ij,jk -> ik",
+    *operands: QuantumNetwork,
+    backend=None,
+    surpress_warning=False,
```
renatomello, line 938: an argument that has a default value should come after all the ones that don't

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +567,500 @@ def _run_checks(self, partition, system_output, pure):
...
+                if index < 0:
+                    continue
+                found += 1
+                if found > 1:
+                    if is_input != operands[inds[i]].system_input[index]:
+                        pass
+                    else:
+                        warning(
+                            f"Index {ind} connects two {'input' if is_input else 'output'} systems."
+                        )
```
renatomello, line 1011:
```suggestion
                if found > 1 and is_input == operands[inds[i]].system_input[index]:
                    warning(
                        f"Index {ind} connects two {'input' if is_input else 'output'} systems."
                    )
```

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +567,500 @@ def _run_checks(self, partition, system_output, pure):
...
+            to return the identity operator. If ``None``, defaults to the backend defined
+            when initializing the :class:`qibo.quant
+    """
+
+    def __init__(self, dim: int, backend=None):
+
+        super().__init__(
+            np.eye(dim, dtype=np.complex128), [dim, dim], pure=True, backend=backend
+        )
+
```
renatomello, line 1050:
```suggestion
        identity = np.eye(dim, dtype=complex)
        identity = backend.cast(identity, dtype=identity.dtype)
        super().__init__(identity, [dim, dim], pure=True, backend=backend)

```

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -629,53 +567,500 @@ def _run_checks(self, partition, system_output, pure):
...
-        braket = "[a-z]"
-        pattern_two = re.compile(braket * 2 + "," + braket * 2 + "->" + braket * 2)
-        pattern_four = re.compile(braket * 4 + "," + braket * 2 + "->" + braket * 2)
+    def __init__(self, dim: int, backend=None):
 
-        return bool(re.match(pattern_two, subscripts)), bool(
-            re.match(pattern_four, subscripts)
+        super().__init__(
+            np.eye(dim, dtype=np.complex128), [dim], [True], pure=False, backend=backend
         )
```
renatomello, line 1066:
```suggestion
        identity = np.eye(dim, dtype=complex)
        identity = backend.cast(identity, dtype=identity.dtype)
        super().__init__(identity, [dim], [True], pure=False, backend=backend)
```

renatomello:
@Canoming can you merge `master` to this branch and fix the conflicts? Also, can you mark which of the suggested changes were already implemented? Thanks.

Canoming:
> @Canoming can you merge `master` to this branch and fix the conflicts? Also, can you mark which of the suggested changes were already implemented? Thanks.

Hi, in my last commit, all suggested changes are implemented. The only left over issue is  [the error class in tensorflow cannot be captured](https://github.com/qiboteam/qibo/pull/1244#discussion_r1665717446)

renatomello:
> > @Canoming can you merge `master` to this branch and fix the conflicts? Also, can you mark which of the suggested changes were already implemented? Thanks.
> 
> Hi, in my last commit, all suggested changes are implemented. The only left over issue is [the error class in tensorflow cannot be captured](https://github.com/qiboteam/qibo/pull/1244#discussion_r1665717446)

There are the new conflicts with `master` to fix still

src/qibo/quantum_info/quantum_networks.py
renatomello: @Canoming This question still remains: we don't need this `Number` class for typing, do we?

src/qibo/quantum_info/quantum_networks.py
renatomello: @Canoming This also wasn't resolved

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -44,161 +45,227 @@ class QuantumNetwork:
...
-            backend = self._backend
+        Returns the order to reshape a tensor to an operator.
+        Given a tenosr of 2n leads, the order is [0, 2, 4, ..., 1, 3, 5, ...].
+        """
+        order = list(range(0, n * 2, 2)) + list(range(1, n * 2, 2))
+        return order
 
-        return backend.cast(self._matrix, dtype=self._matrix.dtype)
+    @staticmethod
+    def _order_operator2tensor(n: int):
```
renatomello, line 76:
```suggestion
    def _order_operator_to_tensor(n: int):
```

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -44,161 +45,227 @@ class QuantumNetwork:
...
+        Given a operator of 2n systems, the order is [0, n, 1, n+1, 2, n+2, ...].
+        """
+        order = list(sum(zip(list(range(0, n)), list(range(n, n * 2))), ()))
+        return order
 
-    def is_pure(self):
-        """Returns bool indicading if the Choi operator of the network is pure."""
-        return self._pure
+    @classmethod
+    def _operator2tensor(cls, operator, partition: List[int]):
```
renatomello, line 85:
```suggestion
    def _operator_to_tensor(cls, operator, partition: List[int]):
```

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -44,161 +45,227 @@ class QuantumNetwork:
...
-            backend (:class:`qibo.backends.abstract.Backend`, optional): Backend to be used
-                to return the Choi operator. If ``None``, defaults to the backend defined
-                when initializing the :class:`qibo.quantum_info.quantum_networks.QuantumNetwork`
-                object. Defaults to ``None``.
+        self.dims = reduce(mul, self.partition) if len(self.partition) > 0 else 1
 
-        Returns:
-            ndarray: Choi matrix of the quantum network.
+    @staticmethod
+    def _order_tensor2operator(n: int):
```
renatomello, line 67:
```suggestion
    def _order_tensor_to_operator(n: int):
```

src/qibo/quantum_info/quantum_networks.py
```diff
@@ -44,161 +45,227 @@ class QuantumNetwork:
...
+                ValueError,
+                "``partition`` does not match the shape of the input matrix. "
+                + f"Cannot reshape matrix of size {operator.shape} to partition {partition}",
+            )
+        except RuntimeError:
+            raise_error(
+                ValueError,
+                "``partition`` does not match the shape of the input matrix. "
+                + f"Cannot reshape matrix of size {operator.shape} to partition {partition}",
+            )
```
renatomello, line 115: There is a much simpler way of implementing this. Just check if `partition` matches `operator.shape`. If not, raise an error. If it does, the calculation is done. None of these `try / except` are needed.

renatomello (changes requested):

src/qibo/quantum_info/quantum_networks.py
renatomello: @Canoming also not resolved

renatomello (changes requested):

src/qibo/quantum_info/quantum_networks.py
renatomello: renaming `n` was also not resolved

src/qibo/quantum_info/quantum_networks.py
renatomello: + missing docstring

Canoming:
For the `Number` type mentioned in https://github.com/qiboteam/qibo/pull/1244#discussion_r1680485459, we include it here as some backends implement non-standard classes for float numbers. It may raise a type error if a `quantum_network` is multiplied by those float numbers.

renatomello:
> For the `Number` type mentioned in [#1244 (comment)](https://github.com/qiboteam/qibo/pull/1244#discussion_r1680485459), we include it here as some backends implement non-standard classes for float numbers. It may raise a type error if a `quantum_network` is multiplied by those float numbers.

I see. Which backend exactly? Do you have an example code?

Canoming:
> > For the `Number` type mentioned in [#1244 (comment)](https://github.com/qiboteam/qibo/pull/1244#discussion_r1680485459), we include it here as some backends implement non-standard classes for float numbers. It may raise a type error if a `quantum_network` is multiplied by those float numbers.
> 
> I see. Which backend exactly? Do you have an example code?

I encountered the issue before, but I don't remember the exact scenario. It was `Pytorch` that caused the problem, where the output of some of its functions will inherit the data type of the `torch.Tensor`, and return a number of type, such as `torch.float32`, instead of `float`.

renatomello:
> I encountered the issue before, but I don't remember the exact scenario. It was `Pytorch` that caused the problem, where the output of some of its functions will inherit the data type of the `torch.Tensor`, and return a number of type, such as `torch.float32`, instead of `float`.

That's fine. That happens everywhere but we don't explicitly check if it's a `float` or a `np.float64` or a `torch.float64`, etc. Just type it as `float` and the rest is implicit. No need to use the `Number`class.

renatomello (approved):

Issue #1117, opened 2023-12-04
Revert ecosystem dependencies


AleCandido:
Currently, the dependencies in the Qibo ecosystem are the following:

```mermaid
flowchart LR
  qibocal --> qibolab --> qibo
  qibojit --> qibo
```

But then the users will need a separate installation of all the various tools, even just to run circuits.
Moreover, we usually leave to the user the burden of installing dependencies for the backends (including TensorFlow and NumPy). I will open a separate issue for that.

A better organization could be the following:

```mermaid
flowchart LR
  qibocal --> qibolab
  qibo --> qibolab --> qibo.core
  qibo --> qibojit --> qibo.core
  qibo --> qibo.core
```

Indeed, the user will often interact just with `qibo` (as frontend), or `qibocal`, for different tasks. Everything else is just going to implement backends.

In particular, `qibo.core` will only contain the current `qibo.backends` and `qibo.gates`, since they are the building blocks used by the backends.

Other relevant points:
- we can have a new repo for the core, but we don't have to (we could even keep here, it's just a matter of reshuffling workflows), but
- we will need separate PyPI packages, for the proper dependency management
- we do **not** need a breaking release (i.e. Qibo 0.3), since we could keep exposing everything through as before (just exporting from qibo.core).

alecandido:
> * The current dependency structure is already pretty good, no need to invert it.

Well, it's partially inverted even in your diagram, since no one would depend on Qibo any longer...

Btw, sorry for the misleading part, but `qibo.core` is not intended to be part of Qibo, I will update the OP to `qibo-core`.

alecandido:
> * Hopefully, the dependency can be stripped down to `backends` and `gates` only (I didn't check all imports one by one, so I am not sure if there is anything else important).

Yes, that's the main goal. Better stated this way.

> * User directly decides what to install and what not. E.g. they decide to install `qibojit` along with `qibo`. In this case concrete backend instances can be supplied to `qibo` in a dependency injection manner.

I'd like it as well, but I refrained from proposing it.
This is more "developer mode", in which you know what you need.
Instead, I wanted `qibo` to be a hat on top of everything. It could even be a metapackage, eventually, just depending on the other packages, but containing no code (e.g. `qiboinfo` will contain all th Qinfo stuffs, `qiboml` all the variational, and `qibo-models` all the pre-built objects). However, no need to go this far in one shot.

[...]

PR #1248, opened 2024-03-05
Integrate `CliffordBackend` and `stim`

renatomello:
Checklist:
- [x] Tests are passing.


pyproject.toml
```diff
@@ -70,6 +70,7 @@ matplotlib = "^3.7.0"
 qibojit = { git = "https://github.com/qiboteam/qibojit.git" }
 tensorflow = { version = "^2.14.1,<2.16", markers = "sys_platform == 'linux'" }
 qibotn = { git = "https://github.com/qiboteam/qibotn.git" }
+stim = "^1.12.0"
```
alecandido, line 73: `stim` is a very light dependency, but I'm becoming more and more cautious about adding new deps.

@renatomello it's fine to add it, but I'd like to split the `quantum_info` module soon (as soon as we'll have `qibo-core` available), if possible, to isolate its dependencies. Are you okay with that?

This is not related to this PR.

renatomello: what do you mean by split?

alecandido: Distributing in a separate package, like they are thinking for `qiboml`. It doesn't have to be a separate repo.

The problem is that, even if you want to just use the basic `NumpyBackend`, at the moment you're forced to get all the dependencies.

I proposed to have pip extras for each module, but after discussing it seems more intuitive to do:
```sh
pip install qibo qiboml qibo_info
```
than
```sh
pip install qibo[ml,info]
```
(I'm not sure myself how much pip extras are known to users, and I got convinced by the people involved in the discussion)

renatomello: I don't agree with that. The info module comes with very fundamental functions and should come with "raw" `qibo`.

This is the same in the other packages that also contain an info module (e.g. [Qiskit](https://github.com/Qiskit/qiskit/tree/f4b50fbb2d3fe7871589fae898f6eadbad7c959b/qiskit)) as well as in others where the same capabilities are spread out throughout their code.

alecandido: Well, then you should split `Clifford` out of `quantum_info` (or just the `stim` part).

Ideally, `qibo` should have mandatory dependencies only on `numpy` (`scipy` is also acceptable).
E.g., right now it is impossible to install Qibo on Mac with compiling a wheel with CMake and additional dependencies, because of `qdldl` (a transitive dependency of `cvxpy`). Moreover, even if you want to have a minimal execution, you will be downloading >200MB of TensorFlow, that is also preventing us from allowing Python 3.12.

Dependencies are an issue, and they should be carefully reviewed. That's why whatever add further dependencies should be optional, one way or the other.

renatomello: > Well, then you should split `Clifford` out of `quantum_info` (or just the `stim` part).
> 
> Ideally, `qibo` should have mandatory dependencies only on `numpy` (`scipy` is also acceptable). E.g., right now it is impossible to install Qibo on Mac with compiling a wheel with CMake and additional dependencies, because of `qdldl` (a transitive dependency of `cvxpy`). Moreover, even if you want to have a minimal execution, you will be downloading >200MB of TensorFlow, that is also preventing us from allowing Python 3.12.
> 
> Dependencies are an issue, and they should be carefully reviewed. That's why whatever add further dependencies should be optional, one way or the other.

- `stim` is an optional dependency. 
- `tensorflow` being in main `qibo` or not does not affect `quantum_info` because the module is backend-agnostic. 
- `cvxpy` is also an optional dependency.

alecandido: TensorFlow was a broader example, and it will in QiboML. I'm not concerned about `quantum_info` alone, the general problem are Qibo deps.

`stim` is currently an undeclared dependency, because you only added it to the test group, so it will never be installed by the user. `cvxpy` is an optional dependency with no associated extra, so it is once more useless for the end user (but it is flagged by Poetry anyhow when you are adding Qibo as a dep to another project).

> Right now `qibo` only depends on `stim` and `cvxpy` for tests, which is kind of inevitable with respect to the development side, but it's completely optional to an user.

Well, not really, since if you want to test the `NumpyBackend` you also need `cvxpy`. Splitting to a different package would decouple the two.

It could be in this repo as usual, just in `src/qibo_info` (or `src/quantum_info` as well, if that name is still available), just a different entry on PyPI.

> Moreover, if you have ideas to rewrite the convex optimization in the `diamond_norm` function without depending on `cvxpy`, I'm open to suggestions.

The goal is not to have a different implementation, that one is perfectly fine (I didn't review myself, but I'm not going to do it now). But features that require complex tools are not basic, even for end users (since they will need those tools).
I can also accept that extras are not intuitive, but requiring an additional package as a plugin should be fairly acceptable.

My concern is coming from the maintenance of the tools we are using, since Qibo is growing, and requiring many pieces to work all at the same time is more and more constraining.
Here is the status of `qdldl` maintenance:
https://github.com/osqp/qdldl/graphs/contributors
and here an illustrative cartoon about my concerns:
https://xkcd.com/2347/
:P

renatomello: > `stim` is currently an undeclared dependency, because you only added it to the test group, so it will never be installed by the user. 

As it should be. If the user calls `stim` without having it installed, the code explicitly raises an error telling the user to install `stim`. Also in the documentation is is clear the `stim` is a separate package that should be installed separately.

`cvxpy` is an optional dependency with no associated extra, so it is once more useless for the end user (but it is flagged by Poetry anyhow when you are adding Qibo as a dep to another project).

It could be an undeclared dependency too, it makes no difference to the implementation.

> > Right now `qibo` only depends on `stim` and `cvxpy` for tests, which is kind of inevitable with respect to the development side, but it's completely optional to an user.
> 
> Well, not really, since if you want to test the `NumpyBackend` you also need `cvxpy`. Splitting to a different package would decouple the two.

i.e. it only depends on `cvxpy` for test, like I mentioned before.
 
> It could be in this repo as usual, just in `src/qibo_info` (or `src/quantum_info` as well, if that name is still available), just a different entry on PyPI.

I already said my opinion on this. In any case, this is both a separate issue and something that more people should opine on it besides just the both of us.

alecandido: > I already said my opinion on this. In any case, this is both a separate issue and something that more people should opine on it besides just the both of us.

The discussion is openly ongoing both in #1117 and https://github.com/qiboteam/qibo-core

src/qibo/backends/clifford.py
```diff
@@ -28,12 +27,26 @@ class CliffordBackend(NumpyBackend):
...
+            try:
+                import stim  # pylint: disable=C0415
 
-            engine = _get_engine_name(_check_backend(engine))
+                self._stim = stim
+            except ModuleNotFoundError:  # pragma: no cover
+                raise_error(
+                    ModuleNotFoundError,
+                    "`engine` was set as `stim`, but module was not found. Please install `stim`.",
+                )
```
alecandido, line 39: Why not to leave the original error? It's not the exact same, but it's very similar to this one.

src/qibo/backends/clifford.py
```diff
@@ -150,6 +172,28 @@ def execute_circuit(self, circuit, initial_state=None, nshots: int = 1000):
...
+            for gate in circuit.queue:
+                circuit_stim.append(gate.__class__.__name__, list(gate.qubits))
+
+            quad_1, quad_2, quad_3, quad_4, x_phases, z_phases = (
+                self._stim.Tableau.from_circuit(  # pylint: disable=no-member
+                    circuit_stim
+                ).to_numpy()
+            )
+            symplectic_matrix = np.block([[quad_1, quad_2], [quad_3, quad_4]])
+            symplectic_matrix = np.c_[symplectic_matrix, np.r_[x_phases, z_phases]]
```
alecandido, line 188: In principle, it should be possible to do just:
```suggestion
            symplectic_matrix = np.block([[quad_1, quad_2], [quad_3, quad_4], [x_phases, z_phases]])
```
or similarly (I'm not sure about the role of `quad_x`, to which blocks they map to).

I'd say it is more intuitive.

src/qibo/backends/clifford.py
```diff
@@ -150,6 +172,28 @@ def execute_circuit(self, circuit, initial_state=None, nshots: int = 1000):
...
             :class:`qibo.quantum_info.clifford.Clifford`: Object storing to the final results.
         """
+        from qibo.quantum_info.clifford import Clifford  # pylint: disable=C0415
+
+        if self.platform == "stim":
+            circuit_stim = self._stim.Circuit()  # pylint: disable=E1101
+            for gate in circuit.queue:
+                circuit_stim.append(gate.__class__.__name__, list(gate.qubits))
+
+            quad_1, quad_2, quad_3, quad_4, x_phases, z_phases = (
```
alecandido, line 182: Maybe, instead of `quad_x`, you could use `x_stab`, `z_destab`, ..., to make it clearer what they are.

(or `xn`, `x2n`, ..., or whatever has a more direct connection).

alecandido (approved):
It should be alright, just a few comments to polish and improve readability.

src/qibo/backends/clifford.py
renatomello: This would concatenate `[x_phases, z_phases]` as the last row, but they are the last column.

src/qibo/backends/clifford.py
renatomello: Done

BrunoLiegiBastonLiegi (commented):
Thanks @renatomello looks good. There is just one thing that is missing in my opinion, which is in the `Clifford` object let `stim` take care of the samples generation.

tests/test_backends_clifford.py
```diff
@@ -223,20 +223,26 @@ def test_set_backend(backend):
...
+
     noise = NoiseModel()
     noise.add(PauliError([("X", 0.5)]), gates.X)
     noise.add(DepolarizingError(0.1), gates.CZ)
+
+    nqubits = 3
+
+    c = random_clifford(nqubits, density_matrix=True, seed=seed, backend=backend)
+    c.add(gates.M(*range(nqubits)))
+    c_copy = c.copy()
```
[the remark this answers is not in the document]
renatomello: Fix in #1250

src/qibo/backends/clifford.py
```diff
@@ -150,6 +165,28 @@ def execute_circuit(self, circuit, initial_state=None, nshots: int = 1000):
...
+            )
+            symplectic_matrix = np.block([[x_destab, z_destab], [x_stab, z_stab]])
+            symplectic_matrix = np.c_[symplectic_matrix, np.r_[x_phases, z_phases]]
+
+            return Clifford(
+                symplectic_matrix,
+                measurements=circuit.measurements,
+                nshots=nshots,
+                _backend=self,
+            )
```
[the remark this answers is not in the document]
renatomello: Done

renatomello:
> Thanks @renatomello looks good. There is just one thing that is missing in my opinion, which is in the `Clifford` object let `stim` take care of the samples generation.

There is a way to do it, but it's not as direct as the circuit execution, so maybe it'd be better to do it in a separate PR.

PR #1252, opened 2024-03-07
Introduce pip extras

### LINKED ISSUE #1251: Tensorflow dependency
I noticed that doing
```py
pip install qibo
```
in a clean environment installs tensorflow, even though this used to be an optional dependency. Is this expected behavior, or is there something wrong?

alecandido:
Closes #1251

They were obviously missing when the dependencies have been made optional.


scarrazza (approved):

stavros11 (approved):
Thanks @alecandido. I cannot tell anything about nix but confirmed that tensorflow is no longer installed by default in this branch.

alecandido:
Currently, I'm the only one using Nix (though it's there for everyone, and you could have a reproducible shell).
However, the baseline is that I'm updating that as well, to avoid outdated content in our repo, but the core of the PR is just Poetry.

PR #1255, opened 2024-03-08
Simplified `_Queue` append

BrunoLiegiBastonLiegi:
This PR removes the `qibo.models.Circuit._Queue.append()` method that was overloading the `list.append()`. The generation of the moments and the calculation of the number of measurements in the circuit is therefore delayed to when it's actually needed. This was mainly motivated by the `random_clifford` function that appends a lot of gates to the circuit, but I believe it is in general useful for speeding up all the functions that make heavily use of the `Circuit.add()` method, even though the difference in most cases is probably negligible.  

Checklist:
- [x] Tests are passing.


src/qibo/models/circuit.py
```diff
@@ -88,24 +85,27 @@ def from_fused(self):
...
 
-        # calculate moment index for this gate
-        idx = max(self.moment_index[q] for q in qubits)
-        for q in qubits:
-            if idx >= len(self.moments):
-                # Add a moment
-                self.moments.append(len(self.moments[-1]) * [None])
-            self.moments[idx][q] = gate
-            self.moment_index[q] = idx + 1
+    def moments(self):
```
stavros11, line 91: Thanks @BrunoLiegiBastonLiegi. Maybe we could define this as a `@property` since it doesn't have any arguments and also to keep the same interface as before. 

In this particular case we could even give a try to [`@cached_property`](https://docs.python.org/3/library/functools.html#functools.cached_property) which could speed things up in cases where calculation of moments is slow and the property is called multiple times (however I have not really tested that). Another thing to check in that case is how it behaves if another gate is added after the cached property is called, if the cache is updated or no (I am also not sure).

src/qibo/models/circuit.py
```diff
@@ -88,24 +85,27 @@ def from_fused(self):
...
-    def append(self, gate: gates.Gate):
-        super().append(gate)
-        if gate.qubits:
-            qubits = gate.qubits
-        else:  # special gate acting on all qubits
-            qubits = tuple(range(self.nqubits))
-
-        if isinstance(gate, gates.M):
-            self.nmeasurements += 1
+    def nmeasurements(self):
```
stavros11, line 88: Same here, that can be a `@property` and even `@cached_property` could also work better.

src/qibo/models/circuit.py
BrunoLiegiBastonLiegi: Yes I thought about the `cached_property` too, but I tested it and it doesn't work if the queue gets updated.

BrunoLiegiBastonLiegi:
Thanks @stavros11, made them properties once again.

src/qibo/models/circuit.py
```diff
@@ -88,24 +85,29 @@ def from_fused(self):
...
+
+    @property
+    def moments(self):
+        moments = [self.nqubits * [None]]
+        moment_index = self.nqubits * [0]
+        for gate in self:
+            if not isinstance(gate, gates.CallbackGate):
+                qubits = gate.qubits
+            else:  # special gate acting on all qubits
+                qubits = tuple(range(self.nqubits))
```
renatomello, line 100:
```suggestion
            qubits = (
                gate.qubits
                if not isinstance(gate, gates.CallbackGate)
                else tuple(range(self.nqubits))
            )
```

src/qibo/models/circuit.py
```diff
@@ -88,24 +85,29 @@ def from_fused(self):
...
-            self.moments[idx][q] = gate
-            self.moment_index[q] = idx + 1
+    @property
+    def nmeasurements(self):
+        return len([1 for gate in self if isinstance(gate, gates.M)])
+
+    @property
+    def moments(self):
+        moments = [self.nqubits * [None]]
+        moment_index = self.nqubits * [0]
```
renatomello, line 95: Before, users had access to `moment_index` and now they don't. I don't know if this is the intended behaviour or not.

renatomello (approved):

src/qibo/models/circuit.py
```diff
@@ -88,24 +85,30 @@ def from_fused(self):
...
-        idx = max(self.moment_index[q] for q in qubits)
-        for q in qubits:
-            if idx >= len(self.moments):
-                # Add a moment
-                self.moments.append(len(self.moments[-1]) * [None])
-            self.moments[idx][q] = gate
-            self.moment_index[q] = idx + 1
+    @property
+    def nmeasurements(self):
+        return len([1 for gate in self if isinstance(gate, gates.M)])
```
renatomello, line 90:
```suggestion
        return len(self.gates_of_type(gates.M))
```

stavros11: Is this possible? I think `_Queue` doesn't have a `gates_of_type` method. It may still be possible to avoid the `for` loop using [`filter`](https://docs.python.org/3/library/functions.html#filter).

stavros11 (approved):

PR #1260, opened 2024-03-13
Add `entangling_layer` to `models.encodings`

renatomello:
Checklist:
- [x] Tests are passing.

### DIFF
--- a/doc/source/api-reference/qibo.rst
+++ b/doc/source/api-reference/qibo.rst
@@ -386,6 +386,14 @@ of the :math:`d`-dimensional array is sampled from a Gaussian distribution
 .. autofunction:: qibo.models.encodings.unary_encoder_random_gaussian
 
 
+Entangling layer
+""""""""""""""""
+
+Generates a layer of nearest-neighbour two-qubit gates, assuming 1-dimensional connectivity.
+
+.. autofunction:: qibo.models.encodings.entangling_layer
+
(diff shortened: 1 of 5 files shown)


Simone-Bordoni (approved):

MatteoRobbiati (commented):
Super fine with this kind of feature! 

I think it would be nicer to start thinking how to move these features (or, better, if we want this) to the [`qiboml`](https://github.com/qibogang/qiboml) package. 

Personally, I think this is exactly that kind of feature (I mean the encodings in general) which would fit perfectly with it. 

I would say this because it could be crucial to understand how to organize the code structure and because I am currently working on the variational model and, consequently, on the encoding and decoding strategy.

MatteoRobbiati:
> Data encoders are very important for QML, but not exclusive to QML. For instance, it's important for "regular" QFT and / or Quantum Phase Estimation applications (e.g. the example of unary encoding application to option pricing included in `qibo`'s documentation, or the paper(s) in which the current implementation of the unary encoder was based on.). After all, a data encoder is nothing but a state preparation, so the applications are broader than QML.

Thanks for this reply! I see, this perfectly makes sense. So the best way will be to integrate the encodings into `qiboml`, picking them up from Qibo. My only concern is not to make redundancy between the two packages (but at this point this has to be done in `qiboml`).

PR #1269, opened 2024-03-16
Dbi cost functions

wrightjandrew:
Features include:

### Changes in `double_bracket.py` which has new cost functions
- off diagonal norm $||\sigma( H)||$
see  https://github.com/qiboteam/qibo/blob/4d4c45295499fd7ccf559bdff63ab494a0bc0c3e/src/qibo/models/dbi/double_bracket.py#L152
-  least squares $||H-D||-||D||$
see https://github.com/qiboteam/qibo/blob/4d4c45295499fd7ccf559bdff63ab494a0bc0c3e/src/qibo/models/dbi/double_bracket.py#L166
- energy fluctuation $(\langle \psi | H^2 |\psi\rangle - \langle \psi | H |\psi\rangle^2)^{1/2}$ where $|\psi\rangle$ is the reference state (e.g. arising from a circuit that approximates an eigenstate preparation so that the energy fluctuation vanishes)
https://github.com/qiboteam/qibo/blob/4d4c45295499fd7ccf559bdff63ab494a0bc0c3e/src/qibo/models/dbi/double_bracket.py#L223

These are used via the loss function
https://github.com/qiboteam/qibo/blob/4d4c45295499fd7ccf559bdff63ab494a0bc0c3e/src/qibo/models/dbi/double_bracket.py#L211-L221

### Scheduling strategies for find the local minimum given a bracket $W=[D,H]$.

These are implemented in `utils_scheduling.py` and include:
- `grid_search_step`
- `hyperopt_step`
- `polynomial_step`
- `simulated_annealing_step`

### Gradient descent methods for finding $D$ operators

This is done in `utils_gradients.py`. There are 2 ways of finding gradients defined by how we parametrize $D$.
We parametrize the ansatz for $D$ based on two possible choices of representing a matrix 
- the computational basis representation of the matrix $A = \sum_{i,j=0}^{2^L} A_{i,j} |i\rangle\langle j|$ where $L$ is the number of qubits. 
- the Pauli operator basis using the orthogonality in the Hilbert-Schmidt scalar product $A = \sum_{\mu,\nu} A_{\mu,\nu} \otimes_{k=1}^L (Z^{\mu_k}X^{\nu_k})$ where $A_{\mu,\nu} = \langle A, \otimes_{k=1}^L (Z^{\mu_k}X^{\nu_k})\rangle_{HS}$

More specifically we have the cases for parametrizing diagonal matrices $D$ as
- via coupling strengths of local Pauli operators $D(B,J) = \sum_i B_i Z_i +\sum_{i,j} J_{i,j} Z_i Z_j$
- via  matrix elements in the computational basis $D(d) = \sum_i d_i |i\rangle\langle i|$. 

The comutational basis ansatz is there for completeness, once an operator $D$ is found it has to be compiled and so in any case a Pauli basis representation will be used. We recommend using gradient descent directly in the Pauli basis representation. In principle the full optimization in the computational basis could allow for finding operators that outperform an Ising model ansatz but then the operator may be inefficient to compile.

For these parametrizations there is the associated gradient descent method to find the numerical values of the parameters $(B,J)$ or $(d)$ 
See https://github.com/qiboteam/qibo/blob/dbi_cost_functions/examples/dbi/dbi_cost_functions_and_d_gradients_tutorial.ipynb for results
- For $D(B,0)$ ie the 1-local magnetic field we have the default
`gradient_descent_dbr_pauli_basis` - the naming convention reflects that we perform the gradient descent aiming to optimize a double-bracket rotation and we parametrized the ansatz in the Pauli basis
- For the element wise search of the $D(d)$ operator we have
`d_opt, loss_opt, grad_opt, diags_opt = gradient_descent_dbr_computational_basis(dbi, params,  100,lr=1e-2, d_type = d_ansatz_type.element_wise)`

 

Checklist:
- [ ] Tests are passing.

Still some work to be done but just wanted to try how the pull request works and see if I'm doing things correctly


src/qibo/models/dbi/double_bracket.py
```diff
@@ -114,47 +146,29 @@ def backend(self):
         """Get Hamiltonian's backend."""
         return self.h0.backend
 
-    def hyperopt_step(
+    def least_squares(self, D: np.array):
+        """Least squares cost function."""
+        H = self.h.matrix
+        return -np.real(np.trace(H@D)-0.5*(np.linalg.norm(H)**2+np.linalg.norm(D)**2))
```
marekgluza, line 152: For a given Hamiltonian we can skip the evaluation of $||H||$ and of $||D||$ which for 10 qubits can already cost some computation time

src/qibo/models/dbi/utils_scheduling.py
```diff
@@ -0,0 +1,260 @@
...
+    if d is None:
+        d = dbi_object.diagonal_h_matrix
+
+    if n > n_max:
+        raise ValueError(
+            "No solution can be found with polynomial approximation. Increase `n_max` or use other scheduling methods."
+        )
+    if coef is None:
+        if cost == "off_diagonal_norm":
+            coef = off_diagonal_norm_polynomial_expansion_coef(dbi_object, d, n)
```
marekgluza, line 136: Please througout don't use the string comparison and use the enum type class created above

src/qibo/models/dbi/double_bracket.py
```diff
@@ -20,6 +25,26 @@ class DoubleBracketGeneratorType(Enum):
...
+    """Define the DBI scheduling strategies."""
+
+    hyperopt = hyperopt_step
+    """Use hyperopt package."""
+    grid_search = grid_search_step
+    """Use greedy grid search."""
+    polynomial_approximation = polynomial_step
+    """Use polynomial expansion (analytical) of the loss function."""
+
+class DoubleBracketCostFunction(Enum):
```
marekgluza, line 42:
```suggestion
class DoubleBracketCostFunctionType(Enum):
```

because auto() is used, this means throughout the name should be changed

src/qibo/models/dbi/double_bracket.py
```diff
@@ -191,4 +210,17 @@ def energy_fluctuation(self, state):
         Args:
             state (np.ndarray): quantum state to be used to compute the energy fluctuation with H.
         """
-        return self.h.energy_fluctuation(state)
+        state_vector = np.zeros(len(self.h.matrix))
+        state_vector[state] = 1.0
```
marekgluza, line 223: what is this doing?

src/qibo/models/dbi/utils.py
```diff
@@ -163,3 +151,27 @@ def cs_angle_sgn(dbi_object, d):
         )
     )
     return np.sign(norm)
+
+
+def off_diagonal_norm_polynomial_expansion_coef(dbi_object, d, n):
+    if d is None:
+        d = dbi_object.diagonal_h_matrix
+    # generate Gamma's where $\Gamma_{k+1}=[W, \Gamma_{k}], $\Gamma_0=H
+    W = dbi_object.commutator(d, dbi_object.sigma(dbi_object.h.matrix))
```
marekgluza, line 161:
```suggestion
    W = dbi_object.commutator(d, dbi_object.h.matrix)
```

src/qibo/models/dbi/utils.py
```diff
@@ -163,3 +151,27 @@ def cs_angle_sgn(dbi_object, d):
         )
     )
     return np.sign(norm)
+
+
+def off_diagonal_norm_polynomial_expansion_coef(dbi_object, d, n):
```
marekgluza, line 157: @Sam-XiaoyueLi  says this is also used for the $\partial_{B_i}(f_{loss})$ so it's not just for scheduling. Maybe worth discussing some more

src/qibo/models/dbi/utils_scheduling.py
```diff
@@ -0,0 +1,260 @@
...
+        elif cost == "least_squares":
+            coef = least_squares_polynomial_expansion_coef(dbi_object, d, n)
+        elif cost == "energy_fluctuation":
+            coef = energy_fluctuation_polynomial_expansion_coef(dbi_object, d, n, dbi_object.state)
+        else:
```
marekgluza, line 143: please move this function to utils.py

src/qibo/models/dbi/utils_scheduling.py
```diff
@@ -0,0 +1,260 @@
...
+        for s in real_positive_roots:
+            if dbi_object.loss(s, d) < dbi_object.loss(sol, d):
+                sol = s
+        return sol
+    # solution does not exist, return None
+    else:
+        return None
+
+
+def off_diagonal_norm_polynomial_expansion_coef(dbi_object, d, n):
```
marekgluza, line 162: please make sure this function is not doubled in utils.py and have all expansion coef funcs in utils.py

src/qibo/models/dbi/utils_scheduling.py
```diff
@@ -0,0 +1,260 @@
...
+    exp_list = np.array([1 / math.factorial(k) for k in range(n + 1)])
+    # coefficients
+    coef = np.empty(n)
+    for i in range(n):
+        coef[i] = np.real(exp_list[i]*np.trace(d@Gamma_list[i+1]))
+    coef = list(reversed(coef))
+    return coef
+
+#TODO: add a general expansion formula not stopping at 3rd order
+def energy_fluctuation_polynomial_expansion_coef(dbi_object, d, n, state):
```
marekgluza, line 200: please make sure not doubled and moved to utils.py

src/qibo/models/dbi/utils_scheduling.py
```diff
@@ -0,0 +1,260 @@
...
+    for k in range(n + 1):
+        for j in range(n + 1):
+            power = k + j
+            product_matrix = c1[k] @ c2[j]
+            trace_coefficients[power] += 2 * np.trace(product_matrix)
+    # coefficients from high to low (n:0)
+    coef = list(reversed(trace_coefficients[: n + 1]))
+    return coef
+
+def least_squares_polynomial_expansion_coef(dbi_object, d, n):
```
marekgluza, line 185: please move to utils.py and make sure not doubled

Sam-XiaoyueLi:
Please add test files to cover the changes you've made. To check the coverage in VS code, you may consider using Coverage Gutters (extension) and display coverage.

src/qibo/models/dbi/utils_scheduling.py
```diff
@@ -0,0 +1,260 @@
...
+    A[i,i] = 1
+    dGamma = [commutator(A,H)]
+    derivative += np.real(np.trace(Gamma_list[0]@A)+np.trace(dGamma[0]@d+Gamma_list[1]@A)*s)
+    for n in range(2,4):
+        dGamma.append(dGamma_diDiagonal(dbi_object,d,H,n,i,dGamma,Gamma_list))
+        derivative += np.real(np.trace(dGamma[-1]@d + Gamma_list[n]@A)*s**n/math.factorial(n))
+
+    return derivative
+
+def gradientDiagonal(dbi_object,d,H):
```
marekgluza:
```suggestion
def gradientDiagonalEntries(dbi_object,d,H):
```

would be better to have more descriptive names

src/qibo/models/dbi/utils_scheduling.py
```diff
@@ -0,0 +1,260 @@
...
+    # Gradient of potential function with respect to diagonal elements of D (full-diagonal ansatz)
+    grad = np.zeros(len(d))
+    for i in range(len(d)):
+        derivative = dpolynomial_diDiagonal(dbi_object,d,H,i)
```
marekgluza: this is not very clear, what is potential function?

src/qibo/models/dbi/utils_scheduling.py
```diff
@@ -0,0 +1,260 @@
...
+    coef[1] = np.real(2*variance(Gamma_list[1],state)+2*covariance(Gamma_list[0],Gamma_list[2],state))
+    coef[2] = np.real(covariance(Gamma_list[0], Gamma_list[3],state)+3*covariance(Gamma_list[1], Gamma_list[2],state))
+    coef = list(reversed(coef))
+    return coef
+
+def dGamma_diDiagonal(dbi_object, d, H, n, i,dGamma, Gamma_list):
+    # Derivative of gamma with respect to diagonal elements of D (full-diagonal ansatz)
+    A = np.zeros(d.shape)
+    A[i,i] = 1
+    B = commutator(commutator(A,H),Gamma_list[n-1])
```
marekgluza: I think @andrea-pasquale wants pep standards so all variables have to be lower case? :crying_cat_face: 

`B` could become some other longer name if that won't be very clear

src/qibo/models/dbi/utils_scheduling.py
```diff
@@ -0,0 +1,260 @@
...
+    coef[0] = np.real(2*covariance(Gamma_list[0], Gamma_list[1],state))
+    coef[1] = np.real(2*variance(Gamma_list[1],state)+2*covariance(Gamma_list[0],Gamma_list[2],state))
+    coef[2] = np.real(covariance(Gamma_list[0], Gamma_list[3],state)+3*covariance(Gamma_list[1], Gamma_list[2],state))
+    coef = list(reversed(coef))
+    return coef
+
+def dGamma_diDiagonal(dbi_object, d, H, n, i,dGamma, Gamma_list):
+    # Derivative of gamma with respect to diagonal elements of D (full-diagonal ansatz)
+    A = np.zeros(d.shape)
+    A[i,i] = 1
```
marekgluza:
```suggestion
    a = np.eye(d.shape[0])
```

src/qibo/models/dbi/utils_scheduling.py
```diff
@@ -0,0 +1,337 @@
...
+        + 3 * covariance(Gamma_list[1], Gamma_list[2], state)
+    )
+    coef = list(reversed(coef))
+    return coef
+
+
+def dGamma_diDiagonal(dbi_object, d, H, n, i, dGamma, Gamma_list):
+    # Derivative of gamma with respect to diagonal elements of D (full-diagonal ansatz)
+    A = np.zeros(d.shape)
+    A[i, i] = 1
```
marekgluza, line 220:
```suggestion
    a = np.eye(d.shape[0])
```

@andrea-pasquale can you confirm if all variables should be lower-case letters because of some PEP standards?

marekgluza (changes requested):

src/qibo/models/dbi/utils_scheduling.py
andrea-pasquale: Yes, in qibo all variables should have lower case letters.

src/qibo/models/dbi/utils_scheduling.py
andrea-pasquale: https://github.com/qiboteam/qibo/pull/1269#discussion_r1537411824

src/qibo/models/dbi/double_bracket.py
```diff
@@ -114,47 +151,31 @@ def backend(self):
         """Get Hamiltonian's backend."""
         return self.h0.backend
 
-    def hyperopt_step(
+    def least_squares(self, D: np.array):
+        """Least squares cost function."""
+        H = self.h.matrix
```
andrea-pasquale, line 156:
```suggestion
        h = self.h.matrix
```
It's better to use lower case letter for variables :)

src/qibo/models/dbi/double_bracket.py
```diff
@@ -114,47 +151,31 @@ def backend(self):
...
+        return -np.real(
+            np.trace(H @ D) - 0.5 * (np.linalg.norm(H) ** 2 + np.linalg.norm(D) ** 2)
+        )
```
andrea-pasquale, line 159: Although we don't test on GPU in GitHub all DBI features should be GPU compatible (qibojit).
Here are the instructions on how to install [qibojit](https://qibo.science/qibo/stable/getting-started/installation.html#installing-qibojit).
After having installed qibojit (cupy) properly by running tests locally you should be able to see if tests are passing or not also for GPU.

src/qibo/models/dbi/double_bracket.py
```diff
@@ -114,47 +151,31 @@ def backend(self):
...
+        if scheduling is None:
+            scheduling = self.scheduling
+        step = scheduling(self, d=d, **kwargs)
+        if (
+            step is None
+            and scheduling == DoubleBracketScheduling.polynomial_approximation
+        ):
+            kwargs["n"] = kwargs.get("n", 3)
+            kwargs["n"] += 1
+            # if n==n_max, return None
```
andrea-pasquale, line 176: Please provide some comments to understand what is happening here.

src/qibo/models/dbi/double_bracket.py
```diff
@@ -171,24 +192,44 @@ def loss(self, step: float, d: np.array = None, look_ahead: int = 1):
...
+        elif self.cost == DoubleBracketCostFunction.energy_fluctuation:
+            loss = self.energy_fluctuation(self.state)
 
         # set back the initial configuration
         self.h = h_copy
 
         return loss
 
-    def energy_fluctuation(self, state):
+    def energy_fluctuation(self, state=None):
```
andrea-pasquale, line 208:
```suggestion
    def energy_fluctuation(self, Optional[state]=None):
```

src/qibo/models/dbi/double_bracket.py
andrea-pasquale: +1

src/qibo/models/dbi/double_bracket.py
```diff
@@ -171,24 +192,44 @@ def loss(self, step: float, d: np.array = None, look_ahead: int = 1):
...
+        if state is None:
+            state = self.state
+        state_vector = np.zeros(len(self.h.matrix))
+        state_vector[state] = 1.0
+        return np.real(self.h.energy_fluctuation(state_vector))
+
+    def sigma(self, h: np.array):
+        return h - self.backend.cast(np.diag(np.diag(self.backend.to_numpy(h))))
+
+    def generate_Gamma_list(self, n: int, d: np.array):
```
andrea-pasquale, line 229:
```suggestion
    def generate_gamma_list(self, n: int, d: np.array):
```

src/qibo/models/dbi/double_bracket.py
```diff
@@ -171,24 +192,44 @@ def loss(self, step: float, d: np.array = None, look_ahead: int = 1):
...
+        r"""Computes the n-nested Gamma functions, where $\Gamma_k=[W,...,[W,[W,H]]...]$, where we take k nested commutators with $W = [D, H]$"""
+        W = self.commutator(d, self.sigma(self.h.matrix))
+        Gamma_list = [self.h.matrix]
```
andrea-pasquale, line 232:
```suggestion
        gamma_list = [self.h.matrix]
```
Also for the other cases.

src/qibo/models/dbi/utils.py
```diff
@@ -75,26 +77,25 @@ def select_best_dbr_generator(
     dbi_object: DoubleBracketIteration,
     d_list: list,
     step: Optional[float] = None,
-    step_min: float = 1e-5,
-    step_max: float = 1,
-    max_evals: int = 200,
     compare_canonical: bool = True,
+    scheduling: DoubleBracketScheduling = None,
```
andrea-pasquale, line 81:
```suggestion
    scheduling: Optional[DoubleBracketScheduling] = None,
```

src/qibo/models/dbi/utils_scheduling.py
```diff
@@ -0,0 +1,337 @@
+import math
+from copy import deepcopy
+from functools import partial
+from typing import Optional
+
+import hyperopt
+import numpy as np
+
+error = 1e-3
```
andrea-pasquale, line 9:
```suggestion
ERROR= 1e-3
```
In this case since it is a global constant it should be all upper case letters :)

src/qibo/models/dbi/utils_scheduling.py
andrea-pasquale: +1

tests/test_models_dbi_utils_scheduling.py
```diff
@@ -0,0 +1,30 @@
...
+from qibo.models.dbi.double_bracket import (
+    DoubleBracketGeneratorType,
+    DoubleBracketIteration,
+    DoubleBracketScheduling,
+)
+from qibo.models.dbi.utils_scheduling import polynomial_step
+from qibo.quantum_info import random_hermitian
+
+NSTEPS = 1
+seed = 10
```
andrea-pasquale, line 16:
```suggestion
SEED = 10
```

src/qibo/models/dbi/double_bracket.py
```diff
@@ -48,10 +78,16 @@ def __init__(
         self,
         hamiltonian: Hamiltonian,
         mode: DoubleBracketGeneratorType = DoubleBracketGeneratorType.canonical,
+        scheduling: DoubleBracketScheduling = DoubleBracketScheduling.grid_search,
+        cost: DoubleBracketCostFunction = DoubleBracketCostFunction.off_diagonal_norm,
+        state: int = 0,
```
andrea-pasquale, line 83:
```suggestion
        state: int = 0,
```
What is `state` supposed to be here? And why is it an `int`?

src/qibo/models/dbi/double_bracket.py
andrea-pasquale: Not really, you can put any name.
I would even shorted the name and opt for something like this.
```suggestion
class DoubleBracketCost(Enum):
```

src/qibo/models/dbi/double_bracket.py
```diff
@@ -114,47 +151,31 @@ def backend(self):
         """Get Hamiltonian's backend."""
         return self.h0.backend
 
-    def hyperopt_step(
+    def least_squares(self, D: np.array):
```
andrea-pasquale, line 154:
```suggestion
    def least_squares(self, d: np.array):
```

andrea-pasquale (changes requested):
Thanks @wrightjandrew @Sam-XiaoyueLi for the PR.
I have a few suggestions mainly related to style. I still need to have a proper look at `utils_scheduling.py`. In the meantime you can start addressing some of my comments.

src/qibo/models/dbi/double_bracket.py
wrightjandrew: While the norm of H can be removed (just a constant), norm of D should be preserved simply due to the fact that if not the minimum of this function will always be a 0 diagonal. We could maybe optimize the computation of this norm as it is essentially the same as a 2**n vector instead of a 2**n by 2**n matrix and maybe there is a more efficient formula if we consider the magnetic field parametrization

examples/dbi/dbi_costs.ipynb
marekgluza, line 1: can you please make a more descriptive name? Thanks! :)

wrightjandrew: Yes I should also update that notebook, as many of the functions were changed and other functionalities were added

src/qibo/models/dbi/double_bracket.py
andrea-pasquale: This is expected, here is a possible blueprint

https://github.com/qiboteam/qibo/blob/d0c0ba2c67350b634b7f419f640dbcea9cf732a1/src/qibo/models/dbi/double_bracket.py#L222-L223

andrea-pasquale: Basically the idea is that if you perform operation with numpy you should make sure that you are playing with a numpy element (which explains `backend.to_numpy`). Once you are finish with the numpy manipulations make sure to call `backend.cast` to convert back to whatever backend you were using in the first place.

tests/test_models_dbi_utils_scheduling.py
andrea-pasquale: I think this change is still pending.

andrea-pasquale:
Thanks for the update @wrightjandrew, there are still some comments to be implemented. 
Also, tests are not passing. Let me know if you need any help with those or with qibojit.

src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -368,12 +368,12 @@ def __init__(self, form=None, nqubits=None, symbol_map={}, backend=None):
...
-            log.warning(
-                "Calculating the dense form of a symbolic Hamiltonian. "
-                "This operation is memory inefficient."
-            )
-            self.dense = self.calculate_dense()
+        #if self._dense is None:
+        #    log.warning(
+        #        "Calculating the dense form of a symbolic Hamiltonian. "
+        #        "This operation is memory inefficient."
+        #    )
```
andrea-pasquale, line 375: You can silence warning by using the environment variable `QIBO_LOG_LEVEL`

src/qibo/models/dbi/utils.py
```diff
@@ -162,3 +101,54 @@ def cs_angle_sgn(dbi_object, d):
...
+
+def generate_Pauli_operators(nqubits, symbols_pauli, positions):
+    # generate matrix of an nqubit-pauli operator with `symbols_pauli` at `positions`
+    if isinstance(positions, int):
+        return SymbolicHamiltonian(
+            symbols_pauli(positions), nqubits=nqubits
+        ).dense.matrix
+    else:
+        terms = [symbols_pauli(pos) for pos in positions]
+        return SymbolicHamiltonian(math.prod(terms), nqubits=nqubits).dense.matrix
```
andrea-pasquale, line 154: Please provide docstrings for those functions

src/qibo/models/dbi/utils_analytical.py
```diff
@@ -0,0 +1,215 @@
+import math
+from typing import Optional
+
+import numpy as np
+
+from qibo.models.dbi.utils import commutator, covariance, variance
+
+
+def dGamma_di_Pauli(dbi_object, n: int, Z_i: np.array, d: np.array):
```
andrea-pasquale, line 9:
```suggestion
def dGamma_di_Pauli(dbi_object, n: int, Z_i: np.array, d: np.array):
```
According to PEP8 for functions it is better to use lower case letters.

andrea-pasquale (changes requested):
Thanks @wrightjandrew for the updates.
I still need to check some `utils` files and the notebooks.
Concerning `utils_*` we could also think about turning it into a module. Something like
```sh
utils/
    __init__.py
    analytical.py
    gradients.py
    scheduling.py
```
I will continue reviewing the other files

src/qibo/models/dbi/double_bracket.py
```diff
@@ -184,11 +210,27 @@ def energy_fluctuation(self, state):
...
 
         Args:
             state (np.ndarray): quantum state to be used to compute the energy fluctuation with H.
         """
-        return self.h.energy_fluctuation(state)
+        h_np = self.backend.cast(np.diag(np.diag(self.backend.to_numpy(self.h.matrix))))
+        h2 = h_np @ h_np
+        a = state.conj() @ h2 @ state
+        b = state.conj() @ h_np @ state
+        return (np.sqrt(np.real(a - b**2))).item()
```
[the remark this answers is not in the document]
marekgluza: Next, @wrightjandrew isolate the problem and share notebook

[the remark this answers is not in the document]
wrightjandrew: Defining a state as: 
state = np.zeros(2**nqubits) gives: ValueError: shapes (8,) and (1,8) not aligned: 8 (dim 0) != 1 (dim 0)
while,
state = np.zeros((1,2**nqubits)) gives: ValueError: matmul: Input operand 1 has a mismatch in its core dimension 0, with gufunc signature (n?,k),(k,m?)->(n?,m?) (size 1 is different from 8)
and, 
state = np.zeros((2**nqubits,1)) gives: ValueError: shapes (8,1) and (8,1) not aligned: 1 (dim 1) != 8 (dim 0)

src/qibo/models/dbi/double_bracket.py
```diff
@@ -20,6 +19,38 @@ class DoubleBracketGeneratorType(Enum):
...
+    energy_fluctuation = auto()
+    """Use energy fluctuation as cost function."""
+
+
+from qibo.models.dbi.utils_scheduling import (
+    grid_search_step,
+    hyperopt_step,
+    polynomial_step,
+    simulated_annealing_step,
+)
```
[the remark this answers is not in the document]
wrightjandrew: The issue lies on a circular import as qibo.models.dbi.utils_scheduling needs the DoubleBracketCost class (defined before) and double.bracket.py needs the step functions which are defined on qibo.models.dbi.utils_scheduling. I'm not too knowledgeable with how imports work so please let me know what would be the best solution (as you said it may be best to change the structure of the code)

src/qibo/hamiltonians/hamiltonians.py
wrightjandrew: Should we leave it up to the user or should I impose it in the part of the code that raises this warning? It appears whenever I create a diagonal operator created using Pauli-Zs which happens a lot when training it.

src/qibo/models/dbi/double_bracket.py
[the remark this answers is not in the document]
andrea-pasquale: If there is a circular import there should be a way to disentangle the code. 
If I understood correctly the problem relies on the fact that `polynomial_step` depends on the cost function.
For a quick patch you could:
- remove the argument `cost` (so that we get rid of the type hint)
- when checking for the corresponding `coef` you could use strings instead of the instance of the class, to do this the `Enum` needs to inherit also from string
Something like this:
```py
from enum import Enum

class Type(str, Enum):
    a = "a"
    b = "b"

test = Type.a
test == "a' # returns True
```

andrea-pasquale (commented):
Thanks for the updates.
I'm sorry for the late review, I was busy with other projects.
For the notebooks a general comment, we recently changed the way we set the backend in qibo. Now `set_backend("qibojit", "numba")` will not work anymore you need to specify the second argument `set_backend("qibojit", platform="numba")`.
Also I feel like there are too many notebooks, maybe some of them we should keep for internal debugging purposes. Having a look at the other folders under example I don't know if it is the best idea to use just notebooks. I would prefer if we switch to an approach where we document the code properly through sphinx and code examples. We could discuss about this later.
Regarding the PR I still have a few suggestions. Make sure that there are no conflicts and that tests are passing. Let me know if you need any help.

src/qibo/hamiltonians/hamiltonians.py
andrea-pasquale: In a sense it is already left to the user through the `QIBO_LOG_LEVEL` env variable...

src/qibo/models/dbi/double_bracket.py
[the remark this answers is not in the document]
andrea-pasquale: I'm not able to reproduce the error.
The following code works fine.
```py
import numpy as np
from qibo import set_backend, hamiltonians
set_backend("numpy")
ham = hamiltonians.XXZ(3)
state = np.ones(2**3)
ham.energy_fluctuation(state)
```

src/qibo/models/dbi/utils.py
```diff
@@ -162,3 +101,54 @@ def cs_angle_sgn(dbi_object, d):
...
+
+def decompose_into_Pauli_basis(h_matrix: np.array, pauli_operators: list):
+    """finds the decomposition of hamiltonian `h_matrix` into Pauli-Z operators"""
+    nqubits = int(np.log2(h_matrix.shape[0]))
+
+    decomposition = []
+    for Z_i in pauli_operators:
+        expect = np.trace(h_matrix @ Z_i) / 2**nqubits
+        decomposition.append(expect)
+    return decomposition
```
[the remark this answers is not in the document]
marekgluza: @renatomello we have the following problem:
Let $D$ be a diagonal operator and $Z_i$ for $i=1,...,2^{L}$ all products of pauli-Z operators on $L$ qubits.
We want to write $D=\sum_i D_i Z_i$ and have a funtion that computs $D_i = \langle D, Z_i\rangle$

Can you comment if `quantum_info` has this functionality?
As I understand this file doesn't have it
https://github.com/qiboteam/qibo/blob/master/src/qibo/quantum_info/basis.py

src/qibo/models/dbi/utils_gradients.py
```diff
@@ -0,0 +1,202 @@
...
+from qibo.models.dbi.utils import commutator
+from qibo.models.dbi.utils_scheduling import polynomial_step
+
+
+class d_ansatz_type(Enum):
+
+    element_wise = auto()
+    local_1 = auto()
+    # local_2 = auto() # for future implementation
+    # ising = auto() # for future implementation
```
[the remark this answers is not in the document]
marekgluza: @wrightjandrew let's just remove it

src/qibo/models/dbi/double_bracket.py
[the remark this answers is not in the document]
wrightjandrew: You are right, nonetheless, I found that the function doesn't work when we do a step of the dbi (i.e. we do dbi_object(s,d) and then we do dbi_object.energy_fluctuation). After a dbi step, we change the matrix of the Hamiltonian but not the entire Hamiltonian object associated with the dbi class. I also changed it so we update the entire Hamiltonian as follows (line 134 in double_bracket.py):

self.h.matrix = operator @ self.h.matrix @ operator_dagger
self.h = Hamiltonian(nqubits=int(np.log2(len(self.h.matrix))), matrix=self.h.matrix)

but the issue seems to persist. Am I correctly updating the Hamiltonian object?

src/qibo/models/dbi/utils.py
wrightjandrew: @Sam-XiaoyueLi when you have the time, please take a look at the docstrings I made to see if they accurately describe the functions :)

marekgluza:
[No action required] Note for future: The next commit will remove the analytical approaches to optimizing $D$. We're not opening new branches because we found no advantage over the numerical methods which will feature in the merged version.

andrea-pasquale (approved):

marekgluza (approved):

marekgluza:
@scarrazza that was a long one! :) Many thanks among others to @wrightjandrew for the initiative on adding cost functions that help gradient descent and to @Sam-XiaoyueLi for suggesting the new optimizers and implementing the strategies + tests (there's a lot of things that went into this multi branch PR). Also thanks @andrea-pasquale for helping us finalize the tests!

PR #1272, opened 2024-03-18
`QulacsBackend`

BrunoLiegiBastonLiegi:
This adds `qulacs` as an optional backend.

Checklist:
- [ ] Tests are passing.


BrunoLiegiBastonLiegi:
@scarrazza some preliminary results for the qulacs backend, it looks indeed comparable with our `qibojit (numba)` 

qulacs_vs_numba-1

scarrazza:
Thanks @BrunoLiegiBastonLiegi, could you please include some GPU curves too?

BrunoLiegiBastonLiegi:
Yes I am trying to install the `qulacs-gpu` package on the cluster, but it requires a little bit of machinery. The `pip install qulacs-gpu` doesn't work, thus I am not sure how we could set that as an optional dependency.

scarrazza (changes requested):
Looks good, however we are missing documentation.

BrunoLiegiBastonLiegi:
I added a small paragraph in the documentation about `qulacs`, there is not much to say as at this level we are supporting the basic features of `qulacs` only. What still needs to updated is the scheme in the getting started section that summarizes all the available backends. @MatteoRobbiati can I ask you hand with that?

MatteoRobbiati:
I just opened a [PR in the Xmind-diagrams repo](https://github.com/qiboteam/xmind-diagrams/pull/29) adding Qulacs to diagrams.

BrunoLiegiBastonLiegi:
Apparently there is some problem with `qulacs` and `python3.12`, even though it should be supporting `python > 3.7` from what stated in the readme.

scarrazza:
@alecandido could you please try to install and run tests from this branch in your mac?

BrunoLiegiBastonLiegi:
I tried on my mac (`darwin`) with `python3.9` and it builds correctly

scarrazza:
Did you run with poetry?

BrunoLiegiBastonLiegi:
Yep `poetry install --with tests`

alecandido:
Sorry for the delay: I'm also able to install with Poetry and run tests. And just one of them is failing.
```py
        log.error(message)
>       raise exception(message)
E       ValueError: Backend qulacs is not available. To check which backends are installed use `qibo.list_available_backends()`.

src/qibo/config.py:46: ValueError
--------------------------------------------------------------------------------------------- Captured stderr call ---------------------------------------------------------------------------------------------
[Qibo 0.2.9|ERROR|2024-06-03 15:35:25]: Backend qulacs is not available. To check which backends are installed use `qibo.list_available_backends()`.
---------------------------------------------------------------------------------------------- Captured log call -----------------------------------------------------------------------------------------------
ERROR    qibo.config:config.py:45 Backend qulacs is not available. To check which backends are installed use `qibo.list_available_backends()`.
```

BrunoLiegiBastonLiegi:
Yeah thanks for spotting this, I forgot to add `qulacs` to the `QIBO_NATIVE_BACKENDS` global var. It should be fixed now.

BrunoLiegiBastonLiegi:
By the way, I think the problem with `3.9` is always related to some missing dependecies, namely the `c++ boost` library. Could it be that the `qulacs` wheel they distribute for `python3.9` is missing it? However, that would be weird, beacuse I remember we did not have problem with `3.9` before.

alecandido:
The problem is a distribution problem: Qulacs is not distributing py3.9 wheels for MacOS ARM
image
The asymmetry on their side is causing the asymmetry in our CI (which, only for py3.9 ARM, is attempting to build the wheel).

The missing library should have not been a difference among different Python versions (at least to first order approximation).

src/qibo/backends/qulacs.py
```diff
@@ -0,0 +1,81 @@
...
+        qasm_str = re.sub("^//.+\n", "", circuit.to_qasm())
+        qasm_str = re.sub(r"creg\s.+;", "", qasm_str)
+        qasm_str = re.sub(r"measure\s.+;", "", qasm_str)
+        circ = self.converter.convert_QASM_to_qulacs_circuit(qasm_str.splitlines())
+        return circ
+
+    def execute_circuit(
+        self,
+        circuit: "qibo.Circuit",
+        nshots: int = 1000,
```
stavros11, line 48: I think you need to add
```suggestion
        circuit: "qibo.Circuit",
        initial_state=None,
        nshots: int = 1000,
```
here, otherwise executing any circuit using the `GlobalBackend` interface, for example:
```py
import qibo

qibo.set_backend("qulacs")


c = Circuit(2)
c.add(gates.H(0))
c.add(gates.CNOT(0, 1))

print(c())
```
will lead to
```sh
TypeError: QulacsBackend.execute_circuit() takes from 2 to 3 positional arguments but 4 were
```

It is also weird that this is not captured in the tests, given that this interface is the most basic one, even advertised as the minimum example in our README.

src/qibo/backends/__init__.py
```diff
@@ -8,10 +8,11 @@
 from qibo.backends.npmatrices import NumpyMatrices
 from qibo.backends.numpy import NumpyBackend
 from qibo.backends.pytorch import PyTorchBackend
+from qibo.backends.qulacs import QulacsBackend
 from qibo.backends.tensorflow import TensorflowBackend
 from qibo.config import log, raise_error
 
-QIBO_NATIVE_BACKENDS = ("numpy", "tensorflow", "pytorch")
+QIBO_NATIVE_BACKENDS = ("numpy", "tensorflow", "pytorch", "qulacs")
```
alecandido, line 15: https://github.com/qiboteam/qibo/pull/1368

tests/test_measurements.py
```diff
@@ -435,7 +435,7 @@ def test_measurement_basis_list(backend):
     c.add(gates.H(2))
     c.add(gates.X(3))
     c.add(gates.M(0, 1, 2, 3, basis=[gates.X, gates.Z, gates.X, gates.Z]))
-    result = c(nshots=100)
+    result = backend.execute_circuit(c, nshots=100)
```
alecandido, line 438: I like a lot the explicit backend usage, but why do you need it for this PR?

src/qibo/backends/qulacs.py
```diff
@@ -0,0 +1,81 @@
...
+        super().__init__()
+        import qulacs  # pylint: disable=import-error
+        from qulacs import (  # pylint: disable=no-name-in-module, import-error
+            QuantumCircuitSimulator,
+            converter,
+        )
+
+        self.qulacs = qulacs
+        self.simulator = QuantumCircuitSimulator
+        self.converter = converter
```
alecandido, line 22: Why do you need to attribute the converter to the instance?

src/qibo/backends/qulacs.py
```diff
@@ -0,0 +1,81 @@
...
+
+class QulacsBackend(NumpyBackend):
+
+    def __init__(self):
+        super().__init__()
+        import qulacs  # pylint: disable=import-error
+        from qulacs import (  # pylint: disable=no-name-in-module, import-error
+            QuantumCircuitSimulator,
+            converter,
+        )
```
alecandido, line 18: Why not top-level?

You could cut the `import` chain by scoping the `import` of this backend inside `qibo/backends/__init__.py`, but it'd make sense that if you `import` the Qulacs backend, you will import Qulacs as well :P

src/qibo/backends/qulacs.py
```diff
@@ -0,0 +1,81 @@
...
+            circuit (:class:`qibo.models.circuit.Circuit`): Input circuit to convert.
+
+        Returns:
+            qulacs.QuantumCircuit: The converted qulacs circuit.
+        """
+        qasm_str = re.sub("^//.+\n", "", circuit.to_qasm())
+        qasm_str = re.sub(r"creg\s.+;", "", qasm_str)
+        qasm_str = re.sub(r"measure\s.+;", "", qasm_str)
+        circ = self.converter.convert_QASM_to_qulacs_circuit(qasm_str.splitlines())
+        return circ
```
alecandido, line 43: If `converter` were not an attribute of `self` (but just an imported variable), this could be a top-level function, no need of a method :)

src/qibo/backends/qulacs.py
```diff
@@ -0,0 +1,81 @@
...
+    def __init__(self):
+        super().__init__()
+        import qulacs  # pylint: disable=import-error
+        from qulacs import (  # pylint: disable=no-name-in-module, import-error
+            QuantumCircuitSimulator,
+            converter,
+        )
+
+        self.qulacs = qulacs
+        self.simulator = QuantumCircuitSimulator
```
alecandido, line 21: Same for this: why an attribute?

alecandido (commented):
Thanks @BrunoLiegiBastonLiegi for the effort!

src/qibo/backends/qulacs.py
BrunoLiegiBastonLiegi: Thanks, I'll raise an error in case `initial_state != None` and test it

tests/test_measurements.py
BrunoLiegiBastonLiegi: Because otherwise only the `GlobalBackend` is tested here, if I am not mistaken. Anyway, this was failing with the `QulacsBackend`, but was not supposed to be testing it, even though it did due to the `GlobalBackend` usage.

src/qibo/backends/qulacs.py
BrunoLiegiBastonLiegi: I did not remember why I had not imported it top level, so, I tried again, and the problem is that the import in `/qibo/backends/__init__.py` will break if `qulacs` is not installed, [https://github.com/qiboteam/qibo/actions/runs/9663894252/job/26657235334?pr=1272](https://github.com/qiboteam/qibo/actions/runs/9663894252/job/26657235334?pr=1272)

better to have it inside then, I would say.

alecandido: Definitely, I agree.

But we can make it non-top here:
https://github.com/qiboteam/qibo/blob/94d5c89cd8bf13e969f9abf432854a8d8b3e18c9/src/qibo/backends/__init__.py#L43-L44
and break the chain.

BrunoLiegiBastonLiegi: ah ok I see, thus you are proposing of adding the imports to `__init__.py`?

``` python
elif backend == "qulacs":
    import qulacs  # pylint: disable=import-error
    from qulacs import (  # pylint: disable=no-name-in-module, import-error
        QuantumCircuitSimulator,
        converter,
     )
     return QulacsBackend()
```

It looks just a little bit weird to me that you import things here that you use in `qulacs.py`, but I am fine with this in case.

BrunoLiegiBastonLiegi: ah yeah, of course, that was so simple indeed 🤦‍♂️

tests/test_measurements.py
stavros11: > I actually searched, and only a few tests are actually using `set_backend()`, most of them just to test it.

This is fine, but there are also tests that are using `circuit()` without `set_backend()`. These are also invoking the `GlobalBackend` to execute the circuit and are even worse, because the `GlobalBackend` will be whetever was set beforehand (by the last `set_backend()` of a previous test). These may also be a bit harder to search for, because they can appear in various ways: `circuit()`, `circuit.execute()`, `c()`, `whatever_circuit_name()`, `c(nshots=100)` (like the one fixed here), etc.. An easy way would be to remove `Circuit.__call__` and `Circuit.execute` methods and see which tests fail.

src/qibo/backends/qulacs.py
```diff
@@ -0,0 +1,87 @@
+import re
+
+import numpy as np
+import qulacs  # pylint: disable=import-error
+from qulacs import (  # pylint: disable=no-name-in-module, import-error
```
stavros11, line 5: Do we need this exceptions because otherwise pylint fails for mac with py3.9 in the CI? If yes, I guess that's fine, otherwise I am not sure if they are needed given that CI is using the test dependencies which include qulacs.

stavros11 (approved):
Thanks @BrunoLiegiBastonLiegi, looks good to me and seems to work after the latest fixes.

tests/test_models_qcnn.py
```diff
@@ -13,6 +13,7 @@
 
 def test_classifier_circuit2():
     """ """
+    set_backend("numpy")
```
stavros11, line 16: Following the discussion above, this is also not very good practice because it will also set numpy as the `GlobalBackend` for all upcoming tests. However, since everything should work with numpy, it will probably not cause any issue. Maybe we could just open an issue for the test discussion and postpone for now.

BrunoLiegiBastonLiegi: Yeah, I agree, however the `QCNN` model doesn't seem to support the specification of the execution backend and always uses the `GlobalBackend`, I opened an issue about this already #1362. The tests were originally using the `NumpyBackend` (the global one) and thus I set it explicitely here.

src/qibo/backends/qulacs.py
BrunoLiegiBastonLiegi: Yes, if I remember correctly `pylint` was complaining in `macos-python3.9`

tests/test_measurements.py
alecandido: > By the way, I think that this discussion is not very relevant to this PR and maybe should be converted to an issue.

I agree about the issue. If you wish, feel free to open yourself, you're definitely the most expert.

> An easy way would be to remove `Circuit.__call__` and `Circuit.execute` methods and see which tests fail.

Another option is to keep `Circuit.__call__`, but require a `backend` argument to be passed.

stavros11: > I agree about the issue. If you wish, feel free to open yourself, you're definitely the most expert.

#1374

scarrazza: Do we have a list of tests relying on a random state of the `GlobalBackend`?

stavros11: > Do we have a list of tests relying on a random state of the `GlobalBackend`?

These are the tests that execute circuits using the `circuit.execute` interface, which invokes the `GlobalBackend()`:

* `test_backends_global::test_circuit_execution` (sets to numpy)
* `test_measurements::test_measurement_compiled_circuit` (does not set)
* `test_models_circuit_execution::test_compiled_execute` (does not set)
* Various tests in `test_models_qcnn` are using the `GlobalBackend` but setting it beforehand to qibojit or numpy without reseting it after.
* `test_models_variational::test_vqe`  (does not set)
* `test_models_variational::test_custom_loss` (does not set)

src/qibo/backends/qulacs.py
alecandido: I'd say this is definitely the best way you could do it. At least is confined within the `qibo.backends.qulacs` module, though not restricted to the offending macos+py3.9 combo.

If done in the workflow, it would be positioned in a larger scope, though you may be able to ignore specifically that file for that combo.

src/qibo/backends/qulacs.py
```diff
@@ -0,0 +1,87 @@
...
+    circ = converter.convert_QASM_to_qulacs_circuit(qasm_str.splitlines())
+    return circ
+
+
+class QulacsBackend(NumpyBackend):
+
+    def __init__(self):
+        super().__init__()
+
+        self.qulacs = qulacs
```
alecandido, line 40: Sorry to be annoying: the motivation for `self.np` is that there is a (non-standard) NumPy interface adopted even by some other packages, so you could hope to swap `self.np = numpy` with `self.np = jax.numpy` and hope it will work.

However, I don't see much the point of `self.qulacs`...

alecandido (approved):
Just one last tiny comment. Other than that, it is good for me.

Thanks @BrunoLiegiBastonLiegi!

src/qibo/backends/qulacs.py
BrunoLiegiBastonLiegi: At the moment, there isn't, I just thought that in the future maybe one might need to call `qulacs` primitives from the backend in files where `qulacs` is not explicitely imported.

alecandido: Well, if you can import `qulacs` somewhere, you can import it everywhere. That's mainly up to Python's module system, but it already guarantees that it will always be the same `qulacs` (unless explicitly unloaded)

alecandido:
@BrunoLiegiBastonLiegi whenever the CI passes, feel free to merge :)

(just make sure before that your branch is up-to-date with `master`, other than lacking conflicts - better to test merges in the PRs than in `master`)

alecandido:
@scarrazza could you withdraw your review?

scarrazza (approved):

Issue #1266, opened 2024-03-14
Gradients for Pytorch backend


Simone-Bordoni:
Add automatic differentiation support to pytorch backend after #1202 is merged.

renatomello:
What's the status here?

Simone-Bordoni:
I just created the PR, I will try to finish this week. I noticed that the only problem regards the parameter shift rule.

PR #1276, opened 2024-03-19
Autodifferentiation support for `PytorchBackend`

Simone-Bordoni:
Complete pytorch backend, close issue #1266 

Checklist:
- [x] Tests are passing.


Simone-Bordoni:
The only differentiation method that was not working for the pytorch backend was the parameter shift rule.
However now i have unmuted the test and removed the check on the backend and everything is working correctly for the pytorch backend. Probably the problem was linked to some error in the initial implementation of the pytorch backend and was fixed by correcting all the other tests.

renatomello:
> The only differentiation method that was not working for the pytorch backend was the parameter shift rule. However now i have unmuted the test and removed the check on the backend and everything is working correctly for the pytorch backend. Probably the problem was linked to some error in the initial implementation of the pytorch backend and was fixed by correcting all the other tests.

`PSR` is the only thing that was missing. There's also all the classes inside `qibo.models.variational`. They need to work properly with the `optmizers`. RIght now the tests are not passing, they are being skipped.

Simone-Bordoni:
> > The only differentiation method that was not working for the pytorch backend was the parameter shift rule. However now i have unmuted the test and removed the check on the backend and everything is working correctly for the pytorch backend. Probably the problem was linked to some error in the initial implementation of the pytorch backend and was fixed by correcting all the other tests.
> 
> There's also all the classes inside `qibo.models.variational`. They need to work properly with the `optmizers`. RIght now the tests are not passing, they are being skipped.

ok, I forgot these tests, I will try to fix them tomorrow

Simone-Bordoni:
I have encountered a problem while coding the SGD training procedure of the VQE for the pytorch backend.
The problem is that when executing a circuit with parametrized gates that require gradients the parameters of the gates are initially casted as numpy and this is not compatible with the gradients (see failed test).
Fixing the issue requires modifying numpy_matrices.py, as this module is of fundamental importance @renatomello @BrunoLiegiBastonLiegi @MatteoRobbiati maybe we can discuss how to modify it in order to make it work correctly with torch.

Simone-Bordoni:
> Do you mean this cast? https://github.com/qiboteam/qibo/blob/7b4489e6c95e5844796e8c08b6dbaacc9e28984e/src/qibo/backends/npmatrices.py#L76C1-L76C72
> 
> Because this is overloaded by the `TorchMatrices` object I believe.

That cast is fine. The problem is in the two lines before:
`
cos = self.np.cos(theta / 2.0) + 0j
isin = -1j * self.np.sin(theta / 2.0)
`
when theta is a torch tensor with requires_grad = True you can't compute cos/sin using numpy.

BrunoLiegiBastonLiegi:
> > Do you mean this cast? https://github.com/qiboteam/qibo/blob/7b4489e6c95e5844796e8c08b6dbaacc9e28984e/src/qibo/backends/npmatrices.py#L76C1-L76C72
> > Because this is overloaded by the `TorchMatrices` object I believe.
> 
> That cast is fine. The problem is in the two lines before: `cos = self.np.cos(theta / 2.0) + 0j isin = -1j * self.np.sin(theta / 2.0)` when theta is a torch tensor with requires_grad = True you can't compute cos/sin using numpy.

but then why isn't `self.np = torch` for `TorchMatrices` as it is for the `TorchBackend`?

Simone-Bordoni:
> > > Do you mean this cast? https://github.com/qiboteam/qibo/blob/7b4489e6c95e5844796e8c08b6dbaacc9e28984e/src/qibo/backends/npmatrices.py#L76C1-L76C72
> > > Because this is overloaded by the `TorchMatrices` object I believe.
> > 
> > 
> > That cast is fine. The problem is in the two lines before: `cos = self.np.cos(theta / 2.0) + 0j isin = -1j * self.np.sin(theta / 2.0)` when theta is a torch tensor with requires_grad = True you can't compute cos/sin using numpy.
> 
> but then why isn't `self.np = torch` for `TorchMatrices` as it is for the `TorchBackend`?

I can try to fix things this way. With Renato we just wanted to have the matrices defined with numpy and then casted to the correct backend.

renatomello:
> > > Do you mean this cast? https://github.com/qiboteam/qibo/blob/7b4489e6c95e5844796e8c08b6dbaacc9e28984e/src/qibo/backends/npmatrices.py#L76C1-L76C72
> > > Because this is overloaded by the `TorchMatrices` object I believe.
> > 
> > 
> > That cast is fine. The problem is in the two lines before: `cos = self.np.cos(theta / 2.0) + 0j isin = -1j * self.np.sin(theta / 2.0)` when theta is a torch tensor with requires_grad = True you can't compute cos/sin using numpy.
> 
> but then why isn't `self.np = torch` for `TorchMatrices` as it is for the `TorchBackend`?

This solution should be fine

Simone-Bordoni:
The autodifferentiation for pytorch backend is finally ready, good luck with the review!

Simone-Bordoni:
I should have finished fixing the tests with the DBI, many problems were linked to the deepcopy function that can't be used on leave graph pytorch tensors. Hopefully this PR will be ready to merge

tests/test_gates_gates.py
```diff
@@ -877,10 +876,9 @@ def test_generalized_fsim(backend):
     matrix = np.eye(4, dtype=target_state.dtype)
     matrix[1:3, 1:3] = rotation
     matrix[3, 3] = np.exp(-1j * phi)
-    matrix = backend.cast(matrix, dtype=matrix.dtype)
-    target_state[:4] = np.dot(matrix, target_state[:4])
-    target_state[4:] = np.dot(matrix, target_state[4:])
-    backend.assert_allclose(final_state, target_state)
+    target_state[:4] = np.matmul(matrix, target_state[:4])
+    target_state[4:] = np.matmul(matrix, target_state[4:])
+    backend.assert_allclose(final_state, target_state, atol=1e-6)
```
renatomello, line 881: This test was failing on GPUs, but I fixed it

tests/test_gates_gates.py
```diff
@@ -1545,33 +1545,30 @@ def test_controlled_fsim(backend):
...
     matrix[3, 3] = np.exp(-1j * phi)
-    matrix = backend.cast(matrix, dtype=matrix.dtype)
     ids = [56, 57, 60, 61]
-    target_state[ids] = np.dot(matrix, target_state[ids])
+    target_state[ids] = np.matmul(matrix, target_state[ids])
     ids = [58, 59, 62, 63]
-    target_state[ids] = np.dot(matrix, target_state[ids])
-    backend.assert_allclose(final_state, target_state)
+    target_state[ids] = np.matmul(matrix, target_state[ids])
+    backend.assert_allclose(final_state, target_state, atol=1e-6)
```
renatomello, line 1554: Same thing about GPUs for this test

src/qibo/backends/npmatrices.py
```diff
@@ -1,3 +1,5 @@
+import cmath
```
renatomello, line 1: It seems like `cmath` is only needed for the `exp` function, not the other ones

src/qibo/gates/channels.py
```diff
@@ -97,7 +97,9 @@ def to_choi(self, nqubits: Optional[int] = None, order: str = "row", backend=Non
             kraus_op.append(gate)
             kraus_op = kraus_op.matrix(backend)
             kraus_op = vectorization(kraus_op, order=order, backend=backend)
-            super_op += coeff * np.outer(kraus_op, np.conj(kraus_op))
+            super_op = super_op + coeff * backend.np.outer(
+                kraus_op, backend.np.conj(kraus_op)
+            )
```
renatomello, line 102:
```suggestion
            super_op += coeff * backend.np.outer(
                kraus_op, backend.np.conj(kraus_op)
            )
```

src/qibo/models/variational.py
```diff
@@ -93,7 +93,10 @@ def minimize(
             # TODO: check if we can use this shortcut
             # dtype = getattr(self.hamiltonian.backend.np, self.hamiltonian.backend._dtypes.get('DTYPE'))
             dtype = self.hamiltonian.backend.np.float64
-            loss = lambda p, c, h: dtype(loss_func(p, c, h))
+            if str(dtype) == "torch.float64":
+                loss = lambda p, c, h: loss_func(p, c, h).item()
+            else:
+                loss = lambda p, c, h: dtype(loss_func(p, c, h))
```
renatomello, line 99:
```suggestion
            loss = (
                lambda p, c, h: loss_func(p, c, h).item() 
                if str(dtype) == "torch.float64" 
                else lambda p, c, h: dtype(loss_func(p, c, h))
            )
            
```

src/qibo/optimizers.py
```diff
@@ -272,7 +271,54 @@ def sgd(
...
+            log.warning(
+                "PyTorch does not support compilation of the optimization graph."
+            )
+        return _sgd_torch(
+            loss, initial_parameters, args, sgd_options, backend, callback=callback
+        )
+    else:
+        raise_error(
+            RuntimeError, "SGD optimizer requires Tensorflow or PyTorch backend."
+        )
```
renatomello, line 295:
```suggestion
    if backend.name == "tensorflow":
        return _sgd_tf(
            loss,
            initial_parameters,
            args,
            sgd_options,
            compile,
            backend,
            callback=callback,
        )
    
    if backend.name == "pytorch":
        if compile:
            log.warning(
                "PyTorch does not support compilation of the optimization graph."
            )
        return _sgd_torch(
            loss, initial_parameters, args, sgd_options, backend, callback=callback
        )

    raise_error(
        RuntimeError, "SGD optimizer requires Tensorflow or PyTorch backend."
    )
```

src/qibo/quantum_info/basis.py
```diff
@@ -100,25 +100,28 @@ def pauli_basis(
     else:
         basis_full = basis_single
 
+    basis_full = backend.cast(basis_full)
```
renatomello, line 103:
```suggestion
    basis_full = backend.cast(basis_full, dtype=basis_full[0].dtype)
```

src/qibo/quantum_info/basis.py
```diff
@@ -100,25 +100,28 @@ def pauli_basis(
...
             row = vectorization(row, order=order, backend=backend)
-            row_indexes = list(np.flatnonzero(row))
+            row_indexes = list(np.flatnonzero(backend.to_numpy(row)))
             indexes.append(row_indexes)
             basis.append(row[row_indexes])
             del row
     elif vectorize and not sparse:
         basis = [
-            vectorization(matrix, order=order, backend=backend) for matrix in basis_full
+            vectorization(backend.cast(matrix), order=order, backend=backend)
```
renatomello, line 115:
```suggestion
            vectorization(backend.cast(matrix, dtype=matrix.dtype), order=order, backend=backend)
```

src/qibo/quantum_info/basis.py
```diff
@@ -100,25 +100,28 @@ def pauli_basis(
...
         basis_full = basis_single
 
+    basis_full = backend.cast(basis_full)
+
     if vectorize and sparse:
         basis, indexes = [], []
         for row in basis_full:
             row = vectorization(row, order=order, backend=backend)
-            row_indexes = list(np.flatnonzero(row))
+            row_indexes = list(np.flatnonzero(backend.to_numpy(row)))
```
renatomello, line 109:
```suggestion
            row_indexes = (
                backend.np.nonzero(row).flatten()
                if isinstance(backend, PyTorchBackend)
                else list(np.flatnonzero(row))
            )
```

src/qibo/quantum_info/basis.py
```diff
@@ -100,25 +100,28 @@ def pauli_basis(
...
         basis = [
-            vectorization(matrix, order=order, backend=backend) for matrix in basis_full
+            vectorization(backend.cast(matrix), order=order, backend=backend)
+            for matrix in basis_full
         ]
     else:
         basis = basis_full
 
-    basis = backend.cast(basis, dtype=backend.dtype)
+    basis = backend.cast(basis)
```
renatomello, line 121:
```suggestion
    basis = backend.cast(basis, dtype=basis[0].dtype)
```

src/qibo/quantum_info/basis.py
```diff
@@ -100,25 +100,28 @@ def pauli_basis(
...
 
     if normalize:
-        basis /= np.sqrt(2**nqubits)
+        basis = basis / np.sqrt(2**nqubits)
 
     if vectorize and sparse:
         indexes = backend.cast(indexes)
```
renatomello, line 127:
```suggestion
        indexes = backend.cast(indexes, dtype=indexes[0][0].dtype)
```

src/qibo/quantum_info/basis.py
```diff
@@ -267,12 +269,12 @@ def pauli_to_comp_basis(
...
         backend=backend,
     )
-    unitary = np.transpose(unitary)
+    unitary = unitary.T
 
     if sparse:
         elements, indexes = [], []
         for row in unitary:
-            index_list = list(np.flatnonzero(row))
+            index_list = list(np.flatnonzero(backend.to_numpy(row)))
```
renatomello, line 277:
```suggestion
            index_list = (
                backend.np.nonzero(row).flatten()
                if isinstance(backend, PyTorchBackend)
                else list(np.flatnonzero(row))
            )
```

src/qibo/models/encodings.py
```diff
@@ -156,6 +156,8 @@ def unary_encoder(data, architecture: str = "tree"):
     circuit_rbs, _ = _generate_rbs_pairs(nqubits, architecture=architecture)
     circuit += circuit_rbs
 
+    if data.__class__.__name__ == "Tensor":
+        data = data.detach().numpy()
```
renatomello, line 160: This shouldn't be the solution because the same input data might be something that needs gradients later on.

src/qibo/quantum_info/entropies.py
```diff
@@ -56,17 +56,19 @@ def shannon_entropy(prob_dist, base: float = 2, backend=None):
...
+    )
 
     shan_entropy = -backend.np.sum(prob_dist * log_prob)
 
     # absolute value if entropy == 0.0 to avoid returning -0.0
-    shan_entropy = np.abs(shan_entropy) if shan_entropy == 0.0 else shan_entropy
+    shan_entropy = backend.np.abs(shan_entropy) if shan_entropy == 0.0 else shan_entropy
 
-    return complex(shan_entropy).real
+    return complex(backend.to_numpy(shan_entropy)).real
```
renatomello, line 71:
```suggestion
    return np.real(backend.to_numpy(shan_entropy))
```

src/qibo/quantum_info/entropies.py
```diff
@@ -511,6 +511,8 @@ def relative_von_neumann_entropy(
         float: Relative (von-Neumann) entropy :math:`S(\\rho \\, \\| \\, \\sigma)`.
     """
     backend = _check_backend(backend)
+    state = backend.cast(state)
+    target = backend.cast(target)
```
renatomello, line 515:
```suggestion
    state = backend.cast(state, dtype=state.dtype)
    target = backend.cast(target, dtype=target.dtype)
```

src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -139,7 +139,7 @@ def expectation(self, state, normalize=False):
         )
 
     def expectation_from_samples(self, freq, qubit_map=None):
-        obs = self.matrix
+        obs = self.backend.to_numpy(self.matrix)
```
renatomello, line 142: same comment I made in person is valid here

src/qibo/quantum_info/random_ensembles.py
```diff
@@ -230,7 +230,7 @@ def random_unitary(dims: int, measure: Optional[str] = None, seed=None, backend=
 
     if measure == "haar":
         unitary = random_gaussian_matrix(dims, dims, seed=local_state, backend=backend)
-        Q, R = np.linalg.qr(unitary)
+        Q, R = np.linalg.qr(backend.to_numpy(unitary))
```
renatomello, line 233: Same here

src/qibo/transpiler/unitary_decompositions.py
```diff
@@ -71,10 +71,8 @@ def calculate_psi(unitary, magic_basis=magic_basis, backend=None):
     )
     # construct and diagonalize UT_U
     ut_u = backend.np.transpose(u_magic, (1, 0)) @ u_magic
-    # When the matrix given to np.linalg.eig is a diagonal matrix up to machine precision the decomposition
-    # is not accurate anymore. decimals = 20 works for random 2q Clifford unitaries. This problem only happens with Tensorflow backend.
     if backend.name == "tensorflow":
```
renatomello:
```suggestion
    if backend.__class__.__name__ == "TensorflowBackend":
```

src/qibo/quantum_info/entropies.py
```diff
@@ -56,17 +57,19 @@ def shannon_entropy(prob_dist, base: float = 2, backend=None):
 
     total_sum = backend.np.sum(prob_dist)
 
-    if np.abs(total_sum - 1.0) > PRECISION_TOL:
+    if np.abs(backend.to_numpy(total_sum) - 1.0) > PRECISION_TOL:
```
renatomello, line 60: I'm quite sure that this would work since `total_sum` is just a `foat`.
```suggestion
    if np.abs(float(total_sum) - 1.0) > PRECISION_TOL:
```
This is valid everywhere else this type of structure appears

src/qibo/quantum_info/metrics.py
```diff
@@ -229,60 +226,68 @@ def fidelity(state, target, check_hermitian: bool = False, backend=None):
...
-            ):
-                eigenvalues, eigenvectors = np.linalg.eigh(state)
-            else:
-                eigenvalues, eigenvectors = np.linalg.eig(state)
+            eigenvalues, eigenvectors = backend.calculate_eigenvectors(
+                state,
+                hermitian=(
+                    check_hermitian is False
+                    or _check_hermitian_or_not_gpu(state, backend=backend)
+                ),
```
renatomello, line 243: I think it's best to put this `bool` in a variable first just for the sake of code clarity

src/qibo/quantum_info/metrics.py
```diff
@@ -229,60 +226,68 @@ def fidelity(state, target, check_hermitian: bool = False, backend=None):
...
-            ):
-                eigenvalues, eigenvectors = np.linalg.eigh(fid)
-            else:
-                eigenvalues, eigenvectors = np.linalg.eig(fid)
+            eigenvalues, eigenvectors = backend.calculate_eigenvectors(
+                fid,
+                hermitian=(
+                    check_hermitian is False
+                    or _check_hermitian_or_not_gpu(state, backend=backend)
+                ),
```
renatomello, line 265: Same here

src/qibo/quantum_info/metrics.py
```diff
@@ -362,8 +369,13 @@ def bures_distance(state, target, check_hermitian: bool = False, backend=None):
...
-    distance = np.sqrt(
-        2 * (1 - np.sqrt(fidelity(state, target, check_hermitian, backend=backend)))
+    backend = _check_backend(backend)
+    distance = backend.np.sqrt(
+        2
+        * (
+            1
+            - backend.np.sqrt(fidelity(state, target, check_hermitian, backend=backend))
+        )
     )
```
renatomello, line 379: Same here regarding the `sqrt` of `fidelity`

src/qibo/quantum_info/utils.py
```diff
@@ -469,25 +477,30 @@ def pqc_integral(circuit, power_t: int, samples: int, backend=None):
...
-def _hadamard_transform_1d(array):
+def _hadamard_transform_1d(array, backend=None):
     # necessary because of tf.EagerTensor
     # does not accept item assignment
-    array_copied = np.copy(array)
+    backend = _check_backend(backend)
+    if backend.name == "pytorch":
+        copy_func = backend.np.copy
+    else:
+        copy_func = np.copy
```
renatomello, line 494:
```suggestion
    copy_func = backend.np.copy if backend.__class__.__name__ == "PyTorchBackend" else np.copy
```

src/qibo/transpiler/unitary_decompositions.py
```diff
@@ -71,15 +71,14 @@ def calculate_psi(unitary, magic_basis=magic_basis, backend=None):
     )
     # construct and diagonalize UT_U
     ut_u = backend.np.transpose(u_magic, (1, 0)) @ u_magic
-    # When the matrix given to np.linalg.eig is a diagonal matrix up to machine precision the decomposition
-    # is not accurate anymore. decimals = 20 works for random 2q Clifford unitaries.
-    if backend.__class__.__name__ == "TensorflowBackend":
+    if backend.__class__.__name__ != "PyTorchBackend":
+        # eig seems to have a different behavior based on backend/hardware,
+        # use np.round to increase precision seems to fix the issue
```
renatomello, line 76: `round` is not being used anymore but it's still in the comments

tests/test_backends.py
```diff
@@ -128,3 +128,16 @@ def test_list_available_backends():
...
+    if backend.name != "pytorch":
+        pytest.skip("Check gradients activation only for pytorch backend.")
+    assert backend.gradients
+    assert backend.matrices.requires_grad
+    backend.requires_grad(False)
+    assert not backend.gradients
+    assert not backend.matrices.requires_grad
+    backend.requires_grad(True)
+    assert backend.gradients
+    assert backend.matrices.requires_grad
```
renatomello, line 143: Maybe add a test asserting the values of the gradients?

src/qibo/transpiler/unitary_decompositions.py
Simone-Bordoni: I had to put back the np.round to make tests pass

renatomello: This seems like a bigger underlying problem that should be solved in a different PR. Maybe an issue here should be opened about it.

Simone-Bordoni: The problem was known before and we did not manage to find a more elegant solution.
Here the old discussion in qibolab:  https://github.com/qiboteam/qibolab/pull/501
If you want I can opens a new issue here in Qibo

src/qibo/models/encodings.py
```diff
@@ -435,10 +435,20 @@ def _generate_rbs_angles(data, nqubits: int, architecture: str):
         list: list of phases for RBS gates.
     """
     if architecture == "diagonal":
-        phases = [
-            math.atan2(np.linalg.norm(data[k + 1 :]), data[k])
-            for k in range(len(data) - 2)
-        ]
+        if data.__class__.__name__ == "Tensor":
+            import torch  # pylint: disable=C0415
```
BrunoLiegiBastonLiegi, line 439: this pattern here is repeated several times, maybe at this point it could be useful to define a helper function to set the engine, something like:

``` python
def _check_torch(array: "ndarray") -> "engine":
    if array.__class__.__name__ == "Tensor":
        import torch

        return torch
    else:
        return np
```

and then always call methods from the engine

``` python
engine = _check_torch(array)
engine.method_needed()
```

src/qibo/quantum_info/basis.py
```diff
@@ -100,28 +100,37 @@ def pauli_basis(
...
+
     if vectorize and sparse:
         basis, indexes = [], []
         for row in basis_full:
             row = vectorization(row, order=order, backend=backend)
-            row_indexes = list(np.flatnonzero(row))
+            row_indexes = (
+                backend.np.nonzero(row).flatten()
+                if backend.__class__.__name__ == "PyTorchBackend"
+                else list(np.flatnonzero(row))
```
BrunoLiegiBastonLiegi, line 112: maybe adding to the torch backend the `flatnonzero` function could simplify this, namely in the `__init__` defining:
``` python
setattr(self.np, "flatnonzero", lambda x: self.np.nonzero(x).flatten())
```

src/qibo/quantum_info/basis.py
```diff
@@ -267,12 +275,16 @@ def pauli_to_comp_basis(
...
+    unitary = unitary.T
 
     if sparse:
         elements, indexes = [], []
         for row in unitary:
-            index_list = list(np.flatnonzero(row))
+            index_list = (
+                backend.np.nonzero(row).flatten()
+                if backend.__class__.__name__ == "PyTorchBackend"
+                else list(np.flatnonzero(row))
```
BrunoLiegiBastonLiegi, line 286: same as above

src/qibo/quantum_info/utils.py
```diff
@@ -469,25 +477,29 @@ def pqc_integral(circuit, power_t: int, samples: int, backend=None):
...
 
 
-def _hadamard_transform_1d(array):
+def _hadamard_transform_1d(array, backend=None):
     # necessary because of tf.EagerTensor
     # does not accept item assignment
-    array_copied = np.copy(array)
+    backend = _check_backend(backend)
+    copy_func = (
+        backend.np.copy if backend.__class__.__name__ == "PyTorchBackend" else np.copy
```
BrunoLiegiBastonLiegi, line 492: in which case `backend.np.copy` is not working here?

BrunoLiegiBastonLiegi (commented):
Thanks @Simone-Bordoni I went through all the source files, I have just some minor suggestions. I'll move to the tests next.

src/qibo/backends/pytorch.py
```diff
@@ -8,17 +8,28 @@
...
-        self.torch = torch
+        self.np = torch
         self.dtype = dtype
+        self.requires_grad = requires_grad
 
     def _cast(self, x, dtype):
-        return self.torch.as_tensor(x, dtype=dtype)
+        flattened = [item for sublist in x for item in sublist]
+        tensor_list = [self.np.as_tensor(i, dtype=dtype) for i in flattened]
+        return self.np.stack(tensor_list).reshape(len(x), len(x))
```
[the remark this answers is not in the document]
Simone-Bordoni: Unfortunately as_tensor is not keeping the gradient with respect to the parameters of the tensor while stack is doing the work.

src/qibo/models/dbi/utils.py
```diff
@@ -263,3 +268,27 @@ def covariance(a, b):
...
+
+    # Manually copy each attribute
+    new.h = copy(dbi_object.h)
+    new.h0 = copy(dbi_object.h0)
+    new.mode = deepcopy(dbi_object.mode)
+    new.scheduling = deepcopy(dbi_object.scheduling)
+    new.cost = deepcopy(dbi_object.cost)
+    new.ref_state = (
+        deepcopy(dbi_object.ref_state) if dbi_object.ref_state is not None else None
+    )
```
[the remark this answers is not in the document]
Simone-Bordoni: deepcopy doesn't work with pytorch tensors when using gradients (when tensors are graph leaves), this is why i need to manually copy h and h0

src/qibo/backends/pytorch.py
```diff
@@ -39,17 +53,23 @@ def __init__(self):
...
         self.np.transpose = self.np.permute
+        self.np.copy = self.np.clone
         self.np.expand_dims = self.np.unsqueeze
         self.np.mod = self.np.remainder
         self.np.right_shift = self.np.bitwise_right_shift
 
+    def requires_grad(self, requires_grad):
+        """Enable or disable gradient calculation."""
+        self.gradients = requires_grad
+        self.matrices.requires_grad = requires_grad
```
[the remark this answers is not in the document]
Simone-Bordoni: I think that it may be useful to disable the gradients directly from the backend @renatomello what do you think?

src/qibo/gates/gates.py
```diff
@@ -2515,17 +2515,19 @@ def __init__(
                 all_function = torch.all
                 conj_function = torch.conj
                 transpose_function = torch.transpose
+                abs_function = torch.abs
             else:
                 diag_function = np.diag
                 all_function = np.all
                 conj_function = np.conj
                 transpose_function = np.transpose
+                abs_function = np.abs
```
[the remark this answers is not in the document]
Simone-Bordoni: I am adding the function in gates

tests/test_models_variational.py
```diff
@@ -101,12 +101,10 @@ def myloss(parameters, circuit, target):
...
     """Performs a VQE circuit minimization test."""
-    if backend.name == "pytorch":
-        pytest.skip("Skipping VQE test for pytorch backend.")
-    if (method == "sgd" or compile) and backend.name != "tensorflow":
+    if (method == "sgd" or compile) and (backend.name not in ("tensorflow", "pytorch")):
         pytest.skip("Skipping SGD test for unsupported backend.")
-    if method != "sgd" and backend.name == "tensorflow":
-        pytest.skip("Skipping scipy optimizers for tensorflow.")
+    if method != "sgd" and backend.name in ("tensorflow", "pytorch"):
+        pytest.skip("Skipping scipy optimizers for tensorflow and pytorch.")
```
BrunoLiegiBastonLiegi, line 107: In this test as well, since we only test torch and tf, maybe rather than using the backend fixture do
``` python
@pytest.mark.parametrize(engine, [PyTorchBackend(), TensorflowBackend()])
def test_vqe(engine, ...):
```

tests/test_models_variational.py
```diff
@@ -127,6 +125,8 @@ def test_vqe(backend, method, options, compile, filename):
     hamiltonian = hamiltonians.XXZ(nqubits=nqubits, backend=backend)
     np.random.seed(0)
     initial_parameters = np.random.uniform(0, 2 * np.pi, 2 * nqubits * layers + nqubits)
+    if backend.name == "pytorch":
+        initial_parameters = backend.np.tensor(initial_parameters, requires_grad=True)
```
BrunoLiegiBastonLiegi, line 129:
```suggestion
    backend.cast_parameters(initial_parameters, requires_grad=True)
```
isn't this working?

tests/test_models_variational.py
```diff
@@ -268,11 +268,14 @@ def test_qaoa_errors(backend):
 
 @pytest.mark.parametrize(test_names, test_values)
 def test_qaoa_optimization(backend, method, options, dense, filename):
-    if method == "sgd" and backend.name != "tensorflow":
+    if (method == "sgd") and (backend.name not in ["tensorflow", "pytorch"]):
         pytest.skip("Skipping SGD test for unsupported backend.")
+    if method != "sgd" and backend.name in ("tensorflow", "pytorch"):
+        pytest.skip("Skipping scipy optimizers for tensorflow and pytorch.")
```
BrunoLiegiBastonLiegi, line 274: same as above

tests/test_quantum_info_random.py
```diff
@@ -234,7 +234,11 @@ def test_random_statevector(backend, seed):
     # tests if random statevector is a pure state
     dims = 4
     state = random_statevector(dims, seed=seed, backend=backend)
-    backend.assert_allclose(abs(purity(state) - 1.0) < PRECISION_TOL, True)
+    print(state)
+    print(purity(state, backend=backend))
```
BrunoLiegiBastonLiegi, line 238:
```suggestion
```

tests/test_models_variational.py
Simone-Bordoni: here we are testing tensorflow and pytorch for the sgd method but then we are also testing scipy optimizers for the other backends

tests/test_models_variational.py
Simone-Bordoni: also here we are using all backends, check the second condition

src/qibo/backends/npmatrices.py
```diff
@@ -71,20 +75,25 @@ def M(self):  # pragma: no cover
         raise_error(NotImplementedError)
 
     def RX(self, theta):
+        theta = self._cast_parameter(theta)
         cos = self.np.cos(theta / 2.0) + 0j
         isin = -1j * self.np.sin(theta / 2.0)
         return self._cast([[cos, isin], [isin, cos]], dtype=self.dtype)
```
BrunoLiegiBastonLiegi, line 81:
```suggestion
        theta = self._cast_parameter(theta)
        cos = self.np.cos(theta / 2.0) + 0j
        isin = -1j * self.np.sin(theta / 2.0)
        return self.np.stack(([cos, isin, isin, cos])).reshape(2, 2)
```
I tested this locally and it seems to be working without the need of the convoluted workaround in `_cast`. I would say that updating all the parametrized gate this way might be better.

renatomello: This will create problems with flexible gates e.g. the new gate in https://github.com/qiboteam/qibo/pull/1373 (and also maybe `gates.Unitary`? I didn't test it but it seems so).

BrunoLiegiBastonLiegi: What is the problem with the `GeneralizedRBS`? From what I see, you always build the matrix and cast it to the appropriate array type, thus `stack` should be fine as well, am I missing something? In general, anyways, this does not have to be applied to all the gates, you could always handle special cases with the current `_cast` in case it is needed (although at the moment no example comes to my mind). I tested the two alternative methods on `(2,2)` matrices and doing directly the `stack` seems to be 30 to 40% faster, clearly on a single gate call doesn't make any difference, but if you pile them up maybe....

renatomello: > What is the problem with the `GeneralizedRBS`? From what I see, you always build the matrix and cast it to the appropriate array type, thus `stack` should be fine as well, am I missing something? 

`GeneralizedRBS` and `Unitary` do not have a fixed matrix representation, so one cannot hardcode their flattened versions  like in your suggestion above.

> In general, anyways, this does not have to be applied to all the gates, you could always handle special cases with the current `_cast` in case it is needed (although at the moment no example comes to my mind). 

Sure, @Simone-Bordoni could add an exception just for `Unitary` right now (if actually needed) and then we handle `GeneralizedRBS` in #1373

BrunoLiegiBastonLiegi: Ok nevermind, I tested numpy with the same direct `stack` and it seems to be way slower, not worth improving pytorch by a little bit while worsening numpy. At this point, either you rewrite the parametric gates in the `TorchMatrices` to maintain the small speed up at the cost of some code duplication, or you just live with a suboptimal pytorch matrix generation.

renatomello: I would just leave it as is. I know having two cast functions is not the prettiest. However, `_cast_parameter` is only needed because of `torch` and `tensorflow` and these two will be moved to `qiboml` soon, right?

BrunoLiegiBastonLiegi (approved):

renatomello (approved):

PR #1277, opened 2024-03-19
Add raw property to Unitary gate serialization

Jacfomg:
To save the circuits from RB https://github.com/qiboteam/qibocal/pull/735 with the inverse gate we would need to serialize the unitary gate as well.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/gates/gates.py
+++ b/src/qibo/gates/gates.py
@@ -4,7 +4,12 @@ from typing import List
 import numpy as np
 
 from qibo.config import PRECISION_TOL, raise_error
-from qibo.gates.abstract import Gate, ParametrizedGate
+from qibo.gates.abstract import (
+    REQUIRED_FIELDS,
+    REQUIRED_FIELDS_INIT_KWARGS,
+    Gate,
+    ParametrizedGate,
+)
 from qibo.parameter import Parameter
 
 
@@ -2357,3 +2362,32 @@ class Unitary(ParametrizedGate):
     def _dagger(self):
         ud = np.conj(np.transpose(self.parameters[0]))
         return self.__class__(ud, *self.target_qubits, **self.init_kwargs)
+
+    @property
+    def raw(self) -> dict:
+        """Serialize to dictionary.
+
+        The values used in the serialization should be compatible with a
+        JSON dump (or any other one supporting a minimal set of scalar
+        types). Though the specific implementation is up to the specific
+        gate.
+        """
+        encoded = self.__dict__
+
+        encoded_simple = {
+            key: value for key, value in encoded.items() if key in REQUIRED_FIELDS
+        }
+
+        encoded_simple["init_kwargs"] = {
+            key: value
+            for key, value in encoded_simple["init_kwargs"].items()
+            if key in REQUIRED_FIELDS_INIT_KWARGS
+        }
+
+        for value in encoded_simple:
+            if isinstance(encoded[value], set):
+                encoded_simple[value] = list(encoded_simple[value])
+
+        encoded_simple["_class"] = type(self).__name__
+
+        return encoded_simple


renatomello:
It seems you need this for RB protocol(s). Wouldn't it be easier to dump the Clifford gates directly?

Jacfomg:
Right now, they calculated the inverse gate as:

```python
circuit.add(gates.Unitary(circuit.unitary(), *range(circuit.nqubits)).dagger())
```
to only dump clifford gates I suppose we should sample also build the circuit with cliffords instead of U3s and calculate the inverse as a clifford as well. Then they will get to be U3s later so we can convert them to pulses but that wouldn't matter for the dumping I guess.

renatomello:
>I suppose we should sample also build the circuit with cliffords instead of U3s 

You could use `quantum_info.random_clifford` for that. For one qubit, it returns a combination (IIRC up to 4) of elementary Clifford gates that combine to give all 24 single-qubit Cliffords.

>and calculate the inverse as a clifford as well. 
That will also be a combination of up to 4 elementary Cliffords.

>Then they will get to be U3s later so we can convert them to pulses but that wouldn't matter for the dumping I guess.

If you need an $U_{3}$, you can call `random_clifford(1).fuse()`. Also, you don't need to dump a matrix. These are only elementary Clifford gates. The only parameter needed is their name, so just a string.

alecandido:
Actually, I would agree with @renatomello concerning the RB, since it is definitely better to use the most efficient representation.

But this PR could be useful in any case: if Qibo allows defining arbitrary unitary gates, we should also be able to serialize them.
Then, we could advise the users to make limited use of it, when alternatives are available (as in this case), but supporting half the way is not the best option.

Is anything missing from this PR? (other than testing)

renatomello:
> Actually, I would agree with @renatomello concerning the RB, since it is definitely better to use the most efficient representation.
> 
> But this PR could be useful in any case: if Qibo allows defining arbitrary unitary gates, we should also be able to serialize them. Then, we could advise the users to make limited use of it, when alternatives are available (as in this case), but supporting half the way is not the best option.
> 
> Is anything missing from this PR? (other than testing) @Jacfomg

I understand the consistency argument. However, dumping a matrix to a `dict` is not a great solution overall.

alecandido:
> I understand the consistency argument. However, dumping a matrix to a `dict` is not a great solution overall.

Well, we can dump it as bytes with `np.save`. Bytes in JSON are the same bytes you would save as a standalone file. If needed, we could even encode it as `base64` (supported by standard library).

It is true in general that it is better to avoid serializing a whole matrix, but if you have no other choice, that's just your best option.

Jacfomg:
I will take a look for coding the RB with the `quantum_info.random_clifford` to see if we would need anything else.

Jacfomg:
> > I understand the consistency argument. However, dumping a matrix to a `dict` is not a great solution overall.
> 
> Well, we can dump it as bytes with `np.save`. Bytes in JSON are the same bytes you would save as a standalone file. If needed, we could even encode it as `base64` (supported by standard library).
> 
> It is true in general that it is better to avoid serializing a whole matrix, but if you have no other choice, that's just your best option.

If the agreement is this one, I will not close it. Yes, it is lacking testing if we want to merge it.

Jacfomg:
> I will take a look for coding the RB with the `quantum_info.random_clifford` to see if we would need anything else.

After talking with Ingo, he explained that using this function would not generate things as he wants so I will not use for RB related stuff.

alecandido:
> After talking with Ingo, he explained that using this function would not generate things as he wants so I will not use for RB related stuff.

In which way the serialization is connected to what is good for the RB?

I agree with @renatomello that using `Unitary` could be an overkill, but I don't see how serializing them could be different from what Ingo wants (it doesn't seem related all...).

Jacfomg:
> > After talking with Ingo, he explained that using this function would not generate things as he wants so I will not use for RB related stuff.
> 
> In which way the serialization is connected to what is good for the RB?
> 
> I agree with @renatomello that using `Unitary` could be an overkill, but I don't see how serializing them could be different from what Ingo wants (it doesn't seem related all...).

It was not about serialization but about using the `quantum_info.random_clifford` function that @renatomello suggested. It will not sample as he believes it's best.

alecandido:
> It was not about serialization but about using the `quantum_info.random_clifford` function that @renatomello suggested. It will not sample as he believes it's best.

Ok, this is perfectly acceptable :)

scarrazza:
@Jacfomg @renatomello @alecandido what is the plan with this PR?

alecandido:
My understanding is that this is not needed for the purpose it was designed, and I know that the present state of the code is adding nothing to what is already there (the implemented `.raw` is a copy-paste of the parent one, and the dump of the tensor is not yet addressed).

I still believe it would be useful to serialize whole circuits, whatever their content is. But we are already doing this in `qibo-core`, so I won't personally work on this PR.

It is true that `qibo-core` is doing plenty of other things, and it will still take a while to be viable for actual usage (i.e. ~2-3 months at least, considering the person power available).
While if you need remote execution for unitaries this could be fixed in a couple of hours (with list conversion, or copying the code for the array dump in JSON from Qibolab 0.2). Still, if there is no need, I won't commit that time.

Not sure about what the others are planning.

PR #1282, opened 2024-03-25
`MetaBackend`

### LINKED ISSUE #1096: List available backends
It would be nice if Qibo were able to check if a backend is available, and have a function to survey all the registered backends, and list those available.

Possibly, it could be implemented even trying to load it, and catching the failure. But if each backend had a property `available` it would be even better.
(since the backend might know about the required conditions for availability, and we'd avoid catching a generic error)

BrunoLiegiBastonLiegi:
This PR implements a `MetaBackend` object that will call the different projects' (`qibojit`, `qibolab`, etc...) `MetaBackend` to load their backends. This will replace the current `construct_backend` function and will also give access to a `list_available()` function to list all the backends available in the system.

Closes #1096.

Checklist:
- [ ] Tests are passing.


src/qibo/backends/__init__.py
```diff
@@ -8,66 +9,61 @@
...
+            return TensorflowBackend()
+        elif backend == "pytorch":
+            return PyTorchBackend()
+        elif backend == "clifford":
+            engine = kwargs.pop("platform", None)
+            kwargs["engine"] = engine
+            return CliffordBackend(**kwargs)
+        elif backend in QIBO_NON_NATIVE_BACKENDS:
+            module = import_module(backend.replace("-", "_"))
+            return getattr(module, "MetaBackend").load(**kwargs)
```
alecandido, line 42: I'm not sure whether I'd nest that here.

Instead, we could make a separate function, and the Qibo `MetaBackend` could be just one like anyone else, providing a nice example of how to implement one.

Eventually, we could discover backends with some runtime mechanism, without relying on something like
https://github.com/qiboteam/qibo/blob/f43a6a1a7eea95c0da1ecb65a9c6d346ecfe0a0a/src/qibo/backends/__init__.py#L13
(e.g. by having installed additional packages prefixed with `qibo-backend-` or something like that).
But I'd keep it simple for this PR, just properly lay the ground.

src/qibo/backends/__init__.py
```diff
@@ -8,66 +9,61 @@
...
+            except:  # pragma: no cover
+                available = False
+            available_backends[backend] = available
+        for backend in QIBO_NON_NATIVE_BACKENDS:
+            try:
+                module = import_module(backend.replace("-", "_"))
+                available = getattr(module, "MetaBackend")().list_available()
+            except:
+                available = False
+            available_backends.update({backend: available})
```
alecandido, line 65: Same as above.

src/qibo/backends/__init__.py
```diff
@@ -200,3 +196,11 @@ def _check_backend(backend):
...
 
     return backend
+
+
+def list_available_backends():
+    return MetaBackend().list_available()
+
+
+def construct_backend(backend, **kwargs):
+    return MetaBackend.load(backend, **kwargs)
```
alecandido, line 206: These are the functions that could implement the mechanism above.

tests/test_models_circuit_backpropagation.py
```diff
@@ -8,9 +8,9 @@
 
 def construct_tensorflow_backend():
     try:
-        from qibo.backends import construct_backend
+        from qibo.backends import MetaBackend
 
-        backend = construct_backend("tensorflow")
+        backend = MetaBackend.load("tensorflow")
```
alecandido, line 13: Here, and in the other tests, I'd keep the `construct_backend` function.

The reason is partly the one above, but even that we can simply avoid to break, and keep using the old function. Since, in this case, it costs nothing to us, I'd just do that.

alecandido (commented):
A suggestion about how to rearrange the non-native backends. But other than that, it looks perfect.

Thanks @BrunoLiegiBastonLiegi to have taken care to implement this idea everywhere :D

BrunoLiegiBastonLiegi:
We have the same problem here as in #1239. Namely, since now in main `qibo` depends on `qibotn/py3.12` and `qibojit/py3.12`, I cannot resolve the conflicts without breaking something. For `qibojit` there is a simple solution which is changing the `qibo` dependency to `qibojit/main` as the corresponding `qibojit` PR has already been merged, however for `qibotn` the `py3.12` PR does not seem to be mergeable soon.

alecandido:
True, this py3.12 upgrade is being a bit of a nightmare (I hope `qibo-core` will land soon...)

> however for `qibotn` the `py3.12` PR does not seem to be mergeable soon.

Here I disagree a bit, since https://github.com/qiboteam/qibotn/pull/48 should be as good as `main`, and py3.12 compatible.
I just need the two approvals, thus I'm adding you as well as a reviewer :P

stavros11 (approved):

pyproject.toml
```diff
@@ -69,8 +69,8 @@ pylint = "^3.1.0"
 matplotlib = "^3.7.0"
 tensorflow = { version = "^2.16.1", markers = "sys_platform == 'linux'" }
 torch = "^2.1.1"
-qibojit = { git = "https://github.com/qiboteam/qibojit.git", branch="clifford_simulator_numba" }
-qibotn = { git = "https://github.com/qiboteam/qibotn.git" }
+qibojit = { git = "https://github.com/qiboteam/qibojit.git", branch="list_backends" }
```
stavros11, line 72: I think https://github.com/qiboteam/qibojit/pull/172 can be merged. So maybe we should merge the PR in qibojit first and point this back to main before merging?

poetry.lock
```diff
@@ -3567,17 +3558,17 @@ tests = ["coverage[toml] (==5.0.4)", "pytest (>=6.0.0,<7.0.0)"]
 
 [[package]]
 name = "pylint"
-version = "3.1.0"
+version = "3.2.0"
```
alecandido, line 3561: I fear the problem is coming from here, but not sure. I'd try to pin Pylint to `3.1.0` in `pyproject.toml` and regenerate the lock.

If this is the actual reason, we could merge with the Pylint version pinned, but we should also consider fixing the errors.

alecandido: I applied the suggestion myself in a83e0c2, let's what happens

alecandido: Ok, hypothesis confirmed:
https://github.com/qiboteam/qibo/actions/runs/9099789800/job/25013165174#step:10:16
```py
src/qibo/backends/__init__.py:21:40: E0602: Undefined variable 'Backend' (undefined-variable)
src/qibo/backends/__init__.py:205:44: E0602: Undefined variable 'Backend' (undefined-variable)
```

Now the error is sensible, since related to this PR, and it should be fixed (it's just a missing import).

However, it could be worth to fix the previous ones as well.

alecandido:
@BrunoLiegiBastonLiegi the issue with Pylint 3.2 is not completely trivial, thus I opened #1328 for it.

At this point, it's not worth to stop this any longer. Feel free to merge whenever you wish :)

poetry.lock
alecandido: #1328

BrunoLiegiBastonLiegi:
I was waiting for `qibotn` and `qibo-cloud-backends` to be merged before merging this one, just to avoid breaking the `set_backend` for those two.

PR #1284, opened 2024-03-27
Fix to classical registers in to_qasm

### LINKED ISSUE #1283: Change to Circuit.measurements broke to_qasm()
**Describe the bug**
This bug was discovered from the Mitiq integration. Currently, things are working properly with version 0.2.4, however in both 0.2.5 and 0.2.6 the to_qasm() method in qibo>models>circuit.py builds the QASM file incorrectly when there are multiple classical measurement registers. Checking the code, the to_qasm() method did not change, but instead the measurements part of the circuit implementation changed.

Here is a quick reproduction of the issue:
```
(mitiqqibo) zzzzz@zzzzz mitiq_bg % pip list | grep qibo
qibo                            0.2.6
(mitiqqibo) zzzzz@zzzzz mitiq_bg % python
Python 3.11.8 (main, Feb 26 2024, 15:36:12) [Clang 14.0.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from qibo.models.circuit import Circuit as QiboCircuit
>>> qasm = 'OPENQASM 2.0;\ninclude "qelib1.inc";\nqreg q[2];\ncreg m0[2];\ny q[0];\nx q[1];\nmeasure q[0] -> m0[0];\nmeasure q[1] -> m0[1];'
>>> qibo_cirq = QiboCircuit.from_qasm(qasm)
>>> qasm_final = qibo_cirq.to_qasm()
>>> print(qasm + "\n\n" + qasm_final)
OPENQASM 2.0;
include "qelib1.inc";
qreg q[2];
creg m0[2];
y q[0];
x q[1];
measure q[0] -> m0[0];
measure q[1] -> m0[1];

// Generated by QIBO 0.2.6
OPENQASM 2.0;
include "qelib1.inc";
qreg q[2];
creg m0[1];
creg m0[1];
y q[0];
x q[1];
measure q[0] -> m0[0];
measure q[1] -> m0[0];
```

Here are screenshots showing the difference in the measurements property of the Qibo circuit:
0.2.4
image

0.2.6
image

**To Reproduce**
Run the following test code:
```
from qibo.models.circuit import Circuit as QiboCircuit

qasm = 'OPENQASM 2.0;\ninclude "qelib1.inc";\nqreg q[2];\ncreg m0[2];\ny q[0];\nx q[1];\nmeasure q[0] -> m0[0];\nmeasure q[1] -> m0[1];'

qibo_cirq = QiboCircuit.from_qasm(qasm)
qasm_final = qibo_cirq.to_qasm()

print(qasm)
print(qasm_final)```



**Expected behavior**
Calling to_qasm() should generate a valid QASM file.

### PR DESCRIPTION
**Pull request by BrunoLiegiBastonLiegi**
Small fix to how the classical registers and measurements are converted to qasm.
This should close #1283 

Checklist:
- [ ] Tests are passing.

### CODE REVIEW

**Comment by stavros11** [comment]
I think the issue with #1283 is in `from_qasm`. The fix implemented here fixes the issue at the QASM level, however the Qibo circuit created by `from_qasm` is still a bit problematic because it contains two different registers with the same name:
```py

qasm = 'OPENQASM 2.0;\ninclude "qelib1.inc";\nqreg q[2];\ncreg m0[2];\ny q[0];\nx q[1];\nmeasure q[0] -> m0[0];\nmeasure q[1] -> m0[1];'

qibo_cirq = Circuit.from_qasm(qasm)

for gate in qibo_cirq.queue:
    if isinstance(gate, gates.M):
        print(gate, gate.qubits, gate.register_name)
```
prints
```
<qibo.gates.measurements.M object at 0x78a057949f10> (0,) m0
<qibo.gates.measurements.M object at 0x78a078565610> (1,) m0
```
instead of
```
<qibo.gates.measurements.M object at 0x78a057949f10> (0, 1) m0
```
that I would expect.

This has undesired effects if the circuit is simulated:
```py
result = qibo_cirq(nshots=10000)
```
then
```
In [3]: result.frequencies(registers=False)
Out[3]: Counter({'11': 10000})

In [4]: result.frequencies(registers=True)
Out[4]: {'m0': Counter({'1': 10000})}  # one of the registers is ignored because you cannot have the same key twice
```

Allowing different measurements with the same `register_name` in the circuit is kind of debatable, personally I would raise an error if the same `register_name` is used more than once, but others may disagree. However, either way I think that the given `qasm` should be translated to a single measurement `gates.M(0, 1, register_name="m0")` instead of two as it does currently.

BrunoLiegiBastonLiegi:
Ah I see. When I rewrote the `from_qasm` I explicitely modified the behaviour of the registers because I thought that doing something like:

```
c.add(gates.M(0, register="m0"))
c.add(gates.M(1, register="m0"))
```

should not raise an error. In my opinion, it is perfectly fine to, in some sense, 'append' measurements to a register. Therefore, now I see three different options:
 - Fix `frequencies`, and most likely `samples` as well, to handle this possibility
 - Implement some kind of `merge_registers` routine that merges measurements associated to the same register, but this is problematic for the `MeasurementResult` I believe, because the results:
```
m0 = c.add(gates.M(0, register="m0"))
m1 = c.add(gates.M(1, register="m0"))
``` 
are not gonna be associated to the merged measurement anymore.
 - Drop the support for this and revert the changes 

Probably the last one is indeed the easiest and less bug prone option...

stavros11:
> Ah I see. When I rewrote the `from_qasm` I explicitely modified the behaviour of the registers because I thought that doing something like:
> 
> ```
> c.add(gates.M(0, register="m0"))
> c.add(gates.M(1, register="m0"))
> ```
> 
> should not raise an error. In my opinion, it is perfectly fine to, in some sense, 'append' measurements to a register. Therefore, now I see three different options:

Thanks for explaining, that makes sense. I agree that it is fine to provide this possibility, but of course as it is implemented now there is this side-effect when asking for `frequencies(registers=True)` and `samples(registers=True)` where only the latest measurement is included in the returned results and the rest are ignored.

> * Implement some kind of `merge_registers` routine that merges measurements associated to the same register, but this is problematic for the `MeasurementResult` I believe, because the results:
> 
> ```
> m0 = c.add(gates.M(0, register="m0"))
> m1 = c.add(gates.M(1, register="m0"))
> ```
> 
> are not gonna be associated to the merged measurement anymore.

I have not thought in detail, but it seems possible to make `m1` point to the same object as `m0`, but this may require some manipulations inside `Circuit.add`. The general idea is that `Circuit` (I believe) has access to all registers/measurement gates, so when you do `c.add(gates.M(1, register="m0"))` you can check if `"m0"` register already exists, add a new qubit to it and return the existing result object again. Of course results need to be mutable for that, but I believe they already are.

> * Drop the support for this and revert the changes
> 
> Probably the last one is indeed the easiest and less bug prone option...

I agree with this. The easiest solution is to just raise an error and not support reusing register names and only fix the qasm issue in this PR. We could also open an issue in case we wish to implement this feature later. From the user side doing 
```py
c.add(gates.M(0, register="m0"))
c.add(gates.M(1, register="m0"))
```
is not much different than
```py
c.add(gates.M(0, 1, register="m0"))
```
but there may be an application that I am missing that the former could be useful.

BrunoLiegiBastonLiegi:
The only case that comes to my mind is whether you would like to measure different qubits at different times but in the same classical register. Thus, for instance something like this:

```python
c.add(gates.H(0))
c.add(gates.M(0, register="m0"))
c.add(gates.H(1))
c.add(gates.CX(1,0))
c.add(gates.M(1, register="m0"))
```

but this is very niche, and probably there's no real reason to have both the measurements in the same register anyway.

MatteoRobbiati (approved):
Thanks! It works for me.

PR #1288, opened 2024-03-28
Coverage fix

BrunoLiegiBastonLiegi:
This restores the 100% coverage.

Checklist:
- [x] Tests are passing.


BrunoLiegiBastonLiegi:
@stavros11 Can I ask for your help to double check whether some lines that I commented out are still useful?
Namely, here 
https://github.com/qiboteam/qibo/blob/21c766e12290c2a2818132c4ef092119dab67e19/src/qibo/models/circuit.py#L704C2-L707C63
I don't think we allow for different measurements to be in the same register, right?
And here
https://github.com/qiboteam/qibo/blob/21c766e12290c2a2818132c4ef092119dab67e19/src/qibo/gates/abstract.py#L87C1-L90C1
I think this was needed because either the `qubits`, `target_qubits` or `control_qubits` attribute of `Gate` was originally initialized as a set, however, I believe I recently changed this in #1259, and thus it is no longer needed. Can you double check that?

stavros11:
> Namely, here https://github.com/qiboteam/qibo/blob/21c766e12290c2a2818132c4ef092119dab67e19/src/qibo/models/circuit.py#L704C2-L707C63 I don't think we allow for different measurements to be in the same register, right?

That is correct, based on the error here: https://github.com/qiboteam/qibo/blob/21c766e12290c2a2818132c4ef092119dab67e19/src/qibo/models/circuit.py#L671-L675
so it should be fine to drop the commented lines.
Not entirely related but https://github.com/qiboteam/qibo/blob/21c766e12290c2a2818132c4ef092119dab67e19/src/qibo/models/circuit.py#L708
doesn't seem to be doing much other than copying the `registers` dictionary, right? It shouldn't be different than simply
```py
return registers
```

> And here https://github.com/qiboteam/qibo/blob/21c766e12290c2a2818132c4ef092119dab67e19/src/qibo/gates/abstract.py#L87C1-L90C1 I think this was needed because either the `qubits`, `target_qubits` or `control_qubits` attribute of `Gate` was originally initialized as a set, however, I believe I recently changed this in #1259, and thus it is no longer needed. Can you confirm that?

I think I was not the author of `raw` and these particular lines, but I believe you are right, this should only be there to accommodate for `_control_qubits` being a `set`. Since this is no longer the case, it should be safe to drop.

scarrazza:
@BrunoLiegiBastonLiegi I believe this is almost ready to go, could you please merge master / fix the lock file?

tests/test_hamiltonians_from_symbols.py
```diff
@@ -10,6 +12,17 @@
 from qibo.symbols import I, Symbol, X, Y, Z
 
 
+@pytest.mark.parametrize("symbol", [I, X, Y, Z])
+def test_symbols_pickling(symbol):
+    symbol = symbol(int(np.random.randint(4)))
+    dumped_symbol = pickle.dumps(symbol)
+    new_symbol = pickle.loads(dumped_symbol)
+    for attr in ("target_qubit", "name", "_gate"):
+        print(attr)
```
stavros11, line 21:
```suggestion
```

stavros11 (approved):

renatomello:
> @BrunoLiegiBastonLiegi I believe this is almost ready to go, could you please merge master / fix the lock file?

Generating the updated lock file is taking upwards of 5min. Is that normal?

renatomello (approved):

PR #1302, opened 2024-04-23
Quantum dynamic programming model

marekgluza:
In https://arxiv.org/abs/2403.09187 by [Jeongrak Son](https://arxiv.org/search/quant-ph?searchtype=author&query=Son,+J), [Marek Gluza](https://arxiv.org/search/quant-ph?searchtype=author&query=Gluza,+M), [Ryuji Takagi](https://arxiv.org/search/quant-ph?searchtype=author&query=Takagi,+R), [Nelly H. Y. Ng](https://arxiv.org/search/quant-ph?searchtype=author&query=Ng,+N+H+Y) a framework for quantum recursion implementation using memory has been proposed.

This PR is adding a model into Qibo which parallels that paper and allows to explore backend-agnostic implementations of quantum dynamic programming (QDP).
More specifically, even though QDP is general this PR focuses on functioning on the level of established procedures, in particular density-matrix exponentiation using quantum instruction qubit recycling by means of (quantum) measurement emulation, see https://arxiv.org/abs/2001.08838.
The generalization to hermitian-map exponentiation is anticipated and can easily follow by inheritance of the classess provided.

src/qibo/models/qdp/dynamic_programming.py is the key class file
- the main base class is AbstractQuantumDynamicProgramming, which has predefines basic functions (such as measure a qubit for feedback reset) and 2 `@abstractmethod` which the user has to define: `memory_usage_query_circuit` and `memory_call_circuit`

A memory-usage query circuit is a `qibo.Circuit` whose unitary approximates $U =e^{i N}\in U(\mathcal H\otimes \mathcal H)$ implements the memory-usage query channel
$$\mathcal E_{s}^{\mathcal N,\rho}(\sigma) = Tr_{1}\left[e^{-iN s}\left(\rho\otimes\sigma\right)e^{iN s}\right]$$
instructed on $\rho$ and operating on $\sigma$.

A memory_call_circuit uses a Trotter-Suzuki decomposition of the desired dynamic unitary channel by means of memory usage queries
$$\mathcal E_{\text{QDP}}^{\mathcal N,\rho,M} := \left(\mathcal E_{1/M}^{\mathcal N,\rho}\right)^{M} = e^{i\mathcal N(\rho)} (\sigma) e^{-i\mathcal N(\rho)}+  O\left(1/{M}\right)$$
Here
   - the channel is dynamic because it is specified only during runtime when the quantum instruction becomes known
   - $\mathcal N(\rho) = \mathcal N(\rho)^\dagger$ is hermicity preserving and specifies how the quantum memory data is going to be used in the algorithm
   - $N$ is the partial transpose of the Choi matrix of $\mathcal N$

The approach take is to assume it is known how to compile unitaries generated by $N$ and then use this theory to implement the memory-call channel. We separate the nomenclature because tracing out generically will reduce purity of the working state $\sigma \mapsto \sigma'$. Instead we make only small rotations and repeat them many times - asymptotically implementing a memory call, i.e. calling quantum information from memory and then based on this input revealed on runtime modifying the state of the working registers. 

The implementation intricacies begin when constructively answering how to implement multiple memory-usage queries.
The specific issue is how to in practice bring in multiple quantum instruction states into play. There are 3 main child classes of AbstractQuantumDynamicProgramming, which have the `memory_call_circuit` defined based on the abstract memory_usage_query_circuit. They apply different methods to create the required number of copies of instruction qubits:

1. QDPSequentialInstruction: Instruction qubit is discarded after each use and we move to the next instruction qubit 
2. QDPMeasurementReset: Instruction qubit is reset after each use for reusing.
3. QDPMeasurementEmulation: Using quantum measurement emulation as laid out by [Kjaergaard (2020)](https://arxiv.org/abs/2001.08838), where we apply either a rotation gate or identity gate with probability 0.5 to emulate a measurement. 

Ultimately, any protocol withing the QDP framework that is to be executed on a Qibo backend is specified by `memory_usage_query_circuit` after inheriting from one of these 3 'transpiling' classes (transpiling in the abstract sense the logistics of providing quantum instructions; transpiling into a fixed architecture layout will be treated elsewhere). 

To evaluate the function, in this PR, we demonstrate DME (density matrix exponentiation), which appies $N$ in steps small angle given by the $SWAP$ gate which for 2 qubits is just the 2 qubit Heisenberg Hamiltonian after appropriate rescaling.
We use QDPSequentialInstruction protocol for generating copies of instruction state $\rho$.

Checklist:
- [ ] Tests are passing.


khanhuyengiang:
The qdp brank is now moved to https://github.com/qiboteam/qibo-qdp

alecandido:
@khanhuyengiang any reason to fork Qibo instead of just starting a new project?

khanhuyengiang:
@alecandido The initial `qdp` branch was originally implemented to be part of `qibo.models`, and we want to keep part of that work. I'll be refactoring the code soon to transform it into a standalone project.

alecandido:
Just as an advice, whenever that will be done, I'd suggest you to recreate the repo from scratch, since you don't need to keep the whole history of a repo that already exists, and which you will just consume as a dependency.

In any case, there is nothing bad if temporarily it happens to be more convenient to you in this way :)

PR #1327, opened 2024-05-17
Added global unitary folding for get_noisy_circuits

mho291:
Checklist:
- [ ] Tests are passing.


src/qibo/models/error_mitigation.py
```diff
@@ -51,51 +51,77 @@ def get_gammas(noise_levels, analytical: bool = True):
...
-        raise_error(
-            ValueError,
-            "Provide a circuit with at least 2 qubits when using the 'CNOT' insertion gate. "
-            + "Alternatively, try with the 'RX' insertion gate instead.",
-        )
 
-    i_gate = gates.CNOT if insertion_gate == "CNOT" else gates.RX
+    if global_unitary_folding:  # pragma: no cover
+        circuit_no_meas = circuit.__class__(**circuit.init_kwargs)
+        circuit_meas = circuit.__class__(**circuit.init_kwargs)
```
MatteoRobbiati, line 74: Are these copies of the circuit? If yes, I would use the `circuit.copy` method.

src/qibo/models/error_mitigation.py
```diff
@@ -51,51 +51,77 @@ def get_gammas(noise_levels, analytical: bool = True):
...
+        for gate in circuit.queue:
+            if gate.name != "measure":
+                circuit_no_meas.add(gate)
+            else:
+                circuit_meas.add(gate)
+
+        noisy_circuit = circuit.__class__(**circuit.init_kwargs)
```
MatteoRobbiati, line 81: See previous comment.

src/qibo/models/error_mitigation.py
```diff
@@ -51,51 +51,77 @@ def get_gammas(noise_levels, analytical: bool = True):
...
-                    noisy_circuit.add(gates.CNOT(control, target))
-            elif gate.init_kwargs["theta"] == theta:
-                qubit = gate.qubits[0]
-                for _ in range(num_insertions):
-                    noisy_circuit.add(gates.RX(qubit, theta=theta))
-                    noisy_circuit.add(gates.RX(qubit, theta=-theta))
+        for gate in circuit.queue:
+            noisy_circuit.add(gate)
+
+            if isinstance(gate, i_gate):
```
MatteoRobbiati, line 112: Why are you applying $U$ and $U^{\dagger}$ only when $U$ is found?
Is this a common practice? Otherwise, we could be more general and ask the user to select the position in the queue (or in the graph, speaking more `qibo-core` friendly) where to apply the gate-inverse mechanism.

MatteoRobbiati (changes requested):
Thanks for working on this!
 
Some comments follows, together with a more general comment here: 
I think the copy of the circuit can be done faster: you could create a circuit of the same size of the target similarly to what you are doing, adding only non-M gates and, later, you can append the `circuit.measurements` gates, which is the part of `circuit.queue` containing the measurements only.

src/qibo/models/error_mitigation.py
mho291: Hi Matteo, the purpose of lines L73 to L79, which reads,
```python
circuit_no_meas = circuit.__class__(**circuit.init_kwargs)
circuit_meas = circuit.__class__(**circuit.init_kwargs)
for gate in circuit.queue:
    if gate.name != "measure":
        circuit_no_meas.add(gate)
    else:
        circuit_meas.add(gate)
```
is to duplicate the input `circuit`. As its name suggests, `circuit_no_meas` is the input `circuit` without the measurements. Likewise, `circuit_meas` is the input `circuit` with measurements.

The reason I opted to have lines L73 to L79 is so that we can then construct our `noisy_circuit` by first creating a copy of `circuit_no_meas`.

Then we perform global unitary folding in lines L83 to L86, which reads,
```python
for _ in range(num_insertions):
    noisy_circuit += circuit_no_meas.invert() + circuit_no_meas

    noisy_circuit += circuit_meas
```

Would you advise to use `circuit.copy` or `Circuit(**circuit.init_kwargs)` to copy the circuit? What does the latter do?

Thanks very much!

MatteoRobbiati: What I mean is that you don't need two copies of the circuit to do what you aim to do:

```python

# build your circuit
c = Circuit(4)
for q in range(4):
    c.add(gates.H(q))
c.add(gates.CNOT(0,1))
c.add(gates.CNOT(3,1))
c.add(gates.M(0,3))

# initialize a copy with same arguments (dense, accelerators, etc)
copy_c = Circuit(**c.init_kwargs)

# populate the copy with non-M gates
for g in c.queue:
    if not isinstance(g, gates.M):
        copy_c.add(g)

# start building noisy circuit with U^+ and U
noisy_c = copy_c.invert() + copy_c

# add measurements according to original circuit
for m in c.measurements:
    noisy_c.add(m)
```

src/qibo/models/error_mitigation.py
MatteoRobbiati: Maybe @AlejandroSopena do you have a clearer picture of the ZNE implementation?

src/qibo/models/error_mitigation.py
mho291: Right! Thanks! I will use your implementation in the code. But we will need to add one more line before we build the noisy circuit so that our final circuit is `U (U^+ U)^n` for `n` repetitions of `(U^+ U)`.

```python
# initialize `noisy_c` as `copy_c`.
noisy_c = copy_c

# start building noisy circuit with U^+ and U
noisy_c = copy_c.invert() + copy_c

# add measurements according to original circuit
for m in c.measurements:
    noisy_c.add(m)
```

src/qibo/models/error_mitigation.py
AlejandroSopena: @MatteoRobbiati , what you propose can be implemented. In fact, it is common to fold some gates at random. However, if the fidelity of each gate is different, then the expected value in the zero-noise limit cannot be calculated as a linear superposition of expected values at different noise levels as we do. We would need to be implement  a numerical fit (see the comment in my review).

src/qibo/models/error_mitigation.py
```diff
@@ -51,51 +51,82 @@ def get_gammas(noise_levels, analytical: bool = True):
...
     """Standalone function to generate the noisy circuit with the inverse gate pairs insertions.
 
     Args:
         circuit (:class:`qibo.models.circuit.Circuit`): circuit to modify.
-        num_insertions (int): number of insertion gate pairs to add.
+        num_insertions (int): number of insertion gate pairs / global unitary folds to add.
+        global_unitary_folding (bool): Default True. Global unitary folding is carried out by default.
+            Global unitary folding takes precedence over insertion_gate.
         insertion_gate (str, optional): gate to be used in the insertion.
             If ``"RX"``, the gate used is :math:``RX(\\pi / 2)``.
```
AlejandroSopena, line 65:
```suggestion
        num_insertions (int): number of insertion gate pairs / global unitary folds to add.
        global_unitary_folding (bool): If ``True``, noise is increased by global unitary folding.
            If ``False``, local unitary folding is used. Defaults to ``True``.
        insertion_gate (str, optional): gate to be folded in local unitary folding.
            If ``RX``, the gate used is :math:`RX(\\pi / 2)`. Otherwise, it is the ``CNOT`` gate.
```

src/qibo/models/error_mitigation.py
```diff
@@ -51,51 +51,82 @@ def get_gammas(noise_levels, analytical: bool = True):
...
-            Default is ``"CNOT"``.
+            Otherwise, the other gate is ``"CNOT"``.
```
AlejandroSopena, line 66:
```suggestion
```

src/qibo/models/error_mitigation.py
```diff
@@ -51,51 +51,82 @@ def get_gammas(noise_levels, analytical: bool = True):
...
-        raise_error(
-            ValueError,
-            "Provide a circuit with at least 2 qubits when using the 'CNOT' insertion gate. "
-            + "Alternatively, try with the 'RX' insertion gate instead.",
-        )
 
-    i_gate = gates.CNOT if insertion_gate == "CNOT" else gates.RX
+    from qibo import Circuit  # pylint: disable=import-outside-toplevel
+
+    if global_unitary_folding:  # pragma: no cover
```
AlejandroSopena, line 74:
```suggestion
    if global_unitary_folding:
```

src/qibo/models/error_mitigation.py
```diff
@@ -51,51 +51,82 @@ def get_gammas(noise_levels, analytical: bool = True):
...
+
+        # Add (U^+ U)s
+        for _ in range(num_insertions):
+            noisy_circuit += copy_c.invert() + copy_c
+
+        # Add measurements according to the original circuit
+        for m in circuit.measurements:
+            noisy_circuit.add(m)
+
+    else:  # pragma: no cover
```
AlejandroSopena, line 93:
```suggestion
    else: 
```

src/qibo/models/error_mitigation.py
```diff
@@ -51,51 +51,82 @@ def get_gammas(noise_levels, analytical: bool = True):
...
+            )
 
-    theta = np.pi / 2
-    noisy_circuit = circuit.__class__(**circuit.init_kwargs)
+        i_gate = gates.CNOT if insertion_gate == "CNOT" else gates.RX
 
-    for gate in circuit.queue:
-        noisy_circuit.add(gate)
+        theta = np.pi / 2
+        noisy_circuit = circuit.__class__(**circuit.init_kwargs)
```
AlejandroSopena, line 112:
```suggestion
        noisy_circuit = Circuit(**circuit.init_kwargs)
```

src/qibo/models/error_mitigation.py
```diff
@@ -51,51 +51,82 @@ def get_gammas(noise_levels, analytical: bool = True):
...
+            noisy_circuit.add(gate)
+
+            if isinstance(gate, i_gate):
+                if insertion_gate == "CNOT":
+                    control = gate.control_qubits[0]
+                    target = gate.target_qubits[0]
+                    for _ in range(num_insertions):
+                        noisy_circuit.add(gates.CNOT(control, target))
+                        noisy_circuit.add(gates.CNOT(control, target))
+                # elif gate.init_kwargs["theta"] == theta:
```
AlejandroSopena, line 124:
```suggestion
```

src/qibo/models/error_mitigation.py
```diff
@@ -51,51 +51,82 @@ def get_gammas(noise_levels, analytical: bool = True):
...
+                elif insertion_gate == "RX":
+                    qubit = gate.qubits[0]
```
AlejandroSopena, line 126:
```suggestion
                    qubit = gate.qubits[0]
                    theta = gate.init_kwargs["theta"]
```

src/qibo/models/error_mitigation.py
```diff
@@ -51,51 +51,82 @@ def get_gammas(noise_levels, analytical: bool = True):
...
-            "Provide a circuit with at least 2 qubits when using the 'CNOT' insertion gate. "
-            + "Alternatively, try with the 'RX' insertion gate instead.",
-        )
 
-    i_gate = gates.CNOT if insertion_gate == "CNOT" else gates.RX
+    from qibo import Circuit  # pylint: disable=import-outside-toplevel
+
+    if global_unitary_folding:  # pragma: no cover
+
+        # Create a copy of input circuit without measurements
```
AlejandroSopena, line 76: To be consistent with the rest of the code, it is better to remove the comments.

AlejandroSopena (commented):
Thanks @mho291. 

I agree that the implementation of ZNE is not as general as it could be and that it can be extended by implementing global unitary folding. We have used local unitary folding because this way we have finer control over the circuit noise, meaning the noise increases less with each noise level compared to global unitary folding. 

This allows using Richardson extrapolation to obtain an expected value in the zero-noise limit. In other words, we construct a function where the expected value in the zero-noise limit is a linear combination (with analytically known coefficients) of the expected values at different noise levels ([line 215](https://github.com/qiboteam/qibo/blob/ZNE_get_noisy_circuit/src/qibo/models/error_mitigation.py#215)). This is equivalent to expressing the expected value of an observable as a function of the noise as an n-degree polynomial where n + 1 is the number of noise levels. The "problem" with this approach is that we need n + 1 noise levels to perform an n-degree polynomial fit (see [arxiv:1612.02058](https://arxiv.org/abs/1612.02058)). When the data points that are experimentally accessible all reflect a fairly high amount of noise, it is not possible to reach a sufficient number of noise levels. In global unitary folding, this is usually the case as the noise increases faster. Then, one must fit the data, typically to a lower-degree polynomial or an exponential decay (see [arxiv:2005.10921](https://arxiv.org/abs/2005.10921)) 

I agree with implementing global unitary folding, but with the current method of calculating the expected value in the zero-noise limit, it might not be useful in practice.

AlejandroSopena:
`global_unitary_folding` is missing in the docstrings of `ZNE()`.

mho291:
> `global_unitary_folding` is missing in the docstrings of `ZNE()`.

Thanks @AlejandroSopena for reviewing the code. I will amend according to your suggestions but will take some time as I am now away. Nonetheless I'd like to share some results using global unitary folding on IQM! 

We used ZNE to extrapolate and estimate the ideal value of the approximation ratio, which is the cut_size/max_cut using QAOA for a fairly large graph of 10 nodes (hence 10 qubits). The QAOA only had one layer and the circuit was at a depth of 21. We used global unitary folding.

3reg_graph_ZNE_approxratio

I was pretty impressed at the ZNE's use of Richardson extrapolation! It does seem like we cannot push the ZNE any further beyond maybe 6 or 7 global unitary foldings because the plateau is imminent, so you're right that we might need to use a different fit for the data if the circuit depth is too large.

mho291:
Hi @MatteoRobbiati and @AlejandroSopena. Sorry for the late reply, I just returned to work after mandatory national service. Just wondering if we can expedite the merging of this PR. We're in the midst of writing up an AWS blogpost for our Qibo-AWS work where we have results based on global unitary folding. Unfortunately our team cannot release the blogpost to the public unless this PR merged and we have a deadline to meet. Thanks so much!

AlejandroSopena:
> > `global_unitary_folding` is missing in the docstrings of `ZNE()`.
> 
> Thanks @AlejandroSopena for reviewing the code. I will amend according to your suggestions but will take some time as I am now away. Nonetheless I'd like to share some results using global unitary folding on IQM!
> 
> We used ZNE to extrapolate and estimate the ideal value of the approximation ratio, which is the cut_size/max_cut using QAOA for a fairly large graph of 10 nodes (hence 10 qubits). The QAOA only had one layer and the circuit was at a depth of 21. We used global unitary folding.
> 
> 3reg_graph_ZNE_approxratio
> 
> I was pretty impressed at the ZNE's use of Richardson extrapolation! It does seem like we cannot push the ZNE any further beyond maybe 6 or 7 global unitary foldings because the plateau is imminent, so you're right that we might need to use a different fit for the data if the circuit depth is too large.

These results are very good! This is a good example where global unitary folding works with Richardson extrapolation because the initial circuit is shallow, so there is still room to increase the noise. Even so, an exponential decay would also work with fewer noise levels.

AlejandroSopena (approved):
Thanks for the changes @mho291

scarrazza (approved):

PR #1328, opened 2024-05-19
Pylint 3.2

alecandido:
With Pylint 3.2 the [E0606](https://pylint.pycqa.org/en/latest/user_guide/messages/error/possibly-used-before-assignment.html) error is being raised in a few places, i.e.:

> Emitted when a local variable is accessed before its assignment took place in both branches of an if/else switch.

As it was found in https://github.com/qiboteam/qibo/pull/1282#discussion_r1601976588.
The current solution has been to block Pylint updates, but this is not great.

I tried to fix the error in a few places, by various mechanisms, but there is a common source for this: the `raise_error` function.

Unfortunately, Pylint is not being able to detect that the `raise_error` function is just raising an error, and so the respective branches in the various `if`s will never return.
I tried to tell it through the [`typing.NoReturn`](https://docs.python.org/3/library/typing.html#typing.NoReturn) hint (better called `Never`, but only from py3.11 on...), but it didn't work.

The simple option is to keep applying fixes like the one showcased here for the few missing cases, and just upgrade Pylint.
The other two, more drastic, options are:
- give up on `raise_error`, and just stick to the built-in `raise`
- consider using a different linter, like [`ruff`](https://github.com/astral-sh/ruff)

Both of them are feasible, but both of them would require non-negligible effort (the first is spread throughout the whole codebase, and dependent packages - though we are already considering removing in other places because of #1117 - and the second requires workflows upgrades, and possibly further lints to fix).

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -57,13 +57,13 @@ trio = ["trio (>=0.23)"]
 
 [[package]]
 name = "astroid"
-version = "3.1.0"
+version = "3.2.1"
 description = "An abstract syntax tree for Python with inference support."
 optional = false
 python-versions = ">=3.8.0"
 files = [
-    {file = "astroid-3.1.0-py3-none-any.whl", hash = "sha256:951798f922990137ac090c53af473db7ab4e70c770e6d7fae0cec59f74411819"},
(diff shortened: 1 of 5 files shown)


alecandido:
Not worth to keep this PR. In case, let's reopen from scratch.

PR #1336, opened 2024-05-29
random_clifford casting qubits to python int

scarrazza:
Related to the bug observed in https://github.com/qiboteam/qibo-client/issues/39. It looks like the qubit numbers are stored as numpy.int64 instead of python int, therefore serialization with json does not work from the qibo-client side. Could you please confirm this patch is acceptable?

Checklist:
- [ ] Tests are passing.


renatomello (approved):
I don't think this alters anything important in the function, but @BrunoLiegiBastonLiegi can correct me if I'm wrong

BrunoLiegiBastonLiegi (approved):

PR #1344, opened 2024-06-04
Speed up `circuit.unitary()` using sparse matrices

andrea-pasquale:
This PR proposes to speed up `circuit.unitary()` which returns the unitary corresponding to the circuit by performing the multiplication of all matrices gates using sparse matrices instead. I think that it makes sense given that most of times we are dealing with single and two qubit gates, which could be seen as sparse operators for large number of qubits.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/backends/numpy.py
+++ b/src/qibo/backends/numpy.py
@@ -2,6 +2,7 @@ import collections
 import math
 
 import numpy as np
+from scipy import sparse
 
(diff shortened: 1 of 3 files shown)


MatteoRobbiati (approved):
Thanks @andrea-pasquale. It worked for me with 10q.

PR #1346, opened 2024-06-05
Fix CMA callback

andrea-pasquale:
CMA callback fixed by calling it for each evaluation and not for each iteartion.
Should fix tests randomly failing, see #1344 for example.

Checklist:
- [x] Tests are passing.


MatteoRobbiati (approved):
Thanks a lot @andrea-pasquale.

PR #1349, opened 2024-06-06
Move some notebooks to `tutorials` repository

MatteoRobbiati:
I think the [tutorials](https://github.com/qiboteam/tutorials) repository could be a nice place where to delve deeper into details of topics (e.g. in the case of the well documented DBI module) in form of notebooks. 

I just opened a PR there: https://github.com/qiboteam/tutorials/pulls. 
If someone else would prefer to take care of this feel, free to do it. Otherwise we can proceed like I suggested.

I would suggest to leave here at least the intro notebook and the readme file for completeness.  

Checklist:
- [ ] Tests are passing.


marekgluza (commented):
@MatteoRobbiati I would leave it to @andrea-pasquale  to decide -  we were aiming to document the module using notebooks.

We could move them for now and then later pull them in but @andrea-pasquale moving them will not break the current docs?

andrea-pasquale (changes requested):
Thanks @MatteoRobbiati, also the documentation needs to be updated propertly by removing the links to those notebooks.

andrea-pasquale (approved):

marekgluza (approved):

PR #1361, opened 2024-06-13
Raise error when initial state shape is wrong

### LINKED ISSUE #646: Unpredictable behaviour when setting initial state for circuit.light_cone()
When I repeatedly execute the following code:

```
from qibo import gates
import numpy as np

n = 12

def lightconed_circuit(q):

  c = Circuit(n, density_matrix=True)
  
  dt = 0.1
  
  for i in range(0, n-1, 2):
    c.add(gates.RXX(i, (i+1)%n, theta=dt))
    c.add(gates.RYY(i, (i+1)%n, theta=dt))
    
  for i in range(1, n-1, 2):
    c.add(gates.RXX(i, (i+1)%n, theta=dt))
    c.add(gates.RYY(i, (i+1)%n, theta=dt))

  sc = c.light_cone(q)

  return sc


sc = lightconed_circuit(6)

print(sc[0].draw())
print()
# makes 4 qubit circuit
reduced_q = 4

initial_state = np.zeros(2**reduced_q)
initial_state[1] = 1
print(sc[0](initial_state).state())
```

The output seems to widely vary, from a sensible looking answer, to an array full of nan's or entries all ~e+250

1:
image

2:
image

3:
image

4:
image

Am I using this functionality incorrectly?

stavros11:
Implements the fix described in https://github.com/qiboteam/qibo/issues/646#issuecomment-2160875608 and closes #646.

Note that without this, the qibojit-numba backend will proceed with execution and return wrong results, sometimes with a non-normalized final state or segmentation fault.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/backends/numpy.py
+++ b/src/qibo/backends/numpy.py
@@ -400,6 +400,16 @@ class NumpyBackend(Backend):
                 return self.execute_circuit(initial_state + circuit, None, nshots)
         elif initial_state is not None:
             initial_state = self.cast(initial_state)
+            if circuit.density_matrix:
+                valid_shape = 2 * (2**circuit.nqubits,)
+            else:
+                valid_shape = (2**circuit.nqubits,)
+            if tuple(initial_state.shape) != valid_shape:
+                raise_error(
+                    ValueError,
+                    f"Given initial state has shape {initial_state.shape} instead of "
+                    f"the expected {valid_shape}.",
+                )
 
         if circuit.repeated_execution:
             if circuit.measurements or circuit.has_collapse:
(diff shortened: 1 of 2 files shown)


src/qibo/backends/numpy.py
```diff
@@ -400,6 +400,16 @@ def execute_circuit(self, circuit, initial_state=None, nshots=1000):
                 return self.execute_circuit(initial_state + circuit, None, nshots)
         elif initial_state is not None:
             initial_state = self.cast(initial_state)
+            if circuit.density_matrix:
+                valid_shape = 2 * (2**circuit.nqubits,)
+            else:
+                valid_shape = (2**circuit.nqubits,)
```
renatomello, line 406:
```suggestion
            valid_shape = 2 * (2**circuit.nqubits,) if circuit.density_matrix else (2**circuit.nqubits,)
```

renatomello (approved):

scarrazza (approved):
Thanks @stavros11.

PR #1365, opened 2024-06-18
Migrate ML backends to Qiboml

MatteoRobbiati:
Checklist:
- [ ] Tests are passing.


tests/conftest.py
```diff
@@ -27,7 +27,10 @@
 
 
 def get_backend(backend_name):
-    if "-" in backend_name:
+    if "qiboml" in backend_name:
+        _, name = backend_name.split("-")
+        platform = None
+    elif "-" in backend_name:
         name, platform = backend_name.split("-")
```
BrunoLiegiBastonLiegi, line 34:
```suggestion
    if "-" in backend_name:
        name, platform = backend_name.split("-")
        name = platform if "qiboml" in backend_name else name
```

BrunoLiegiBastonLiegi (approved):
Thanks, this is fine, my only concern is whether it is better to specialize the backend loading for `qiboml` as you are doing here, which allows not to brake all the previous code that was using `set_backend("tensorflow")`, but creates an asymmetry with the other backend providers, such as `qibojit`. Or, whether it is preferable to modify the `qiboml` metabackend to have a platform argument selecting the actual backend out of the ones `qiboml` provides, this would simplify this PR and maintain the symmetry with the other repos, but will involve setting the backends as `set_backend("qiboml", platform="tensorflow")`. What do you think @alecandido ?

tests/conftest.py
MatteoRobbiati: I agree with this doubt. 
My decision was in the end to keep the old interface, but it generates an inconsistency with Qibojit indeed. 
My point was that it would be more natural to refer to e.g. `tensorflow` as a backend, like it is `numpy`, rather than a platform. 
But I am completely open to different povs.

alecandido:
@BrunoLiegiBastonLiegi I'd just go for the uniform scheme, no need to make Qiboml different from the other providers.

We should keep a single instruction, just for `"tensorflow"`, that will check explicitly for it, and, in case, replace the backend with `"qiboml"`, and the platform with `"tensorflow"`.
This should be just a transition device, until we decide to break for good (thus, we'll open an issue about, and schedule for removal).

Moreover, I'd even drop all the `QIBO_NON_NATIVE_BACKENDS` specification, just rely on the package name (as we're already doing).
If the package is not available, we could catch the error and re-raise, hinting that is not installed, or misspelled (that could even be non-existing).
This will allow future backend providers to avoid at all the need of editing `qibo` source.

Issue #1354, opened 2024-06-11
Inconsistencies on gate definitions coming from  `Gate.controlled_by` and `Circuit.gates_of_type`


renatomello:
There are some inconsistencies coming from these two methods, and I'll list them below:

1. First, apparently controlled gates cannot receive more controls by using the `Gate.controlled_by` method, e.g.
```python
In [1]: from qibo import Circuit, gates, set_backend
   ...: set_backend("numpy")
   ...: 
   ...: gates.CRY(0, 1, 0.1).controlled_by(2)

[Qibo 0.2.8|INFO|2024-06-11 08:25:51]: Using numpy backend on /CPU:0
[Qibo 0.2.8|ERROR|2024-06-11 08:25:51]: Cannot use `controlled_by` method on gate <qibo.gates.gates.CRY object at 0x7cef2f413a60> because it is already controlled by (0,).
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In[1], line 4
      1 from qibo import Circuit, gates, set_backend
      2 set_backend("numpy")
----> 4 gates.CRY(0, 1, 0.1).controlled_by(2)

File ~/.local/lib/python3.10/site-packages/qibo/gates/abstract.py:311, in Gate.check_controls.<locals>.wrapper(self, *args)
    309 def wrapper(self, *args):
    310     if self.control_qubits:
--> 311         raise_error(
    312             RuntimeError,
    313             "Cannot use `controlled_by` method "
    314             + f"on gate {self} because it is already "
    315             + f"controlled by {self.control_qubits}.",
    316         )
    317     return func(self, *args)

File ~/.local/lib/python3.10/site-packages/qibo/config.py:46, in raise_error(exception, message)
     39 """Raise exception with logging error.
     40 
     41 Args:
     42     exception (Exception): python exception.
     43     message (str): the error message.
     44 """
     45 log.error(message)
---> 46 raise exception(message)

RuntimeError: Cannot use `controlled_by` method on gate <qibo.gates.gates.CRY object at 0x7cef2f413a60> because it is already controlled by (0,).
```
Meanwhile, it is possible to create a $2$-controlled $RY$ gate using 
```python
In [3]: gates.RY(1, 0.1).controlled_by(0, 2)
Out[3]: <qibo.gates.gates.RY at 0x7cef2f2f8cd0>
``` 
[...]

PR #1367, opened 2024-06-20
Fix `Gate.matrix` and `Circuit.unitary` in the presence of `Gate.controlled_by` method

### LINKED ISSUE #1279: `Circuit.unitary` breaks in the presente of two-qubit gate initialized by `Gate.controlled_by`
Code to reproduce bug:

```python

circuit = Circuit(2)
circuit.add(gates.H(0).controlled_by(1))

circuit.unitary()
```

renatomello:
This closes #1279 and #1354

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/backends/numpy.py
+++ b/src/qibo/backends/numpy.py
@@ -110,25 +110,12 @@ class NumpyBackend(Backend):
         _matrix = getattr(self.matrices, name)
         if callable(_matrix):
             _matrix = _matrix(2 ** len(gate.target_qubits))
-
-        num_controls = len(gate.control_qubits)
-        if num_controls > 0:
-            _matrix = block_diag(
-                np.eye(2 ** len(gate.qubits) - len(_matrix)), self.to_numpy(_matrix)
-            )
-
         return self.cast(_matrix, dtype=_matrix.dtype)
 
     def matrix_parametrized(self, gate):
         """Convert a parametrized gate to its matrix representation in the computational basis."""
         name = gate.__class__.__name__
         _matrix = getattr(self.matrices, name)(*gate.parameters)
-
-        num_controls = len(gate.control_qubits)
-        if num_controls > 0:
-            _matrix = block_diag(
-                np.eye(2 ** len(gate.qubits) - len(_matrix)), self.to_numpy(_matrix)
-            )
         return self.cast(_matrix, dtype=_matrix.dtype)
 
     def matrix_fused(self, fgate):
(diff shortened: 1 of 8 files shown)


src/qibo/gates/abstract.py
```diff
@@ -322,12 +319,24 @@ def wrapper(self, *args):
     def controlled_by(self, *qubits: int) -> "Gate":
         """Controls the gate on (arbitrarily many) qubits.
 
+        To see how this method affects the underlying matrix representation of a gate,
+        please see the documentation of :meth:`qibo.gates.Gate.matrix`.
+
+        .. note::
+            Some gate classes default to another gate class depending on the number of controls
+            present. For instance, an :math:`1`-controlled :class:`qibo.gates.X` gate
+            will default to a :class:`qibo.gates.CNOT` gate, while a :math:`1`-controlled
```
stavros11, line 328:
```suggestion
            will default to a :class:`qibo.gates.CNOT` gate, while a :math:`2`-controlled
```
?

src/qibo/backends/numpy.py
```diff
@@ -127,6 +127,13 @@ def matrix_fused(self, fgate):
             # small tensor calculations
             # explicit to_numpy see https://github.com/qiboteam/qibo/issues/928
             gmatrix = self.to_numpy(gate.matrix(self))
+            # add controls if controls were instantiated using
+            # the ``Gate.controlled_by`` method
+            num_controls = len(gate.control_qubits)
+            if num_controls > 0:
+                gmatrix = block_diag(
+                    np.eye(2 ** len(gate.qubits) - len(gmatrix)), gmatrix
+                )
```
stavros11, line 136: I haven't tested, but are you sure that this is correct in all cases? My main concern is that it is not using the `control_qubits` indices in any way other then length. Shouldn't for example `gmatrix` be different for 
```py
Unitary(matrix, 0).controlled_by(1, 2, 3)
```
compared to 
```py
Unitary(matrix, 2).controlled_by(0, 1, 3)
```
In the second case you would have to put identities on both sides.

tests/test_gates_gates.py
```diff
@@ -1574,23 +1562,6 @@ def test_controlled_unitary(backend):
...
-    gate = gates.Unitary(matrix, 1).controlled_by(0)
-
-    target_state = apply_gates(backend, [gate], nqubits, initial_state)
-
-    u = backend.control_matrix(gate)
-    u = backend.cast(u, dtype=u.dtype)
-
-    final_state = np.dot(u, initial_state)
-
-    backend.assert_allclose(final_state, target_state)
```
stavros11, line 1591: Instead of completely dropping this test, I would suggest to port it to the new interface you propose above for calculating the matrix of the controlled gate using `circuit.unitary()`. You could also do this for various examples, such as a `Unitary` gate and the example you provided in the docstring above:
```py
nqubits = 5
circuit = Circuit(nqubits)
circuit.add(gates.SWAP(3, 4).controlled_by(0, 1, 2))
print(circuit.unitary()) # check if this is the matrix you expect
```
Switching the order of controls and targets will also reveal if the issue in my previous comment is true or not.

src/qibo/backends/numpy.py
renatomello: I have tested this locally before, and it works fine because the only thing the new lines is doing is to get the matrix into the right shape for a gate with controls and target qubits. Then the inclusion of identities and assigment of correct order of qubits is done the same way as before.

tests/test_gates_gates.py
stavros11: Indeed, I tried a few examples and it seems to work as expected, thanks. It would still be good to have some test of the `circuit.unitary()` when the circuit contains a non-trivial `.controlled_by` gate, since we advertise that feature in the docs.

stavros11 (approved):

scarrazza (approved):
Thanks @renatomello.

Issue #1318, opened 2024-05-01
Unroller duplicates gates when measuring in specific basis


andrea-pasquale:
When applying a measurement there is the possibility to specify in which basis we want to perform the measurement using the [basis](https://github.com/qiboteam/qibo/blob/24acee037138d995b4ea5acfb56d9f32831415ae/src/qibo/models/circuit.py#L660-L662) argument.

However, when such circuit is transpiled the "rotation gate" appears twice and only once it is correctly transpiled.
Here is an example:

```py
from qibo import Circuit, gates
from qibo.transpiler.unroller import NativeGates, Unroller

unroller = Unroller(native_gates=NativeGates.default())

nqubits = 2
circuit = Circuit(nqubits)
circuit.add(gates.M(i, basis=gates.X) for i in range(nqubits))

print(circuit.draw())
# which gives
# q0: ─H─M─
# q1: ─H─M─
transpiled_circuit = unroller(circuit)
print(transpiled_circuit.draw())
# which gives
# q0: ─Z─GPI2─H─M─
# q1: ─Z─GPI2─H─M─
```
In the transpiled circuit `H` appears twice, once it is compiled into natives as `Z` and `GPI2` while the second times appears as `H` which will trigger an error since it is not a native gate.
@Edoardo-Pedicillo I believe that this is not related to PRs that you recently open since I am using `master`.

Edoardo-Pedicillo:
> @Edoardo-Pedicillo I believe that this is not related to PRs that you recently open since I am using `master`.

I confirm it is not related

Issue #1363, opened 2024-06-17
Transpiler moving measurements on wrong qubit


BrunoLiegiBastonLiegi:
I am testing `GST` #1106 with the transpiler and I see that sometimes the measurements get messed up.
For instance, given this input circuit:
```
q0: ─I─H─M─
```
and trying to transpile it with the following pipeline:
``` python
    # define transpiler
    connectivity = nx.Graph()
    # star connectivity
    connectivity.add_edges_from([(0, 2), (1, 2), (2, 3), (2, 4)])
    transpiler = Passes(
        connectivity=connectivity,
        passes=[
            Preprocessing(connectivity),
            Random(connectivity),
            Sabre(connectivity),
            Unroller(NativeGates.default()),
        ],
        int_qubit_names=True,
    )
```
results in this transpiled circuit:
```
q0: ──────────
q2: ─M────────
q1: ─I─Z─GPI2─
q3: ──────────
q4: ──────────
```

PR #1106, opened 2023-11-24
Gate set tomography

mho291:
One qubit calibration (empty circuit tomography)
Two qubit calibration (empty circuit tomography)
One qubit operator(s) gate set tomography
Two qubits operator(s) gate set tomography

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [x] Tests are passing.
- [x] Coverage does not decrease.
- [x] Documentation is updated.
- [ ] Testing with QiboLab backend.

PR #1369, opened 2024-06-21
Fix transpiler bugs about moved measurements and duplicated gates

Simone-Bordoni:
Solve bugs in #1363 and #1318 

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/transpiler/placer.py
+++ b/src/qibo/transpiler/placer.py
@@ -360,7 +360,7 @@ class Random(Placer):
                 final_cost = cost
 
         final_layout = dict(zip(dict_keys, list(final_mapping.values())))
-        return {key: value for key, value in sorted(final_layout.items())}
+        return dict(sorted(final_layout.items()))
 
     def _cost(self, graph: nx.Graph, gates_qubits_pairs: list):
         """
--- a/src/qibo/transpiler/placer.py
+++ b/src/qibo/transpiler/placer.py
@@ -46,11 +46,20 @@ def assert_mapping_consistency(layout: dict, connectivity: nx.Graph = None):
             qubits of the original connectivity graph. Defaults to ``None``.
     """
     values = sorted(layout.values())
+    physical_qubits = list(layout.keys())
     if connectivity is not None:
-        ref_keys = ["q" + str(i) for i in list(connectivity.nodes)]
+        if isinstance(physical_qubits[0], str):
+            ref_keys = ["q" + str(i) for i in list(connectivity.nodes)]
+        else:
+            ref_keys = list(connectivity.nodes)
     else:
-        ref_keys = ["q" + str(i) for i in range(len(values))]
-    if list(layout.keys()) != ref_keys:
+        if isinstance(physical_qubits[0], str):
+            ref_keys = ["q" + str(i) for i in range(len(values))]
+        else:
+            ref_keys = list(range(len(values)))
+    print(ref_keys)
+    print(list(layout.keys()))
+    if physical_qubits != ref_keys:
         raise_error(
             PlacementError,
             "Some physical qubits in the layout may be missing or duplicated.",
--- a/src/qibo/transpiler/placer.py
+++ b/src/qibo/transpiler/placer.py
@@ -334,13 +334,15 @@ class Random(Placer):
             graph = nx.relabel_nodes(self.connectivity, mapping)
             cost = self._cost(graph, gates_qubits_pairs)
             if cost == 0:
-                return dict(zip(dict_keys, list(mapping.values())))
+                final_layout = dict(zip(dict_keys, list(mapping.values())))
+                return {key: value for key, value in sorted(final_layout.items())}
             if cost < final_cost:
                 final_graph = graph
                 final_mapping = mapping
                 final_cost = cost
 
-        return dict(zip(dict_keys, list(final_mapping.values())))
+        final_layout = dict(zip(dict_keys, list(final_mapping.values())))
+        return {key: value for key, value in sorted(final_layout.items())}
 
     def _cost(self, graph: nx.Graph, gates_qubits_pairs: list):
         """
--- a/src/qibo/transpiler/placer.py
+++ b/src/qibo/transpiler/placer.py
@@ -47,7 +47,7 @@ def assert_mapping_consistency(layout: dict, connectivity: nx.Graph = None):
             qubits of the original connectivity graph. Defaults to ``None``.
     """
     values = sorted(layout.values())
-    physical_qubits = list(layout.keys())
+    physical_qubits = list(layout)
     nodes = (
         list(range(len(values))) if connectivity is None else list(connectivity.nodes)
     )
@@ -208,9 +208,9 @@ class Custom(Placer):
             qubits of the original connectivity graph. Defaults to ``None``.
     """
 
-    def __init__(self, map: Union[list, dict], connectivity: nx.Graph = None):
+    def __init__(self, initial_map: Union[list, dict], connectivity: nx.Graph = None):
         self.connectivity = connectivity
-        self.map = map
+        self.initial_map = initial_map
 
     def __call__(self, circuit=None):
         """Return the custom placement if it can be applied to the given circuit (if given).
@@ -221,24 +221,30 @@ class Custom(Placer):
         Returns:
             (dict): physical to logical qubit mapping.
         """
-        if isinstance(self.map, dict):
+        if isinstance(self.initial_map, dict):
             pass
-        elif isinstance(self.map, list):
+        elif isinstance(self.initial_map, list):
             if self.connectivity is not None:
-                self.map = dict(
-                    zip(["q" + str(i) for i in self.connectivity.nodes()], self.map)
+                self.initial_map = dict(
+                    zip(
+                        ["q" + str(i) for i in self.connectivity.nodes()],
+                        self.initial_map,
+                    )
                 )
             else:
-                self.map = dict(
-                    zip(["q" + str(i) for i in range(len(self.map))], self.map)
+                self.initial_map = dict(
+                    zip(
+                        ["q" + str(i) for i in range(len(self.initial_map))],
+                        self.initial_map,
+                    )
                 )
         else:
             raise_error(TypeError, "Use dict or list to define mapping.")
         if circuit is not None:
-            assert_placement(circuit, self.map, connectivity=self.connectivity)
+            assert_placement(circuit, self.initial_map, connectivity=self.connectivity)
         else:
-            assert_mapping_consistency(self.map, connectivity=self.connectivity)
-        return self.map
+            assert_mapping_consistency(self.initial_map, connectivity=self.connectivity)
+        return self.initial_map
 
 
 class Subgraph(Placer):
--- a/src/qibo/transpiler/placer.py
+++ b/src/qibo/transpiler/placer.py
@@ -57,8 +57,6 @@ def assert_mapping_consistency(layout: dict, connectivity: nx.Graph = None):
             ref_keys = ["q" + str(i) for i in range(len(values))]
         else:
             ref_keys = list(range(len(values)))
-    print(ref_keys)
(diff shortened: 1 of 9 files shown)


Simone-Bordoni:
@BrunoLiegiBastonLiegi I have solved the bug reported in #1363. I will try to rapidly fix the Issue in #1318, in the meantime you can start the review of this part (I was not reardering the initial qubit map and this created some issues with the random placer). I have also added the seed for the random placer

renatomello:
@Simone-Bordoni Can the PR name be more descriptive?

src/qibo/transpiler/router.py
```diff
@@ -176,10 +176,13 @@ def __init__(
...
-        self._graph_qubits_names = [int(key[1:]) for key in initial_layout.keys()]
-        self._circuit_logical = list(range(len(initial_layout)))
-        self._physical_logical = list(initial_layout.values())
+        # Order the initial layout based on the hardware qubit name (this may create problems for a custom layout)
+        self.initial_layout = {
+            key: value for key, value in sorted(initial_layout.items())
+        }
+        self._graph_qubits_names = [int(key[1:]) for key in self.initial_layout.keys()]
+        self._circuit_logical = list(range(len(self.initial_layout)))
+        self._physical_logical = list(self.initial_layout.values())
```
BrunoLiegiBastonLiegi, line 185: Do you mean that this is going to fail when the qubits names are not just `q0, q1, ...`? In that case I would at least raise a warning when `int_qubit_names == False` https://github.com/qiboteam/qibo/blob/c68e58bc2c0ba87a286f35544f97d69efab76d40/src/qibo/transpiler/pipeline.py#L193

Ideally, I think it would be better to internally always only support integer qubit identifiers, and, in case, to convert them to the user specified ones before returning them to any user facing layer.

tests/test_transpiler_pipeline.py
```diff
@@ -283,3 +283,25 @@ def test_custom_passes_wrong_pass():
...
+        ],
+        int_qubit_names=True,
+    )
+    circuit = Circuit(1)
+    circuit.add(gates.I(0))
+    circuit.add(gates.H(0))
+    circuit.add(gates.M(0))
+    transpiled_circuit, _ = transpiler(circuit)
+    assert transpiled_circuit.queue[2].qubits == (1,)
+    assert transpiled_circuit.queue[3].qubits == (1,)
```
BrunoLiegiBastonLiegi, line 307: It would be better to have a more meaningful name for this test, and maybe as a further assertion, test that by executing the initial circuit and the transpiled one the results are compatible, for example by comparing the frequencies or the probabilities (by tracing out the unused qubits of the transpiled circuit).

tests/test_transpiler_unroller.py
```diff
@@ -111,3 +111,11 @@ def test_unroller(natives_1q, natives_2q):
...
         native_gates=natives_1q | natives_2q | NativeGates.RZ | NativeGates.Z,
     )
+
+
+def test_measurements_non_comp_basis():
+    unroller = Unroller(native_gates=NativeGates.default())
+    circuit = Circuit(1)
+    circuit.add(gates.M(0, basis=gates.X))
+    transpiled_circuit = unroller(circuit)
+    assert isinstance(transpiled_circuit.queue[2], gates.M)
```
BrunoLiegiBastonLiegi, line 121: maybe test also that the basis of the measurement is still `X`

src/qibo/transpiler/router.py
Simone-Bordoni: My comment on the code was not correct, it should say "ordering is necessary to avoid problems with custom layouts".
Now all the placers are creating an initial map using hardware qubit names "q0", "q1", "q2"... so there is no possibility to use an initial map with integer qubit names unless you define one by yourself without using a placer

tests/test_transpiler_unroller.py
Simone-Bordoni: After transpiling the measurement gate should be in the compuational basis. It is a task of the unroller to translate the gates used to rotate the state before the measurement. I added a check and explicited it with a comment.

src/qibo/transpiler/placer.py
```diff
@@ -46,11 +46,18 @@ def assert_mapping_consistency(layout: dict, connectivity: nx.Graph = None):
...
+        else:
+            ref_keys = list(connectivity.nodes)
     else:
-        ref_keys = ["q" + str(i) for i in range(len(values))]
-    if list(layout.keys()) != ref_keys:
+        if isinstance(physical_qubits[0], str):
+            ref_keys = ["q" + str(i) for i in range(len(values))]
+        else:
+            ref_keys = list(range(len(values)))
+    if physical_qubits != ref_keys:
```
BrunoLiegiBastonLiegi, line 60: I think you can avoid repeating this twice by doing something like this:
``` python
nodes = list(range(len(values))) if connectivity is None else list(connectivity.nodes)
if isinstance(physical_qubits[0], str):
    ref_keys = ["q" + str(i) for i in nodes]
else:
    ref_keys = nodes
```

BrunoLiegiBastonLiegi (approved):
Thanks @Simone-Bordoni  I tested this with #1106 and it seems to be working

src/qibo/transpiler/placer.py
```diff
@@ -306,9 +309,10 @@ class Random(Placer):
             Defaults to :math:`100`.
     """
 
-    def __init__(self, connectivity, samples: int = 100):
+    def __init__(self, connectivity, samples: int = 100, seed=42):
```
renatomello, line 312: `samples` is missing from the docstring

src/qibo/transpiler/placer.py
```diff
@@ -319,6 +323,7 @@ def __call__(self, circuit):
         Returns:
             (dict): physical-to-logical qubit mapping.
         """
+        random.seed(self.seed)
```
renatomello, line 326: Instead of using the `random` package for this, I'd recommend allowing `numpy.random.Generator`s. You can take a look on how its done in other modules here in `qibo` (e.g. `quantum_info.random_ensembles`).

src/qibo/transpiler/placer.py
```diff
@@ -332,13 +337,15 @@ def __call__(self, circuit):
...
+                final_layout = dict(zip(dict_keys, list(mapping.values())))
+                return {key: value for key, value in sorted(final_layout.items())}
             if cost < final_cost:
                 final_graph = graph
                 final_mapping = mapping
                 final_cost = cost
 
-        return dict(zip(dict_keys, list(final_mapping.values())))
+        final_layout = dict(zip(dict_keys, list(final_mapping.values())))
+        return {key: value for key, value in sorted(final_layout.items())}
```
renatomello, line 352: You could do this:

```suggestion
        return dict(sorted(final_layout.items()))
```
However, why are we sorting a dictionary? It's not supposed to require being a sorted data structure.

src/qibo/transpiler/placer.py
```diff
@@ -332,13 +337,15 @@ def __call__(self, circuit):
             graph = nx.relabel_nodes(self.connectivity, mapping)
             cost = self._cost(graph, gates_qubits_pairs)
             if cost == 0:
-                return dict(zip(dict_keys, list(mapping.values())))
+                final_layout = dict(zip(dict_keys, list(mapping.values())))
+                return {key: value for key, value in sorted(final_layout.items())}
```
renatomello, line 344:
```suggestion
                return dict(sorted(final_layout.items()))
```

Same comment here.

src/qibo/transpiler/router.py
```diff
@@ -176,10 +176,13 @@ def __init__(
...
         else:
             self.circuit_blocks = CircuitBlocks(circuit, index_names=True)
-        self.initial_layout = initial_layout
-        self._graph_qubits_names = [int(key[1:]) for key in initial_layout.keys()]
-        self._circuit_logical = list(range(len(initial_layout)))
-        self._physical_logical = list(initial_layout.values())
+        # Order the initial layout based on the hardware qubit names to avoid problems in custom layouts
+        self.initial_layout = {
+            key: value for key, value in sorted(initial_layout.items())
+        }
```
renatomello, line 182: Same thing here.

```suggestion
        self.initial_layout = dict(sorted(initial_layout.items()))
```

src/qibo/transpiler/unroller.py
```diff
@@ -142,7 +142,11 @@ def translate_gate(
     Returns:
         (list): List of native gates that decompose the input gate.
     """
-    if isinstance(gate, (gates.M, gates.I, gates.Align)):
+    if isinstance(gate, (gates.I, gates.Align)):
+        return gate
+    elif isinstance(gate, gates.M):
```
renatomello, line 147:
```suggestion
    if isinstance(gate, (gates.I, gates.Align)):
        return gate
    
    if isinstance(gate, gates.M):
```

src/qibo/transpiler/unroller.py
```diff
@@ -142,7 +142,11 @@ def translate_gate(
...
+        gate.basis_gates = len(gate.basis_gates) * [gates.Z]
+        gate.basis = []
         return gate
     elif len(gate.qubits) == 1:
         return _translate_single_qubit_gates(gate, native_gates)
```
renatomello, line 152:
```suggestion
        gate.basis = []
        return gate
    
    if len(gate.qubits) == 1:
        return _translate_single_qubit_gates(gate, native_gates)
```

renatomello (commented):
A couple of other things: 

1. the function `translate_gate` in `unroller.py` receives the variable `gate` as an input, but redefines what `gate` is in a loop.
2. the `Custom.__call__` method in `placer.py` receives the variable `map` as an input, but `map` is a `python` built-in function. This may lead to issues, so this variable should be renamed.

src/qibo/transpiler/placer.py
```diff
@@ -46,11 +47,14 @@ def assert_mapping_consistency(layout: dict, connectivity: nx.Graph = None):
             qubits of the original connectivity graph. Defaults to ``None``.
     """
     values = sorted(layout.values())
-    if connectivity is not None:
-        ref_keys = ["q" + str(i) for i in list(connectivity.nodes)]
-    else:
-        ref_keys = ["q" + str(i) for i in range(len(values))]
-    if list(layout.keys()) != ref_keys:
+    physical_qubits = list(layout.keys())
```
andrea-pasquale, line 50:
```suggestion
    physical_qubits = list(layout)
```

andrea-pasquale (commented):
Thanks @Simone-Bordoni.
I can confirm that this PR fixes #1318.
While testing the `state_tomography` in Qibocal, I noticed that even without this PR is now failing with this error:
```sh
File "/home/andreapasquale/qibocal/src/qibocal/protocols/state_tomography.py", line 114, in _acquisition
    _, results = execute_transpiled_circuit(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/andreapasquale/qibocal/src/qibocal/auto/transpile.py", line 60, in execute_transpiled_circuit
    transpiled_circ, _ = transpiler(new_circuit)
                         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/andreapasquale/qibo/src/qibo/transpiler/pipeline.py", line 265, in __call__
    return circuit, final_layout
                    ^^^^^^^^^^^^
UnboundLocalError: cannot access local variable 'final_layout' where it is not associated with a value
```
I believe that the error was introduced by https://github.com/qiboteam/qibo/commit/8d29397fdc241bc34c082f643d53869ba0055406. It should be related to the fact that we are returning `final_layout` even if it is defined only in a `if` statement. Probably in qibocal we fall in one of the cases where `final_layout` is not defined.
I believe that this was related to the changes that @Edoardo-Pedicillo introduced in https://github.com/qiboteam/qibo/pull/1301.
Given that the error by itself is not related to this PR feel free to address the problem here or in a separate PR and I will open an issue pointing to this comment.

andrea-pasquale (approved):

PR #1370, opened 2024-06-21
Matplotlib circuit drawer

sergiomtzlosa:
A new class named MPLDrawer to draw the circuit using matplotlib, this is a code integration and enhancement from https://github.com/rpmuller/PlotQCircuit.

The circuit now has a function called draw_mpl() to invoke matplotlib on drawing.

Checklist:
- [ ] Tests are passing.


renatomello:
I think this is a great idea, but `models` is not the best submodule to insert this functionality.

sergiomtzlosa:
> I think this is a great idea, but `models` is not the best submodule to insert this functionality.

Thanks a lot for your feedback @renatomello I will move the `MPLDrawer` to a new single module called `imaging` as I think this will grow up in the future...

renatomello:
> > I think this is a great idea, but `models` is not the best submodule to insert this functionality.
> 
> Thanks a lot for your feedback @renatomello I will move the `MPLDrawer` to a new single module called `imaging` as I think this will grow up in the future...

Maybe we should hear some other opinions first, e.g. @scarrazza @MatteoRobbiati @alecandido ?

src/qibo/imaging/mpldrawer.py
```diff
@@ -0,0 +1,481 @@
...
+# Simplified Plotting Routines for Quantum Circuits
+# https://github.com/rpmuller/PlotQCircuit
+#
+import matplotlib
+import numpy as np
+
+
+class MPLDrawer:
+    def __init__(self):
+        pass
```
alecandido, line 11: This class has no explicit attributes, and they are not even set later on (which is a good thing).

Just remove these three lines, deindent everything else, and remove `self` from functions parameters and methods calls.
It will work in the exact same way, saving a bunch of overhead.

src/qibo/imaging/mpldrawer.py
alecandido, line 1: *Imaging* has usually a different meaning: https://en.wikipedia.org/wiki/Imaging

I would suggest calling the subpackage `graphics` or `ui`.

alecandido, line 1: Every subpackage requires an `__init__.py` file, please provide it.

src/qibo/__init__.py
```diff
@@ -32,4 +32,5 @@
     set_batch_size,
     set_metropolis_threshold,
 )
+from qibo.imaging.mpldrawer import MPLDrawer
```
alecandido, line 35: I would avoid the explicit re-export.

src/qibo/imaging/mpldrawer.py
```diff
@@ -0,0 +1,481 @@
...
+            return self._plot_quantum_circuit(gates_plot, inits, labels, scale=scale)
+        else:
+            return self._plot_lines_circuit(inits, labels, scale=scale)
+
+    @staticmethod
+    def save_fig(fig, path_file):
+        """Save matpltlib image to file
+        path_file   Path to file with filename and extension
+        """
+        return fig.savefig(path_file, bbox_inches="tight")
```
alecandido, line 481: This function is just calling a function, and it is invoked only once.

Just drop it and replace its invocation with a direct call to `fig.savefig()`.

src/qibo/models/circuit.py
```diff
@@ -1269,6 +1270,15 @@ def _update_draw_matrix(self, matrix, idx, gate, gate_symbol=None):
...
         return matrix, idx
 
+    def draw_mpl(self, scale=0.6, cluster_gates=True, save_file=None):
+
+        ax = MPLDrawer().plot_qibo_circuit(self, scale, cluster_gates)
+
+        if save_file:
+            MPLDrawer.save_fig(ax.figure, save_file)
+
+        return ax
```
alecandido, line 1280: Do not make this function a method of `Circuit`, to avoid making `Circuit` depending on the `mpldrawer` module.

Moreover, this function is just calling another function, and optionally (with a flag) a second one.
This is definitely not modular, and it leads to concatenate the arguments of the two functions.

Instead, just encourage direct usage of the functions above. Even by choosing simpler names.
E.g.:
```py
fig, ax = qibo.ui.plot(circuit, scale=0.6)
fig.savefig("mycircuit.png")
```
is perfectly acceptable as user's code.

src/qibo/models/circuit.py
```diff
@@ -8,6 +8,7 @@
 from qibo import gates
 from qibo.config import raise_error
 from qibo.gates.abstract import Gate
+from qibo.imaging.mpldrawer import MPLDrawer
```
alecandido, line 11: As said before, avoid importing the drawing module in `circuit.py`.

The motivation is that `mpldrawer` is a UI module, more or less an extension. While `circuit` is one of the core elements.
So, it should definitely go the other way round. And importing without a similar criterion largely increases the chances of circular imports.

alecandido (commented):
Thanks for the contribution!

I didn't enter the details of the plotting functions, but the results look pretty nice.
I'm not completely sure why you're drawing a double line after the measurement gates (we also have non-collapsing measurements, and the collapsing ones could be intermediate, so you may control a gate with a measured qubit - not sure how much this double line is appropriate...), but the rest looks great :)

Instead, I provided some suggestions about code simplifications and improvements.

alecandido:
@sergiomtzlosa you have about 30 commits with the same message (more exactly 28, of which 26 in two consecutive days).

I'm not against small commits at all, but:
- try to make commits that corresponds more or less to meaningful units
- such that you can give them a descriptive commit message to each of them

If you can not describe the change, try to group your commits differently, e.g. completing a *unit* before saving (where a "unit" could be even something temporary, or a step in a bug hunt, not necessarily code ready for production), or reconsider if that code is needed at all (if you don't know how to describe its role, there are chances that it's not that useful after all).

sergiomtzlosa:
Thanks a lot for this valuable feedback. I will have a look at them...

src/qibo/ui/mpldrawer.py
```diff
@@ -0,0 +1,537 @@
...
+    else:
+        return _plot_lines_circuit(inits, labels, scale=scale)
+
+
+def plot(circuit, scale=0.6, cluster_gates=True, style=None, save_file=None):
+
+    ax = _plot_qibo_circuit(circuit, scale, cluster_gates, style)
+
+    if save_file:
+        ax.figure.savefig(save_file, bbox_inches="tight")
```
alecandido, line 535: As in the previous comment: this part could be completely omitted, and left up to the user (you're trading one function call for one parameter).

And without this, the `plot` function would just be the function `_plot_qibo_circuit`. So just rename that function :)

src/qibo/imaging/mpldrawer.py
sergiomtzlosa: Thanks! I have removed the class and now there is a `ui` submodule with cleaner code

src/qibo/__init__.py
sergiomtzlosa: done!

src/qibo/models/circuit.py
sergiomtzlosa: Thanks @alecandido !!! this is done!

src/qibo/imaging/mpldrawer.py
sergiomtzlosa: I removed this function by a direct call to `savefig`

src/qibo/imaging/mpldrawer.py
alecandido: Done in ffdf9b0b48439d174623131be8aaf5d3392ac72f

alecandido: Whenever you complete something, feel free to add a comment like the former, and resolve the conversation.

src/qibo/ui/mpldrawer.py
sergiomtzlosa: thanks! I removed the function and leave the option to the user

MatteoRobbiati (commented):
Thanks a lot for this work @sergiomtzlosa. It is for sure something which can improve the usability of Qibo in some contexts. 
Before doing a proper review to this, I just want to open a small discussion the best way to set up this in Qibo. Some time ago I had a discussion with @alecandido where we explored the possibility to add some representation of the circuit through HTML or Matplotlib indeed. At that time we discussed it as a method of the circuit class. 

It is also true we typically avoid to add plotting functions in the already existing Qibo files, thus the approach you propose could make sense.  

Anyway, thanks a lot for this work :)

src/qibo/ui/plot_styles.py
alecandido, line 1: You could even convert this whole file as a single dictionary constant:
```py
STYLE = {
  "garnacha": {
    "facecolor": ...,
  },
  "fardelejo": {
    "facecolor": ...,
  },
  ...: ...
}
```

sergiomtzlosa: thanks! it is now converted to constant

sergiomtzlosa:
> Thanks a lot for this work @sergiomtzlosa. It is for sure something which can improve the usability of Qibo in some contexts. Before doing a proper review to this, I just want to open a small discussion the best way to set up this in Qibo. Some time ago I had a discussion with @alecandido where we explored the possibility to add some representation of the circuit through HTML or Matplotlib indeed. At that time we discussed it as a method of the circuit class.
> 
> It is also true we typically avoid to add plotting functions in the already existing Qibo files, thus the approach you propose could make sense.
> 
> Anyway, thanks a lot for this work :)

Thanks everybody for all comments! Regarding your question, IMHO, to draw the circuit using a function from inside the circuit is a fast way to do it as it is something you would like to see on the screen, it is very convenient but I understand that you may want to detach the UI functions from the qibo core. I do not have a last opinion on this because both approaches are right.

sergiomtzlosa:
sorry for these last-minute commits, now I will not push any further commits, you can review the code.

src/qibo/ui/symbols.py
alecandido, line 1: Since there is no logic in here, this could even by a JSON file (that we have to remember to include the package)

src/qibo/ui/plot_styles.py
alecandido, line 1: A more idiomatic way of defining styles would be to use the matplotlib style sheets:
https://matplotlib.org/stable/users/explain/customizing.html#using-style-sheets

This will also make all graphics produced in the session consistent with the chosen style (it could be easily reset with `plt.style.use('default')`).

This file could be converted as well to a JSON file, and loaded by the `mpldrawer` to set the style in a single shot (with `plt.style.use`).
If you don't want to affect all Matplotlib graphics (which could not be bad, but not universally appreciated), you can make a wrapper function to save the current style, set the chosen style sheet, plot, and restore the previous one.

sergiomtzlosa: Thanks @alecandido, I will have a look at this now

src/qibo/ui/mpldrawer.py
```diff
@@ -0,0 +1,721 @@
...
+    # for i in measured:
+    #    j = measured[i]
+    #    _line(
+    #        ax,
+    #        gate_grid[j],
+    #        gate_grid[-1] + scale,
+    #        wire_grid[i] + dy,
+    #        wire_grid[i] + dy,
+    #        plot_params,
+    #    )
```
alecandido, line 429: Drop dead code before merging

src/qibo/ui/mpldrawer.py
```diff
@@ -0,0 +1,721 @@
...
+
+    if len(temp_gates) > 0:
+        cluster_gates.append(temp_gates)
+
+    if len(temp_mgates) > 0:
+        cluster_gates.append(temp_mgates)
+
+    temp_gates = []
+    temp_mgates = []
+
```
alecandido, line 562:
```suggestion
```

Unused variable definitions (this is not changing what you appended, but shadowing the previous definitions - and then they are never used).

src/qibo/ui/mpldrawer.py
```diff
@@ -0,0 +1,721 @@
...
+    dir_path = path.dirname(file_path)  # full path of the directory from file
+    final_path = path.join(dir_path, filename)  # absolute file path of given file
+    return final_path
+
+
+def _check_list_str(list, str):
+    for item in list:
+        if item in str:
+            return True
+    return False
```
alecandido, line 577:
```suggestion
    return any(item in str for item in list)
```
Doing the exact same, and more readable (it reads more like English than like an algorithm).

src/qibo/ui/mpldrawer.py
```diff
@@ -0,0 +1,721 @@
...
+
+
+def _build_path(filename):
+    file_path = path.abspath(__file__)  # full path of current file
+    dir_path = path.dirname(file_path)  # full path of the directory from file
+    final_path = path.join(dir_path, filename)  # absolute file path of given file
+    return final_path
+
+
+def _check_list_str(list, str):
```
alecandido, line 573: Do not use `list` and `str` as names, they are shadowing builtins (and pretty vague, that is the reason why they are used as names for builtins).

You could use something like:
```suggestion
def _check_list_str(substrings, string):
```
or whatever you prefer.

src/qibo/ui/mpldrawer.py
```diff
@@ -0,0 +1,721 @@
...
+    temp_mgates = []
+
+    return cluster_gates
+
+
+def _build_path(filename):
+    file_path = path.abspath(__file__)  # full path of current file
+    dir_path = path.dirname(file_path)  # full path of the directory from file
+    final_path = path.join(dir_path, filename)  # absolute file path of given file
+    return final_path
```
alecandido, line 570:
```suggestion
    return path.join(dir_path, filename)  # absolute file path of given file
```
It doesn't need a name if it's returned right after (the role is clear: the output of the function).

src/qibo/ui/mpldrawer.py
```diff
@@ -0,0 +1,721 @@
...
+            init_label = (
+                r"$\rm{\sqrt{X}}^{\dagger}$" if is_dagger else r"$\rm{\sqrt{X}}$"
+            )
+
+        if (
+            len(gate._control_qubits) > 0
+            and "C" in init_label[0]
+            and "CNOT" != init_label
+        ):
+            init_label = gate.draw_label.upper()
```
alecandido, line 603: I guess they are all `elif`, since if one is true, the other ones are not. No need to repeat the checks if you already matched one of the conditions

src/qibo/ui/mpldrawer.py
```diff
@@ -0,0 +1,721 @@
...
+
+from qibo import gates
+
+from .FusedGateBarrier import FusedGateEndBarrier, FusedStartGateBarrier
+
+global STYLE
+STYLE = {}
+
+global SYMBOLS
+SYMBOLS = {}
```
alecandido, line 19: `global` keyword is only needed in a local scope, here is useless.
```suggestion
STYLE = {}

SYMBOLS = {}
```

alecandido: Moreover, it is almost always a bad idea to use global variables, only use global constants.

You can load them top-level (and never write to them), or pass them as parameters.
```py
from pathlib import Path

UI = Path(__file__).parent
STYLE = json.loads((UI / "styles.json").read_text())
SYMBOLS = json.loads((UI / "symbols.json").read_text())
```

alecandido (commented):
I didn't review carefully all of `mpldrawer`, especially I didn't go through the details of the drawing process.

However, here are some general comments on the notebook, and just technical ones on the drawer.

examples/circuit-draw-mpl/qibo-draw-circuit-matplotlib.ipynb
alecandido, line 1: Often, you assign `ax, fig = plot(...)`, but then you do not use `ax, fig`. I would drop them.

alecandido: A tiny suggestion about notebook aesthetics: when you `plot()`, and your plot function returns, the notebook also shows the interpreter-oriented representation of the output (something like `(<Axes: >, <Figure size 2160x240 with 1 Axes>)`) which is not especially menaingful.

If you add a `;` at the end of your statement, the return value would not be shown :)

src/qibo/ui/mpldrawer.py
```diff
@@ -0,0 +1,687 @@
...
+
+    return cluster_gates
+
+
+def _build_path(filename):
+    file_path = path.abspath(__file__)  # full path of current file
+    dir_path = path.dirname(file_path)  # full path of the directory from file
+    return path.join(dir_path, filename)  # absolute file path of given file
+
+
```
alecandido, line 554:
```suggestion
```
Now unused :)

src/qibo/ui/mpldrawer.py
```diff
@@ -0,0 +1,687 @@
...
+from qibo import gates
+
+from .FusedGateBarrier import FusedEndGateBarrier, FusedStartGateBarrier
+
+UI = Path(__file__).parent
+STYLE = json.loads((UI / "styles.json").read_text())
+SYMBOLS = json.loads((UI / "symbols.json").read_text())
+
+plot_params = {
```
alecandido, line 20:
```suggestion
PLOT_PARAMS = {
```
Naming convention for constants.

src/qibo/ui/mpldrawer.py
```diff
@@ -0,0 +1,687 @@
...
+
+    if style is not None:
+        if type(style) is dict:
+            plot_params.update(style)
+        else:
+            plot_params.update(
+                STYLE[style] if style in list(STYLE.keys()) else STYLE["default"]
+            )
+    else:
+        plot_params.update(STYLE["default"])
```
alecandido, line 638:
```suggestion
	params = PLOT_PARAMS.copy()
    if not isinstance(style, dict):
        try:
            style = STYLE.get(style)
        except AttributeError:
            style = STYLE["default"]
    params.update(style)
```

It could even a function on its own `_plot_params(style: Union[dict, str, None]) -> dict`.

src/qibo/ui/mpldrawer.py
```diff
@@ -0,0 +1,687 @@
...
+        if cluster_gates:
+            gates_cluster = _make_cluster_gates(gates_plot)
+            ax = _plot_quantum_schedule(gates_cluster, inits, labels, scale=scale)
+            return ax, ax.figure
+
+        ax = _plot_quantum_circuit(gates_plot, inits, labels, scale=scale)
+        return ax, ax.figure
+    else:
+        ax = _plot_lines_circuit(inits, labels, scale=scale)
+        return ax, ax.figure
```
alecandido, line 687: It would be ideal if a circuit with no gates would not require a dedicated function. I.e.:
- for sure `all_gates` would correctly be `[]` after the `for` loop (which would not enter)
- `_process_gates([])` will also return `[]`
- `_plot_quantum_circuit([], ...)` should be able to do the final job :)

alecandido: Maybe, it might be sufficient to promote some `ng` to `nq+ng` in `_plot_quantum_circuit()`

alecandido (approved):
There are things that may still be improved. But that will be the case forever ^^

As the first implementation of this feature in Qibo, it is definitely acceptable.

Thanks a lot @sergiomtzlosa! :)

src/qibo/ui/mpldrawer.py
sergiomtzlosa: Following your suggestion, I also review the code in `_plot_quantum_schedule` to be just a single call to `_plot_quantum_circuit` this way I can get rid of `_plot_lines_circuit` and make `_plot_quantum_circuit` the main internal function to draw the circuits

src/qibo/ui/mpldrawer.py
```diff
@@ -0,0 +1,654 @@
...
+    return measured
+
+
+def _draw_gates(
+    ax, l, labels, gate_grid, wire_grid, plot_params, measured={}, schedule=False
+):
+    for i, gate in _enumerate_gates(l, schedule=schedule):
+        _draw_target(ax, i, gate, labels, gate_grid, wire_grid, plot_params)
+        if len(gate) > 2:  # Controlled
+            _draw_controls(
```
MatteoRobbiati, line 146: We could probably improve this method by adding an option to draw gates involving $\geq 2$ qubits but without forcing the procedure to draw controls. What I mean is that there are in Qibo general unitaries which we may want to draw as longer gates touching more than one qubit (rectangles instead of squares). Of course we can keep this for a future work/PR :)

ATM those general unitaries are quite confusing (in my opinion) because we draw them adding an "U" (or a square in this plotting procedure) for each qubit involved in the process.

src/qibo/ui/mpldrawer.py
```diff
@@ -0,0 +1,654 @@
...
+
+    return style
+
+
+def plot(circuit, scale=0.6, cluster_gates=True, style=None):
+    """Main matplotlib plot function for Qibo circuit
+    circuit         A Qibo circuit to plot (type: qibo.models.circuit.Circuit)
+    scale           Scale the ouput plot
+    cluster_gates   Group single-qubit gates
+    style           Style applied to the circuit (built-in styles: garnacha, fardelejo, quantumspain, color-blind and cachirulo)
```
MatteoRobbiati, line 606: I don't know how good it will be, but it would be very nice to have a style based on  [SciencePlots](https://github.com/garrettj403/SciencePlots), to produce circuit diagrams standardized with the default LaTeX font. Also for this, we can keep it for future PR.

src/qibo/ui/FusedGateBarrier.py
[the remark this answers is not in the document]
alecandido: In case of doubts, always refer to the authoritative [PEP 8](https://peps.python.org/pep-0008/)

src/qibo/ui/FusedGateBarrier.py
```diff
@@ -0,0 +1,32 @@
...
+
+        super().__init__()
+        self.name = (
+            "FusedStartGateBarrier"
+            + str(nfused)
+            + ("" if not equal_qbits else "@EQUAL")
+        )
+        self.draw_label = ""
+        self.control_qubits = (q_ctrl,)
+        self.target_qubits = (q_trgt,) if q_ctrl != q_trgt else ()
```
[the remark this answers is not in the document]
sergiomtzlosa: Hi,

Those attributes are holding the drawing configuration array for matplotlib, so they are needed, these attributes are used in different purpose from the ones in built-in qibo gates

src/qibo/ui/mpldrawer.py
sergiomtzlosa: I think this could be a deep change on the code and we should talk about it because the unitary symbol is also comfusing when using the `print()` function, maybe the change should be to use another symbol than  a boxed "U" in my case. Maybe this is suitable for another PR as the issue  below.

If this is fine, we can close it by now...

src/qibo/ui/mpldrawer.py
sergiomtzlosa: This is a good enhance but I think it should be better to keep it for a future PR

MatteoRobbiati (commented):
Thanks a lot for the modifications @sergiomtzlosa.
If it is fine with you I would prefer to align the docstrings style with the rest of Qibo :) The docstrings here are a bit different from the "canonical" way we do them. You can take [this](https://github.com/qiboteam/qibo/blob/f20f4330fe62ce7bc022b687cf6559f2fbc1b50c/src/qibo/derivative.py#L15) or [this](https://github.com/qiboteam/qibo/blob/f20f4330fe62ce7bc022b687cf6559f2fbc1b50c/src/qibo/measurements.py#L177) as examples (I just selected them randomly). 

Finally, I was wondering if we aim to add the `mpldrawer` to the documentation. In that case we have to add it to the [doc/source](https://github.com/qiboteam/qibo/tree/master/doc/source) files. I think it would be nice to prepare a simple example to show the users how to use the drawer :)

sergiomtzlosa:
Hi, 

Thanks a lot for your support and advices, I have uploaded the documentation and gave and example on the doc/source path, please, let me know if this is enough of not.

Thanks!

PR #1371, opened 2024-06-24
Remove `TrotterCircuit` cache

### LINKED ISSUE #1357: Question about `SymbolicHamiltonian.circuit`
I assume it's a bug because it has cost me 2x a couple of days of work to trouble shoot the following issue appearing deep in the `dbi` module code:

```py
import numpy as np
from copy import deepcopy

from qibo import symbols, hamiltonians

d = hamiltonians.SymbolicHamiltonian( symbols.Z(0))
 
a = d.circuit(0.1)
b = d.circuit(0.1)
print(np.linalg.norm( d.circuit(0.2).unitary() - (a+b).unitary()))

a = deepcopy(d).circuit(0.1)
b = deepcopy(d).circuit(0.1)
print(np.linalg.norm( d.circuit(0.2).unitary() - (a+b).unitary()))
```
Output:
```
0.28237154359997657
0.0
```

TLDR: If you assign the return value of `SymbolicHamiltonian.circuit` and then want to use the `Circuit` instance e.g. for further addition then noise comes out. Correct result can be obtained by running on deepcopies.

stavros11:
Fixes #1357. 

I believe the `TrotterCircuit` class is implementing a cache of the circuit implementing the Trotter decomposition returned by `hamiltonian.circuit(dt)` , to avoid recreating it if the users decides to use a different `dt`. However, this creates a ("secret") link between the circuits returned by `hamiltonian.circuit` that can confuse users, such as in #1357, and also complicates the code.

If allocating a circuit is not considered very costly, this cache probably does not provide much in terms of performance, therefore I decided to remove it to have cleaner code (and behavior).

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/tests/test_hamiltonians_trotter.py
+++ b/tests/test_hamiltonians_trotter.py
@@ -3,7 +3,7 @@
 import numpy as np
 import pytest
 
-from qibo import hamiltonians
+from qibo import hamiltonians, symbols
 from qibo.backends import NumpyBackend
 from qibo.quantum_info import random_hermitian, random_statevector
 
@@ -141,6 +141,14 @@ def test_trotter_hamiltonian_three_qubit_term(backend):
     target_state = np.dot(u[1], np.dot(u[0], initial_state))
     target_state = np.dot(u[0], np.dot(u[1], target_state))
     backend.assert_allclose(final_state, target_state)
+
+
+def test_symbolic_hamiltonian_circuit_different_dts(backend):
+    """Issue: https://github.com/qiboteam/qibo/issues/1357."""
+    ham = hamiltonians.SymbolicHamiltonian(symbols.Z(0))
+    a = ham.circuit(0.1)
+    b = ham.circuit(0.1)
+    assert np.linalg.norm(ham.circuit(0.2).unitary() - (a + b).unitary()) == 0
 
 
 def test_old_trotter_hamiltonian_errors():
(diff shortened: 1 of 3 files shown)


tests/test_hamiltonians_trotter.py
```diff
@@ -143,6 +143,14 @@ def test_trotter_hamiltonian_three_qubit_term(backend):
     backend.assert_allclose(final_state, target_state)
 
 
+def test_symbolic_hamiltonian_circuit_different_dts(backend):
+    """Issue: https://github.com/qiboteam/qibo/issues/1357."""
+    ham = hamiltonians.SymbolicHamiltonian(symbols.Z(0))
+    a = ham.circuit(0.1)
+    b = ham.circuit(0.1)
+    assert np.linalg.norm(ham.circuit(0.2).unitary() - (a + b).unitary()) == 0
```
alecandido, line 151:
```suggestion
    assert np.allclose(ham.circuit(0.2).unitary(), (a + b).unitary())
```
?

alecandido (approved):
Thanks, I agree with your analysis, which is confirmed by the test you added.

marekgluza:
> If allocating a circuit is not considered very costly

- [x] In my applications it was not costly and I was doing deepcopy all the time.

marekgluza (approved):
Thank you @stavros11 for this implementation of a solution.

When I was reading the code before I was wondering what was the use of `TrotterCircuit` and removing it resolves that question too.

tests/test_hamiltonians_trotter.py
marekgluza: suggestion looks good to me too

stavros11: 784db2c0e3c5c6aad5de18be5574aa575814960c

stavros11:
Thank you for reviewing. I updated the test in 784db2c0e3c5c6aad5de18be5574aa575814960c. I will merge after the CI passes.

PR #1373, opened 2024-06-26
Add `gates.GeneralizedRBS` gate

renatomello:
This needs https://github.com/qiboteam/qibojit/pull/183

Checklist:
- [x] Tests are passing.


src/qibo/gates/gates.py
```diff
@@ -2469,6 +2469,76 @@ def __init__(self, q0, q1, q2, theta, trainable=True):
...
+    """The generalized Reconfigurable Beam Splitter gate.
+
+    Corresponds to the following matrix:
+
+    .. math::
+
+        \begin{pmatrix}
+            I &           &  &             &  \\
+              & e^{i\\phi}\\cos\theta &      & e^{-i\\phi}\\sin\theta   &  \\
+              &  & I'       &    &  \\
```
alecandido, line 2482: Why do you need an explicit identity here?

This gate is always acting on just two qubits, can't you just define the matrix on the subspace of the two qubits involved?
I would keep this similar to the swap gate, if possible...

src/qibo/backends/npmatrices.py
```diff
@@ -489,6 +489,28 @@ def DEUTSCH(self, theta):
             dtype=self.dtype,
         )
 
+    def GeneralizedRBS(self, qubits_in, qubits_out, theta, phi):
```
alecandido, line 492: I'm definitely missing something.

Similarly to the docs, the non-trivial elements (non-zero off-diagonal, non-one on-diagonal) are just 4, corresponding to only two qubits.

Since there is already a `GIVENS` gate (and the `RBS` alias), couldn't you use the same approach there, just changing the values for the matrix elements?

src/qibo/backends/npmatrices.py
```diff
@@ -489,6 +489,28 @@ def DEUTSCH(self, theta):
...
+        integer_out = int(integer_out, 2)
+
+        matrix = [
+            [1 + 0j if l == k else 0j for l in range(2**bitstring_length)]
+            for k in range(2**bitstring_length)
+        ]
+        matrix[integer_in][integer_in] = self.np.exp(-1j * phi) * self.np.cos(theta)
+        matrix[integer_in][integer_out] = -self.np.exp(1j * phi) * self.np.sin(theta)
+        matrix[integer_out][integer_in] = self.np.exp(-1j * phi) * self.np.sin(theta)
+        matrix[integer_out][integer_out] = self.np.exp(1j * phi) * self.np.cos(theta)
```
alecandido, line 510: I'd say that there is no need to recompute the elements twice. I know it's not that expensive, but it should be sufficiently simple to compute them only once, by defining variables like:
```py
ep = self.np.exp(1j * phi)
emp = self.np.exp(-1j * phi)
ct = self.np.cos(theta)
st = self.np.sin(theta)
```
(or whatever other name you like)

src/qibo/gates/gates.py
renatomello: This is not a two-qubit matrix. I forgot to include that in the docstring.

src/qibo/backends/npmatrices.py
renatomello: Again. this is not a two-qubit gate. I'll expand the docstring to make it clearer.

alecandido: Ok, fine for docstrings. I missed that point.

Here, what I'd suggest you is to define the gate on the two-states subspace, and then replace the entries in the larger matrix for that:
```py
matrix = np.eye(bitstring_length, dtype=np.complex128)
submatrix = np.array(
  [[exp * cos, -exp * sin],
   [conj * sin, -conj * cos]]
)
matrix[[integer_in, integer_out]][:,[integer_in, integer_out]] = submatrix
```

We could even apply the `submatrix` directly in the simulators, but this would require extra effort in each of them to support this gate. So, much better to define the full matrix, but I'd make as apparent as possible that is acting on just two states.

renatomello: > It's already defined like that, but using `list` instead of `numpy.ndarray` because every gate is defined like that in other to avoid casting issues w.r.t. `tensorflow` and `torch`. These changes were necessary in the past. If you look closely, `numpy` is not even imported in that submodule because of that.

alecandido: I wrote `np.`, but you could use `self.np.`, which used almost everywhere else.

In any case, you're casting them at the end anyhow, and defined as lists or defined as arrays the cast can work in both cases (all Python array frameworks support the cast from NumPy).

To be fair, I would even define all of them as arrays, and then just cast in the simulator, when you retrieve the matrix (you're executing this operation anyhow, every time that you invoke the methods).

My motivation for asking you to do this for the gRBS gate is that it's the only one for which the matrix is not easily readable for the code. So, it's to reduce future human effort of people that will need to maintain it.
But for sure, casting everything to NumPy is not this PR's business.

If you're strongly motivated to avoid the usage of NumPy[*], even as `self.np.`, just ignore this thread.

[*]: notice that the only primitive above that is not already used somewhere else is the indexing, since `self.np.eye` is used for the identity, and the `np.array()` call could be replaced by `self._cast()` - but I can not guarantee that exactly the indexing is consistent in all frameworks, if the other things are consistent (i.e. `np.eye`, `np.sin`, `np.conj`, ...), so I'll not take this responsibility

renatomello: > I wrote `np.`, but you could use `self.np.`, which used almost everywhere else.
> 
> In any case, you're casting them at the end anyhow, and defined as lists or defined as arrays the cast can work in both cases (all Python array frameworks support the cast from NumPy).

I'm quite sure it's not done like that right now because `tensorflow` doesn't support item assignments.

> To be fair, I would even define all of them as arrays, and then just cast in the simulator, when you retrieve the matrix (you're executing this operation anyhow, every time that you invoke the methods).

We also moved on from doing that because of issues with `tensorflow`and `torch` gradients.

> If you're strongly motivated to avoid the usage of NumPy[*], even as `self.np.`, just ignore this thread.

Reasons above.

> [*]: notice that the only primitive above that is not already used somewhere else is the indexing, since `self.np.eye` is used for the identity, and the `np.array()` call could be replaced by `self._cast()` - but I can not guarantee that exactly the indexing is consistent in all frameworks, if the other things are consistent (i.e. `np.eye`, `np.sin`, `np.conj`, ...), so I'll not take this responsibility

The `self.np.eye` in the identity gate works only because no item assignment is done after the identity is created, which doesn't create `tensorflow` issues.

alecandido: > We also moved on from doing that because of issues with `tensorflow`and `torch` gradients.

I guess these functions should be enough to convert constant arrays to constant tensors:

https://www.tensorflow.org/api_docs/python/tf/convert_to_tensor
https://pytorch.org/docs/stable/generated/torch.from_numpy.html

It should only be done consistently, as `self._cast()` is consistently applied.

Thanks for the discussion and providing the historical background. For sure, it's not something for this PR to handle, and soon not even for this repo (after #1365 those backends won't be in `qibo` any longer, and we're even considering moving the matrices' definition to `qibo-core`, cf. https://github.com/qiboteam/qibo-core/issues/20).

renatomello: > > We also moved on from doing that because of issues with `tensorflow`and `torch` gradients.
> 
> I guess these functions should be enough to convert constant arrays to constant tensors:
> 
> https://www.tensorflow.org/api_docs/python/tf/convert_to_tensor https://pytorch.org/docs/stable/generated/torch.from_numpy.html
> 
> It should only be done consistently, as `self._cast()` is consistently applied.

The issue wasn't about converting the arrays to tensors. It's that sometimes one would start with a tensor, then a numpy function would be applied to the tensor. Even if one converts the array back to a tensor, gradient taping (and the torch equivalent) is already broken.

renatomello: > Thanks for the discussion and providing the historical background. For sure, it's not something for this PR to handle, and soon not even for this repo (after #1365 those backends won't be in `qibo` any longer, and we're even considering moving the matrices' definition to `qibo-core`, cf. [qiboteam/qibo-core#20](https://github.com/qiboteam/qibo-core/issues/20)).

As a matter of record for posterity: It's not to be addressed on this PR, as you said. However, regardless of where the matrix representations are defined, the aforementioned `tensorflow` and `torch` particularities will still persist.

alecandido: I see, the problem is that you're not working with constant arrays, so you need the gate derivative, and you're obtaining it by propagating the derivative through the gate definition.

For working with multi-framework, you may define the derivative analytically, since the derivative of a gate is the derivative of its entries (usually pretty simple functions).
However, an alternative approach is to use a single differentiation backend, just exposing gradients of the circuits to all the backends. This is something that will be possible in Qiboml, and should be discussed there.

Simone-Bordoni (approved):

src/qibo/transpiler/decompositions.py
```diff
@@ -402,6 +402,38 @@ def _u3_to_gpi2(t, p, l):
 )
 
 
+def _decomposition_generalized_RBS(ins, outs, theta, phi, controls):
```
MatteoRobbiati, line 405: Adding here some docstrings could improve the readability for who is not familiar with the paper.

MatteoRobbiati (approved):
Thanks, LGTM!

src/qibo/transpiler/decompositions.py
renatomello: The docstring would be basically redundant w.r.t. the one of `gates.GeneralizedRBS`, which is why I chose to not repeat it.

renatomello:
@MatteoRobbiati the PR in `qibojit` needs to be merged before this one

PR #1380, opened 2024-07-05
Generalization of `models.unary_encoder(nqubits, architecture="tree")` for all `nqubits`

renatomello:
This function was previously implemented only for when `nqubits` was a power of 2. Now it works for general `nqubits`.

Checklist:
- [x] Tests are passing.


Simone-Bordoni (approved):

src/qibo/models/encodings.py
```diff
@@ -395,27 +385,33 @@ def _generate_rbs_pairs(nqubits: int, architecture: str, **kwargs):
...
+            new_registers, per_depth = [], []
+            for register in registers:
+                limit = (
+                    int(len(register) / 2)
+                    if len(register) % 2 == 0
+                    else int(len(register) / 2) + 1
+                )
+                new_registers.append(register[:limit])
+                new_registers.append(register[limit:])
+                if len(register[:limit]) + len(register[limit:]) >= 2:
```
BrunoLiegiBastonLiegi, line 405: isn't this the same as
```suggestion
                if len(register) >= 2:
```

src/qibo/models/encodings.py
```diff
@@ -395,27 +385,33 @@ def _generate_rbs_pairs(nqubits: int, architecture: str, **kwargs):
...
+                    per_depth.append((register[:limit][0], register[limit:][0]))
```
BrunoLiegiBastonLiegi, line 406: and this should simply be 
```suggestion
                    per_depth.append((register[0], register[limit]))
```
if I am understanding this correctly

src/qibo/models/encodings.py
```diff
@@ -395,27 +385,33 @@ def _generate_rbs_pairs(nqubits: int, architecture: str, **kwargs):
...
+            pairs_rbs.append(per_depth)
+            registers = new_registers
+            new_registers = []
```
BrunoLiegiBastonLiegi, line 409:
```suggestion
```
this should be initialized at the beginning of the next iteration anyway

renatomello:
This is not gonna be as simple as I initially thought. So I'm gonna close this for now and recreate the PR if and when we have a general solution.

PR #1382, opened 2024-07-07
Expectation values as native backend arrays

BrunoLiegiBastonLiegi:
This updates the expectation values calculation, that are now computed and returned as native arrays of the backend used. This, in turns, begs for updating some of the error mitigation methods accordingly. In particular, as `torch` was having problems with the `scipy.curve_fit` I had to implement a custom `_curve_fit` based on the `torch.optim.LBFGS` optimizer.

Checklist:
- [x] Tests are passing.


src/qibo/models/error_mitigation.py
```diff
@@ -302,6 +306,28 @@ def sample_training_circuit_cdr(
...
+            optimizer.zero_grad()
+            output = model(xdata, *params)
+            loss_val = loss(output, ydata)
+            loss_val.backward(retain_graph=True)
+            return loss_val
+
+        optimizer.step(closure)
+        return params
+    else:
+        return curve_fit(model, xdata, ydata, p0=params)[0]
```
renatomello, line 328:
```suggestion
    return curve_fit(model, xdata, ydata, p0=params)[0]
```

src/qibo/models/error_mitigation.py
```diff
@@ -302,6 +306,27 @@ def sample_training_circuit_cdr(
     return sampled_circuit
 
 
+def _curve_fit(
+    backend, model, params, xdata, ydata, lr=1, max_iter=int(1e2), tolerance_grad=1e-5
+):
+    if backend.name == "pytorch":
+        loss = lambda pred, target: backend.np.mean((pred - target) ** 2)
+        optimizer = backend.np.optim.LBFGS(
```
MatteoRobbiati, line 314: LBFGS should be a gradient-based optimizer, thus I am not sure it will work well in case we use shots. 
It is true here we just fit the plot of noisy vs exact, so it is probably fine,  but, just in case, I would use Powell instead.

MatteoRobbiati (approved):
Thanks! I just left a comment about the optimizer choice.

src/qibo/models/error_mitigation.py
BrunoLiegiBastonLiegi: pytorch doesn't seem to have any implementation for the Powell method unfortunately. In any case here we are just fitting a given function on some `(x, y)` points which should be independent on how the points were generated, with shots or not.

src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -564,7 +564,9 @@ def expectation_from_samples(self, freq, qubit_map=None):
             if len(term.factors) != len(set(term.factors)):
                 raise_error(NotImplementedError, "Z^k is not implemented since Z^2=I.")
         keys = list(freq.keys())
-        counts = np.array(list(freq.values())) / sum(freq.values())
+        counts = self.backend.cast(list(freq.values()), self.backend.precision) / sum(
```
renatomello, line 567: Why the pass `dtype=backend.precision`? Wouldn't it be `dtype=self.backend.dtype`?

src/qibo/models/error_mitigation.py
```diff
@@ -302,6 +306,27 @@ def sample_training_circuit_cdr(
...
+            [params], lr=lr, max_iter=max_iter, tolerance_grad=tolerance_grad
+        )
+
+        def closure():
+            optimizer.zero_grad()
+            output = model(xdata, *params)
+            loss_val = loss(output, ydata)
+            loss_val.backward(retain_graph=True)
+            return loss_val
```
renatomello, line 323: It's better to define this outside of `_curve_fit` and have its own docstring?

src/qibo/models/error_mitigation.py
MatteoRobbiati: > In any case here we are just fitting a given function on some `(x, y)` points which should be independent on how the points were generated, with shots or not.

Yes indeed, you are right @BrunoLiegiBastonLiegi.

> It should be documented somewhere that the optimizer for torch is different from the others.

I think all the backends are still supporting our optimizers, but yes, I would definitely mention that each framework (in particular QML ones) can work with their own optimizers too and in some cases this choice is necessary (like here).

src/qibo/hamiltonians/hamiltonians.py
BrunoLiegiBastonLiegi: No because `backend.dtype` specifies the type of the states and matrices, which is going to be `complex128/64`, whereas the `backend.precision` only specifies `"single"` or `"double"` which casts to `float32/64` that is what we need for expectation values.

renatomello (approved):

PR #1389, opened 2024-07-12
Symbolic Hamiltonian with constants

### LINKED ISSUE #1383: Expectation for observable does not take into account constants
I was fiddling around with a QAOA MaxCut Hamiltonian by feeding it into the ZNE as observable and stumbled upon some weird stuff.

The QAOA MaxCut Hamiltonian has the form $H = \sum_{<i,j> \in E} 0.5 ( 1 - Z_i Z_j)$, where the $Z_i$ and $Z_j$ are $Z$ operators on the $i$-th and $j$-th qubits. 

For a simple fully connected 3 nodes graph (thus 3 qubits), $H$ has the form $H = 0.5 (1-Z(0) Z(1)) + 0.5  (1-Z(1) Z(2)) + 0.5 (1-Z(2) Z(0))$. We can simplify this to be $H = 1.5 - 0.5 Z(0) Z(1) - 0.5 Z(1) Z(2) - 0.5 Z(2) Z(0)$. 

To compute the expectation value of the observable, we input observable $H = 1.5 - 0.5 Z(0) Z(1) - 0.5 Z(1) Z(2) - 0.5 Z(2) Z(0)$. The answer turns out wrong as the computation doesn't take into account the constant, which in this case is $1.5$. It does take into account the $0.5$ for each $Z(i) Z(j)$. 

Therefore to properly compute H, we need to do two separate computations: First to compute $0.5 Z(0) Z(1) - 0.5 Z(1) Z(2) - 0.5 Z(2) Z(0)$, then we subtract it from $1.5$. 
 
I was just wondering if this should be an issue because it's not intuitive to do two separate computations for a Hamiltonian that has constants.

alecandido:
Fixes #1383

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/hamiltonians/hamiltonians.py
+++ b/src/qibo/hamiltonians/hamiltonians.py
@@ -328,8 +328,8 @@ class SymbolicHamiltonian(AbstractHamiltonian):
             self.nqubits = nqubits
 
     @property
-    def dense(self):
-        """Creates the equivalent :class:`qibo.hamiltonians.MatrixHamiltonian`."""
+    def dense(self) -> "MatrixHamiltonian":
+        """Creates the equivalent Hamiltonian matrix."""
         if self._dense is None:
             log.warning(
                 "Calculating the dense form of a symbolic Hamiltonian. "
@@ -380,8 +380,9 @@ class SymbolicHamiltonian(AbstractHamiltonian):
 
     @property
     def terms(self):
-        """List of :class:`qibo.core.terms.HamiltonianTerm` objects
-        of which the Hamiltonian is a sum of.
+        """List of terms of which the Hamiltonian is a sum of.
+
+        Terms will be objects of type :class:`qibo.core.terms.HamiltonianTerm`.
         """
         if self._terms is None:
             # Calculate terms based on ``self.form``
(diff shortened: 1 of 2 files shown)


alecandido:
@mho291 I started adding a minimal test that reproduces your issue.

I'm now going to add the patch that should make that test passing. Whenever you have time, confirm that the test is close enough to what you expect (then I will also ask to check the full fix).

alecandido:
For the time being, it's a quick patch (plus some docstrings beautifications).

I wanted to refactor a bit more the function. But I took a look at the rest of the module, and there are many things to refactor. So, it's not worth to do it now, better to allocate some time and do it consistently.

src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -566,8 +568,8 @@ def expectation_from_samples(self, freq, qubit_map=None):
                 if subk.count(1) % 2 == 1:
                     expval_k = -1
                 expval_q += expval_k * counts[i]
-            expval += expval_q * float(coeff[j])
-        return expval
+            expval += expval_q * self.terms[j].coefficient.real
```
alecandido, line 571: Any clue why `SymbolicTerm` coefficient are stored as complex values?
https://github.com/qiboteam/qibo/blob/c3031c606464650856d6d278b799eeee4a1f49fe/src/qibo/hamiltonians/terms.py#L138

@andrea-pasquale?

stavros11: I think `terms` here may not necessarily be hermitian. A term is a product of symbols and it happens that the symbols we provide are the Paulis which are all hermitian, however the user can also define their own symbol using custom matrix. So there may be a case where you have non-hermitian terms and complex coefficients resulting to a hermitian Hamiltonian.

The constant, on the other hand, probably needs to be real in every case, since it is multiplying the idenity that is hermitian.

stavros11 (approved):

mho291:
Thanks @alecandido ! I have checked the code and verified it works. Will approve the changes.

mho291 (approved):

PR #1390, opened 2024-07-12
Move `partial_trace` and `matrix_exponentiation` from `backend` methods to standalone functions in `quantum_info.linalg_operations`

### LINKED ISSUE #1168: Promote partial trace (states and density matrices) from Backend to Quantum Information Module
There are some features in the Backend that could be promoted into some other modules as they are useful in a variety of situations. Partial traces are one of them. But maybe more are also good to promote.

renatomello:
Closes #1168 

Checklist:
- [x] Tests are passing.


src/qibo/quantum_info/linalg_operations.py
```diff
@@ -0,0 +1,227 @@
...
+        if backend_class_name == "TensorflowBackend":
+            return backend.tf.linalg.expm(matrix)
+
+        if backend_class_name == "PyTorchBackend":
+            return backend.np.linalg.matrix_exp(matrix)
+
+        if is_sparse:
+            expm = expm_sparse_scipy
+        elif backend_class_name in ["CupyBackend", "CuQuantumBackend"]:
+            from cupyx.scipy.linalg import expm as expm_cupy  # pylint: disable=E0401
```
alecandido, line 216: If a function explicitly references individual backends, then it should be part of the backend definition.
Backends may change, from content to even the name. And we should be free to do so without scanning the whole Qibo for explicit mentions.

If you want to expose a function in this module, to have the feature together with other similar ones, then keep the `calculate_matrix_exp` in the backends, and call that one inside here.

src/qibo/backends/abstract.py
```diff
@@ -66,7 +66,7 @@ def cast(self, x, copy=False):  # pragma: no cover
         raise_error(NotImplementedError)
 
     @abc.abstractmethod
-    def issparse(self, x):  # pragma: no cover
+    def is_sparse(self, x):  # pragma: no cover
```
alecandido, line 69: I'm not using this myself, and I expect no user is directly using it. But did you scan all examples and other relevant locations (i.e. Qibojit) to make sure this won't break anything?

alecandido (approved):
A small comment about a renaming, other than that I'm fine with the proposed changes.

src/qibo/backends/abstract.py
renatomello: I did it here but not in `qibojit`. I'll create the corresponding PR there.

andrea-pasquale (approved):

PR #1395, opened 2024-07-19
Fix `DeprecationWarning` for `numpy > 1.25`

renatomello:
Checklist:
- [x] Tests are passing.


alecandido (approved):

alecandido:
At some point, we should also take into account migrating to NumPy 2.0, now that is out.

A much relevant improvement (to us) could be the following:
https://numpy.org/doc/stable/release/2.0.0-notes.html#array-api-compatible-functions-aliases
https://numpy.org/doc/stable/release/2.0.0-notes.html#array-api-compatible-functions-for-numpy-linalg
https://numpy.org/doc/stable/release/2.0.0-notes.html#array-api-standard-support-in-the-main-namespace

It was already documented for v1, and there are compatibility packages available
https://numpy.org/doc/stable/reference/array_api.html
but now it is available by default, and it should improve the compatibility with the other array frameworks (i.e. TensorFlow, PyTorch, JAX, CuPy, ...), since the Python API has been standardized (not all of them adopted it already, but they might do it).

Issue #1396, opened 2024-07-22
Looping Behavior in Qibo SABRE


csookim:
There is a bug in the SABRE algorithm, and I modified `router.py` to fix it. Is it okay to make a PR? My code is in a forked repo. Below is the explanation.

csookim:
# Looping Behavior in Qibo SABRE

## Introduction

When the SABRE router encounters a two-qubit gate that cannot be executed due to hardware constraints, it adds a sequence of SWAPs to make the two-qubit gate executable. SABRE generates candidate SWAPs and selects one based on a defined formula. This process is repeated until the two-qubit gate becomes executable. However, since SABRE is a heuristic algorithm, the process can sometimes get stuck in a loop.

## Problem

In some cases, SABRE adds a repeated sequence of similar SWAPs. For example, a 10-qubit circuit with the following connectivity:

```python
qubit_array = [(7, 2), (6, 0), (5, 6), (4, 8), (3, 5), (9, 1)]
```

is routed with 156 SWAPs using the Trivial placer and SABRE router. (After modification, the number of SWAPs is reduced to 17.) This example is illustrated in the [notebook](https://github.com/csookim/qibo_transpiler_cskim/blob/bff207ebc519313701bf0f724a757aeee243a6e6/cskim/docs/transpiler_general/21-1_qibo_sabre_looping.ipynb).


## Solution

To solve this, the number of SWAPs is managed until every routing is made. [Notebook](https://github.com/csookim/qibo_transpiler_cskim/blob/bff207ebc519313701bf0f724a757aeee243a6e6/cskim/docs/transpiler_general/21-1_qibo_sabre_looping.ipynb)

```python
...        
    self.circuit.update(best_candidate)   # SWAP is added
    self._added_swaps += 1
...
```

`_added_swaps` is counted after one SWAP is picked from the candidate SWAPs. If routing is completed (a two-qubit gate becomes executable), `_added_swaps` is reset to 0.

```python
...
    if self._added_swaps > 1.5 * longest_path:  # threshold is arbitrary
        self.circuit = deepcopy(self._saved_circuit)
        self._route_to_nearest_gate()
...
```

If `_added_swaps` exceeds the threshold, the nearest two-qubit gate is routed manually. The two-qubit gate with the minimum distance between the qubits is selected, and qubit 1 is moved adjacent to qubit 2.

## Tests

When routing 30 random CZ circuits with the Random placer and original SABRE router, there are cases where more than 100 SWAPs are added.

[...]

PR #1398, opened 2024-07-22
Fix Looping Behavior in Qibo SABRE

csookim:
This PR addresses the issue #1396.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/transpiler/router.py
+++ b/src/qibo/transpiler/router.py
@@ -687,7 +687,7 @@ class Sabre(Router):
             # If the number of added swaps is too high, the algorithm is stuck.
             # Reset the circuit to the last saved state and make the nearest gate executable by manually adding SWAPs.
             if (
-                self._temporary_added_swaps > 1.5 * longest_path
+                self._temporary_added_swaps > self.swap_threshold * longest_path
             ):  # threshold is arbitrary
                 self.circuit = deepcopy(self._saved_circuit)
                 self._route_to_nearest_gate()
--- a/src/qibo/transpiler/router.py
+++ b/src/qibo/transpiler/router.py
@@ -690,7 +690,7 @@ class Sabre(Router):
                 self._temporary_added_swaps > self.swap_threshold * longest_path
             ):  # threshold is arbitrary
                 self.circuit = deepcopy(self._saved_circuit)
-                self._route_to_nearest_gate()
+                self._shortest_path_routing()
 
         circuit_kwargs = circuit.init_kwargs
         circuit_kwargs["wire_names"] = list(initial_layout.keys())
@@ -900,7 +900,7 @@ class Sabre(Router):
         self._temporary_added_swaps = 0
         self._saved_circuit = deepcopy(self.circuit)
 
-    def _route_to_nearest_gate(self):
+    def _shortest_path_routing(self):
         """Route a gate in the front layer using the shortest path. This method is executed when the standard SABRE fails to find an optimized solution.
 
         Method works in-place.
--- a/src/qibo/transpiler/router.py
+++ b/src/qibo/transpiler/router.py
@@ -793,7 +793,6 @@ class Sabre(Router):
         self.circuit.update(best_candidate)
         self._added_swaps += 1
 
-
     def _compute_cost(self, candidate: int):
         """Compute the cost associated to a possible SWAP candidate."""
         temporary_circuit = CircuitMap(
@@ -891,7 +890,7 @@ class Sabre(Router):
         self._delta_register = [1.0 for _ in self._delta_register]
         self._added_swaps = 0
         self._saved_circuit = deepcopy(self.circuit)
-    
+
     def _route_to_nearest_gate(self):
         """Route the circuit to the nearest gate by adding SWAPs.
 
@@ -905,16 +904,21 @@ class Sabre(Router):
             Q1 = self.circuit.get_physical_qubits(block)[0]
             Q2 = self.circuit.get_physical_qubits(block)[1]
             path = nx.bidirectional_shortest_path(self.connectivity, Q1, Q2)
-            
+
             # Between the gates in the front layer, the one requiring the minimum #SWAPs is selected
             if len(path) < min_distance:
                 min_distance = len(path)
                 shortest_path = path
 
         # Q1 is moved
-        swaps = [(self.circuit.physical_to_logical(shortest_path[i]),
-                  self.circuit.physical_to_logical(shortest_path[i+1])) for i in range(len(shortest_path)-2)]
-        
+        swaps = [
+            (
+                self.circuit.physical_to_logical(shortest_path[i]),
+                self.circuit.physical_to_logical(shortest_path[i + 1]),
+            )
+            for i in range(len(shortest_path) - 2)
+        ]
+
         for swap in swaps:
             self.circuit.update(swap)
 
--- a/src/qibo/transpiler/router.py
+++ b/src/qibo/transpiler/router.py
@@ -922,7 +922,9 @@ class Sabre(Router):
         )
 
         # Q1 is moved
-        shortest_path = [self.circuit.physical_to_logical(q) for q in shortest_path[:-1]]
+        shortest_path = [
+            self.circuit.physical_to_logical(q) for q in shortest_path[:-1]
+        ]
         swaps = list(zip(shortest_path[:-1], shortest_path[1:]))
 
         for swap in swaps:
--- a/src/qibo/transpiler/router.py
+++ b/src/qibo/transpiler/router.py
@@ -626,6 +626,11 @@ class Sabre(Router):
             Defaults to math:`10^{-3}`.
         seed (int, optional): seed for the candidate random choice as tiebraker.
             Defaults to ``None``.
+        swap_threshold (float, optional): limits the number of added SWAPs in every routing iteration.
+            This threshold is multiplied by the length of the longest path in the circuit connectivity.
+            If the number of added SWAPs exceeds the threshold before a gate is routed,
+            the circuit is reset to the last saved state.
+            Defaults to :math:`1.5`.
 
     References:
         1. G. Li, Y. Ding, and Y. Xie,
@@ -639,12 +644,14 @@ class Sabre(Router):
         lookahead: int = 2,
         decay_lookahead: float = 0.6,
         delta: float = 0.001,
+        swap_threshold: float = 1.5,
         seed: Optional[int] = None,
     ):
         self.connectivity = connectivity
         self.lookahead = lookahead
         self.decay = decay_lookahead
         self.delta = delta
+        self.swap_threshold = swap_threshold
         self._delta_register = None
         self._dist_matrix = None
         self._dag = None
@@ -652,7 +659,7 @@ class Sabre(Router):
         self.circuit = None
         self._memory_map = None
         self._final_measurements = None
-        self._added_swaps = 0
+        self._temporary_added_swaps = 0
         self._saved_circuit = None
         random.seed(seed)
 
@@ -679,7 +686,7 @@ class Sabre(Router):
 
             # If the number of added swaps is too high, the algorithm is stuck.
             # Reset the circuit to the last saved state and make the nearest gate executable by manually adding SWAPs.
-            if self._added_swaps > 1.5 * longest_path:  # threshold is arbitrary
+            if self._temporary_added_swaps > 1.5 * longest_path:  # threshold is arbitrary
                 self.circuit = deepcopy(self._saved_circuit)
                 self._route_to_nearest_gate()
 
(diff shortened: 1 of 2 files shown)


alecandido:
Is this ready for review? Or still work in progress?

alecandido:
> Ready for review. Thanks.

Then, let's appoint some reviewers during the Qibo meeting ;)

Unless you or @marekgluza have already someone to propose.

Simone-Bordoni:
I will take care of the review. I have tried to fix this issue for many month without success.
I made the algorithm using the paper https://arxiv.org/pdf/1809.02573 and I didn't manage to find this correction made by qiskit (https://github.com/Qiskit/qiskit/blob/e5ee1a8d97753ccca09b0c15be7b7ab0c1c84b90/crates/accelerate/src/sabre/route.rs#L559).
Thank you for finding this out!

src/qibo/transpiler/router.py
```diff
@@ -875,6 +888,39 @@ def _execute_blocks(self, blocklist: list):
...
+        Method works in-place.
+        """
+
+        min_distance = float("inf")
+        shortest_path = None
+
+        for block in self._front_layer:
+            Q1 = self.circuit.get_physical_qubits(block)[0]
+            Q2 = self.circuit.get_physical_qubits(block)[1]
+            path = nx.bidirectional_shortest_path(self.connectivity, Q1, Q2)
```
Simone-Bordoni, line 906:
```suggestion
            path = nx.bidirectional_shortest_path(self.connectivity, self.circuit.get_physical_qubits(block)[0], self.circuit.get_physical_qubits(block)[1])
```
Or don't use capital letters for q1, q2

src/qibo/transpiler/router.py
```diff
@@ -875,6 +888,39 @@ def _execute_blocks(self, blocklist: list):
...
+
+            # Between the gates in the front layer, the one requiring the minimum #SWAPs is selected
+            if len(path) < min_distance:
```
Simone-Bordoni, line 909: Maybe we can first use a cycle to find the shortest path using "self._dist_matrix" and then compute the actual shortest path. In this way we call "nx.bidirectional_shortest_path()" only once.

src/qibo/transpiler/router.py
```diff
@@ -664,13 +667,22 @@ def __call__(self, circuit: Circuit, initial_layout: dict):
...
         while self._dag.number_of_nodes() != 0:
             execute_block_list = self._check_execution()
             if execute_block_list is not None:
                 self._execute_blocks(execute_block_list)
             else:
                 self._find_new_mapping()
 
+            # If the number of added swaps is too high, the algorithm is stuck.
+            # Reset the circuit to the last saved state and make the nearest gate executable by manually adding SWAPs.
+            if self._added_swaps > 1.5 * longest_path:  # threshold is arbitrary
```
Simone-Bordoni, line 682: We can leave this treshold as a parameter (in the __init__) that can be defined by the user with 1.5 as default value.

src/qibo/transpiler/router.py
```diff
@@ -651,6 +652,8 @@ def __init__(
         self.circuit = None
         self._memory_map = None
         self._final_measurements = None
+        self._added_swaps = 0
```
Simone-Bordoni, line 655: This variable name can be a bit misleading. The transpiler has the method added_swaps() that returns self.circuit._swaps which is the total number of swaps added during the procedure. I would suggest to call this variable "temporary_added_swaps", "cycle_added_swaps" or something like that to avoid confusion with the total number of swaps.

src/qibo/transpiler/router.py
```diff
@@ -875,6 +888,39 @@ def _execute_blocks(self, blocklist: list):
         self._update_front_layer()
         self._memory_map = []
         self._delta_register = [1.0 for _ in self._delta_register]
+        self._added_swaps = 0
+        self._saved_circuit = deepcopy(self.circuit)
+
+    def _route_to_nearest_gate(self):
+        """Route the circuit to the nearest gate by adding SWAPs.
```
Simone-Bordoni, line 895:
```suggestion
        """Route a gate in the front layer using the shortest path. This method is executed when the standard SABRE fails to find an optimized solution.
```

Simone-Bordoni (changes requested):

Simone-Bordoni:
The implemented solution uses the shortest path routing strategy to address the prolem. This is the same euristic strategy employed in the shortest_paths router. In this router all the possible shortest paths are considered and between them the one that also matches the connectivity of the maximum amount of other gates is chosen. Integrating the shortest_paths router in the SABRE may improve its performance (in terms of added_swaps) with a small time performance reduction.
I think we may have a meet to discuss this possibility in a future PR.

src/qibo/transpiler/router.py
```diff
@@ -664,13 +674,24 @@ def __call__(self, circuit: Circuit, initial_layout: dict):
...
             execute_block_list = self._check_execution()
             if execute_block_list is not None:
                 self._execute_blocks(execute_block_list)
             else:
                 self._find_new_mapping()
 
+            # If the number of added swaps is too high, the algorithm is stuck.
+            # Reset the circuit to the last saved state and make the nearest gate executable by manually adding SWAPs.
+            if (
+                self._temporary_added_swaps > 1.5 * longest_path
```
Simone-Bordoni, line 690:
```suggestion
                self._temporary_added_swaps > self.swap_threshold * longest_path
```

Simone-Bordoni:
Can you also add a test checking that the method "_route_to_nearest_gate()" alone is correctly working?
I also suggest changing the name of the method to "_shortest_path_routing()"

Simone-Bordoni (approved):

tests/test_transpiler_router.py
```diff
@@ -186,6 +186,63 @@ def test_routing_with_measurements():
     )
 
 
+def test_sabre_looping():
```
BrunoLiegiBastonLiegi, line 189: Is this test checking that the looping problem of sabre is fixed by the shortest path approach? The final `assert > 50` might be ok in the majority of the cases, however, I see that a bit risky. If possible, I would try to fix a setting (by fixing the seed most likely) where we have complete control over all the moving parts: make sure the standard sabre get stuck in the loop, know in advance the precise number of swaps you get with a given threshold, etc... 
Or keep this random but, for testing, check that the number of swaps inserted with the old sabre (`threshold = inf`) is larger than the new one (`threshold=1.5`) rather than just comparing to an arbitrary value 50.
I would say that the optimal solution is actually having both the tests, the fixed one and the random one, if possible.

BrunoLiegiBastonLiegi (commented):
Thanks for fixing this, everything looks almost ready to go, I have just some minor comments.

src/qibo/transpiler/router.py
```diff
@@ -875,6 +897,42 @@ def _execute_blocks(self, blocklist: list):
...
+        )
+
+        # Q1 is moved
+        swaps = [
+            (
+                self.circuit.physical_to_logical(shortest_path[i]),
+                self.circuit.physical_to_logical(shortest_path[i + 1]),
+            )
+            for i in range(len(shortest_path) - 2)
+        ]
```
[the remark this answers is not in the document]
csookim: Thank you for your suggestion, but I think your suggestion differs from the original code in two ways.

- Since it is enough to move q1 adjacent to q2, it doesn't require a swap between q1 and q2. So it would be something like:

```py
... shortest_path[0:-2], shortest_path[1: -1] ...
```

- Since the `shortest_path` array contains physical qubit numbers and the swap occurs on logical qubit numbers, `physical_to_logical` should be applied to all elements of `shortest_path`.

Therefore, the equivalent code would be:

```py
swaps = [
    (
        self.circuit.physical_to_logical(a), 
        self.circuit.physical_to_logical(b)
    ) 
    for a, b in zip(shortest_path[0:-2], shortest_path[1:-1])
]
```

Is it better to change the code to the one above?

src/qibo/transpiler/router.py
```diff
@@ -875,6 +897,42 @@ def _execute_blocks(self, blocklist: list):
         self._update_front_layer()
         self._memory_map = []
         self._delta_register = [1.0 for _ in self._delta_register]
+        self._temporary_added_swaps = 0
+        self._saved_circuit = deepcopy(self.circuit)
```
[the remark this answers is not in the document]
csookim: `self.circuit` is a CircuitMap, so the `copy()`cannot be used.

src/qibo/transpiler/router.py
```diff
@@ -664,13 +674,24 @@ def __call__(self, circuit: Circuit, initial_layout: dict):
...
                 self._execute_blocks(execute_block_list)
             else:
                 self._find_new_mapping()
 
+            # If the number of added swaps is too high, the algorithm is stuck.
+            # Reset the circuit to the last saved state and make the nearest gate executable by manually adding SWAPs.
+            if (
+                self._temporary_added_swaps > self.swap_threshold * longest_path
+            ):  # threshold is arbitrary
+                self.circuit = deepcopy(self._saved_circuit)
```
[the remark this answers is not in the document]
csookim: `self.circuit` is a CircuitMap, so the `copy()`cannot be used.

src/qibo/transpiler/router.py
```diff
@@ -664,13 +674,24 @@ def __call__(self, circuit: Circuit, initial_layout: dict):
             (:class:`qibo.models.circuit.Circuit`, dict): routed circuit and final layout.
         """
         self._preprocessing(circuit=circuit, initial_layout=initial_layout)
+        self._saved_circuit = deepcopy(self.circuit)
```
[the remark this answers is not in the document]
csookim: `self.circuit` is a CircuitMap, so the `copy()`cannot be used.

tests/test_transpiler_router.py
csookim: Thanks. Please review the update.

src/qibo/transpiler/router.py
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: Ah yeah sorry, then better to first convert with `physical_to_logical` and then build the pairs in order to avoid converting the same element twice:
``` python
shortest_path = [self.circuit.physical_to_logical(q) for q in shortest_path[:-1]]
swaps = zip(shortest_path[:-1], shortest_path[1:])
```

tests/test_transpiler_router.py
```diff
@@ -203,15 +202,22 @@ def test_sabre_looping():
 
     placer = Trivial(connectivity=chip)
     initial_layout = placer(loop_circ)
-    router = Sabre(connectivity=chip)
-    routed_circuit, final_layout = router(loop_circ, initial_layout=initial_layout)
+    router_old = Sabre(connectivity=chip, swap_threshold=np.inf)  # Old
+    router_new = Sabre(connectivity=chip)  # New
```
BrunoLiegiBastonLiegi, line 206: Thanks, that's exactly what I meant, I would only change the names because `router_old` and `router_new` are meaningful names for us that we have followed the discussion, but in the future some other people might get confused. I would suggest something like, `router_no_threshold` (for the old one) and `router` for the new one, or something similar.

BrunoLiegiBastonLiegi (approved):
Thanks again for fixing this @csookim, it is all good from my side. As soon as all the tests finishes I believe we can merge and close this.

csookim:
@Simone-Bordoni @BrunoLiegiBastonLiegi Thanks for the review. Will this be merged at the qibo meeting, or can I merge it myself?

BrunoLiegiBastonLiegi:
I don't think we need to wait, but just to be sure we can ask @scarrazza.

Simone-Bordoni:
> @Simone-Bordoni @BrunoLiegiBastonLiegi Thanks for the review. Will this be merged at the qibo meeting, or can I merge it myself?

Please wait until tomorrow.
I would like to have a Quick Look after the changes.

tests/test_transpiler_router.py
```diff
@@ -186,6 +186,73 @@ def test_routing_with_measurements():
...
+    router_no_threshold = Sabre(
+        connectivity=chip, swap_threshold=np.inf
+    )  # Without reset
+    router_threshold = Sabre(connectivity=chip)  # With reset
+    routed_no_threshold, _ = router_no_threshold(
+        loop_circ, initial_layout=initial_layout
+    )
+    routed_threshold, _ = router_threshold(loop_circ, initial_layout=initial_layout)
+
+    def count_swaps(circuit):
```
Simone-Bordoni, line 214: You can use router.added_swaps() that returns the number of swaps added during the process without defining this function

Simone-Bordoni (changes requested):

tests/test_transpiler_router.py
```diff
@@ -186,6 +186,73 @@ def test_routing_with_measurements():
...
+        count = 0
+        for gate in circuit.queue:
+            if isinstance(gate, gates.SWAP):
+                count += 1
+        return count
+
+    count_no_threshold = count_swaps(routed_no_threshold)
+    count_threshold = count_swaps(routed_threshold)
+
+    assert count_no_threshold > count_threshold
```
Simone-Bordoni, line 224: Can you also add an “assert_circuit_equivalence” see other tests for reference

Simone-Bordoni (approved):

PR #1408, opened 2024-07-31
Router sabre

csookim:
This PR is an addition to #1398.

Since `deepcopy()` makes the compiler slower, it needs to be changed.

[code](https://github.com/qiboteam/qibo/blob/9db9b666f67563122a5ab716ae689bfe579d333f/src/qibo/transpiler/router.py#L901)

Line 901 takes over 30% of the routing time.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/transpiler/router.py
+++ b/src/qibo/transpiler/router.py
@@ -659,8 +659,9 @@ class Sabre(Router):
         self.circuit = None
         self._memory_map = None
         self._final_measurements = None
-        self._temporary_added_swaps = 0
-        self._saved_circuit = None
+        # self._temporary_added_swaps = 0
+        # self._saved_circuit = None
+        self._temp_added_swaps = []
         random.seed(seed)
 
     def __call__(self, circuit: Circuit, initial_layout: dict):
@@ -689,7 +690,11 @@ class Sabre(Router):
             if (
                 self._temporary_added_swaps > self.swap_threshold * longest_path
             ):  # threshold is arbitrary
-                self.circuit = deepcopy(self._saved_circuit)
+                # self.circuit = deepcopy(self._saved_circuit)
+                while self._temp_added_swaps:
+                    swap = self._temp_added_swaps.pop()
+                    self.circuit.update(swap) # -1
+                self._temp_added_swaps = []
                 self._shortest_path_routing()
 
         circuit_kwargs = circuit.init_kwargs
@@ -799,8 +804,10 @@ class Sabre(Router):
 
         for qubit in self.circuit.logical_to_physical(best_candidate, index=True):
             self._delta_register[qubit] += self.delta
-        self.circuit.update(best_candidate)
-        self._temporary_added_swaps += 1
+        # self.circuit.update(best_candidate)
+        # self._temporary_added_swaps += 1
+        self._temp_added_swaps.append(best_candidate)
+
 
     def _compute_cost(self, candidate: int):
         """Compute the cost associated to a possible SWAP candidate."""
@@ -897,8 +904,9 @@ class Sabre(Router):
         self._update_front_layer()
         self._memory_map = []
         self._delta_register = [1.0 for _ in self._delta_register]
-        self._temporary_added_swaps = 0
-        self._saved_circuit = deepcopy(self.circuit)
+        # self._temporary_added_swaps = 0
+        # self._saved_circuit = deepcopy(self.circuit)
+        self._temp_added_swaps = []
 
     def _shortest_path_routing(self):
         """Route a gate in the front layer using the shortest path. This method is executed when the standard SABRE fails to find an optimized solution.
--- a/src/qibo/transpiler/router.py
+++ b/src/qibo/transpiler/router.py
@@ -8,7 +8,7 @@ import numpy as np
 from qibo import gates
 from qibo.config import log, raise_error
 from qibo.models import Circuit
-from qibo.transpiler._exceptions import ConnectivityError
+from qibo.transpiler._exceptions import ConnectivityError, TranspilerPipelineError
 from qibo.transpiler.abstract import Router
 from qibo.transpiler.blocks import Block, CircuitBlocks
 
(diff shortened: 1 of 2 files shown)


csookim:
### Previous approach

- Save the circuit after the gate is routed. 

```py
self._temporary_added_swaps: int
self._saved_circuit: CircuitMap
```

https://github.com/qiboteam/qibo/blob/9db9b666f67563122a5ab716ae689bfe579d333f/src/qibo/transpiler/router.py#L900-L901

- Revert to the saved circuit when the number of SWAPs exceeds the threshold.

https://github.com/qiboteam/qibo/blob/9db9b666f67563122a5ab716ae689bfe579d333f/src/qibo/transpiler/router.py#L689-L693

### Updated approach

- Save the sequence of SWAPs. 

```py
self._temp_added_swaps: list
```

https://github.com/qiboteam/qibo/blob/a7a1c62850bd4e720a3cc0f90cc39df0cdc44519/src/qibo/transpiler/router.py#L817-L818

- Undo all of the temporary SWAPs when the number of SWAPs exceeds the threshold.

https://github.com/qiboteam/qibo/blob/a7a1c62850bd4e720a3cc0f90cc39df0cdc44519/src/qibo/transpiler/router.py#L701-L708

### Discussion

- The undo functionality is currently implemented in `CircuitMap.update()`. Would it be better to create a new function `undo()` for this purpose?
- `TranspilerPipelineError` is raised when the gate to be undone doesn't match the last added SWAP gate. Is there a more suitable error type for this situation?

Simone-Bordoni:
Thank you for this work, the code looks nice.
Regarding the two points you asked in the discussion:
- For the undo I think it would be better to have a separate function.
- Regarding the error type I think it would be better to raise ConnectivityError. In my Idea Connectivity error is raised every time there is some problem during routing while TranspilerPipelineError regards an error occurring in defining the pipeline for example if you forget to define the Placer.

Simone-Bordoni (approved):

csookim:
@Simone-Bordoni , I have a question about the `def test_circuit_map()` function.

https://github.com/qiboteam/qibo/blob/a2f74249c40e1d76eada8b8b795b5155e1ad6374/tests/test_transpiler_router.py#L264

The function includes the following code:
```py
...
initial_layout = {"q0": 2, "q1": 0, "q2": 1, "q3": 3}
...
circuit_map.update((0, 2)) # first swap
...
circuit_map.update((1, 2)) # second swap
assert routed_circuit.queue[5].qubits == (2, 0)    # routed_circuit.queue[5] is the second swap
```

After the first swap, the layout should be converted to:
```py
{"q0": 0, "q1": 2, "q2": 1, "q3": 3}
```
since `update()` operates on logical qubits. This would be equivalent to swapping physical qubits q0 and q1.

After the second swap, the swap should be on logical qubits (1, 2). So, the layout will be changed to:
```py
{"q0": 0, "q1": 1, "q2": 2, "q3": 3}
```

Does this mean that the qubits affected by the swap are q2 and q1?
```py
assert routed_circuit.queue[5].qubits == (2, 1)
```

Since `routed_circuit.queue[i].qubits` contains physical qubits, I think the qubits should be (2, 1). Could you help me understand if there's something I’m misunderstanding here?

Thank you for your time.

src/qibo/transpiler/router.py
```diff
@@ -257,6 +257,32 @@ def update(self, swap: tuple):
...
+                ConnectivityError,
+                "The last block does not match the swap to undo.",
+            )
+        self._routed_blocks.remove_block(last_swap_block)
+        self._swaps -= 1
+
+        idx_0, idx_1 = self._circuit_logical.index(
+            swap[0]
+        ), self._circuit_logical.index(swap[1])
+        self._circuit_logical[idx_0], self._circuit_logical[idx_1] = swap[1], swap[0]
```
BrunoLiegiBastonLiegi, line 284: This looks ok code-wise, the only thing that seems a bit weird to me is that the `undo` requires for you to pass the `swap` you wish to undo. I mean, since in principle you undo the last one, you wouldn't need to pass explicitely the swap, but just look for the last one added in a routed block, right? And, indeed, here `swap` is only used to check that the last swap you find is the one you expected.

csookim: I used `swap` to double-check if the undo function works correctly. I'll remove the checking step.

Also, could you take a look at the question I posted above?

BrunoLiegiBastonLiegi: Sorry for the delay. Double checking could always be useful, indeed, but, if you want to keep it, I wouldn't make it mandatory at least. In the sense that you could have the `swap=None` argument as optional. In any case, this check should be covered in the tests in principle. 
Regarding the question above, instead, I had a look at it last week. Now, the best person to answer this is probably @Simone-Bordoni, but I believe he is currently on a leave. Thus, I will provide you with my guess, which however may be incorrect. By looking at the `update` function https://github.com/qiboteam/qibo/blob/a2f74249c40e1d76eada8b8b795b5155e1ad6374/src/qibo/transpiler/router.py#L241
it seems that you update the `circuit_logical` mapping, not the physical one (`physical_logical`), meaning that in the layout dict, you do not swap the values, but rather the keys.
Therefore you start from:
```
{"q0": 0, "q1": 2, "q2": 1, "q3": 3}
```
after the `(0, 2)` swap you get:
```
{"q2": 0, "q1": 2, "q0": 1, "q3": 3}
```
and after `(1, 2)`:
```
{"q1": 0, "q2": 2, "q0": 1, "q3": 3}
```
This means that you end up with `circuit_logical = [1, 2, 0, 3]`, which is indeed checked in the test:
``` python
assert circuit_map._circuit_logical == [1, 2, 0, 3]
```
I hope this helped, assuming it is correct. In general, though, this small review of the transpiler code that I did to find the answer, made me think that probably, both, the naming scheme could be improved to make it clearer, and some further documentation should be added to the docstrings. Otherwise, everytime we go back to the code, we risk to waste time in re-understanding everything.

csookim: Thanks for your response. I noticed that while optimizing SABRE and a better naming scheme seems to be needed. I also found and implemented a more efficient way to manage the mapping. I think it would be best to wait until @Simone-Bordoni returns to discuss the new mapping scheme.

Simone-Bordoni: I am currently on leaves. If you want to implement a new mapping scheme you can do the PR and I will review it when I will be back.

BrunoLiegiBastonLiegi (approved):

PR #1415, opened 2024-08-07
Cast samples to int32

### LINKED ISSUE #1412: Complex numbers in samples and frequencies

scarrazza:
Attempt to fix #1412. In principle we could replace the default dtype in backends to something more useful (e.g. type(x)) however this is not trivial because this function was designed for state vectors. 
PS. I prefer to `dtype=np.uint8` however torch does not like it.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -24,13 +24,13 @@ files = [
 
 [[package]]
 name = "antlr4-python3-runtime"
-version = "4.13.1"
-description = "ANTLR 4.13.1 runtime for Python 3"
+version = "4.13.2"
+description = "ANTLR 4.13.2 runtime for Python 3"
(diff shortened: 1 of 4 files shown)


scarrazza:
Nevermind, I have just realized #1414.

PR #1417, opened 2024-08-08
Fix root cause of bug reported in PR #1412

renatomello:
The root cause of the issue raised in #1412 was a `cast` to `numpy` inside `NumpyBackend.samples_to_binary`. That is what was turning the subsequent `cast` into a requirement for `torch`. Without the former, the latter could be removed.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/backends/numpy.py
+++ b/src/qibo/backends/numpy.py
@@ -645,9 +645,7 @@ class NumpyBackend(Backend):
         return self.cast(shots, dtype=shots[0].dtype)
 
     def samples_to_binary(self, samples, nqubits):
-        ### This is faster just staying @ NumPy.
         qrange = np.arange(nqubits - 1, -1, -1, dtype=np.int32)
-        samples = self.to_numpy(samples)
         return np.mod(np.right_shift(samples[:, None], qrange), 2)
 
     def samples_to_decimal(self, samples, nqubits):
--- a/src/qibo/measurements.py
+++ b/src/qibo/measurements.py
@@ -167,11 +167,12 @@ class MeasurementResult:
             # calculate samples for the whole circuit so that
             # individual register samples are registered here
             self.circuit.final_state.samples()
+
         if binary:
             return self._samples
-        else:
-            qubits = self.measurement_gate.target_qubits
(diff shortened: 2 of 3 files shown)


src/qibo/backends/numpy.py
```diff
@@ -645,9 +645,7 @@ def aggregate_shots(self, shots):
         return self.cast(shots, dtype=shots[0].dtype)
 
     def samples_to_binary(self, samples, nqubits):
-        ### This is faster just staying @ NumPy.
         qrange = np.arange(nqubits - 1, -1, -1, dtype=np.int32)
-        samples = self.to_numpy(samples)
         return np.mod(np.right_shift(samples[:, None], qrange), 2)
```
alecandido, line 649: Aren't all the other operations in here still NumPy? Or is Torch automatically converting NumPy calls?

I expected `self.np.` usage, but instead there is `np.` everywhere...

renatomello: I tried `self.np` and it creates other issues because `torch.arange` and `right_shift` work differently

alecandido: But is `np.right_shift()` able to work on Torch tensors? (i.e. taking Torch tensors as input and returning them as output)

I see it could be vaguely possible, but it's still good to know, because I wouldn't have been confident to say it's actually able.

renatomello:
```python
In [6]: import numpy as np
   ...: import torch
   ...: 
   ...: x = torch.Tensor(torch.arange(10))
   ...: qrange = np.arange(10 - 1, -1, -1, dtype=np.int32)
   ...: type(np.right_shift(x[:, None], qrange))
Out[6]: torch.Tensor
```

alecandido: Nice, thanks :)

alecandido: It seems that by now this starts being reliable

https://numpy.org/doc/stable/user/basics.interoperability.html#operating-on-foreign-objects-without-converting
https://numpy.org/neps/nep-0018-array-function-protocol.html#nep18

alecandido (approved):

scarrazza (approved):

Issue #1377, opened 2024-07-02
Gate Set Tomography wishlist


BrunoLiegiBastonLiegi:
As the first version of GST is about to be merged, I thought it could have been useful to collect somewhere all the desired improvements and features to add on top of the very first implementation. 

- [ ] Allow for user-defined gates for state preparation, measurement bases and observable, while providing some default configuration

PR #1419, opened 2024-08-14
SABRE + CircuitMap Optimization

csookim:
To improve Sabre's performance, several implementations have been updated. I’ve added comments to the code and each comment number corresponds to the explanation numbers provided below. Please let me know if any clarifications are needed. Below is the code explanation.

Checklist:
- [x] Tests are passing.

After the code review:
- [x] Delete comments.
- [x] Update test functions.
- [ ] Optimize `_swap_candidates(), _check_execution(), _shortest_path_routing()` (remove `get_physical_qubits(block_num)`).
- [ ] Optimize `ShortestPaths`.


csookim:
# SABRE + CircuitMap Optimization

## Previous Implementation

- The mapping scheme could be clearer.
- Using `list.index()` in logical-to-physical and circuit-to-physical mapping slowed down the routing.
- Frequently using the function that finds a gate's target qubits by block number was inefficient.

## New Implementation

The code has been updated in three key areas:

### 1. Qubit Mapping

- SABRE operates using `connectivity`, `initial_layout`, and `circuit`.
- `initial_layout` is a dictionary where `"q{physical_qubit_number}": {logical_qubit_number}` pairs define the initial mapping.
- `physical_qubit_number` corresponds to the node number in `connectivity`.
- The `initial_layout` and `connectivity` can have arbitrary `physical_qubit_numbers`, when using restricted connectivity.

#### 1-1. Layout Relabel

- To simplify routing, `initial_layout` and `connectivity` are relabeled before routing begins.
- `physical_qubit_number` is reassigned to a range from `0` to `len(connectivity.nodes) - 1`.
- For example, the layout `{"q4":1, "q6":2, "q3":0}` might be relabeled to `{"q0":1, "q1":2, "q2":0}`.
- This relabeling map is saved, and the final layout is reverted to the original labeling when routing is complete.
- All operations in `Sabre` are performed using the updated `physical_qubit_number`.

https://github.com/qiboteam/qibo/blob/0111a6e49bc27b8e5b8102f5eb6615a093ade35c/src/qibo/transpiler/router.py#L731-L739

https://github.com/qiboteam/qibo/blob/0111a6e49bc27b8e5b8102f5eb6615a093ade35c/src/qibo/transpiler/router.py#L702-L707

#### 1-2 Physical-Logical Mapping

- Due to the layout labeling, there are two sets of numbers ranging from `0` to `len(connectivity.nodes()) - 1` in `initial_layout`.
- Physical to logical mapping (`p2l`) and logical to physical mapping (`l2p`) are implemented using two arrays: `self._p2l` and `self._l2p`.
- For example, with a mapping of `{"q0":1, "q1":2, "q2":0}`, `self._p2l` would be `[1, 2, 0]` and `self._l2p` would be `[2, 0, 1]`.
- Routing is performed using these p2l and l2p conversions.

https://github.com/qiboteam/qibo/blob/0111a6e49bc27b8e5b8102f5eb6615a093ade35c/src/qibo/transpiler/router.py#L183-L192

https://github.com/qiboteam/qibo/blob/0111a6e49bc27b8e5b8102f5eb6615a093ade35c/src/qibo/transpiler/router.py#L280-L284
 
### 2. Temporary CircuitMap

- Creating a `CircuitMap` includes unnecessary operations such as block decomposition, circuit object creation, and gate copying, which are not needed for calculating the cost of a candidate swap.
- Enable temporary `CircuitMap` using a `self._temporary` flag.
- For temporary `CircuitMap` instances used only to calculate swap costs, return immediately from `init` after initializing `self._l2p` and `self._p2l`.
- In these temporary `CircuitMap` instances, skip gate additions or removals in the `update()` method.

https://github.com/qiboteam/qibo/blob/0111a6e49bc27b8e5b8102f5eb6615a093ade35c/src/qibo/transpiler/router.py#L194-L196

https://github.com/qiboteam/qibo/blob/0111a6e49bc27b8e5b8102f5eb6615a093ade35c/src/qibo/transpiler/router.py#L843-L849

### 3. Blocks in the `front_layer`

- SABRE calculates the cost of a swap by summing the distances of successive gates.
- To calculate the distance of a successive gate, the target qubit numbers of the gate need to be given.
- Previously, only gate (block) numbers were stored in the `front_layer`.
- To find the target qubits, the code had to iterate through the list of gates using `search_by_index(block)`, which increased runtime.
- The new approach stores both the gate number and the target qubits when creating `self._dag`.

https://github.com/qiboteam/qibo/blob/0111a6e49bc27b8e5b8102f5eb6615a093ade35c/src/qibo/transpiler/router.py#L990-L995

- For example,

```py
self._dag = [(0, {'qubits': (0, 9), 'layer': 0}), (1, {'qubits': (5, 9), 'layer': 1}), (2, {'qubits': (2, 8), 'layer': 0})]
```

- The function `_get_dag_layer(n, qubits=True)` returns a list of target qubits for the gates in layer `n`.
- This allows retrieval of target qubits in $O(1)$ time.

https://github.com/qiboteam/qibo/blob/0111a6e49bc27b8e5b8102f5eb6615a093ade35c/src/qibo/transpiler/router.py#L807-L815

### 4. Additional Improvements

- Use shallow copy when copying integer lists.

## Evaluation

- Test: 100 random 50-CZ gates on 20 qubits.
- Qibo SABRE is 6x faster than before.
- The reason for the reduction in the number of SWAP gates is unknown.

```py
                     Mean   Std
Qiskit Sabre Time    0.0439 0.0075
Qibo Sabre Time      0.0625 0.0179
Qibo Sabre Old Time  0.4252 0.106

Qiskit Sabre Swaps   189.97 21.7295
Qibo Sabre Swaps     196.54 21.8712
Qibo Sabre Old Swaps 279.96 35.739
```

Screenshot 2024-08-14 at 20 49 23

src/qibo/transpiler/router.py
```diff
@@ -230,130 +251,80 @@ def routed_circuit(self, circuit_kwargs: Optional[dict] = None):
...
+            self._routed_blocks.add_block(
+                Block(qubits=swap_p, gates=[gates.SWAP(*swap_p)])
+            )
+            self._swaps += 1
+
+        # 1# update the bidirectional mapping
+        p1, p2 = swap_p
+        l1, l2 = swap_l
+        self._p2l[p1], self._p2l[p2] = l2, l1
+        self._l2p[l1], self._l2p[l2] = p2, p1
```
BrunoLiegiBastonLiegi, line 284: It might be worth making this a small helper function, e.g. `def _update_mappings(self, logical_pair, physical_pair)`

BrunoLiegiBastonLiegi (commented):
Thanks @csookim, I looked into the new `CircuitMap` and it seems better, I just have a handful of comments.

Simone-Bordoni (commented):
Thank you for this improvement, I had a first look at the changes and I agree on the corrections proposed by @BrunoLiegiBastonLiegi
Regarding the performance improvement (number of swaps) I beliave that refactoring the circuit_map probaly solved some bug that we were not able to identify.

src/qibo/transpiler/router.py
```diff
@@ -793,14 +833,31 @@ def _update_front_layer(self):
...
+        # 3# depend on the 'qubits' flag, return the block number or target qubits
+        # 3# return target qubits -> to avoid using get_physical_qubits(block_num)
+        if qubits:
+            layer_qubits = []
+            nodes = self._dag.nodes(data=True)
+            for node in nodes:
+                if node[1]["layer"] == n_layer:
+                    # return target qubits
+                    layer_qubits.append(node[1]["qubits"])
+            return layer_qubits
```
BrunoLiegiBastonLiegi, line 851:
```suggestion
            return [
                node[1]["qubits"]
                for node in self._dag.nodes(data=True) if node[1]["layer"] == n_layer 
            ]
```
this spares a couple of memory assignements

tests/test_transpiler_router.py
```diff
@@ -272,40 +272,40 @@ def test_circuit_map():
...
+    # circuit_map.execute_block(block_list.search_by_index(3))
+    # routed_circuit = circuit_map.routed_circuit()
+    # assert isinstance(routed_circuit.queue[6], gates.CZ)
+    # # circuit to logical map: [1,2,0,3]. initial map: {"q0": 2, "q1": 0, "q2": 1, "q3": 3}.
+    # assert routed_circuit.queue[6].qubits == (0, 1)  # initial circuit qubits (1,2)
+    # assert routed_circuit.queue[7].qubits == (2, 0)  # (0,1)
+    # assert routed_circuit.queue[8].qubits == (1, 3)  # (2,3)
+    # assert len(circuit_map.circuit_blocks()) == 0
+    # # test final layout
+    # assert circuit_map.final_layout() == {"q0": 1, "q1": 2, "q2": 0, "q3": 3}
```
BrunoLiegiBastonLiegi, line 308: If this test is no longer needed we should remove it.

tests/test_transpiler_router.py
```diff
@@ -483,17 +483,17 @@ def test_undo():
     # Two SWAP gates are added
     circuit_map.update((1, 2))
     circuit_map.update((2, 3))
-    assert circuit_map._circuit_logical == [0, 3, 1, 2]
+    # assert circuit_map._circuit_logical == [0, 3, 1, 2]
```
BrunoLiegiBastonLiegi, line 486: don't we want to double check the mapping? (similarly for the two below)

BrunoLiegiBastonLiegi (commented):
Thanks @csookim, this looks almost perfect to me now, I just made some further minor comments.

csookim:
@BrunoLiegiBastonLiegi Thank you for your review. However, if we consider using a custom string for the physical qubit it would result in wide code modifications, such as updates to test code and `placer`.

As far as I've heard, qibolab isn't ready for string-based qubit names. So, I suggest we complete this optimization PR with the `"q" + str(i)` physical qubit names. Then, we can open a new PR focused on establishing a unified qubit naming scheme for both qibo and qibolab. I'd like to hear your thoughts on this approach.

BrunoLiegiBastonLiegi:
Thanks @csookim for the insight about the `wire_names`, indeed as this PR is already adding various optimizations to the `CircuitMap` object, I am ok with merging this with the support for the default naming scheme only (`q0, q1, ...`), as long as we open an issue to remind us about it. So, in the end, up to you, if you feel this is something straightforward we could address it here directly, fine, otherwise we leave it for a future PR. In any case the support to custom qubit names is probably not a widely used feature anyway...

Simone-Bordoni:
> Custom qubit names can be used in Sabre. But the node names in the `connectivity graph` and the physical qubit names in the `layout` must be identical. Layout relabeling and recovery are done within `CircuitMap`.
> 
> ### `CircuitMap`
> * Receive the layout `{'A': 2, 'B': 0, 'C': 1}`.
> * Assign physical qubit numbers to the keys of the layout (`'A' -> 0`, `'B' -> 1`, `'C' -> 2`).
> * Relabel the connectivity graph based on this mapping.
> * Save the qubit names (keys) in `wire_names`.
> * Logical qubit numbers are the values of the layout.
> * Route using physical and logical qubit numbers.
> * Restore the original connectivity graph using the inverse of the mapping (`0 -> 'A'`, `1 -> 'B'`, `2 -> 'C'`).
> 
> Regarding the test files, some `placers` currently use node names that do not match the physical qubit names.
> 
> https://github.com/qiboteam/qibo/blob/21732ad59d31740e8c0641921abe98e49a1ea8c1/src/qibo/transpiler/placer.py#L203-L205
> 
> I believe some parts of the code in `Placer` need to be updated, and the test files should be updated after resolving issue #1429.

I agree on relabelling the graph nodes with the physical qubit names. However, in principle, the two objects should already have the same names. It is better to raise at least a warning when the names do not match, something like "Connectivity graph nodes names and physical qubit names dont match. Renaming connectivity nodes after physical qubits."

Simone-Bordoni:
> Thanks @csookim for the insight about the `wire_names`, indeed as this PR is already adding various optimizations to the `CircuitMap` object, I am ok with merging this with the support for the default naming scheme only (`q0, q1, ...`), as long as we open an issue to remind us about it. So, in the end, up to you, if you feel this is something straightforward we could address it here directly, fine, otherwise we leave it for a future PR. In any case the support to custom qubit names is probably not a widely used feature anyway...

At the moment it is not an important requirement but it may become fundamental for qibolab in the future when it will support hardware qubit names different from integers.

Simone-Bordoni:
@csookim there are a few tests failing in the pipeline and router. If you can fix them then i will proceed with the review

csookim:
> I agree on relabelling the graph nodes with the physical qubit names. However, in principle, the two objects should already have the same names. It is better to raise at least a warning when the names do not match, something like "Connectivity graph nodes names and physical qubit names dont match. Renaming connectivity nodes after physical qubits."

Relabeling is used for Sabre optimization. Sabre receives the layout as `{'A': 2, 'B': 0, 'C': 1}` and the connectivity graph as `A-B-C`. These names are both converted to numerical values for faster routing: `'A' -> 0, 'B' -> 1, 'C' -> 2`.

I believe that the name matching/check should already occur at the `Placer` level.

> @csookim there are a few tests failing in the pipeline and router. If you can fix them then i will proceed with the review

Yes, but to address this, the `Placer` should be updated first to enable same names for connectivity graph nodes and physical qubits.

I'll work on updating `Placer` and the test files.

@BrunoLiegiBastonLiegi Could you delete the reviews that you think have already been addressed?

src/qibo/transpiler/router.py
```diff
@@ -174,34 +174,57 @@ class CircuitMap:
     def __init__(
         self,
         initial_layout: dict,
-        circuit: Circuit,
+        circuit: Optional[Circuit] = None,
```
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: @csookim do you agree that the circuit is not really optional or am I missing something?

src/qibo/transpiler/router.py
```diff
@@ -174,34 +174,57 @@ class CircuitMap:
...
+        temp: Optional[bool] = False,  # 2# for temporary circuit
     ):
+        self.initial_layout = dict(sorted(initial_layout.items()))
+
+        # 1# bidirectional mapping
+        # 1# self._p2l: physical qubit number i -> logical qubit number _p2l[i]
+        # 1# self._l2p: logical qubit number i -> physical qubit number _l2p[i]
+        self._l2p, self._p2l = [0] * len(self.initial_layout), [0] * len(
+            self.initial_layout
+        )
```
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: the `#n` enumeration looks a bit confusing to me still, any reason we want to keep it?

src/qibo/transpiler/router.py
```diff
@@ -174,34 +174,57 @@ class CircuitMap:
...
+        for mapping in self.initial_layout.items():
+            physical_qubit, logical_qubit = int(mapping[0][1:]), mapping[1]
+            self._l2p[logical_qubit] = physical_qubit
+            self._p2l[physical_qubit] = logical_qubit
```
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: Any comment about this?
>looking at the docstring above for initial_layout, it seems that the keys are the logical indices and the values the physical ones, whereas here is reverted, am I missing something?
Furthermore this snippet here seems to be more or less replaced by the set_p2l function below called on initial_layout.

src/qibo/transpiler/router.py
```diff
@@ -735,11 +762,22 @@ def _preprocessing(self, circuit: Circuit, initial_layout: dict):
...
+
+        # 1# To simplify routing, some data is relabeled before routing begins.
+        # 1# physical qubit is reassigned to a range from 0 to len(self.connectivity.nodes) - 1.
+        node_mapping, new_initial_layout = {}, {}
+        for i, node in enumerate(self.connectivity.nodes):
+            node_mapping[node] = i
+            new_initial_layout["q" + str(i)] = initial_layout["q" + str(node)]
+
+        self.connectivity = nx.relabel_nodes(self.connectivity, node_mapping)
+        self.node_mapping_inv = {v: k for k, v in node_mapping.items()}
```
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: Except for the `wire_names`, I believe the case of the helper function is still open

tests/test_transpiler_router.py
BrunoLiegiBastonLiegi: @csookim should we remove this then?

tests/test_transpiler_router.py
BrunoLiegiBastonLiegi: @csookim should we remove this then?

BrunoLiegiBastonLiegi:
I just did, there are some open comments that were not addressed and did not receive any answer

BrunoLiegiBastonLiegi:
@csookim I don't see any new updates or answers to the points that are still open, could you please double check those and the tests that are failing? 
Regarding the custom `wires_names`, as agreed, we could disentagle that from this PR and address it in a separate one, just please open an issue about that as a reminder.

csookim:
@BrunoLiegiBastonLiegi  Could you close these reviews if resolved? If you need further clarification, please let me know.

### 1
> @csookim do you agree that the circuit is not really optional or am I missing something?

### 2

> Any comment about this?
> 
> > looking at the docstring above for initial_layout, it seems that the keys are the logical indices and the values the physical ones, whereas here is reverted, am I missing something?
> > Furthermore this snippet here seems to be more or less replaced by the set_p2l function below called on initial_layout.

### 3
> this is not super important, but still I would prefer to get rid of this ambiguity

BrunoLiegiBastonLiegi:
> @BrunoLiegiBastonLiegi Could you close these reviews if resolved? If you need further clarification, please let me know.

I cannot see any answer to those comments, did you answer them? Please double check, because if you click the button "start a review" instead of "Add a single comment", your answers are going to be pending until you finish the review, and they are going to be visible for you only.

csookim:
> > @BrunoLiegiBastonLiegi Could you close these reviews if resolved? If you need further clarification, please let me know.
> 
> I cannot see any answer to those comments, did you answer them? Please double check, because if you click the button "start a review" instead of "Add a single comment", your answers are going to be pending until you finish the review, and they are going to be visible for you only.

Oh, I didn't know I needed to submit my review. Sorry for the mistake.

src/qibo/transpiler/router.py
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: Ah I see, then I believe it would be better to move the circuit check right after the temporary one:
``` python
        if self._temporary:  # if temporary circuit, no need to store the blocks
            return
        elif circuit is None:
            raise_error(ValueError, "Circuit must be provided for a non temporary `CircuitMap`.")
        self._nqubits = circuit.nqubits  # number of qubits
```

tests/test_transpiler_router.py
BrunoLiegiBastonLiegi: Do you mind providing a short explanation of why it is the case to drop this test and the other checks below?

src/qibo/transpiler/router.py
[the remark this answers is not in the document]
csookim: I updated it.

csookim:
@BrunoLiegiBastonLiegi I've updated the code handling these issues and modified the test functions. 

> Ah I see, then I believe it would be better to move the circuit check right after the temporary one:
> 
> ```python
>         if self._temporary:  # if temporary circuit, no need to store the blocks
>             return
>         elif circuit is None:
>             raise_error(ValueError, "Circuit must be provided for a non temporary `CircuitMap`.")
>         self._nqubits = circuit.nqubits  # number of qubits
> ```

> Except for the `wire_names`, I believe the case of the helper function is still open

Currently, the `Router` operates on physical qubits `qi` in the layout, which correspond to `i` in the connectivity nodes. Once all of your reviews are resolved, I will remove the comments.

BrunoLiegiBastonLiegi (approved):
Thanks @csookim this looks ready to me know, could you just add a reminder to #1377 about the `wire_names`?

csookim:
@Simone-Bordoni  Please review this PR. Thanks.

Simone-Bordoni (approved):

geexie:
> # SABRE + CircuitMap Optimization
> ## Previous Implementation
> * The mapping scheme could be clearer.
> * Using `list.index()` in logical-to-physical and circuit-to-physical mapping slowed down the routing.
> * Frequently using the function that finds a gate's target qubits by block number was inefficient.
> 
> ## New Implementation
> The code has been updated in three key areas:
> 
> ### 1. Qubit Mapping
> * SABRE operates using `connectivity`, `initial_layout`, and `circuit`.
> * `initial_layout` is a dictionary where `"q{physical_qubit_number}": {logical_qubit_number}` pairs define the initial mapping.
> * `physical_qubit_number` corresponds to the node number in `connectivity`.
> * The `initial_layout` and `connectivity` can have arbitrary `physical_qubit_numbers`, when using restricted connectivity.
> 
> #### 1-1. Layout Relabel
> * To simplify routing, `initial_layout` and `connectivity` are relabeled before routing begins.
> * `physical_qubit_number` is reassigned to a range from `0` to `len(connectivity.nodes) - 1`.
> * For example, the layout `{"q4":1, "q6":2, "q3":0}` might be relabeled to `{"q0":1, "q1":2, "q2":0}`.
> * This relabeling map is saved, and the final layout is reverted to the original labeling when routing is complete.
> * All operations in `Sabre` are performed using the updated `physical_qubit_number`.
> 
> https://github.com/qiboteam/qibo/blob/0111a6e49bc27b8e5b8102f5eb6615a093ade35c/src/qibo/transpiler/router.py#L731-L739
> 
> https://github.com/qiboteam/qibo/blob/0111a6e49bc27b8e5b8102f5eb6615a093ade35c/src/qibo/transpiler/router.py#L702-L707
> 
> #### 1-2 Physical-Logical Mapping
> * Due to the layout labeling, there are two sets of numbers ranging from `0` to `len(connectivity.nodes()) - 1` in `initial_layout`.
> * Physical to logical mapping (`p2l`) and logical to physical mapping (`l2p`) are implemented using two arrays: `self._p2l` and `self._l2p`.
> * For example, with a mapping of `{"q0":1, "q1":2, "q2":0}`, `self._p2l` would be `[1, 2, 0]` and `self._l2p` would be `[2, 0, 1]`.
> * Routing is performed using these p2l and l2p conversions.
> 
> https://github.com/qiboteam/qibo/blob/0111a6e49bc27b8e5b8102f5eb6615a093ade35c/src/qibo/transpiler/router.py#L183-L192
> 
> https://github.com/qiboteam/qibo/blob/0111a6e49bc27b8e5b8102f5eb6615a093ade35c/src/qibo/transpiler/router.py#L280-L284
> 
> ### 2. Temporary CircuitMap
> * Creating a `CircuitMap` includes unnecessary operations such as block decomposition, circuit object creation, and gate copying, which are not needed for calculating the cost of a candidate swap.
> * Enable temporary `CircuitMap` using a `self._temporary` flag.
> * For temporary `CircuitMap` instances used only to calculate swap costs, return immediately from `init` after initializing `self._l2p` and `self._p2l`.
> * In these temporary `CircuitMap` instances, skip gate additions or removals in the `update()` method.
> 
> https://github.com/qiboteam/qibo/blob/0111a6e49bc27b8e5b8102f5eb6615a093ade35c/src/qibo/transpiler/router.py#L194-L196
> 
> https://github.com/qiboteam/qibo/blob/0111a6e49bc27b8e5b8102f5eb6615a093ade35c/src/qibo/transpiler/router.py#L843-L849
> 
> ### 3. Blocks in the `front_layer`
> * SABRE calculates the cost of a swap by summing the distances of successive gates.
> * To calculate the distance of a successive gate, the target qubit numbers of the gate need to be given.
> * Previously, only gate (block) numbers were stored in the `front_layer`.
> * To find the target qubits, the code had to iterate through the list of gates using `search_by_index(block)`, which increased runtime.
> * The new approach stores both the gate number and the target qubits when creating `self._dag`.
> 
> https://github.com/qiboteam/qibo/blob/0111a6e49bc27b8e5b8102f5eb6615a093ade35c/src/qibo/transpiler/router.py#L990-L995
> 
> * For example,
> 
> ```python
> self._dag = [(0, {'qubits': (0, 9), 'layer': 0}), (1, {'qubits': (5, 9), 'layer': 1}), (2, {'qubits': (2, 8), 'layer': 0})]
> ```
> 
>  * The function `_get_dag_layer(n, qubits=True)` returns a list of target qubits for the gates in layer `n`.
> * This allows retrieval of target qubits in 
>     O
>     (
>     1
>     )
>    time.
> 
> https://github.com/qiboteam/qibo/blob/0111a6e49bc27b8e5b8102f5eb6615a093ade35c/src/qibo/transpiler/router.py#L807-L815
> 
> ### 4. Additional Improvements
> * Use shallow copy when copying integer lists.
> 
> ## Evaluation
> * Test: 100 random 50-CZ gates on 20 qubits.
> * Line connectivity
> * Qibo SABRE is 6x faster than before.
> * The reason for the reduction in the number of SWAP gates is unknown.
> 
> ```python
>                      Mean   Std
> Qiskit Sabre Time    0.0439 0.0075
> Qibo Sabre Time      0.0625 0.0179
> Qibo Sabre Old Time  0.4252 0.106
> 
> Qiskit Sabre Swaps   189.97 21.7295
> Qibo Sabre Swaps     196.54 21.8712
> Qibo Sabre Old Swaps 279.96 35.739
> ```
> 
> Screenshot 2024-08-14 at 20 49 23

Hello, @csookim! Amazing work! I'm currently experimenting with SABRE in Qibo. May I know what benchmark was used to make the plot above? Do you have some data about circuit parameters as well like number of extra gates, etc? Which gate set do you use? Do you transpile for some specific architecture or Cliffird+T dataset?

I'm also wondering what transpilation pipeline setup is used for best SABRE performance? I use 
```python
    hw_gates = [gates.GPI2, gates.RZ, gates.Z, gates.CZ]
    natives = NativeGates(0).from_gatelist(hw_gates)
    custom_passes2 = [Preprocessing(), Unroller(native_gates=natives), Sabre(), Unroller(native_gates=natives)]
    return Passes(connectivity=nx.Graph(connectivity), passes=custom_passes2)
```
where connectivity map is one for IQM20

Thank you it advance!

BrunoLiegiBastonLiegi:
Hey @geexie, this is a merged pull request already, thus I doubt you will receive much attention here,. If you have any problem or wish to discuss about something, I'd say it is better for you to open a new issue or a new discussion.

geexie:
> Hey @geexie, this is a merged pull request already, thus I doubt you will receive much attention here,. If you have any problem or wish to discuss about something, I'd say it is better for you to open a new issue or a new discussion.

Hey @BrunoLiegiBastonLiegi! I followed your suggestion and opened a [discussion](https://github.com/qiboteam/qibo/discussions/1685#discussion-8562672). Thank you🙏🏻

PR #1420, opened 2024-08-14
Tuple-related type hints

alecandido:
A quite trivial fix, just because I noticed it while working on Qibolab.

Tuples are immutable containers, thus the plain type hint is dedicated to the fixed size ones. To return a variable size tuple, `tuple[MyType, ...]` is required.
`tuple[MyType]` is describing a 1-element tuple, whose element has type `MyType`.

Conversely, lists and dictionaries are mutable, and their basic and only type hints are not imposing any restriction on the length.


renatomello:
These changes would have to be applied throughout the entire module, no?

alecandido:
> These changes would have to be applied throughout the entire module, no?

Yes, and I'm planning to do a consistent review of types at some point (there are further improvements/fixes to add). But these methods are very common, so I decided to fix these immediately, instead of delaying them with everything else.

Moreover, I checked that most of the usage of `Tuple[]` are consistent, since they refer to fixed length tuples. There might be some further occurrences just in the `quantum_info` module.

alecandido:
@renatomello: for me, this PR is ready as it is. But if you want to do something else, we can discuss (though I will not commit soon to much more, that's why I'd prefer to open a new one later on).

renatomello (approved):

BrunoLiegiBastonLiegi:
I wasn't aware of `Tuple[type, ...]` syntax, thanks for pointing out.

BrunoLiegiBastonLiegi (approved):

Issue #1421, opened 2024-08-19
CNOT gate decomposition should be implemented


csookim:
@sorewachigauyo Paul mentioned that an unroller is needed to decompose the circuit using `CNOT`, `GPI2`, `RZ`, `Z`, and `M`. To achieve this, `CNOT` should be included as a native gate.

Here are the tasks to consider:

- [ ] Add `CNOT` as an attribute of `NativeGates`.
- [ ] Implement `cnot_dec` in `decompositions.py`.
- [ ] Modify the logic of `_translate_two_qubit_gates()` in `unroller.py`.

I believe `cnot_dec` can be implemented based on the existing `cz_dec`.

alecandido:
Pay attention: often CZ, iSWAP, and CNOT are alternative to each other.

`_translate_two_qubit_gates()` is now hardcoding that the interactions are only carried on by CZ and iSWAP
https://github.com/qiboteam/qibo/blob/8fb9af92c8ee4049d6575571ab33edd756196f2d/src/qibo/transpiler/unroller.py#L205-L207
and this is for sure not sufficient.

But it is not even that modular: it would be better to label interacting gates where they are defined (i.e. in the `NativeGates` class, by using non-trivial values), and then associate different actions (so, functions) to the specified set of interacting gates.
If you can make such that a single unroller can work for all the possible native two-qubit gates combinations, so much the better (the amount of combinations explodes as a power set, though it's unlikely to have much more than these three gates in the short term - but that was also the reasoning behind hardcoding the previous two).

csookim:
@alecandido Thank you for your reply. Since `CNOT` is needed for the upcoming chip test at CQT, I made minimal updates to enable using `CNOT` as a native gate for a simple circuit consisting of `CZ`, `CNOT`, `SWAP`, and single-qubit gates. Is this ok for the new PR?

alecandido:
> @alecandido Thank you for your reply. Since `CNOT` is needed for the upcoming chip test at CQT, I made minimal updates to enable using `CNOT` as a native gate for a simple circuit consisting of `CZ`, `CNOT`, `SWAP`, and 1q gates. Is this ok for the new PR?

If you need that urgently, for sure :)

However, we should schedule some time also for taking proper care of the unroller. If not immediately, whenever we can.

PR #1422, opened 2024-08-19
Add `gates.CNOT` as a native gate in gate decomposition

csookim:
This covers issue #1421.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/transpiler/unroller.py
+++ b/src/qibo/transpiler/unroller.py
@@ -4,7 +4,7 @@ from qibo import gates
 from qibo.config import raise_error
 from qibo.models import Circuit
 from qibo.transpiler._exceptions import DecompositionError
-from qibo.transpiler.decompositions import cz_dec, gpi2_dec, iswap_dec, opt_dec, u3_dec
+from qibo.transpiler.decompositions import cz_dec, gpi2_dec, iswap_dec, opt_dec, u3_dec, cnot_dec_temp
 
 
 class NativeGates(Flag):
@@ -32,6 +32,7 @@ class NativeGates(Flag):
     U3 = auto()
     CZ = auto()
     iSWAP = auto()
+    CNOT = auto()       # For testing purposes
 
     @classmethod
     def default(cls):
@@ -239,6 +240,12 @@ def _translate_two_qubit_gates(gate: gates.Gate, native_gates: NativeGates):
             for g_translated in translate_gate(g, native_gates=native_gates):
                 iswap_decomposed.append(g_translated)
         return iswap_decomposed
+    
+    # For testing purposes
+    # No CZ, iSWAP gates in the native gate set
+    # Decompose CNOT, CZ, SWAP gates into CNOT gates
+    if native_gates & NativeGates.CNOT:       
+        return cnot_dec_temp(gate)
 
     raise_error(
         DecompositionError, "Use only CZ and/or iSWAP as native gates"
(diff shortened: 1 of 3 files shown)


csookim:
### Example

```py
circ = Circuit(2)
circ.add(gates.H(0))
circ.add(gates.CNOT(0, 1))
circ.add(gates.SWAP(0, 1))
circ.add(gates.CZ(0, 1))
circ.add(gates.M(0, 1))

gate_list = [gates.GPI2, gates.RZ, gates.Z, gates.M, gates.CNOT]
native_gates = NativeGates(0).from_gatelist(gate_list)

custom_pipeline = Passes([Unroller(native_gates=native_gates)])
unrolled_circuit, _ = custom_pipeline(circ)

print(unrolled_circuit.draw(line_wrap=100))
```

### Output
```py
q0: ─Z─GPI2─o─o─X─o────────o────────M─
q1: ────────X─X─o─X─Z─GPI2─X─Z─GPI2─M─
```

alecandido (approved):
As said in the issue, the unroller should be entirely reviewed, since there are many more combinations that now are not accounted for.

In any case, they may also be unrealistic, and support is not needed in the short term.

For the time being, just fix the docstrings, and it should be ready to go.

src/qibo/transpiler/unroller.py
```diff
@@ -32,6 +39,7 @@ class NativeGates(Flag):
     U3 = auto()
     CZ = auto()
     iSWAP = auto()
+    CNOT = auto()  # For testing purposes
```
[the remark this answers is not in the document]
csookim: Thanks for your review. I have added error msgs and a docstring. Please check them.

alecandido (approved):
@csookim, you didn't have to ask again for a review: I already approved, and the modifications I asked were so minor that I trusted you on how to implement them :)

alecandido:
Btw, why are you deleting Codecov comments? Just out of curiosity

image

alecandido:
> I deleted it to avoid confusion, as the removed one was the coverage of the previous code. Will it be updated automatically?

Indeed, if you leave it there it will edit the same comment, instead of adding a new one :)

src/qibo/transpiler/unroller.py
```diff
@@ -240,6 +249,13 @@ def _translate_two_qubit_gates(gate: gates.Gate, native_gates: NativeGates):
                 iswap_decomposed.append(g_translated)
         return iswap_decomposed
 
+    # For testing purposes
+    # No CZ, iSWAP gates in the native gate set
+    # Decompose CNOT, CZ, SWAP gates into CNOT gates
+    if native_gates & NativeGates.CNOT:
+        return cnot_dec_temp(gate)
```
Simone-Bordoni, line 256: You can make this decomposition work for every gate by first decomposing into CZ and then decomposing CZ into CNOT. This is how it works for iSWAP decomposition (line 238-250).

csookim: Yes, but it adds two H gates next to the CNOT and it requires optimization of 1q gates. Since Paul will use simple circuits for testing, I only added basic decompositions. As @alecandido mentioned, it would be better to entirely review the unroller.

csookim: Yes, but it adds two H gates next to the CNOT and it requires optimization (fusion) of 1q gates. Since Paul will use simple circuits for testing, I only added basic decompositions. As @alecandido mentioned, it would be better to entirely review the unroller.

Simone-Bordoni: It should be a realy easy check to remove two successive H gates, however if you need it only for simple applications it is fine

Simone-Bordoni (approved):

PR #1430, opened 2024-08-28
Fix default backend loading

alecandido:
Extract just the fix from #1425 (but not the complex and failing test).


alecandido:
@scarrazza this is the PR with only the commits fixing the test. I left the other commits in the original #1425, and converted it to draft.

scarrazza:
Thanks @alecandido

scarrazza (approved):

Issue #1426, opened 2024-08-23
Retrieve nested samples during `MeasurementOutcome` dump


scarrazza:
I am trying to dump the circuit results object and reload. 
The qibolab backend is not storing samples therefore the code crashes.

Code example using the latest `module load qibo` and `sbatch -p qw5q_platinum`):
```python
import qibo
c = qibo.Circuit(5)
c.add(qibo.gates.M(0))
r = c(nshots=100)
r.dump("./output.npy")
```
Then, while loading:
```python
import qibo
qibo.result.load_result("output.npy").samples()
```
we get:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/nfs/users/test.test/qibo/src/qibo/result.py", line 341, in samples
    samples = self.backend.sample_shots(self._probs, self.nshots)
  File "/nfs/users/test.test/qibo/src/qibo/backends/numpy.py", line 650, in sample_shots
    range(len(probabilities)), size=nshots, p=probabilities
TypeError: object of type 'NoneType' has no len()
```

scarrazza:
@BrunoLiegiBastonLiegi may I ask you to have a look at this?

alecandido:
Sorry @scarrazza to have stopped here: I actually started implementing, but then the proposed solution didn't seem that good.

The `MeasurementOutcome.to_dict()` method was meant to serialize the object, in a way that it could be recreated.
For this reason, is also serializing the gates:
https://github.com/qiboteam/qibo/blob/386fe25c69e54211b4816702b5f0b619d645b35c/src/qibo/result.py#L417

However, the `M` gate is not customizing the `.raw` property (and thus `.to_json()` method), relying on the general `Gate` behavior, which is only retaining the arguments needed to construct:
https://github.com/qiboteam/qibo/blob/386fe25c69e54211b4816702b5f0b619d645b35c/src/qibo/gates/abstract.py#L75-L89
(you can read the list of the selected arguments in the `REQUIRED_*` constants on top) skipping the `M.result` attribute
https://github.com/qiboteam/qibo/blob/386fe25c69e54211b4816702b5f0b619d645b35c/src/qibo/gates/measurements.py#L59

But on its turn, `MeasurementResult` has no available way to be serialized (which it could be defined).

The `._samples` to `.samples()` above is a bit destructive, because the object reloaded would not be the exact same (you will get the samples separately from the gate, and the gate would still be empty, though there).
[...]

PR #1431, opened 2024-08-29
Serialization of measurements

BrunoLiegiBastonLiegi:
This addresses the bug #1426 by implementing a proper serialization of measurement gates and their result.
Furthermore, the possibility to specify the `basis` of measurement gates with strings (for instance `"Z"` or `["X", "Z", "Y"]`) has been added.
I also made some minor modifications to `MeasurementResult`, and I would also like to remove the `circuit` attribute from it. At the moment it is only used to generate the samples when they are not available. However, as I see it, the `MeasurementResult` should always be associated with the samples.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/gates/measurements.py
+++ b/src/qibo/gates/measurements.py
@@ -115,7 +115,7 @@ class M(Gate):
         gate.
         """
         encoded_simple = super().raw
-        encoded_simple.update(self.result.raw)
+        encoded_simple.update({"measurement_result": self.result.raw})
         return encoded_simple
 
     @staticmethod
--- a/src/qibo/gates/measurements.py
+++ b/src/qibo/gates/measurements.py
@@ -1,5 +1,5 @@
 import json
-from typing import Dict, Optional, Tuple
+from typing import Dict, Optional, Tuple, Union
 
 from qibo import gates
 from qibo.config import raise_error
@@ -23,11 +23,13 @@ class M(Gate):
             performed. Can be used only for single shot measurements.
             If ``True`` the collapsed state vector is returned. If ``False``
             the measurement result is returned.
-        basis (:class:`qibo.gates.Gate`, list): Basis to measure.
-            Can be a qibo gate or a callable that accepts a qubit,
-            for example: ``lambda q: gates.RX(q, 0.2)``
-            or a list of these, if a different basis will be used for each
-            measurement qubit.
+        basis (:class:`qibo.gates.Gate`, str, list): Basis to measure.
+            Can be either:
+            - a qibo gate
+            - the string representing the gate
+            - a callable that accepts a qubit, for example: ``lambda q: gates.RX(q, 0.2)``
+            - a list of the above, if a different basis will be used for each
+              measurement qubit.
             Default is Z.
         p0 (dict): Optional bitflip probability map. Can be:
             A dictionary that maps each measured qubit to the probability
@@ -46,7 +48,7 @@ class M(Gate):
         *q,
         register_name: Optional[str] = None,
         collapse: bool = False,
-        basis: Gate = Z,
+        basis: Union[Gate, str] = Z,
         p0: Optional["ProbsType"] = None,
         p1: Optional["ProbsType"] = None,
     ):
@@ -61,16 +63,24 @@ class M(Gate):
         # relevant for experiments only
         self.pulses = None
         # saving basis for __repr__ ans save to file
+        to_gate = lambda x: getattr(gates, x) if isinstance(x, str) else x
         if not isinstance(basis, list):
-            self.basis_gates = len(q) * [basis]
+            self.basis_gates = len(q) * [to_gate(basis)]
+            basis = len(self.target_qubits) * [basis]
+        elif len(basis) != len(self.target_qubits):
+            raise_error(
+                ValueError,
+                f"Given basis list has length {len(basis)} while "
+                f"we are measuring {len(self.target_qubits)} qubits.",
+            )
         else:
-            self.basis_gates = basis
+            self.basis_gates = [to_gate(g) for g in basis]
 
         self.init_args = q
         self.init_kwargs = {
             "register_name": register_name,
             "collapse": collapse,
-            "basis": self.basis_gates,
+            "basis": [g.__name__ for g in self.basis_gates],
             "p0": p0,
             "p1": p1,
         }
@@ -89,16 +99,8 @@ class M(Gate):
 
         # list of gates that will be added to the circuit before the
         # measurement, in order to rotate to the given basis
-        if not isinstance(basis, list):
-            basis = len(self.target_qubits) * [basis]
-        elif len(basis) != len(self.target_qubits):
-            raise_error(
-                ValueError,
-                f"Given basis list has length {len(basis)} while "
-                f"we are measuring {len(self.target_qubits)} qubits.",
-            )
         self.basis = []
-        for qubit, basis_cls in zip(self.target_qubits, basis):
+        for qubit, basis_cls in zip(self.target_qubits, self.basis_gates):
             gate = basis_cls(qubit).basis_rotation()
             if gate is not None:
                 self.basis.append(gate)
@@ -113,9 +115,6 @@ class M(Gate):
         gate.
         """
         encoded_simple = super().raw
-        encoded_simple.pop("_control_qubits")
-        basis = [g.__name__ for g in encoded_simple["init_kwargs"]["basis"]]
-        encoded_simple["init_kwargs"]["basis"] = basis
         encoded_simple.update(self.result.raw)
         return encoded_simple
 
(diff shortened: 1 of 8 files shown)


src/qibo/measurements.py
```diff
@@ -96,6 +98,11 @@ def __repr__(self):
         nshots = self.nshots
         return f"MeasurementResult(qubits={qubits}, nshots={nshots})"
 
+    @property
+    def raw(self) -> dict:
+        samples = self._samples.tolist() if self.has_samples() else self._samples
+        return {"samples": samples}
```
alecandido, line 104: In this way, you're discarding all other attributes, including public ones (`.measurement_gate`, `.backend`, `.nshots`, `.circuit`).

Isn't it a bit poor for the serialization?

I understand that all the complex objects' references are hard to serialize (and, arguably, they should not be there in the first place...), but at least `.nshots` should be a simple object (not sure if it's worth to serialize the backend through its name...).

src/qibo/measurements.py
```diff
@@ -117,12 +124,12 @@ def add_shot_from_sample(self, sample):
     def has_samples(self):
         return self._samples is not None
 
-    def register_samples(self, samples, backend=None):
+    def register_samples(self, samples):
         """Register samples array to the ``MeasurementResult`` object."""
-        self._samples = samples
+        self._samples = self.backend.cast(samples, self.backend.np.int64)
```
alecandido, line 129: I would avoid casting if not strictly needed...

#1412 #1414 #1417

src/qibo/gates/measurements.py
```diff
@@ -88,20 +99,25 @@ def __init__(
...
+    def raw(self) -> dict:
+        """Serialize to dictionary.
+
+        The values used in the serialization should be compatible with a
+        JSON dump (or any other one supporting a minimal set of scalar
+        types). Though the specific implementation is up to the specific
+        gate.
+        """
+        encoded_simple = super().raw
+        encoded_simple.update(self.result.raw)
```
alecandido, line 118: Here you're "flattening" the dictionary, since what was part of `.result` is now serialized together with the other attributes of the measurement gate.

alecandido (commented):
Some suggestions, but they are not incredibly relevant, as the whole measurement storage is a bit convoluted (with many references everywhere), and a greater action should be taken to simplify it (not now, possibly as part of `qibo-core`).

The relevant part instead it's that a test is missing. It would be nice to have a test failing before this PR, and now working.

alecandido:
> Furthermore, the possibility to specify the `basis` of measurement gates with strings (for instance `"Z"` or `["X", "Z", "Y"]`) has been added.

Try to keep features' implementation as separate as possible (and separate from bug fixes).

src/qibo/measurements.py
BrunoLiegiBastonLiegi: Well, the `nshots` can be recovered from the `samples` and I could actually make it a property. The `backend` I am not really sure how relevant it is, because if you load a new `MeasurementResult` object (which I find unlikely in any case) you are probably better off selecting the backend that you are currently using rather than the one that was in use when the object was dumped. Regarding the `circuit`, on the other hand, I would say it is safe to remove, and I plan doing it if everyone agrees. Now, the problem with the `measurement_gate` is that the `M` gate and the `MeasurementResult` are nested, they both contain an instance of the other one. Therefore, at some point you have to drop something. In this case, I decided to simplify the `MeasurementResult` serialization in favor of a more thorough `M` serialization. Mainly since In my experience with qibo, I had the impression that the `MeasurementResult` object is rarely used directly, and more often you access it through the measurement.

To be honest, as I see it now, the `MeasurementResult` is basically just a wrapper to the samples, and I would even argue that its dependence on the measurement gate is not necessary. In the end, I believe that the `measurement_gate` attribute  is only used to find the qubits (`measurement_gate.target_qubits`), at that point we would be better off building it just starting from the qubit indices.

src/qibo/measurements.py
BrunoLiegiBastonLiegi: Yeah, I agree, the problem though is whether the samples passed to the `register_samples` method are not proper tensors of the `MeasurementResult.backend`. At that point, you either raise an error or cast.

src/qibo/gates/measurements.py
BrunoLiegiBastonLiegi: Yep, I agree on this one as well, but since at the moment the serialization of a `MeasurementResult` is just the `samples`, building the dict as:
```
{
...
...

"measurement_result": {"samples": [0,1,0,...]}
}
```
looked a bit superfluous.

BrunoLiegiBastonLiegi:
Yeah, let's say that rather than a new feature is a side effect of the fix in this case, but I fundamentally agree.

src/qibo/measurements.py
alecandido: > Well, the `nshots` can be recovered from the `samples` and I could actually make it a property.

Great, let's do that

> The `backend` I am not really sure how relevant it is, because if you load a new `MeasurementResult` object (which I find unlikely in any case) you are probably better off selecting the backend that you are currently using rather than the one that was in use when the object was dumped.

Then, I would argue that the same could be done even immediately, dropping the `backend` attribute as well.

> Now, the problem with the `measurement_gate` is that the `M` gate and the `MeasurementResult` are nested, they both contain an instance of the other one.

That's worse than nested (which would be fine): they are cyclic.

> To be honest, as I see it now, the `MeasurementResult` is basically just a wrapper to the samples, and I would even argue that its dependence on the measurement gate is not necessary. In the end, I believe that the `measurement_gate` attribute is only used to find the qubits (`measurement_gate.target_qubits`), at that point we would be better off building it just starting from the qubit indices.

I would argue that the `MeasurementResult` methods could be turned into a library of functions. If it's convenient to group them together in a class, let's make this class containing a single attribute: the `samples`. Everything else can be received by argument to its methods, if needed.

In any case, this is a separate improvement. I'd suggest making it an issue.

src/qibo/gates/measurements.py
alecandido: If that's how you will deserialize, that's even fine. That dictionary should not be consumed manually, but automatically. So, if you have a one attribute class, it should be serialized as a one item dictionary - and it is convenient to do that, since you will need to treat fewer special cases.

Otherwise, you can always consider dropping the one attribute class, and make its methods plain functions (or methods of a parent class, if there is a specific one that is always the only one wrapping that).

src/qibo/measurements.py
alecandido: > Yeah, I agree, the problem though is whether the samples passed to the `register_samples` method are not proper tensors of the `MeasurementResult.backend`. At that point, you either raise an error or cast.

It could be a good reason in favor of dropping `MeasurementResult.backend`...

BrunoLiegiBastonLiegi:
Ok, I should have addressed everything now, except the `MeasurementResult.circuit` attribute that is still there, but I would get rid of it if everyone agrees.

alecandido (approved):
The test is effective and passing.

Thanks @BrunoLiegiBastonLiegi!

marcorossi5:
Hello guys, I've installed qibo from this source branch and checked if the `qibo-webappdaemon` works smoothly.
Unfortunately we are having the following error raised by the following snippet:

```python
with open(circuit_path, "r") as f:
      raw = json.load(f)
c = qibo.Circuit.from_dict(raw)
r = c(nshots=settings.get("nshots"))
print(f"Results: {r.samples()}")
```

```text
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/nfs/users/tiiq/webappdaemon/src/daemon/slurm_circuit_computation.py", line 62, in <module>
    main(args.job_folder)
  File "/nfs/users/tiiq/webappdaemon/src/daemon/slurm_circuit_computation.py", line 55, in main
    print(f"Results: {r.samples()}")
  File "/nfs/users/tiiq/qibo/src/qibo/result.py", line 324, in samples
    qubits = self.measurement_gate.target_qubits
AttributeError: 'NoneType' object has no attribute 'target_qubits'
```

scarrazza:
Do you have a simpler example using only qibo primitives?

BrunoLiegiBastonLiegi:
The output on `sim` should be the correct one, which was updated in this PR. The other one seems to be the old serialization of the `M` gate (the one in main).

Similarly below, the CPU -> QPU loading seems to fail because the `M` object is not able to receive strings as basis arguments, which is an additional feature introduced here. Viceversa, the QPU -> CPU loading fails because the object serialized on the QPU doesn't include attributes that are required to load a measurement now, i.e. the serialization of the `MeasurementResult`. 

I would say that the CPU and QPU are using two different versions qibo:
- CPU -> this branch
- QPU -> main

marcorossi5:
Let me check again the environments. If the branch mismatch is the cause, then we are fine.

scarrazza:
I have just tried your code and it works, no crashes.

marcorossi5:
Yes, I agree. I was loading the wrong environment on the QPU, my bad.
I guess the test is passed then

PR #1434, opened 2024-09-04
Allow `Circuit.draw` to display string directly

renatomello:
This PR proposes enabling the `Circuit.draw` method to print the circuit without having to wrap it inside a `print()` call. Example:

```python
In [1]: from qibo import Circuit, gates, set_backend

In [2]: circuit = Circuit(2)

In [3]: circuit.draw()
q0: ─
q1: ─

In [4]: circuit.draw(output_string=True)
Out[4]: 'q0: ─\nq1: ─'

In [3]: print(circuit.draw(output_string=True))
q0: ─
q1: ─
```

Checklist:
- [x] Tests are passing.


alecandido:
@renatomello following Python standards, I'd propose to use `Circuit.__str__()` to return the stringified representation of the circuit, and we can then convert `Circuit.draw()` to directly print (avoiding extra parameters in both cases).
*Notice that some strings have to be closer to a code representation, but this is not the case for `__str__`, but rather `__repr__`'s role*

However, notice that this would be a breaking change, as all the assignment to the result of `circuit.draw()` will receive `None`.

So, we could also schedule `.draw()` replacement, but for the time being, I'd suggest using a different name for this feature, something like `Circuit.display()` or `Circuit.print()` itself.

renatomello:
> @renatomello following Python standards, I'd propose to use `Circuit.__str__()` to return the stringified representation of the circuit, and we can then convert `Circuit.draw()` to directly print (avoiding extra parameters in both cases). _Notice that some strings have to be closer to a code representation, but this is not the case for `__str__`, but rather `__repr__`'s role_
> 
> However, notice that this would be a breaking change, as all the assignment to the result of `circuit.draw()` will receive `None`.
> 
> So, we could also schedule `.draw()` replacement, but for the time being, I'd suggest using a different name for this feature, something like `Circuit.display()` or `Circuit.print()` itself.

Done. I'd like to push this to be merged before the next release, so then we can deprecate it by release 0.2.13.

src/qibo/models/circuit.py
```diff
@@ -1267,18 +1268,7 @@ def _update_draw_matrix(self, matrix, idx, gate, gate_symbol=None):
...
-        Args:
-            line_wrap (int): maximum number of characters per line. This option
-                split the circuit text diagram in chunks of line_wrap characters.
-            legend (bool): If ``True`` prints a legend below the circuit for
-                callbacks and channels. Default is ``False``.
-
-        Return:
-            String containing text circuit diagram.
-        """
+    def __str__(self, line_wrap: int = 70, legend: bool = False) -> str:
```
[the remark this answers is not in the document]
alecandido: So, sorry for the confusion of my initial proposal, I didn't take into account the arguments.

Let me formulate a complete one here (if needed I could also implement it to speed up the process)
```py
def diagram(self, line_wrap, ...) -> str:
    ...

def __str__(self):
    return self.diagram()

def draw(self, line_wrap, ...) -> str:
    """Just an alias, to avoid breaking compatibility."""
    return self.diagram(line_wrap, ...)

def display(self, line_wrap, ...):
    """Here for compatibility, to be replaced by `draw`."""
    sys.stdout.write(self.diagram(line_wrap, ...)) 
```

Whenever we will decide that it is acceptable to break, we will just drop `display` and replace it with `draw`.

tests/conftest.py
```diff
@@ -14,9 +14,9 @@
     "numpy",
     "tensorflow",
     "pytorch",
-    "qibojit-numba",
-    "qibojit-cupy",
-    "qibojit-cuquantum",
+    # "qibojit-numba",
+    # "qibojit-cupy",
+    # "qibojit-cuquantum",
```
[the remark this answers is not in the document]
renatomello: faster test locally that I forgot to undo

src/qibo/models/circuit.py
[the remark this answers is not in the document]
renatomello: Done

alecandido:
@renatomello a9bca8 is exactly what I meant.

I'm not sure why you're prepending the underscores: I know that `.display()` is temporary, and it will be removed, but you may want to use it in the meanwhile (and we will need to break anyhow to change the meaning of `.draw()`, breaking one or two doesn't make much difference).
Instead, the `.diagram()` call is there to stay, since it will be the only way to directly access the string, even though it won't be used that frequently (ok, the name is random, I was out of ideas and went looking on Thesaurus for inspiration - if you have any better proposal it's perfectly fine).

Other than that, the only bit missing to this PR is to convert the already converted usages, since now `print(c.draw())` has been converted to just `c.draw()`, but it's not (yet) going to work as intended.

renatomello:
> @renatomello a9bca8 is exactly what I meant.
> 
> I'm not sure why you're prepending the underscores: I know that `.display()` is temporary, and it will be removed, but you may want to use it in the meanwhile (and we will need to break anyhow to change the meaning of `.draw()`, breaking one or two doesn't make much difference). Instead, the `.diagram()` call is there to stay, since it will be the only way to directly access the string, even though it won't be used that frequently (ok, the name is random, I was out of ideas and went looking on Thesaurus for inspiration - if you have any better proposal it's perfectly fine).
> 
> Other than that, the only bit missing to this PR is to convert the already converted usages, since now `print(c.draw())` has been converted to just `c.draw()`, but it's not (yet) going to work as intended.

right now `print(circuit.draw())` still works as before and `circuit.display` works in-place, However, `circuit.draw` raises a warning saying that it will work in-place in the future.

doc/source/code-examples/advancedexamples.rst
```diff
@@ -1151,7 +1151,7 @@ Additionally, one can also pass single-qubit readout error probabilities (`reado
     )
 
     print("raw circuit:")
-    print(circuit.draw())
+    circuit.draw()
```
alecandido, line 1154: Like this one.

alecandido (commented):
> right now `print(circuit.draw())` still works as before and `circuit.display` works in-place, However, `circuit.draw` raises a warning saying that it will work in-place in the future.

Yes, that's correct, I perfectly agree with you.

It's just that some statements `print(circuit.draw())` have been replaced by just `circuit.draw()`. While, since you're now preserving the old behavior, they should be unchanged (or invoke `.display()` instead)

doc/source/code-examples/advancedexamples.rst
```diff
@@ -1242,7 +1242,7 @@ Let's see how to use them. For starters, let's define a dummy circuit with some
    circ.add(gates.M(*range(nqubits)))
 
    # visualize the circuit
-   print(circ.draw())
+   circ.draw()
```
renatomello, line 1245:
```suggestion
   print(circ.draw())
```

examples/adiabatic_qml/adiabatic-qml.ipynb
```diff
@@ -568,7 +568,7 @@
     "circ1 = rotcirc.rotations_circuit(t=0.1)\n",
     "circ2 = rotcirc.rotations_circuit(t=0.8)\n",
     "\n",
-    "print(f\"Circuit diagram: {circ1.draw()}\")\n",
+    "print(f\"Circuit diagram: {print(circ1.draw())}\")\n",
```
renatomello, line 571:
```suggestion
    "print(f\"Circuit diagram: {circ1.draw()}\")\n",
```

examples/anomaly_detection/test.py
```diff
@@ -99,7 +99,7 @@ def compute_loss_test(encoder, vector):
     encoder_test = make_encoder(n_qubits, n_layers, trained_params, q_compression)
     encoder_test.compile()
     print("Circuit model summary")
-    print(encoder_test.draw())
+    encoder_test.draw()
```
renatomello, line 102:
```suggestion
    print(encoder_test.draw())
```

alecandido (approved):

src/qibo/models/circuit.py
```diff
@@ -1267,18 +1268,7 @@ def _update_draw_matrix(self, matrix, idx, gate, gate_symbol=None):
...
-        Args:
-            line_wrap (int): maximum number of characters per line. This option
-                split the circuit text diagram in chunks of line_wrap characters.
-            legend (bool): If ``True`` prints a legend below the circuit for
-                callbacks and channels. Default is ``False``.
-
-        Return:
-            String containing text circuit diagram.
-        """
+    def diagram(self, line_wrap: int = 70, legend: bool = False) -> str:
```
BrunoLiegiBastonLiegi, line 1271:
```suggestion
    def diagram(self, line_wrap: int = 70, legend: bool = False) -> str:
    """Builds the string representation of the circuit diagram."""
```

BrunoLiegiBastonLiegi (approved):

src/qibo/models/circuit.py
alecandido:
```suggestion
    def diagram(self, line_wrap: int = 70, legend: bool = False) -> str:
        """Builds the string representation of the circuit diagram."""
```

Issue #1437, opened 2024-09-05
Variable not defined in `mpldrawer`


renatomello:
@MatteoRobbiati @sergiomtzlosa 

PR #1370 introduced the new `ui` module to `master`. There is a helper function in the module that is raising a `lint` error due to an undefined variable. The specific line is [this one](https://github.com/qiboteam/qibo/blob/4f2314ab2cc311894dbb05c862c5c8cc63b75ec0/src/qibo/ui/mpldrawer.py#L604). This is preventing tests from passing, e.g. https://github.com/qiboteam/qibo/actions/runs/10715691273/job/29711648870

MatteoRobbiati:
Thanks for spotting this! @sergiomtzlosa, should we pass the circuit as an argument so that we keep that line or do you think we can do that in a different way?

sergiomtzlosa:
sorry! I did not spot this, I push a fix for this, I just submit a commit to fix this bug

alecandido:
Yeah, unfortunately we didn't notice because the CI was not running, since the PR came from a fork.

However, this may happen for new PRs, even from forks. If that's not the case, please repeat to me or @scarrazza

MatteoRobbiati:
> sorry @renatomello and @MatteoRobbiati ! I did not spot this, I just submit a commit to fix this bug

No problem. Are you going to open a PR for this?

sergiomtzlosa:
yes! here is the pull request https://github.com/qiboteam/qibo/pull/1438

PR #1438, opened 2024-09-06
Matplotlib drawer: fix bug #1437 and add tests

sergiomtzlosa:
Checklist:
- [ ] Tests are passing.


alecandido:
Already noticed that the CI is still not running...

scarrazza:
@alecandido I have the suspicion that they changed `on` tag name to `pull_request_target`: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target

alecandido:
@scarrazza this is the only related change I've found in the last six months

https://github.com/github/docs/commit/9e3e15773248f53207fd12514e480bad15687cab

But maybe I should keep checking, I may have missed something.
The file is the following: 
https://github.com/github/docs/blob/main/content/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows.md

(I believe the workflow runner is not even open source, but GitHub docs are, so we can check those, at least)

scarrazza:
Ok, the problem is the `if` condition in rules.yml: 
https://github.com/qiboteam/qibo/blob/4f2314ab2cc311894dbb05c862c5c8cc63b75ec0/.github/workflows/rules.yml#L13 
This line is forcing a skip due to missing pushes (this PR was opened with a series of pushes already performed). 
We should the "PR opened event" in the `if` condition too.

alecandido:
Correct, thanks for the investigation.

In any case, even the `run-workflow` label is definitely a good solution. For sure for this PR.

MatteoRobbiati (changes requested):
Thanks a lot @scarrazza for spotting the CI problem and @sergiomtzlosa for the fix. 
The lint is now fine. Anyway, there is a problem in the test code we added to the documentation. Please remove the `%matplotlib inline` line from [this example](https://github.com/qiboteam/qibo/blob/master/doc/source/code-examples/examples.rst#how-to-visualize-a-circuit-with-style) and ask me for a new review @sergiomtzlosa :)

sergiomtzlosa:
thanks @alecandido @scarrazza for tune in the CI pipeline, I removed the line!!

MatteoRobbiati (changes requested):
Thanks @sergiomtzlosa for the fix!
Now we miss the tests. I would suggest to create a `test_ui_mlpdrawings.py` file (or something like that) in which we test all the code of your contribution. 
Some dummy tests will be enough to cover all of them. 
A complete report of the covered lines is provided by codecov everytime a new commit is pushed here :)

Let me know if you need some help in writing/defining the tests!

sergiomtzlosa:
Hi, thanks @MatteoRobbiati , just a question, which version should I install to do the test? I had installed pytest and I get this errors on runing the tests:

```
> pytest test_ui_mpldrawer.py
ERROR: usage: __main__.py [options] [file_or_dir] [file_or_dir] [...]
__main__.py: error: unrecognized arguments: --cov=qibo --cov-append --cov-report=xml --cov-report=html
  inifile: ..\qibo\pyproject.toml
  rootdir: ..\qibo`
```

scarrazza:
@sergiomtzlosa you can install `pytest-cov` or simply install this package with `poetry install --with tests`.

scarrazza:
@sergiomtzlosa could you please fix tests? Following my instructions above you should be able to test locally.

MatteoRobbiati:
Thanks for your work @sergiomtzlosa! 
If you want to avoid pre-commit to fix the code cosmetics after your commits, I would suggest you to install it and activate it in your local qibo repo. It will standardize the code following the qibo guidelines.

You can install it using pip via `pip install pre-commit` and then type `pre-commit install` after entering qibo locally. 

You will se pre-commit fixing the code everytime you do a commit. After that, you can add the new modifications to git and commit again with the same commit message. 

As usual, let me know if you need any further help/info!

And again thanks for the work :)

sergiomtzlosa:
Thanks a lot @MatteoRobbiati !! it is a useful hook indeed

sergiomtzlosa:
All code is covered by tests, this PR is ready to merge @MatteoRobbiati

tests/test_ui_mpldrawer.py
```diff
@@ -0,0 +1,342 @@
...
+    c.add(gates.M(1))
+    return c
+
+
+@pytest.mark.parametrize("nqubits", [2, 3])
+def test_plot_circuit(nqubits):
+    """Test for main plot function"""
+    circ = circuit(nqubits)
+    ax, _ = plot_circuit(circ)
+    assert ax != None
```
MatteoRobbiati, line 35: Take this as a general comment, valid also for the following assertions in the code.

It is the first time I need to test a routine which involves plots, so I am investigating on which can be the best way of doing it. I don't think just checking the axes of our Matplotlib object are different from `None` would be enough. Probably, it would be better to define a strategy to check that also the output is what we expect. 
One possible way of doing this could be to generate some reference plots and constructs some correspondent hash codes. This can be done properly selecting the Matplotlib backend (so that we don't have problem with the rendering of the images adopting different systems) and using the `hashlib` library. 

Once we have some reference hashes (I think just a few plots can be enough to test the whole drawer), we can regenerate the plots in the tests and check if the generated hashes correspond with the target ones. 

Do you have some suggestions/ideas @alecandido @scarrazza?

tests/test_ui_mpldrawer.py
```diff
@@ -0,0 +1,342 @@
...
+def test_fused_gates():
+    """Test for FusedStartGateBarrier and FusedEndGateBarrier"""
+    min_q = 0
+    max_q = 1
+    l_gates = 1
+    equal_qbits = True
+    start_barrier = FusedStartGateBarrier(min_q, max_q, l_gates, equal_qbits)
+    end_barrier = FusedEndGateBarrier(min_q, max_q)
+    assert start_barrier.unitary == start_barrier.unitary
+    assert end_barrier.unitary == end_barrier.unitary
```
MatteoRobbiati, line 149: Was the idea here to check whether end and start barriers are the same? I wouldn't take as a valid check to use an assertion on the same object.

src/qibo/ui/mpldrawer.py
```diff
@@ -301,12 +287,8 @@ def _draw_target(ax, i, gate, labels, gate_grid, wire_grid, plot_params):
     x = gate_grid[i]
     target_index = _get_flipped_index(target, labels)
     y = wire_grid[target_index]
-    if not symbol:
-        return
     if name in ["CNOT", "TOFFOLI"]:
         _oplus(ax, x, y, plot_params)
-    elif name == "CPHASE":
```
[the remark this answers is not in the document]
sergiomtzlosa: This gate name does not exists in qibo indeed, the CPAHSE name is inherited code from the original source https://github.com/rpmuller/PlotQCircuit/blob/17ff7f04c3c1de4e47dcbb68cdb4b823a227e2d4/plot_quantum_circuit.py#L104

src/qibo/ui/mpldrawer.py
```diff
@@ -227,20 +227,6 @@ def _draw_controls(ax, i, gate, labels, gate_grid, wire_grid, plot_params, measu
...
-        if ismeasured:
-            dy = 0.04  # TODO: put in plot_params
-            _line(
-                ax,
-                gate_grid[i] + dy,
-                gate_grid[i] + dy,
-                wire_grid[min_wire],
-                wire_grid[max_wire],
-                plot_params,
-            )
```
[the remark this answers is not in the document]
sergiomtzlosa: this is an old code to draw a vertical line from the measure gate to the bottom of the circuit, this function was depreciated at the begining of the development and thus is not valid here, here is the original source https://github.com/rpmuller/PlotQCircuit/blob/17ff7f04c3c1de4e47dcbb68cdb4b823a227e2d4/plot_quantum_circuit.py#L87

tests/test_ui_mpldrawer.py
sergiomtzlosa: The idea is to create a dummy check, I do not bare in mind to do something odd, but you are right, it could be also an assertion on the same object

sergiomtzlosa:
@MatteoRobbiati  I will merge the master branch into this one to get latest changes, are yo agree?

sergiomtzlosa:
> During the meeting we agreed in setting up a bit more robust strategy to test the images. What is done so far is perfectly fine, but we just need to be sure the output of our plots is actually what we expect during the tests. For doing this, we can:
> 
>     1. generate locally the images we are re-generating in the tests;
> 
>     2. convert them into Numpy arrays and store those arrays in a folder;
> 
>     3. during tests, check whether the new plots, converted into arrays, are actually equal to the target arrays.
> 
> 
> Alternatively, we can also think to use the [Matplotlib testing tools.](https://matplotlib.org/devdocs/devel/testing.html#writing-an-image-comparison-test)

Good!!! I will start with the plot arrays...

MatteoRobbiati:
> Good!!! I will start with the plot arrays comparison...

Hi there! Do you need any help here? It would be ideal to fix this for the next review (friday). Let me know about the status/if help is needed @sergiomtzlosa.

sergiomtzlosa:
> > Good!!! I will start with the plot arrays comparison...
> 
> Hi there! Do you need any help here? It would be ideal to fix this for the next review (friday). Let me know about the status/if help is needed @sergiomtzlosa.

Hola @MatteoRobbiati , everything is done on the latest commit, please let me know if somethinh is missing, you can launch CI indeed

sergiomtzlosa:
Hi @MatteoRobbiati , I am wondering why all test are correct , but the macos test fail or got cancelled, please can you gibe some light? Is the CI maybe crashing?

MatteoRobbiati (commented):
Thanks for this @sergiomtzlosa! It sounds like what I had in mind. I was thinking if the PNG images are the best way to store the data, but I think this can work. 

1. PIL is used to test the assertions here, but I think it has to be added to the `pyproject.toml` and update the lock (we need to specify this package has to be installed when performing tests);
2. tests are failing on ios right now. It seems the arrays are not equivalent.. maybe we should consider to use some specific Matplotlib backend?

MatteoRobbiati:
> Hi @MatteoRobbiati , I am wondering why all test are correct , but the macos test fail or got cancelled, please can you gibe some light? Is the CI maybe crashing?

Sorry!! Missed it! Remember also to ask me for review once you are done with modifications :)

sergiomtzlosa:
> > Hi @MatteoRobbiati , I am wondering why all test are correct , but the macos test fail or got cancelled, please can you gibe some light? Is the CI maybe crashing?
> 
> Sorry!! Missed it! Remember also to ask me for review once you are done with modifications :)

Hi @MatteoRobbiati , now I have a Big issue, I cannot fix this until the next week then I cannot finish this for friday, and i would ask for some help to get this before the next friday otherwise i fix it Next week, sorry!

alecandido:
Testing exact equality pixel by pixel is definitely a hard constraint, so you should make sure there is no platform-dependent behavior in matplotlib as well (even driven by the environment, i.e. variables and packages available on a certain system).

A different option could be generating the pictures directly in the CI, and caching them somewhere, separately for each platform, skipping the tests if they are not available.
However, while this option will be more robust from the point of view of platform-dependent behaviors, but it will be much more expensive (time-wise) to instrument...

If you're able to debug the issue, and fix some env var to obtain the same result, so much the better.
Otherwise, I'd consider acceptable even just testing that an image file is produced, whatever the content is, since it will tell us that the code didn't raise/crash (at least).

alecandido (approved):

MatteoRobbiati (approved):
Thanks a lot for this effort @sergiomtzlosa, and again sorry not to have noticed this before. 
Now everything sounds good to me. I re run the tests and everything is passing.

sergiomtzlosa:
Thanks @alecandido @MatteoRobbiati , finally we made It!

Issue #1432, opened 2024-08-30
Shor example performance drop


stavros11:
First, the [Shor example](https://github.com/qiboteam/qibo/tree/master/examples/shor) with the semi-classical approach, which can be executed using:
```sh
python examples/shor/main.py --semiclassical
```
is broken because it is no longer possible (I think after #1039) to execute circuits that only contain collapse measurements using state vectors. The error says to use density matrices, however this is intractable for the typically large circuits of this example. An easy fix is to add a normal (non-collapse) measurement at the end of the circuit, which makes it fall back to repeated execution and returns `MeasurementOutcomes` (instead of full state).

However, even with the above fix, the default N=15 did not finish execution within several minutes of using qibojit (numba). This is much slower than the original benchmark in https://github.com/qiboteam/qibo/pull/277#issuecomment-727641458. I also managed to roll back to qibo 0.1.6rc1 and qibojit 0.0.1 (not very trivial) and the same example runs in a few seconds so clearly something broke in between, or the last measurement I added breaks something (but this seems unlikely based on some profiling I did).

igres26:
I used this example at 0.2.8 and it is fast to execute. So something has been introduced in between these two versions.

alecandido:
#1039 was first included in v0.2.2.

We should take the time of running it with `git bisect` (or manually bisecting versions first), to find when this behavior appeared.

stavros11:
> I used this example at 0.2.8 and it is fast to execute. So something has been introduced in between these two versions.

Are you using the qibojit numba backend for this? If yes, could you please also add the qibojit and numba versions? You can find them using `pip list`. I tried with
```
numba                  0.60.0
numpy                  1.26.4
qibo                   0.2.8
qibojit                0.1.4
```
and the `--semiclassical` case is still slow.

stavros11:
> #1039 was first included in v0.2.2.

To me it seems that the problem was introduced in that version. In particular, the following environment
```
numba                  0.60.0
numpy                  1.26.4
qibo                   0.2.1
qibojit                0.1.5
```
is fast with qibojit (numba) and it becomes slow if I upgrade qibo to v0.2.2 (with the rest staying the same).

I am not sure if it is related to #1039 in particular or some other PR first included in v0.2.2. `git bisect` could help with that.

PR #1439, opened 2024-09-09
Fix Shor example

stavros11:
Fix Shor example, which is outdated since #1039, as reported in #1432.

The measurement in the end is needed because it is no longer possible to execute circuits that contain only collapse measurements using state vectors.

Specifying `nshots=1` is needed because a default `nshots=1000` was introduced in #1039 compared to the `None` being used before. This was also causing the performance issue reported in #1432, because during the repeated execution it was executing the circuit 1000 times instead of one that we need.

Checklist:
- [ ] Tests are passing.


igres26 (approved):
Thanks! This looks like the solution.

MatteoRobbiati (approved):
Thanks.

MatteoRobbiati (approved):

PR #1448, opened 2024-09-16
Add missing `hermitian` argument to `backend.calculate_eigenvalues` and `backend.calculate_eigenvectors` methods

### LINKED ISSUE #1447: Tests failing with `cupy`
Some tests are failing with `cupy` because the methods `calculate_eigenvectors()` and `calculate_eigenvalues()` do not accept `hermitian` as input. For example, [test_fidelity_and_infidelity_and_bures.py](https://github.com/qiboteam/qibo/blob/master/tests/test_quantum_info_metrics.py#L161) fails with the error 
```python
TypeError: CupyBackend.calculate_eigenvectors() got an unexpected keyword argument 'hermitian'
```

renatomello:
Closes #1447 together with https://github.com/qiboteam/qibojit/pull/188. Tests should be run locally to verify that they are passing on GPUs.

The `numpy`, `numba`, `pytorch`, and `tensorflow` backends already had this argument in their methods, but `abstract`, `cupy`, and `cuquantum` did not. This fixes the issue.

Checklist:
- [x] Tests are passing.


src/qibo/backends/abstract.py
```diff
@@ -318,12 +318,16 @@ def calculate_overlap_density_matrix(self, state1, state2):  # pragma: no cover
         raise_error(NotImplementedError)
 
     @abc.abstractmethod
-    def calculate_eigenvalues(self, matrix, k=6):  # pragma: no cover
+    def calculate_eigenvalues(
+        self, matrix, k: int = 6, hermitian: bool = True
+    ):  # pragma: no cover
```
alecandido, line 323: Pay attention: this kind of change affects the public API, and it could be breaking.

In particular, this is mostly not breaking because of the default value (and also because almost everyone subclasses the `NumpyBackend`...), but if there were backends directly subclassing this, respecting its signature, they would now be broken, since they would not support the extra argument (and they would be flagged by linters, e.g. Pylint in the CI).

So, strictly speaking, it is breaking, but by virtue of the `NumpyBackend`-subclassing habit it should be irrelevant.

renatomello: In this case, I'm not adding anything new everywhere, just fixing an inconsistency in the method in the `abstract` class (method that is being overwritten by every backend anyways)

alecandido: Yes, that's way, I added the note about it being irrelevant.

Moreover, Qibo is not that stable anyway, also because the public API is not well-identified.
But, in principle, it's the `AbstractBackend` defining the interface, not the `NumpyBackend`.

I just wanted to flag it, just because I'd like to move towards a more stable package, and a more consistent use of semver (as a guarantee to the Qibo users).
But I agree with you in qualifying this as mostly a bugfix (but if Qibo was more stable and widely used, we should have considered this "bug" as the current interface, to avoid breaking users' code - no matter if it was like that by mistake).

renatomello: It is also a bug fix because that is making some tests fail for `cupy` / `cuquantum`. The main reason for this fix, actually. But yeah we are in agreement.

BrunoLiegiBastonLiegi (approved):

AlejandroSopena (approved):

PR #1450, opened 2024-09-16
Fix `pytorch` gradients

### LINKED ISSUE #1449: `PyTorchBackend` backpropagation compatibility with `torch.nn.Module` and `torch.optim`
I am trying to train a VQC through the `PyTorchBackend` and using the `pytorch` API. Here you can find a small test example.

``` python
import torch
import numpy as np
from qibo.backends import PyTorchBackend
from qibo import hamiltonians
from qibo.symbols import Z

BACKEND = PyTorchBackend()

class QuantumCircuit(torch.nn.Module):

    def __init__(self, encoding: Circuit, circuit: Circuit):
        super().__init__()
        self.encoding = encoding
        self.circuit = circuit
        self.circuit_parameters = torch.nn.Parameter(BACKEND.cast(circuit.get_parameters(), dtype=torch.float64))
        self.circuit.set_parameters(self.circuit_parameters)
        self.observable = hamiltonians.SymbolicHamiltonian(
            sum([Z(int(i)) for i in range(circuit.nqubits)]),
            nqubits=circuit.nqubits,
            backend=BACKEND,
        )
        # self.linear = torch.nn.Linear(2 ** circuit.nqubits, 1).double()
        
    def forward(self, x):
        self.encoding.set_parameters(x * BACKEND.np.pi)
        circuit = self.encoding + self.circuit
        state = BACKEND.execute_circuit(circuit).state()
        return self.observable.expectation(
            state
        )
        #return self.linear(BACKEND.execute_circuit(circuit).probabilities())
        

if __name__ == "__main__":

    nqubits = 3
    encoding = Circuit(nqubits)
    [encoding.add(gates.H(i)) for i in range(nqubits)]
    [encoding.add(gates.RZ(i, theta=0.)) for i in range(nqubits)]
    circuit = Circuit(nqubits)
    for q in range(nqubits):
        circuit.add(gates.RY(q, theta=torch.randn(1) * BACKEND.np.pi))
        circuit.add(gates.RZ(q, theta=torch.randn(1) * BACKEND.np.pi))
    for i, q in enumerate(range(nqubits)[:-2]):
        circuit.add(gates.CNOT(q0=q, q1=i + 1))
    circuit.add(gates.CNOT(q0=nqubits-1, q1=0))
    
    model = QuantumCircuit(encoding, circuit)
    data = torch.randn(100, 3)
    target = (data[:, 0] * BACKEND.np.pi) ** 2
    loss_f = lambda x, y: (x - y)**2
    optimizer = torch.optim.SGD(model.parameters(), lr=1e-2, momentum=0.9)
    for x, y in zip(data, target):
        optimizer.zero_grad()
        out = model(x)
        loss = loss_f(out, y)
        loss.backward()
        optimizer.step()
        print(f"Loss: {loss}")
        print(f"Parameters: {list(model.parameters())}")
```

You first wrap the circuit inside a `torch.nn.Module` object, register the parameters and then, in the forward, you execute the circuit and calculate the desired output, for instance an expectation value. Everything is trained using a `torch.optim` optimizer.
Now, if you run the training it is easy to see that the parameters never get updated. If you try to switch from the expectation value to the probabilities, for instance, by adding a linear layer that maps to the correct shape (2**nqubits --> 1) you can also notice as the parameters of the linear layer get updated, but not those of the circuit.

Therefore, I would say that there is some operation, either in the process of setting/getting the parameters of the circuit or in the execution + post processing (expectation value or probabilities calculation) that brakes the tracing of the tensors.
@Simone-Bordoni in the test did we actually try to run a gradient-based optimization anywhere?

Simone-Bordoni:
Fix #1449 

Checklist:
- [x] Tests are passing.


BrunoLiegiBastonLiegi:
I found a potential problem when you cast the parameters of parametrized gates:
https://github.com/qiboteam/qibo/blob/1b6eb8a358f8f51d0f5a267303809724505cfb53/src/qibo/backends/pytorch.py#L32
here you cast to `self.dtype` which is usually `torch.complex128` but it should actually be `torch.float32/64` since this is a parameter. I don't know whether this is the origin of the problem with the gradients though.

Simone-Bordoni:
> I found a potential problem when you cast the parameters of parametrized gates:
> 
> https://github.com/qiboteam/qibo/blob/1b6eb8a358f8f51d0f5a267303809724505cfb53/src/qibo/backends/pytorch.py#L32
> 
> 
> here you cast to `self.dtype` which is usually `torch.complex128` but it should actually be `torch.float32/64` since this is a parameter. I don't know whether this is the origin of the problem with the gradients though.
> EDIT: nevermind, you necessarily have to cast to complex otherwise when you build the matrix elements pytorch complains about mismatching dtypes...

Actually this may be part of the problem, torch is not suporting backpropagation for complex. However I am rewriting in a better way the whole casting function differentiating the matrix_dtype from the parameters_dtype

Simone-Bordoni:
Now the gradients are passing however their final value is always zero. I have tried visualizing the computational graph (shown below) and all the operations seems to be performed correctly.
I am currently tsting a simple circuit with just one rotation gate (test_torch_gradients.py). 
In the next days I will not be able to work furthermore on this issue, I will resume working on the gradients next week. In the meantime if you find any possible reason for this problemlet me know, it would be very helpful.
loss

src/qibo/backends/pytorch.py
```diff
@@ -121,23 +121,34 @@ def matrix_parametrized(self, gate):
         name = gate.__class__.__name__
         _matrix = getattr(self.matrices, name)
         if name == "GeneralizedRBS":
+            for parameter in ["theta", "phi"]:
+                if not isinstance(gate.init_kwargs[parameter], self.np.Tensor):
+                    gate.init_kwargs[parameter] = self.cast_parameter(
+                        gate.init_kwargs[parameter], trainable=gate.trainable
+                    )
+                elif gate.init_kwargs[parameter].requires_grad == True:
```
renatomello, line 129:
```suggestion
                elif gate.init_kwargs[parameter].requires_grad:
```

src/qibo/backends/pytorch.py
```diff
@@ -121,23 +121,34 @@ def matrix_parametrized(self, gate):
...
-            )
-            _matrix = _matrix(*parameters)
-        print("parameterized matrix:", _matrix)
+            if not isinstance(gate.parameters[0], self.np.Tensor):
+                parameters = tuple(
+                    self.cast_parameter(param, trainable=gate.trainable)
+                    for param in gate.parameters
+                )
+                gate.parameters = parameters
+            elif gate.parameters[0].requires_grad == True:
```
renatomello, line 146:
```suggestion
            elif gate.parameters[0].requires_grad:
```

Simone-Bordoni:
I think that now everything has been fixed. 
The changes required were bigger than expected, I had to add the backend in the gate decompositions and in other parts.
I have added a test to check the correct backpropagation in test_torch_gradients.py so that thay can be easily moved to qiboml.

Simone-Bordoni:
> Thanks @Simone-Bordoni, #1449 is still failing even with this patch, but I think that's more related to a problem with the `circuit.set/get_parameters()`.

In the new version the backend.cast has requires_grad=False by default, you need to modify #1449 and activate the gradients when casting.

BrunoLiegiBastonLiegi:
Like this?
``` python
self.circuit_parameters = torch.nn.Parameter(BACKEND.cast(circuit.get_parameters(), dtype=torch.float64, requires_grad=True))
```
because this fails anyway.

Simone-Bordoni:
> Like this?
> 
> ```python
> self.circuit_parameters = torch.nn.Parameter(BACKEND.cast(circuit.get_parameters(), dtype=torch.float64, requires_grad=True))
> ```
> 
> because this fails anyway.
> 
> EDIT: sorry maybe I should have specified better, but it doesn't fail in the sense that the gradients don't pass, it actually crashes with an error:
> 
> ```
> RuntimeError: Output 0 of UnbindBackward0 is a view and its base or another view of its base has been modified inplace. This view is the output of a function that returns multiple views. Such functions do not allow the output views to be modified inplace. You should replace the inplace operation by an out-of-place one.
> ```

OK, I thought that the problem was the gradient not passing. Is it a different error with respect to what you were getting before this update?

src/qibo/backends/numpy.py
```diff
@@ -398,7 +398,8 @@ def execute_circuit(self, circuit, initial_state=None, nshots=1000):
             else:
                 return self.execute_circuit(initial_state + circuit, None, nshots)
         elif initial_state is not None:
-            initial_state = self.cast(initial_state)
+            if initial_state.dtype != self.dtype:
+                initial_state = self.cast(initial_state)
```
renatomello, line 402: Is this needed in the `numpy` backend?

src/qibo/backends/numpy.py
```diff
@@ -428,8 +429,8 @@ def execute_circuit(self, circuit, initial_state=None, nshots=1000):
             if circuit.density_matrix:
                 if initial_state is None:
                     state = self.zero_density_matrix(nqubits)
+
```
renatomello, line 432:
```suggestion
```

src/qibo/backends/numpy.py
```diff
@@ -438,8 +439,8 @@ def execute_circuit(self, circuit, initial_state=None, nshots=1000):
             else:
                 if initial_state is None:
                     state = self.zero_state(nqubits)
+
```
renatomello, line 442:
```suggestion
```

src/qibo/backends/pytorch.py
```diff
@@ -131,9 +116,50 @@ def cast(
...
+            gate.parameters = tuple(new_parameters)
+        _matrix = _matrix(*gate.parameters)
+        return _matrix
+
+    def check_parameters(self, parameters):
+        """Check if the parameters are torch tensors."""
+        for parameter in parameters:
+            if not isinstance(parameter, self.np.Tensor):
+                return False
+        return True
```
renatomello, line 156:
```suggestion
    def _check_parameters(self, parameters):
        """Check if the parameters are torch tensors."""
        return all(isisntance(param, self.np.Tensor) for param in parameters)
```

src/qibo/backends/pytorch.py
```diff
@@ -131,9 +116,50 @@ def cast(
...
+
+    def cast_parameter(self, x, trainable):
+        if isinstance(x, int) and trainable:
```
renatomello, line 159: Missing docstring

src/qibo/backends/pytorch.py
```diff
@@ -131,9 +116,50 @@ def cast(
...
-
         return x
 
+    def matrix_parametrized(self, gate):
+        """Convert a parametrized gate to its matrix representation in the computational basis."""
+        name = gate.__class__.__name__
+        _matrix = getattr(self.matrices, name)
+        if name == "GeneralizedRBS":
+            for parameter in ["theta", "phi"]:
+                if not isinstance(gate.init_kwargs[parameter], self.np.Tensor):
```
renatomello, line 127: This looks redundant with the `_check_parameter` method

src/qibo/backends/pytorch.py
```diff
@@ -131,9 +116,50 @@ def cast(
...
+                qubits_out=gate.init_args[1],
+                theta=gate.init_kwargs["theta"],
+                phi=gate.init_kwargs["phi"],
+            )
+            return _matrix
+        if not self.check_parameters(gate.parameters):
+            new_parameters = []
+            for parameter in gate.parameters:
+                if not isinstance(parameter, self.np.Tensor):
+                    parameter = self.cast_parameter(parameter, trainable=gate.trainable)
```
renatomello, line 143: Also redundant w.r.t. `_check_parameter`

src/qibo/gates/gates.py
```diff
@@ -2600,7 +2603,7 @@ def __init__(
     @Gate.parameters.setter
     def parameters(self, x):
         shape = self.parameters[0].shape
-        engine = _check_engine(self.parameters[0])
+        engine = _check_engine(x)
         # Reshape doesn't accept a tuple if engine is pytorch.
         x = x[0] if type(x) is tuple else x
```
renatomello, line 2608:
```suggestion
        if isinstance(x, tuple):
            x = x[0]
```

src/qibo/gates/gates.py
```diff
@@ -2633,7 +2637,9 @@ def _dagger(self):
 
 def _check_engine(array):
     """Check if the array is a numpy or torch tensor and return the corresponding library."""
-    if array.__class__.__name__ == "Tensor":
+    if (array.__class__.__name__ == "Tensor") or (
+        type(array) is tuple and array[0].__class__.__name__ == "Tensor"
```
renatomello, line 2641:
```suggestion
        isinstance(array, tuple) and array[0].__class__.__name__ == "Tensor"
```

src/qibo/models/circuit.py
```diff
@@ -1142,6 +1142,9 @@ def from_dict(cls, raw):
     def to_qasm(self):
         """Convert circuit to QASM.
 
+        .. note::
+            This method doesn't support multi-controlled gates and gates with torch Tensors parameters.
```
renatomello, line 1146:
```suggestion
            This method does not support multi-controlled gates and gates with ``torch.Tensor`` as parameters.
```

src/qibo/models/circuit.py
```diff
@@ -1174,6 +1177,12 @@ def to_qasm(self):
 
             qubits = ",".join(f"q[{i}]" for i in gate.qubits)
             if isinstance(gate, gates.ParametrizedGate):
+                for x in gate.parameters:
+                    if x.__class__.__name__ == "Tensor":
```
renatomello, line 1181:
```suggestion
                if any(x.__class__.__name__ == "Tensor" for x in gate.parameters):
```

src/qibo/models/circuit.py
```diff
@@ -1174,6 +1177,12 @@ def to_qasm(self):
...
+                        raise_error(
+                            ValueError,
+                            "OpenQASM does not support gates with torch Tensors parameters.",
+                        )
```
renatomello, line 1185:
```suggestion
                if any(x.__class__.__name__ == "Tensor" for x in gate.parameters):
                    raise_error(
                        ValueError,
                        "OpenQASM does not support gates with torch Tensors parameters.",
                    )
```

src/qibo/models/error_mitigation.py
```diff
@@ -330,6 +330,7 @@ def _curve_fit(
     if backend.name == "pytorch":
         # pytorch has some problems with the `scipy.optim.curve_fit` function
         # thus we use a `torch.optim` optimizer
+        params = params.requires_grad_(True)
```
renatomello, line 333: Should this be forced here without user input?

src/qibo/transpiler/decompositions.py
```diff
@@ -83,7 +80,7 @@ def _u3_to_gpi2(t, p, l):
 gpi2_dec.add(gates.SX, [gates.GPI2(0, 0)])
 gpi2_dec.add(
     gates.RX,
-    lambda gate: [
+    lambda gate, backend: [
```
renatomello, line 85: Why is `backend` needed if it is never used? This goes for all the other `lambda` functions in this file.

src/qibo/transpiler/unroller.py
```diff
@@ -142,17 +147,22 @@ def assert_decomposition(
...
         gate (:class:`qibo.gates.abstract.Gate`): gate to be decomposed.
         native_gates (:class:`qibo.transpiler.unroller.NativeGates`): native gates
             to use in the decomposition.
+        backend (:class:`qibo.backends.Backend`): backend to use for gate matrix.
 
     Returns:
         list: List of native gates that decompose the input gate.
     """
+    if backend is None:
+        backend = _check_backend(backend)
```
renatomello, line 164: `_check_backend` already handles the `None` case

```suggestion
    backend = _check_backend(backend)
```

tests/test_torch_gradients.py
```diff
@@ -0,0 +1,34 @@
...
+        optimizer.zero_grad()
+        c.set_parameters(params)
+        final_state = c().state()
+        fidelity = torch.abs(torch.sum(torch.conj(target_state) * final_state))
+        loss = 1 - fidelity
+        loss.backward()
+        optimizer.step()
+
+    assert initial_loss > loss
+    assert initial_params[0] != params[0]
```
renatomello, line 34: Can you add a test that, given the same circuit and same optimizer (e.g. `SGD`), `torch` and `tensorflow` return the same gradients (up to some precision)?

Simone-Bordoni:
> Yes, if I remember correctly before it used to run without crashing but had the gradients flow problem.

Ok, anyway i think that it would be better to first merge this PR, as it is getting quite big, and then try to find why this specific case is not working correctly.

src/qibo/backends/numpy.py
Simone-Bordoni: If you use a float/int for the initial state it will be casted to complex which is the correct dtype for the computation. Moreover this function is also used by the pytorch backend, so this will make the circuit run even if you are passing a numpy array as initial state when using the pytorch backend.

src/qibo/models/error_mitigation.py
Simone-Bordoni: It can be removed but this is a function to train the parameters, this way the training will work even if the user forgets to activate the gradients. I think it is just a way to help the user.

renatomello: Let's hear @BrunoLiegiBastonLiegi

Simone-Bordoni:
@alecandido

src/qibo/models/circuit.py
```diff
@@ -1174,6 +1177,11 @@ def to_qasm(self):
 
             qubits = ",".join(f"q[{i}]" for i in gate.qubits)
             if isinstance(gate, gates.ParametrizedGate):
+                if any(x.__class__.__name__ == "Tensor" for x in gate.parameters):
+                    raise_error(
+                        ValueError,
+                        "OpenQASM does not support gates with torch Tensors parameters.",
+                    )
```
BrunoLiegiBastonLiegi, line 1184: can't you just do `str(float(x))` below?

src/qibo/models/error_mitigation.py
BrunoLiegiBastonLiegi: This is not a function the user is supposed to face in principle, I mean it doesn't harm to enforce it and shouldn't be expensive, but I wonder where the error was coming from. I guess inside the mitigation functions, and thus I would say it's better to correct there.

src/qibo/transpiler/decompositions.py
```diff
@@ -20,43 +17,63 @@ def add(self, gate, decomposition):
...
         self.decompositions[gate] = decomposition
 
-    def count_2q(self, gate):
+    def count_2q(self, gate, backend):
         """Count the number of two-qubit gates in the decomposition of the given gate."""
         if gate.parameters:
-            decomposition = self.decompositions[gate.__class__](gate)
+            if isinstance(
+                gate,
+                (gates.FusedGate, gates.Unitary, gates.GeneralizedfSim, gates.fSim),
```
BrunoLiegiBastonLiegi, line 25: Why do we need to treat these differently?

BrunoLiegiBastonLiegi (commented):
Thanks @Simone-Bordoni, this looks good. I still have to look into the tests, but I leave you just some thoughts on the source.

src/qibo/backends/pytorch.py
```diff
@@ -131,9 +118,53 @@ def cast(
...
+        return _matrix
+
+    def cast_parameter(self, x, trainable):
+        """Cast a gate parameter to a torch tensor.
+
+        Args:
+            x (Union[int, float, complex]): Parameter to be casted.
+            trainable (bool): If ``True``, the tensor requires gradient.
+        """
+        if isinstance(x, int) and trainable:
```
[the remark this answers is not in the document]
Simone-Bordoni: In the other backends if you use gate.RX(0, 0) this doesn't raise an error. This way the parameter is converted to float and makes the behaviour of the backend the same as the others.

src/qibo/transpiler/decompositions.py
Simone-Bordoni: We need the backend to implement the decomposition

BrunoLiegiBastonLiegi: Yes that I can see myself, but why is that the case? It was not like this before. In general, I don't understand why you need the backend to count how many two qubit gates you have in the decomposition.

Simone-Bordoni: It is not to count 2 qubit gates, it is to decompose the matrix (using u3_decomposition or two_qubit_decomposition) when calling the decomposition.

BrunoLiegiBastonLiegi: Well but the purpose of the function is just to count the 2q gates as the name and the docstring says. Thus you don't even need to instantiate the gate in principle, and even if you do it does not really matter whether the parameters of the gate are `torch.Tensor`s or plain floats, but this is probably more related to the decomposition mechanism in general not specifically to this PR.

Anyway, at least we could try to wrap this snippet:
``` python
            if isinstance(
                gate,
                (gates.FusedGate, gates.Unitary, gates.GeneralizedfSim, gates.fSim),
            ):
                decomposition = self.decompositions[gate.__class__](gate, backend)
            else:
                decomposition = self.decompositions[gate.__class__](gate)
```
in a `_helper_function` since it is repeated several times.

Simone-Bordoni: Not all decomposition have a fixed number of gates. When decomposing a two qubit unitary the number of 2q gates is usually 3 but in some cases it reduces to 2 or even 1. It is necessary to perform the decomposition to know it.

BrunoLiegiBastonLiegi: Yes, of course, but the decomposition is well defined even though the gates that constitute it are not explicitely instantiated nor whether their parameters are consistent with the backend. The decomposition should be backend independent I believe.

src/qibo/models/error_mitigation.py
```diff
@@ -432,6 +431,8 @@ def CDR(
         len(signature(model).parameters) - 1
     )  # first arg is the input and the *params afterwards
     params = backend.cast(local_state.random(nparams), backend.precision)
+    if backend.name == "pytorch":
+        params.requires_grad = True
```
BrunoLiegiBastonLiegi, line 435: All this specific case handling is not ideal though, can't we have `PyTorchBackend.cast` enabling gradients by default? I mean, in general, gradients are the main reason why you would use the backend in the first place.

src/qibo/transpiler/decompositions.py
Simone-Bordoni: Making the decomposition backend independent is not part of this PR. However, I think that the decomposition should be implemented on the selected backend. This would make the gradients pass (in principle) and help in executing qml algorithms directly on hardware. Regarding the function count_2q/1q i agree that the backend is not fundamental, if you prefer i can set backend=NumpyBackend() at the beginning of the function and avoid passing it.

src/qibo/models/error_mitigation.py
Simone-Bordoni: with @renatomello we decided to set require_gradients=False by default (we changed it from the first version) to avoid putting gradients everywhere like in the circuit initial state.

BrunoLiegiBastonLiegi: I see, it's just that `pytorch` cast is now asymmetric compared to other backends, which sometimes forces you to deal with this case handling. One thing, I was wondering is why don't we use [`torch.is_grad_enabled`](https://pytorch.org/docs/stable/generated/torch.is_grad_enabled.html):
``` python
def cast(self, ...):
    requires_grad = self.np.is_grad_enabled()
    ....
```

Simone-Bordoni: but still you will activate gradients for the initial state of the circuit and many other variables that do not require them

BrunoLiegiBastonLiegi: I don't know, in general I am not understanding why it is a problem to have an `initial_state` tensor with `requires_grad=True`, which other cases are you referring to in particular can you provide some examples?

Simone-Bordoni:
@alecandido I have talket to @renatomello regarding the error that arises when running the tests on github (locally they pass). The test "test_backends_torch_gradients" computes the gradient and an update step using both tensorflow and torch and checks that the result is the same. However tensorflow is enabled during the tests only when using ubuntu so the test fails on windows. Is there a way to perform the test only on ubuntu? or how can I solve the problem in another way?

tests/test_backends_torch_gradients.py
```diff
@@ -0,0 +1,93 @@
+import numpy as np
+import pytest
+
+import qibo
+from qibo import gates, models
```
renatomello, line 5:
```suggestion
from qibo import gates, models
from qibo.quantum_info import fidelity
```

tests/test_backends_torch_gradients.py
```diff
@@ -0,0 +1,93 @@
...
+
+    optimizer = optimizer([params])
+    for _ in range(nepochs):
+        optimizer.zero_grad()
+        c.set_parameters(params)
+        final_state = backend.execute_circuit(c).state()
+        fidelity = backend.np.abs(
+            backend.np.sum(backend.np.conj(target_state) * final_state)
+        )
+        loss = 1 - fidelity
```
renatomello, line 36:
```suggestion
        loss = infidelity(final_state, target_state)
```

tests/test_backends_torch_gradients.py
```diff
@@ -0,0 +1,93 @@
...
+    c = models.Circuit(1)
+    c.add(gates.RX(0, params[0]))
+    c.add(gates.RY(0, params[1]))
+
+    initial_params = params.clone()
+    initial_loss = 1 - backend.np.abs(
+        backend.np.sum(
+            backend.np.conj(target_state) * backend.execute_circuit(c).state()
+        )
+    )
```
renatomello, line 26:
```suggestion
    initial_state = backend.execute_circuit(c).state()
    initial_loss = infidelity(initial_state, target_state)
```

BrunoLiegiBastonLiegi:
I think you can use `sys.platform` to check under which system the test is running and skip:

``` python
@pytest.mark.skipif(sys.platform == "win32", reason="Tensorflow not available under windows.")
def test_...
```

src/qibo/models/error_mitigation.py
renatomello: Calculating gradients of immutable objects is both a waste of resources and has no physical meaning.

```In [1]: import numpy as np
   ...: import torch
   ...: 
   ...: from qibo import Circuit, gates, set_backend
   ...: from qibo.backends import PyTorchBackend
   ...: from qibo.quantum_info import infidelity
   ...: 
   ...: set_backend("pytorch")
   ...: backend = PyTorchBackend()
   ...: 
   ...: 
   ...: nqubits, nepochs = 2, 1
   ...: dims = 2**nqubits
   ...: 
   ...: target_state = np.random.rand(dims)
   ...: target_state /= np.linalg.norm(target_state)
   ...: target_state = backend.cast(target_state, requires_grad=True)
   ...: 
   ...: initial_state = np.random.rand(dims)
   ...: initial_state /= np.linalg.norm(initial_state)
   ...: initial_state = backend.cast(initial_state, requires_grad=True)
   ...: 
   ...: params = backend.cast(np.random.rand(nqubits), dtype=float, requires_grad=True)
   ...: 
   ...: optimizer = torch.optim.Adam([params])
   ...: 
   ...: circuit = Circuit(nqubits)
   ...: circuit.add(gates.RX(q, 0.0) for q in range(nqubits))
   ...: 
   ...: for _ in range(nepochs):
   ...:     optimizer.zero_grad()
   ...:     circuit.set_parameters(params)
   ...:     state = circuit(initial_state=initial_state).state()
   ...:     loss = infidelity(state, target_state)
   ...:     loss.backward(retain_graph=True)
   ...:     print(initial_state.grad)
   ...:     print(target_state.grad)
[Qibo 0.2.13|INFO|2024-10-02 10:29:36]: Using pytorch backend on cuda:0
tensor([-0.7136-0.1033j, -0.0798-0.2461j, -0.9491-0.0770j, -1.3825+0.0783j],
       dtype=torch.complex128)
tensor([-0.0810+0.1756j, -0.1235+0.2421j, -1.0581-0.0375j, -1.4742-0.0753j],
       dtype=torch.complex128)
```

BrunoLiegiBastonLiegi: Ok so, for the user it's just a matter of preparing the initial state (which is anyway not a widely used feature in my personal experience) as:
``` python
initial_state = torch.randn(2 ** nqubits, requires_grad=False)
```

In any case, regardless of the philosophical choice between `requires_grad=True` or `False` by default, I believe we should think twice about creating an asymmetry with a widely used backend method such as `backend.cast`

BrunoLiegiBastonLiegi: Yes, that's precisely the asymmetry, you add an argument to the `cast` method that is not present in any other. This means that if you implement a function in `qibo` that needs to cast an array, it will be painless only if the default value is the one you need, otherwise you will have to check if the backend is pytorch, as it happens in this case:
``` python
if backend.name == "pytorch":
    backend.cast(param, requires_grad=True)
else:
    backend.cast(param)
```
Therefore, either we separate the `cast_parameter` and `cast` in all backends, which in most cases, if not all, are just going to coincide, or you rely on some internal variable like the `torch.is_grad_enabled`.

renatomello: Correct me if I'm wrong, but my understanding is that `torch.is_grad_enabled` is about gradients being enabled globally. 
We can't rely on that because what determines if a gate should have grads is its `trainable` attribute (this is important e.g. in seesaw algorithms), and what determines if a specific tensor requires grads is its `requires_grad` attribute.
The solution you're seeing was specifically introduced to remove the enabling of grads globally because this was creating a bunch of issues like the ones in the example I showed.
Creating a `cast_parameter` for every backend does not necessarily solve the issues you're raising because it doesn't cover cases like when a gate like `gates.Unitary` requires grad, or when a statevector / density matrix requires grad.

One way that I could see to cover all issues is the following:

- Remove `requires_grad` from `PyTorchBackend.cast`
- Create `AbstractBackend.cast_parameter`
- Overwrite `cast_parameter` in `PyTorchBackend` (technically already done)
- Explicitly mention in the documentation that if one wants to optimize something that is not a parameterized gate, it should do something like the following:
```python3
backend = PyTorchBackend()

matrix = backend.cast(matrix)
matrix.requires_grad_()
```

BrunoLiegiBastonLiegi: Yes, it is like that, `torch.is_grad_enabled=False` if you are inside a `with torch.no_grad()` context, but I get it that it's a little bit clunky. Ok got it, I am ok with not having gradients enabled by default, but I would still get rid of the `requires_grad` argument of the `cast`, which is set to true just a handful of times
```
doc/source/code-examples/advancedexamples.rst:        torch.rand(2, dtype=torch.float64), requires_grad=True
src/qibo/backends/pytorch.py:            return self.np.tensor(x, dtype=self.parameter_dtype, requires_grad=True)
tests/test_backends_torch_gradients.py:    params = backend.np.rand(2, dtype=backend.np.float64, requires_grad=True)
tests/test_backends_torch_gradients.py:    param = backend.np.tensor([0.1], dtype=backend.np.float64, requires_grad=True)
tests/test_models_variational.py:            initial_parameters, dtype=backend.np.float32, requires_grad=True
tests/test_models_variational.py:            initial_p, dtype=backend.np.float32, requires_grad=True
src/qibo/models/error_mitigation.py:        params.requires_grad = True
src/qibo/models/error_mitigation.py:        params.requires_grad = True
```
and we can easily work around that by moving inside the backend.
@Simone-Bordoni can you please restore this as it was before, namely setting the `requires_grad=True` inside the `_curve_fit` function? Sorry for the confusion. 
I am planning to transform that `_curve_fit` into a backend method in a future PR anyway, such that we don't have to deal with checking the backend used.

src/qibo/models/circuit.py
```diff
@@ -1174,7 +1177,10 @@ def to_qasm(self):
 
             qubits = ",".join(f"q[{i}]" for i in gate.qubits)
             if isinstance(gate, gates.ParametrizedGate):
-                params = (str(x) for x in gate.parameters)
+                if any(x.__class__.__name__ == "Tensor" for x in gate.parameters):
+                    params = (str(x.detach().item()) for x in gate.parameters)
+                else:
+                    params = (str(x) for x in gate.parameters)
```
BrunoLiegiBastonLiegi, line 1183: Simply `str(float(x))` was causing problems with any backend? Because, for `torch` it seems to be fine:
``` python
x = torch.tensor([2.], requires_grad=True)
y = 2 * x **2
str(float(y)) # --> "8.0"
```

tests/test_backends_torch_gradients.py
```diff
@@ -0,0 +1,89 @@
...
+import numpy as np
+import pytest
+
+from qibo import gates, models
+from qibo.quantum_info import infidelity
+
+
+def test_torch_gradients(backend):
+    if backend.name != "pytorch":
+        pytest.skip("Test only valid for PyTorch backend.")
```
BrunoLiegiBastonLiegi, line 12:
```suggestion
def test_torch_gradients():
    backend = PyTorchBackend()
```
no need to use the fixture if only `pytorch` is tested

tests/test_quantum_info_entanglement.py
```diff
@@ -136,11 +136,11 @@ def test_meyer_wallach_entanglement(backend):
     circuit2.add(gates.CNOT(0, 1))
 
     backend.assert_allclose(
-        meyer_wallach_entanglement(circuit1, backend=backend), 0.0, atol=PRECISION_TOL
+        meyer_wallach_entanglement(circuit1, backend=backend), 0.0, atol=1e-6, rtol=1e-6
```
BrunoLiegiBastonLiegi, line 139: why do we need to change the tolerance (reduce?) in a bunch of tests here and below? Is it always related to `pytorch` using `float32`?

BrunoLiegiBastonLiegi (commented):
Thanks @Simone-Bordoni, looks good. The only two things that I am not a fan of are:
- The new `requires_grad` argument added to `cast`. I would remove that and, when possible, move the functions, where it is needed, inside the backend as methods (like the `_curve_fit` case discussed above) . For the cases when it is not possible I would create specific `backend.cast_parameter()`
``` python
# numpy
def cast_parameter(self, x, **kwargs):
    return self.cast(x, **kwargs)
#  torch
def cast_parameter(self, x, **kwargs):
    return self.cast(x, **kwargs).requires_grad_()
```
- The fact that the decomposition needs the backend now. Ideally, I don't think this should be the case, but I don't know in practice if it's possible to avoid.

tests/test_quantum_info_entanglement.py
Simone-Bordoni: Now I should have set torch for using float64 by default. However there are still many cases when it is necessary to set tolerance to 10-6 (usually errors of the order of 10-7 appears). I think this may be related to a slightly different ways to implement the operations.

tests/test_models_hep.py
```diff
@@ -108,4 +108,11 @@ def test_qpdf(backend, ansatz, layers, nqubits, multi_output, output):
...
     params = np.random.rand(model.nparams)
     result = model.predict(params, [0.1])
-    np.testing.assert_allclose(result, output, atol=1e-5)
+    # Pytorch backend has a different tolerance as is by default working with float32
+    if backend.name == "pytorch":
+        atol = 1e-2
+        rtol = 1e-5
+    else:
+        atol = 1e-5
+        rtol = 1e-7
```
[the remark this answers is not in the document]
Simone-Bordoni: Related to the previous comment now the test is passing with the same tolerance so i have removed the "if"

tests/test_gates_gates.py
```diff
@@ -860,6 +860,7 @@ def test_generalized_fsim(backend):
     phi = np.random.random()
     rotation = np.random.random((2, 2)) + 1j * np.random.random((2, 2))
     gatelist = [gates.H(0), gates.H(1), gates.H(2)]
+    gate = gates.GeneralizedfSim(1, 2, rotation, phi)
```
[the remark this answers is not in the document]
Simone-Bordoni: removed, it was a leftover from fixing tests

tests/test_backends_torch_gradients.py
```diff
@@ -0,0 +1,89 @@
...
+    backend.np.manual_seed(42)
+    nepochs = 1001
+    optimizer = backend.np.optim.Adam
```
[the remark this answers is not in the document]
Simone-Bordoni: I prefer to keep the test as simple as possible, LBFGS is a bit different from the standard optimizers.

Simone-Bordoni:
> Thanks @Simone-Bordoni, looks good. The only two things that I am not a fan of are:
> 
> * The new `requires_grad` argument added to `cast`. I would remove that and, when possible, move the functions, where it is needed, inside the backend as methods (like the `_curve_fit` case discussed above) . For the cases when it is not possible I would create specific `backend.cast_parameter()`
> 
> ```python
> # numpy
> def cast_parameter(self, x, **kwargs):
>     return self.cast(x, **kwargs)
> #  torch
> def cast_parameter(self, x, **kwargs):
>     return self.cast(x, **kwargs).requires_grad_()
> ```
> 
> * The fact that the decomposition needs the backend now. Ideally, I don't think this should be the case, but I don't know in practice if it's possible to avoid.

Regarding the decomposition, as it is now you need the backend. 
Regarding the cast_parameter, I can add it to the other backends, however the corrections you are suggesting is not possible as the cast_parameter differs a lot from the cast function:
- cast parameter tries to convert integer to float in torch
- the default dtype for cast parameter is torch float64 (self.parameter_dtype), opposite to cast that has complex 128 (self.dtype) that is used on matrices.
- cast makes different operations on torch trying to make it the same as tensorflow, for example regarding lists of arrays.

So I agree on adding a cast parameter function to all backends but this should not be related to the standard cast()

Simone-Bordoni:
> Thanks @Simone-Bordoni, looks good. The only two things that I am not a fan of are:
> 
> * The new `requires_grad` argument added to `cast`. I would remove that and, when possible, move the functions, where it is needed, inside the backend as methods (like the `_curve_fit` case discussed above) . For the cases when it is not possible I would create specific `backend.cast_parameter()`
> 
> ```python
> # numpy
> def cast_parameter(self, x, **kwargs):
>     return self.cast(x, **kwargs)
> #  torch
> def cast_parameter(self, x, **kwargs):
>     return self.cast(x, **kwargs).requires_grad_()
> ```
> 
> * The fact that the decomposition needs the backend now. Ideally, I don't think this should be the case, but I don't know in practice if it's possible to avoid.

Regarding the decomposition, as it is now you need the backend. 
Regarding the cast_parameter, I can add it to the other backends, however the corrections you are suggesting is not possible as the cast_parameter differs a lot from the cast function:
- cast parameter tries to convert integer to float in torch
- the default dtype for cast parameter is torch float64 (self.parameter_dtype), opposite to cast that has complex 128 (self.dtype) that is used on matrices.
- cast makes different operations on torch trying to make it the same as tensorflow, for example regarding lists of arrays.

So I agree on adding a cast parameter function to all backends but this should not be related to the standard cast()

Simone-Bordoni:
> Regarding the decomposition, as it is now you need the backend. Regarding the cast_parameter, I can add it to the other backends, however the corrections you are suggesting is not possible as the cast_parameter differs a lot from the cast function:
> 
> * cast parameter tries to convert integer to float in torch
> * the default dtype for cast parameter is torch float64 (self.parameter_dtype), opposite to cast that has complex 128 (self.dtype) that is used on matrices.
> * cast makes different operations on torch trying to make it the same as tensorflow, for example regarding lists of arrays.
> 
> So I agree on adding a cast parameter function to all backends but this should not be related to the standard cast()

Actually cast_parameter() was made to cast a single parameter for matrix_parametrized().
An idea could be making it a helper method in the torch backend and creating a new public mthod like cast_parameters() to be used in these situations.
Anyway I don't know if @renatomello agrees on adding this new function to the numpy backend to avoid different backend behaviours.

src/qibo/models/circuit.py
Simone-Bordoni: This is raising an error in test_models_circuit_qasm_cirq.py

BrunoLiegiBastonLiegi: I cannot reproduce the error
``` python
params = (str(float(x)) for x in gate.parameters)
```
this works for me locally, not sure, is the test random? Maybe there are just some random configurations causing problems

Simone-Bordoni: I get this error:
FAILED test_models_circuit_qasm_cirq.py::test_parametrized_gate_cirq[pytorch] - cirq.contrib.qasm_import.exception.QasmException: Syntax error: ','
There should not be anything random

BrunoLiegiBastonLiegi: weird, it does not fail for me, I just checked again. Does it fail here in the CI or locally for you?
@renatomello does it fail for you as well?

Simone-Bordoni: it fails locally, the parameter becomes: "torch.Tensor([....], gradient....)"

Simone-Bordoni: But it also fails here, look at the commits before restoring it.
Screenshot from 2024-10-04 12-38-31

renatomello (approved):

src/qibo/models/error_mitigation.py
BrunoLiegiBastonLiegi: @Simone-Bordoni can you restore this as it was in the beginning, i.e. casting to parameter for `pytorch`inside the `_curve_fit` funcion?

BrunoLiegiBastonLiegi (approved):

PR #1451, opened 2024-09-17
Extend `meyer_wallach_entanglement` to `ndarrays`

AlejandroSopena:
Checklist:
- [x] Tests are passing.


src/qibo/quantum_info/entanglement.py
```diff
@@ -280,7 +294,8 @@ def entangling_capability(circuit, samples: int, seed=None, backend=None):
     for _ in range(samples):
         params = local_state.uniform(-np.pi, np.pi, circuit.trainable_gates.nparams)
         circuit.set_parameters(params)
-        entanglement = meyer_wallach_entanglement(circuit, backend=backend)
+        state = backend.execute_circuit(circuit).state()
```
BrunoLiegiBastonLiegi, line 297: This is going to fail when `circuit.density_matrix=False` and are present either noise channels or collapsing measurements, which is ok but maybe we should point it out in the docstring.

BrunoLiegiBastonLiegi (approved):

src/qibo/quantum_info/entanglement.py
renatomello: I fixed the bug for `density_matrix=True` and added a note for channels.

renatomello (approved):

PR #1452, opened 2024-09-17
Add `total_variation_distance` to `quantum_info.utils`

renatomello:
Checklist:
- [x] Tests are passing.


src/qibo/quantum_info/utils.py
```diff
@@ -339,6 +339,68 @@ def hellinger_shot_error(
...
+    Returns:
+        float: Total variation distance measure.
+    """
+    backend = _check_backend(backend)
+
+    if isinstance(prob_dist_p, list):
+        prob_dist_p = backend.cast(prob_dist_p, dtype=np.float64)
+
+    if isinstance(prob_dist_q, list):
+        prob_dist_q = backend.cast(prob_dist_q, dtype=np.float64)
```
BrunoLiegiBastonLiegi, line 373: I don't know whether the cast is needed in the validation process, but `backend.calculate_norm` is casting already by default.

src/qibo/quantum_info/utils.py
```diff
@@ -339,6 +339,68 @@ def hellinger_shot_error(
...
+
+    if (len(prob_dist_p.shape) != 1) or (len(prob_dist_q.shape) != 1):
+        raise_error(
+            TypeError,
+            "Probability arrays must have dims (k,) but have "
+            + f"dims {prob_dist_p.shape} and {prob_dist_q.shape}.",
+        )
+
+    if (len(prob_dist_p) == 0) or (len(prob_dist_q) == 0):
+        raise_error(TypeError, "At least one of the arrays is empty.")
```
BrunoLiegiBastonLiegi, line 383: Isn't it going to fail anyway when you move to the norm calculation? Just saying, because I am not sure whether it is worth checking for all the possible mistakes an user could make. Surely, the error is more explicit this way, but even  a mismatching shape error raised by `numpy` & co should be clear enough.

Let's say that delegating others to raise the error would gain us some readability both here and in the test below.

renatomello: That would require a bigger refactoring since the entire module has these types of checks, which is out of the scope of this PR

src/qibo/quantum_info/utils.py
renatomello: It is needed because the checks are using the `shape` attribute that lists do not have

src/qibo/quantum_info/utils.py
BrunoLiegiBastonLiegi: ok but why to introduce more of them here that we will want to refactor later?

AlejandroSopena (approved):

BrunoLiegiBastonLiegi (approved):

PR #1454, opened 2024-09-18
Add `matrix_power` to `quantum_info.linalg_operations`

renatomello:
Needs https://github.com/qiboteam/qibojit/pull/189

Checklist:
- [x] Tests are passing.


src/qibo/backends/numpy.py
```diff
@@ -768,6 +769,14 @@ def calculate_matrix_exp(self, a, matrix, eigenvectors=None, eigenvalues=None):
         ud = self.np.transpose(np.conj(eigenvectors))
         return self.np.matmul(eigenvectors, self.np.matmul(expd, ud))
 
+    def calculate_matrix_power(self, matrix, power: Union[float, int]):
+        if not isinstance(power, (float, int)):
+            raise_error(
+                TypeError,
+                f"``power`` must be either float or int, but it is type {type(power)}.",
+            )
```
BrunoLiegiBastonLiegi, line 777: I think `fractional_matrix_power` would raise an error anyway.
```suggestion
```

src/qibo/backends/pytorch.py
```diff
@@ -195,6 +195,11 @@ def calculate_matrix_exp(self, a, matrix, eigenvectors=None, eigenvalues=None):
         ud = self.np.conj(eigenvectors).T
         return self.np.matmul(eigenvectors, self.np.matmul(expd, ud))
 
+    def calculate_matrix_power(self, matrix, power):
+        copied = self.to_numpy(self.np.copy(matrix))
+        copied = super().calculate_matrix_power(copied, power)
+        return self.cast(copied, dtype=copied.dtype)
```
BrunoLiegiBastonLiegi, line 201: This probably brakes gradient flow with `pytorch`, but unfortunately there is no way of avoiding it unless we implement fractional exponentiation on our side...

src/qibo/quantum_info/linalg_operations.py
```diff
@@ -196,3 +196,28 @@ def matrix_exponentiation(
...
+
+    Returns:
+        ndarray: matrix power :math:`A^{\\alpha}`.
+    """
+
+    if not isinstance(power, (float, int)):
+        raise_error(
+            TypeError,
+            f"``power`` must be either float or int, but it is type {type(power)}.",
+        )
```
BrunoLiegiBastonLiegi, line 219: If we decide to keep this check, I would keep it only in the backend function, I don't think you need redundancy here.

src/qibo/backends/pytorch.py
renatomello: The copy is needed specifically because `torch` doesn't allow the `scipy` function to work on a `tensor` that requires gradients. That's why we need to calculate the power on a copy. I'm not sure if the process of copying breaks the graph of the original tensor. Moreover, I'm not expecting this function to be used on something that requires optimization, unless there's some niche case that I'm not aware of.

src/qibo/quantum_info/linalg_operations.py
renatomello: This goes in the same basket as the refactoring comment of the other PR

src/qibo/backends/pytorch.py
BrunoLiegiBastonLiegi: It's not much the copy, but rather the conversion to numpy. scipy functions fail with torch tensors that require gradients because they internally try to convert them to numpy arrays, which is not possible (unless you force it) when they have `requires_grad=True`. But yeah, regardless, in this case there's not much you can do except implementing the whole fractional power exponentiation with torch natives. However, I would maybe add a comment or docstring remind us that this breaks gradient flow.

src/qibo/quantum_info/linalg_operations.py
BrunoLiegiBastonLiegi: Besides the refactor, here you have the same identical check that happens in two functions that are concatenated:
``` python
def internal_function(...):
    check(...)
    ...

def external_f(...):
    check(...)
    return internal_function(...)
```
the check will always happen even if you remove it from the external function, i.e. this one here.

tests/test_quantum_info_operations.py
```diff
@@ -109,3 +111,16 @@ def test_partial_trace(backend, density_matrix):
...
+    nqubits, power = 2, 2
+    dims = 2**nqubits
+
+    state = random_density_matrix(dims, backend=backend)
+
+    power = matrix_power(state, power, backend)
+
+    backend.assert_allclose(
+        float(backend.np.real(backend.np.trace(power))), purity(state, backend=backend)
+    )
```
BrunoLiegiBastonLiegi, line 126:
```suggestion
@pytest.mark.parametrize(power, [2, 2., "2"])
def test_matrix_power(backend, power):
    nqubits = 2
    dims = 2**nqubits

    state = random_density_matrix(dims, backend=backend)

    if isinstance(power, str):
        with pytest.raises(TypeError):
            return matrix_power(state, power, backend)

    power = matrix_power(state, power, backend)

    backend.assert_allclose(
        float(backend.np.real(backend.np.trace(power))), purity(state, backend=backend)
    )
```
this should fix the coverage, if we want to keep the check

renatomello:
@scarrazza this is ready on my side and the corresponding `qibojit` PR was already merged. Tests and coverage are passing.

doc/source/api-reference/qibo.rst
```diff
@@ -1964,6 +1964,12 @@ Matrix exponentiation
 .. autofunction:: qibo.quantum_info.matrix_exponentiation
 
 
+Matrix power
```
MatteoRobbiati, line 1967: Compiling the documentation I cannot see any content under the "Matrix Power" title.

src/qibo/backends/numpy.py
MatteoRobbiati: I agree with @BrunoLiegiBastonLiegi.

src/qibo/backends/abstract.py
```diff
@@ -355,6 +356,21 @@ def calculate_matrix_exp(
         """
         raise_error(NotImplementedError)
 
+    @abc.abstractmethod
+    def calculate_matrix_power(
+        self, matrix, power: Union[float, int]
+    ):  # pragma: no cover
+        """Calculates the (fractional) ``power`` :math:`\\alpha` of ``matrix`` :math:`A`,
```
MatteoRobbiati, line 363:
```suggestion
        """Calculate the (fractional) ``power`` :math:`\\alpha` of ``matrix`` :math:`A`,
```

MatteoRobbiati: I usually refer to [PEP 257](https://peps.python.org/pep-0257/#one-line-docstrings).

src/qibo/quantum_info/linalg_operations.py
```diff
@@ -196,3 +196,21 @@ def matrix_exponentiation(
     backend = _check_backend(backend)
 
     return backend.calculate_matrix_exp(phase, matrix, eigenvectors, eigenvalues)
+
+
+def matrix_power(matrix, power: Union[float, int], backend=None):
+    """Given a ``matrix`` :math:`A` and power :math:`\\alpha`, calculates :math:`A^{\\alpha}`.
```
MatteoRobbiati, line 202:
```suggestion
    """Given a ``matrix`` :math:`A` and power :math:`\\alpha`, calculate :math:`A^{\\alpha}`.
```

MatteoRobbiati (commented):
Thanks a lot. Everything looks good, I just leave some comments below.

doc/source/api-reference/qibo.rst
renatomello: I see it locally

Screenshot from 2024-09-20 14-39-49

renatomello: I see it locally

Screenshot from 2024-09-20 14-39-49
Screenshot from 2024-09-20 14-41-23

MatteoRobbiati (approved):
LGTM

doc/source/api-reference/qibo.rst
MatteoRobbiati: Ok, fine. If @BrunoLiegiBastonLiegi you were able to compile it, i'm fine with the merge.

PR #1459, opened 2024-09-20
Optimizations for the `qibo.quantum_info.basis.pauli_basis` and `vectorization` function

BrunoLiegiBastonLiegi:
This improves the construction of the pauli basis by moving everything to tensor notation and removing loops. Namely the `basis_full` is now constructed via contraction through `einsum`. This should also scale well with GPU backends, weirdly for standard numpy CPU there is no speedup.
These are the results:
```
nqubits    old (numpy)    new (numpy)    new (cupy)
4            ~ 1s            ~ 1s          ~ 2.7s
7           ~ 9.5s          ~ 9.1s         ~ 3.5s
8            killed          memory         memory
```
The GPU always takes ~1s to set up. With 8 qubits, `old` gets killed together with the shell session, whereas `new`, both CPU and GPU, raise an out of memory error. To run 8 qubits you apparently need ~64GB of memory.

<s>To perform the einsum I use all the 48 ascii characters available, which means that we are limited to `48/3=16` qubits, unless other characters can be used in einsum. In any case, the memory requirements for 16 qubits are probably going to be very taxing</s> I am using integers as the indices for the `einsum` now, thus this is not limited anymore by the number of ascii charachters. It may be possible to obtain a speedup with numba as well, but I still have to investigate.

EDIT 1: unfortunately `einsum` is not supported by `numba`, however if you `jit` the old implementation:
``` python
            basis_single = list(product(basis_single, repeat=nqubits))

            @njit(fastmath=True, parallel=True, cache=True)
            def _build_basis(basis_single):
                rows = [basis_single[0][0] for i in range(len(basis_single))]
                for i in prange(len(basis_single)):
                    row = basis_single[i][0]
                    for j in range(len(basis_single[i][1:])):
                        row = np.kron(row, basis_single[i][j])
                    rows[i] = row
                return rows

            basis_full = backend.cast(_build_basis(basis_single))
```
<s>you are able to get a nice speedup ~ 5s for 7 qubits, however this happens from the second call, since the first time you are still dealing with the compilation</s> unfortunately this implementation was yielding wrong results, thus I had to rollback to the `einsum` aprroach. Further investigation is needed to understand if it's possible to parallelize and improve this with `numba`.

EDIT 2:
At some point I realized that a possible bottleneck, or better inefficiency, was due to the `qibo.quantum_info.superoperator_transformation.vectorization` that allowed to be run on a single input only, either a state vector or a density matrix, thus forcing to run loops on each element of the basis (which can grow large very quickly). The impact on the runtime is still marginal for the cpu as for 7 qubits its contribute was around ~1-2s out 10s, but for GPUs this starts becoming relevant. Furthermore, `vectorization` appears to be widely used in the `quantum_info` module, which convinced me to generalize it to accept batches of state vectors or density matrices, lifting therefore the need of explicit loops and rather leveraging tensor primitives directly. This was applied to the `pauli_basis` for now but has to be propagated throughout the whole module.

Checklist:
- [x] Tests are passing.


alecandido:
@BrunoLiegiBastonLiegi you don't have to use ascii characters for `einsum`

> [`einsum`](https://numpy.org/doc/stable/reference/generated/numpy.einsum.html#numpy.einsum) also provides an alternative way to provide the subscripts and operands as `einsum(op0, sublist0, op1, sublist1, ..., [sublistout])`. If the output shape is not provided in this format [`einsum`](https://numpy.org/doc/stable/reference/generated/numpy.einsum.html#numpy.einsum) will be calculated in implicit mode, otherwise it will be performed explicitly. The examples below have corresponding [`einsum`](https://numpy.org/doc/stable/reference/generated/numpy.einsum.html#numpy.einsum) calls with the two parameter methods.

https://numpy.org/doc/stable/reference/generated/numpy.einsum.html#numpy-einsum

In this way, you can use integers instead of characters, that is much better if you are generating the `einsum` input (and you're not limited by the amount of characters).

renatomello:
One way we could move forward is to return `generators` for `n >= 8` and let the user do its memory management by accessing the elements on the fly.

BrunoLiegiBastonLiegi:
> One way we could move forward is to return `generators` for `n >= 8` and let the user do its memory management by accessing the elements on the fly.

mmh yeah, that's a possibility, but only in certain cases. For the sake of parallelization, surely having everything represented by a single tensor contraction is better, but you are limited by memory. 
Maybe, what we could do is defining the generator of a single element of the basis for each backend:
``` python
@cache
def _pauli_basis_element(self, i, nqubits):
    # do things according to backend
```
and then you can build the complete basis as a generator as you suggested 
``` python
def pauli_basis(...):
    return (backend._pauli_basis_element(i, nqubits) for i in range(len))
```
but GPU wise this will be less efficient and it will work, anyway, as long as you don't need the complete basis at the same time.

renatomello:
> > One way we could move forward is to return `generators` for `n >= 8` and let the user do its memory management by accessing the elements on the fly.
> 
> mmh yeah, that's a possibility, but only in certain cases. For the sake of parallelization, surely having everything represented by a single tensor contraction is better, but you are limited by memory. Maybe, what we could do is defining the generator of a single element of the basis for each backend:
> 
> ```python
> @cache
> def _pauli_basis_element(self, i, nqubits):
>     # do things according to backend
> ```
> 
> and then you can build the complete basis as a generator as you suggested
> 
> ```python
> def pauli_basis(...):
>     return (backend._pauli_basis_element(i, nqubits) for i in range(len))
> ```
> 
> but GPU wise this will be less efficient and it will work, anyway, as long as you don't need the complete basis at the same time.

The Hilbert space is too big, unfortunately there's no way around that. I wouldn't spend much time on this since there are bigger priorities in terms of optimization.

BrunoLiegiBastonLiegi:
In the end I was not able to make the `numba` parallelized implementation to work, it was working with `parallel=False` but there was no speed up in that case. Thus I just rolled back to using `einsum` in every case. We are still getting a nice scaling with GPUs. I will get rid of the ascii charachters now.

src/qibo/quantum_info/basis.py
```diff
@@ -93,15 +92,18 @@ def pauli_basis(
 
     pauli_labels = {"I": matrices.I, "X": matrices.X, "Y": matrices.Y, "Z": matrices.Z}
     basis_single = backend.cast([pauli_labels[label] for label in pauli_order])
+    if backend.name == "tensorflow":
+        einsum = np.einsum
+    else:
+        einsum = backend.np.einsum
```
renatomello, line 98:
```suggestion
    einsum = (
        np.einsum 
        if backend.name == "tensorflow" 
        else backend.np.einsum
    )
```

BrunoLiegiBastonLiegi:
Ok I ended up also adding a generalization to the `qibo.quantum_info.superoperator_transformation.vectorization`, which now accepts batches of state vectors or density matrices as well, thus allowing to vectorize the whole basis in one shot rather than looping. In the process, I found a weird behaviour of `tensorflow`. Namely, if you have a complex tensor the elements with zero real part, but still non zero imaginary part, are not recovered by `nonzero`:
``` python
import tensorflow as tf
import tensorflow.experimental.numpy as tnp 

a = tf.Variable([0 + 1j, 1 + 0j, 1 + 1j])
tnp.nonzero(a)
# this finds the last two only
# [<tf.Tensor: shape=(2,), dtype=int64, numpy=array([1, 2])>]
```

src/qibo/quantum_info/basis.py
```diff
@@ -92,43 +91,42 @@ def pauli_basis(
...
 
     pauli_labels = {"I": matrices.I, "X": matrices.X, "Y": matrices.Y, "Z": matrices.Z}
-    basis_single = [pauli_labels[label] for label in pauli_order]
+    basis_single = backend.cast([pauli_labels[label] for label in pauli_order])
+    einsum = np.einsum if backend.name == "tensorflow" else backend.np.einsum
 
     if nqubits > 1:
-        basis_full = list(product(basis_single, repeat=nqubits))
-        basis_full = [reduce(np.kron, row) for row in basis_full]
+        dim = 2**nqubits
```
renatomello, line 98:
```suggestion
```

src/qibo/quantum_info/basis.py
```diff
@@ -92,43 +91,42 @@ def pauli_basis(
...
+        input_indices = [range(3 * i, 3 * (i + 1)) for i in range(nqubits)]
+        output_indices = (i for indices in zip(*input_indices) for i in indices)
+        operands = [basis_single for _ in range(nqubits)]
+        inputs = [item for pair in zip(operands, input_indices) for item in pair]
+        basis_full = einsum(*inputs, output_indices).reshape(4**nqubits, dim, dim)
     else:
+        dim = 2
```
renatomello, line 105:
```suggestion
```

src/qibo/quantum_info/basis.py
```diff
@@ -92,43 +91,42 @@ def pauli_basis(
     backend = _check_backend(backend)
 
     pauli_labels = {"I": matrices.I, "X": matrices.X, "Y": matrices.Y, "Z": matrices.Z}
-    basis_single = [pauli_labels[label] for label in pauli_order]
+    basis_single = backend.cast([pauli_labels[label] for label in pauli_order])
+    einsum = np.einsum if backend.name == "tensorflow" else backend.np.einsum
```
renatomello, line 95:
```suggestion
    dim = 2**nqubits
    basis_single = backend.cast([pauli_labels[label] for label in pauli_order])
    einsum = np.einsum if backend.name == "tensorflow" else backend.np.einsum
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -26,8 +26,9 @@ def vectorization(state, order: str = "row", backend=None):
     .. math::
         |\\rho) = \\sum_{k, l} \\, \\rho_{kl} \\, \\ket{l} \\otimes \\ket{k}
 
+    If ``state`` is a 3-dimensional tensor it is interpreted as a batch of states.
     Args:
-        state: state vector or density matrix.
+        state: state vector, density matrix or batch of those.
```
renatomello, line 31:
```suggestion
        state: statevector, density matrix, batches of statevectors, or batches of density matrices.
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -40,13 +41,13 @@ def vectorization(state, order: str = "row", backend=None):
...
     if (
-        (len(state.shape) >= 3)
+        (len(state.shape) > 3)
         or (len(state) == 0)
         or (len(state.shape) == 2 and state.shape[0] != state.shape[1])
     ):
         raise_error(
             TypeError,
-            f"Object must have dims either (k,) or (k,k), but have dims {state.shape}.",
+            f"Object must have dims either (k,), (k, k), (n, 1, k) or (n, k, k), but have dims {state.shape}.",
```
renatomello, line 50:
```suggestion
            f"Object must have dims either (k,), (k, k), (N, 1, k) or (N, k, k), but have dims {state.shape}.",
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -64,23 +65,32 @@ def vectorization(state, order: str = "row", backend=None):
 
     if len(state.shape) == 1:
         state = backend.np.outer(state, backend.np.conj(state))
+    elif len(state.shape) == 3 and state.shape[1] == 1:
+        state = backend.np.einsum(
+            "aij,akl->aijkl", state, backend.np.conj(state)
+        ).reshape(state.shape[0], state.shape[-1], state.shape[-1])
```
renatomello, line 71:
```suggestion
    dims = state.shape[-1]

    if len(state.shape) == 1:
        state = backend.np.outer(state, backend.np.conj(state))
    elif len(state.shape) == 3 and state.shape[1] == 1:
        state = backend.np.einsum(
            "aij,akl->aijkl", state, backend.np.conj(state)
        ).reshape(state.shape[0], dims, dims)
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -64,23 +65,32 @@ def vectorization(state, order: str = "row", backend=None):
...
 
     if order == "row":
-        state = backend.np.reshape(state, (1, -1))[0]
+        state = backend.np.reshape(state, (-1, state.shape[-1] ** 2))
```
renatomello, line 74:
```suggestion
        state = backend.np.reshape(state, (-1, dims**2))
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -64,23 +65,32 @@ def vectorization(state, order: str = "row", backend=None):
...
     elif order == "column":
-        state = state.T
-        state = backend.np.reshape(state, (1, -1))[0]
+        indices = list(range(len(state.shape)))
+        indices[-2:] = reversed(indices[-2:])
+        state = backend.np.transpose(state, indices)
+        state = backend.np.reshape(state, (-1, state.shape[-1] ** 2))
```
renatomello, line 79:
```suggestion
        state = backend.np.reshape(state, (-1, dims**2))
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -64,23 +65,32 @@ def vectorization(state, order: str = "row", backend=None):
...
+        new_axis = [0]
         for qubit in range(nqubits):
-            new_axis += [qubit + nqubits, qubit]
+            new_axis.extend([qubit + nqubits + 1, qubit + 1])
 
-        state = backend.np.reshape(state, [2] * 2 * nqubits)
+        state = backend.np.reshape(state, [-1] + [2] * 2 * nqubits)
         state = backend.np.transpose(state, new_axis)
-        state = backend.np.reshape(state, (-1,))
+        state = backend.np.reshape(state, (-1, 2 ** (nqubits * 2)))
```
renatomello, line 89:
```suggestion
        state = backend.np.reshape(state, (-1, 2 ** (2 * nqubits)))
```

tests/test_quantum_info_basis.py
```diff
@@ -51,6 +51,8 @@ def test_pauli_basis(
         basis_test = list(product(basis_test, repeat=nqubits))
         basis_test = [reduce(np.kron, matrices) for matrices in basis_test]
 
+    # if order == "system" and nqubits ==2 and vectorize:
+    #    breakpoint()
```
renatomello, line 55:
```suggestion
```

renatomello (approved):

renatomello:
@stavros11 since you looked into #1462, could you look into this too? Thanks.

stavros11 (approved):

PR #1462, opened 2024-09-23
Add `partial_transpose` to `quantum_info.linalg_operations`

renatomello:
Needs to be merged before #1456 

Checklist:
- [x] Tests are passing.

### DIFF
--- a/tests/test_quantum_info_operations.py
+++ b/tests/test_quantum_info_operations.py
@@ -137,12 +137,18 @@ def _werner_state(p, backend):
     return state, transposed
 
 
+@pytest.mark.parametrize("batch", [False, True])
 @pytest.mark.parametrize("statevector", [False, True])
 @pytest.mark.parametrize("p", [1 / 5, 1 / 3, 1.0])
-def test_partial_transpose(backend, p, statevector):
+def test_partial_transpose(backend, p, statevector, batch):
     with pytest.raises(ValueError):
         state = random_density_matrix(3, backend=backend)
         test = partial_transpose(state, [0], backend)
+    with pytest.raises(TypeError):
+        state = np.random.rand(2, 2, 2, 2).astype(complex)
+        state += 1j * np.random.rand(2, 2, 2, 2)
+        state = backend.cast(state, dtype=state.dtype)
+        test = partial_transpose(state, [1], backend=backend)
 
     zero, one = np.array([1, 0], dtype=complex), np.array([0, 1], dtype=complex)
     psi = (np.kron(zero, one) - np.kron(one, zero)) / np.sqrt(2)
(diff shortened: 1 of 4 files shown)


BrunoLiegiBastonLiegi (commented):
I think it might be useful to allow the inputs to both `partial_transpose` and `partial_trace` (this one in a separate PR though) to be batches of states, similarly to what it is done for `vectorization` in #1459. This would make more efficient, for instance, the calculation of the `meyer_wallach_entanglement` 
https://github.com/qiboteam/qibo/blob/165b2046b169c70426c780b7035403bba10807e3/src/qibo/quantum_info/entanglement.py#L196
and similar methods that need to perform these operations on several different states.

renatomello:
> I think it might be useful to allow the inputs to both `partial_transpose` and `partial_trace` (this one in a separate PR though) to be batches of states, similarly to what it is done for `vectorization` in #1459. This would make more efficient, for instance, the calculation of the `meyer_wallach_entanglement`
> 
> https://github.com/qiboteam/qibo/blob/165b2046b169c70426c780b7035403bba10807e3/src/qibo/quantum_info/entanglement.py#L196
> 
> and similar methods that need to perform these operations on several different states.

Done.

tests/test_quantum_info_operations.py
```diff
@@ -113,6 +114,98 @@ def test_partial_trace(backend, density_matrix):
...
+
+        # partial transpose of two-qubit werner state is known analytically
+        target = (1 / 4) * np.array(
+            [
+                [1 - p, 0, 0, -2 * p],
+                [0, p + 1, 0, 0],
+                [0, 0, p + 1, 0],
+                [-2 * p, 0, 0, 1 - p],
+            ],
+            dtype=complex,
```
BrunoLiegiBastonLiegi, line 195:
```suggestion
        state, target = _werner_state(p, backend)
```
is this supposed to be the `_werner_state` above?

tests/test_quantum_info_operations.py
```diff
@@ -113,6 +114,98 @@ def test_partial_trace(backend, density_matrix):
...
+        )
+        target = backend.cast(target, dtype=target.dtype)
+
+        transposed = partial_transpose(state, [1], backend)
+
+        if batch:
+            print(transposed.shape)
```
BrunoLiegiBastonLiegi, line 202:
```suggestion
```

renatomello:
@AlejandroSopena @stavros11 this is ready. Could you review it soon since it is blocking #1456 ?

stavros11 (approved):
Thanks @renatomello. Generally fine with me, but I have not checked validity so I just trust the test.

It took me some time to understand the supported input shapes and I think we are kind of overcomplicating it by supporting 1-dimensional arrays. To me, since this is a partial trace, it would even be sufficient to stick to 2-dimensional. If someone wants to process a 1-dimensional one, they can still do the outer product before calling this (this way it is also a bit more modular). However, I am not the best candidate user of this, so it's not a strong opinion.

Other than that, I understand you are also supporting a "batch" dimension (`N`). There may be a way to generalize this to arbitrary number of batch dimensions (`(..., k, k)`), so that you don't even have to distinguish between batch/no batch, but I haven't thought much about whether this can be easily done and maybe it's not very useful in the end.

renatomello:
> Thanks @renatomello. Generally fine with me, but I have not checked validity so I just trust the test.
> 
> It took me some time to understand the supported input shapes and I think we are kind of overcomplicating it by supporting 1-dimensional arrays. To me, since this is a partial trace, it would even be sufficient to stick to 2-dimensional. If someone wants to process a 1-dimensional one, they can still do the outer product before calling this (this way it is also a bit more modular). However, I am not the best candidate user of this, so it's not a strong opinion.

I think it's fine supporting statevectors because it takes an extra step from the users' hands at the price of a handful of lines of code on our side. So the trade-off is worth IMO.
 
> Other than that, I understand you are also supporting a "batch" dimension (`N`). There may be a way to generalize this to arbitrary number of batch dimensions (`(..., k, k)`), so that you don't even have to distinguish between batch/no batch, but I haven't thought much about whether this can be easily done and maybe it's not very useful in the end.

I like the idea, but for the interest of time I prefer to move forward as is.

Simone-Bordoni (approved):

PR #1464, opened 2024-09-24
Remove deprecation warning, remove old `draw`, and rename `display` as `draw` in `Circuit` class

renatomello:
Checklist:
- [x] Tests are passing.


alecandido:
@renatomello is it here because you want to have it directly in the next release, or is it here just not to forget?

In principle, we should collect breaking changes for 0.3. I would agree that we may be a bit more lenient, but just one/two months after seems a very short notice (especially given the amount of time `.draw()` has been there since...).

renatomello:
> @renatomello is it here because you want to have it directly in the next release, or is it here just not to forget?
> 
> In principle, we should collect breaking changes for 0.3. I would agree that we may be a bit more lenient, but just one/two months after seems a very short notice (especially given the amount of time `.draw()` has been there since...).

The current warning says the change would happen in 0.2.13.

alecandido:
> The current warning says the change would happen in 0.2.13.

Ok, I've not been careful in reviewing that...

It's a small thing, and updating broken third-party code because of that won't be a massive effort. We can make an exception...

Still, I'd like to increasingly improve our stability guarantees. We should definitely lower the number of semver exceptions abruptly, getting to zero as fast as possible...

stavros11 (approved):
Note that `circuit.draw()` (current `circuit.display()`) is missing a newline `\n` at the end. This leads to a bit weird behavior when using in a Python interpreter or if you have another print after it, for example
```py
circuit.draw()
print("something else...")
```
It is also different than `print(str(circuit))` which will include the default Python newline at the end of every `print`.

renatomello:
@stavros11 can you confirm the right behavior before I merge this?

stavros11 (approved):
> @stavros11 can you confirm the right behavior before I merge this?

Yes, now the new line is added. Thanks @renatomello.

PR #1465, opened 2024-09-25
Remove circular dependency between `MeasurementResult` and `M` gate

### LINKED ISSUE #1433: Cyclic dependence of gate `M` and `MeasurementResult`
Currently the gate `M` and the `MeasurementResult` object cyclically depend on each other:
https://github.com/qiboteam/qibo/blob/f20f4330fe62ce7bc022b687cf6559f2fbc1b50c/src/qibo/gates/measurements.py#L59
https://github.com/qiboteam/qibo/blob/f20f4330fe62ce7bc022b687cf6559f2fbc1b50c/src/qibo/measurements.py#L83

I would argue that this is not really necessary and introduces unneeded complications. The `MeasurementResult`, indeed, appears to be using the `.measurement_gate` just to retrieve the qubits (`.measurement_gate.target_qubits`). For this reason, I would propose to remove this dependence and, for instance, build a `MeasurementResult` object just starting from the measured qubit indices, rather than the whole measurement gate (as discussed in #1431 already).

BrunoLiegiBastonLiegi:
This should close #1433

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/gates/measurements.py
+++ b/src/qibo/gates/measurements.py
@@ -57,7 +57,7 @@ class M(Gate):
         self.target_qubits = tuple(q)
         self.register_name = register_name
         self.collapse = collapse
-        self.result = MeasurementResult(self)
+        self.result = MeasurementResult(self.target_qubits)
         # list of measurement pulses implementing the gate
         # relevant for experiments only
         self.pulses = None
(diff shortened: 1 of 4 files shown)


stavros11 (approved):
Thanks @BrunoLiegiBastonLiegi, good catch.

MatteoRobbiati (approved):
Thanks!

PR #1470, opened 2024-10-03
Add `relative_tsallis_entropy` to `quantum_info.entropies`

renatomello:
Checklist:
- [ ] Reviewers confirm new code works as expected.
- [x] Tests are passing.
- [x] Coverage does not decrease.
- [x] Documentation is updated.

PR #1472, opened 2024-10-04
Fix bug in `quantum_info.linalg_operations.matrix_power` for negative powers of singular matrices

renatomello:
Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/backends/numpy.py
+++ b/src/qibo/backends/numpy.py
@@ -501,8 +501,11 @@ class NumpyBackend(Backend):
             and not circuit.measurements
             and not circuit.density_matrix
         ):
-            raise RuntimeError(
-                "The circuit contains only collapsing measurements (`collapse=True`) but `density_matrix=False`. Please set `density_matrix=True` to retrieve the final state after execution."
+            raise_error(
+                RuntimeError,
+                "The circuit contains only collapsing measurements (`collapse=True`) but "
+                + "`density_matrix=False`. Please set `density_matrix=True` to retrieve "
+                + "the final state after execution.",
             )
 
         results, final_states = [], []
--- a/src/qibo/backends/numpy.py
+++ b/src/qibo/backends/numpy.py
@@ -784,9 +784,11 @@ class NumpyBackend(Backend):
         if power < 0.0:
             # negative powers of singular matrices via SVD
             # det with np instead of self.np because of torch
-            determinant = np.linalg.det(matrix)
+            determinant = self.np.linalg.det(matrix)
             if abs(determinant) < precision_singularity:
                 U, S, Vh = self.np.linalg.svd(matrix)
+                # cast needed because of different dtypes in `torch`
+                S = self.cast(S)
                 S_inv = self.np.where(
                     self.np.abs(S) < precision_singularity, 0.0, S**power
                 )
--- a/src/qibo/backends/numpy.py
+++ b/src/qibo/backends/numpy.py
@@ -783,7 +783,6 @@ class NumpyBackend(Backend):
 
         if power < 0.0:
             # negative powers of singular matrices via SVD
-            # det with np instead of self.np because of torch
             determinant = self.np.linalg.det(matrix)
             if abs(determinant) < precision_singularity:
                 U, S, Vh = self.np.linalg.svd(matrix)
--- a/src/qibo/backends/numpy.py
+++ b/src/qibo/backends/numpy.py
@@ -415,8 +415,12 @@ class NumpyBackend(Backend):
             if circuit.measurements or circuit.has_collapse:
                 return self.execute_circuit_repeated(circuit, nshots, initial_state)
             else:
-                raise RuntimeError(
-                    "Attempting to perform noisy simulation with `density_matrix=False` and no Measurement gate in the Circuit. If you wish to retrieve the statistics of the outcomes please include measurements in the circuit, otherwise set `density_matrix=True` to recover the final state."
+                raise_error(
+                    RuntimeError,
+                    "Attempting to perform noisy simulation with `density_matrix=False` "
+                    + "and no Measurement gate in the Circuit. If you wish to retrieve the "
+                    + "statistics of the outcomes please include measurements in the circuit, "
+                    + "otherwise set `density_matrix=True` to recover the final state.",
                 )
 
         if circuit.accelerators:  # pragma: no cover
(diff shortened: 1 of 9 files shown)


renatomello:
This is also ready and it is blocking #1470

src/qibo/backends/numpy.py
```diff
@@ -769,12 +769,31 @@ def calculate_matrix_exp(self, a, matrix, eigenvectors=None, eigenvalues=None):
...
         if not isinstance(power, (float, int)):
             raise_error(
                 TypeError,
                 f"``power`` must be either float or int, but it is type {type(power)}.",
             )
+
+        if power < 0.0:
+            # negative powers of singular matrices via SVD
+            # det with np instead of self.np because of torch
+            determinant = np.linalg.det(matrix)
```
BrunoLiegiBastonLiegi, line 787: is this going to work with `CupyBackend`? 
But in general why don't we use the usual `self.np`?

src/qibo/backends/pytorch.py
```diff
@@ -199,9 +201,29 @@ def calculate_matrix_exp(self, a, matrix, eigenvectors=None, eigenvalues=None):
...
+        power: Union[float, int],
+        precision_singularity: float = 1e-14,
+    ):
+        if power < 0.0:
+            # negative powers of singular matrices via SVD
+            determinant = self.np.linalg.det(matrix)
+            if abs(determinant) < precision_singularity:
+                U, S, Vh = self.np.linalg.svd(matrix)
+                # cast needed because of different dtypes
+                S = self.cast(S)
```
BrunoLiegiBastonLiegi, line 216: if you cast in the `NumpyBackend` as well all this part won't be needed right?

src/qibo/backends/numpy.py
renatomello: It is working locally

Simone-Bordoni (approved):

src/qibo/backends/numpy.py
BrunoLiegiBastonLiegi: > In this specific point of the code the matrix might be a torch.Tensor or np.ndarray

but why is that the case? I mean the `backend.calculate_matrix_power` method should expect native arrays of the backend as inputs, right?

src/qibo/backends/pytorch.py
```diff
@@ -199,9 +201,15 @@ def calculate_matrix_exp(self, a, matrix, eigenvectors=None, eigenvalues=None):
...
-        copied = super().calculate_matrix_power(copied, power)
+    def calculate_matrix_power(
+        self,
+        matrix,
+        power: Union[float, int],
+        precision_singularity: float = 1e-14,
+    ):
+        copied = self.cast(matrix, copy=True)
+        copied = self.to_numpy(copied) if power >= 0.0 else copied.detach()
+        copied = super().calculate_matrix_power(copied, power, precision_singularity)
```
BrunoLiegiBastonLiegi, line 212: This is due to `torch` not having a `fractional_matrix_power` alternative, right? Or is it due to some of the other `linalg` operations? 
I it's the first case, then I'd say you need to copy and cast to `numpy` only for `power > 0.`

renatomello: TBH idk how `detach` works in `torch`, but detaching the original will not affect the gradients? If the answer is no, then yes I can change it.

BrunoLiegiBastonLiegi: No indeed, detach basically discards the gradients. But I am not sure why do yo need to detach them for `power > 0.`. All the linalg operations you perform in that case should be `torch` compatible and should not have problems with the gradients if I am not mistaken, or am I missing something?

renatomello: some (maybe all?) `numpy` and `scipy` functions do not work on `torch` tensors if `requires_grad=True`. This error shows up on tests all the time. In this case, it is because of `scipy.linalg.fractional_matrix_power`.

BrunoLiegiBastonLiegi: Sorry, I meant for `power < 0.` This part in practice
``` python
            # negative powers of singular matrices via SVD
            determinant = self.np.linalg.det(matrix)
            if abs(determinant) < precision_singularity:
                U, S, Vh = self.np.linalg.svd(matrix)
                # cast needed because of different dtypes in `torch`
                S = self.cast(S)
                S_inv = self.np.where(
                    self.np.abs(S) < precision_singularity, 0.0, S**power
                )
                return (
                    self.np.linalg.inv(Vh) @ self.np.diag(S_inv) @ self.np.linalg.inv(U)
                )
```

renatomello: The code only enters this section if `power < 0.0` *and* the matrix is singular. So, for non-singular matrices, the execution still goes through the `scipy` function and thus either still needs the `.detach` or needs to check the determinant at that stage. I chose to just detach every tensor because it seemed it'd be less expensive than calculating the determinant, but I didn't rigorously check it.

BrunoLiegiBastonLiegi: Ah I see you are right sorry, I missed the det check, which you cannot do outside of the `NumpyBackend` method, I mean you could do but it wouldn't make much sense.

BrunoLiegiBastonLiegi (approved):

src/qibo/backends/numpy.py
```diff
@@ -769,12 +769,32 @@ def calculate_matrix_exp(self, a, matrix, eigenvectors=None, eigenvalues=None):
...
+            if abs(determinant) < precision_singularity:
+                U, S, Vh = self.np.linalg.svd(matrix)
+                # cast needed because of different dtypes in `torch`
+                S = self.cast(S)
+                S_inv = self.np.where(
+                    self.np.abs(S) < precision_singularity, 0.0, S**power
+                )
+                return (
+                    self.np.linalg.inv(Vh) @ self.np.diag(S_inv) @ self.np.linalg.inv(U)
+                )
```
stavros11, line 796: I didn't check all the details, so maybe I am wrong, but it seems that maybe we could reduce code repetition by moving this to a function that accepts `self.np` as an argument and then reuse for tensorflow (passing `self.tf`) and maybe even for cupy in qibojit.

I mean something like
```py
def calculate_matrix_negative_power(mod, matrix, ...):
    determinant = mod.linalg.det(matrix)
    ...
```
It can be defined even outside the backend, but then it will need to be imported in the various places, otherwise as `@staticmethod` in the backend.

stavros11 (approved):

PR #1476, opened 2024-10-07
Add `hilbert_schmidt_inner_product` to `quantum_info.metrics`

renatomello:
Checklist:
- [x] Tests are passing.


src/qibo/quantum_info/metrics.py
```diff
@@ -134,8 +134,34 @@ def trace_distance(state, target, check_hermitian: bool = False, backend=None):
...
+            Defaults to ``None``.
+
+    Returns:
+        float: Hilbert-Schmidt inner product :math:`\\braket{A, \\, B}_{\\text{HS}}`.
+    """
+    backend = _check_backend(backend)
+
+    inner_product = backend.np.trace(backend.np.conj(operator_A.T) @ operator_B)
+
+    return float(backend.np.real(inner_product))
```
BrunoLiegiBastonLiegi, line 160: Is there a reason why we return a float instead of a native array of the backend? I'd say it would be better to have an `array(1.)` or a `tensor(1.)` rather than just `1.`.

stavros11 (approved):

BrunoLiegiBastonLiegi (approved):

PR #1479, opened 2024-10-08
Migrating TensorFlowBackend to Qiboml

MatteoRobbiati:
Checklist:
- [x] Tests are passing.


src/qibo/quantum_info/basis.py
```diff
@@ -105,7 +105,7 @@ def pauli_basis(
         basis_full = basis_single
 
     if vectorize and sparse:
-        if backend.name == "tensorflow":
+        if backend.platform == "tensorflow":
             nonzero = np.nonzero
```
renatomello, line 109: Same thing about an alias here

renatomello:
@MatteoRobbiati will the documentation be moved separately?

MatteoRobbiati:
> > > @MatteoRobbiati will the documentation be moved separately?
> > 
> > 
> > I didn't ticked it in the list indeed, It would probably be ideal to adapt the documentation to the migration, what you think about it?
> 
> I agree that the docs should be moved as well. However, if too many files have to be touched, then a separate PR is worth it.

I was mostly thinking about all the examples, e.g. replacing 

`qibo.set_backend("tensorflow")` with `qibo.set_backend(backend="qiboml", platform="tensorflow")` mentioning Qiboml explicitly.

alecandido:
Like @renatomello said, I would also advise migrating the docs separately.

In general, it is better to keep the code and docs synced, but here there is a large migration process involved, and even reviewing the PRs will be harder otherwise...

Moreover, Qiboml is still quite unstable and not yet used by anyone outside Qibo developers. So there should be no problem in missing the docs in `master`. And instead it could ease the migration process, since you could rely on the presence of the migrated backend in the Qibo's removal PR.

alecandido:
> I was mostly thinking about all the examples, e.g. replacing

I would even move the examples in yet a further PR (so, first the actual docs, and eventually the examples).

MatteoRobbiati:
Ok, I am fine with this. It will just be important to update docs and examples before the release, but makes sense to me!

renatomello:
@BrunoLiegiBastonLiegi had to change some things in the `_curve_fit` function because it was not working properly with the GPU backends. Currently, `scipy.optimize` has to `cupy` equivalent, so had to default to `numpy` on that one.

src/qibo/models/error_mitigation.py
```diff
@@ -345,7 +345,16 @@ def closure():
 
         optimizer.step(closure)
         return params
-    return curve_fit(model, xdata, ydata, p0=params)[0]
+
+    if backend.platform in ["cupy", "cuquantum"]:
```
renatomello, line 349:
```suggestion
    if backend.platform in ["cupy", "cuquantum"]:  # pragma: no cover
```

pyproject.toml
```diff
@@ -70,10 +70,11 @@ scikit-learn = "^1.2.1"
 pytest-cov = "^4.0.0"
 pylint = "3.1.0"
 matplotlib = "^3.7.0"
-tensorflow = { version = "^2.16.1", markers = "sys_platform == 'linux'" }
+tensorflow = { version = "^2.16.1", markers = "sys_platform == 'linux' or sys_platform == 'darwin'", optional = true }
```
BrunoLiegiBastonLiegi, line 73:
```suggestion
```
why do we need explicit dependence from `tensorflow`? this should be taken care of by the `qiboml` dependency, as it happens for `qibojit` and `numba`.

src/qibo/backends/__init__.py
```diff
@@ -27,30 +26,36 @@ def load(backend: str, **kwargs) -> Backend:
...
-            qibo.backends.abstract.Backend: The loaded backend.
+            :class:`qibo.backends.abstract.Backend`: Loaded backend.
         """
 
-        if backend == "numpy":
-            return NumpyBackend()
-        elif backend == "tensorflow":
-            return TensorflowBackend()
-        elif backend == "pytorch":
+        possible_backends = ["numpy", "pytorch", "clifford", "qulacs"]
```
BrunoLiegiBastonLiegi, line 35: isn't this basically `QIBO_NATIVE_BACKENDS + ["clifford"]` ?

src/qibo/backends/__init__.py
```diff
@@ -27,30 +26,36 @@ def load(backend: str, **kwargs) -> Backend:
...
+            from qibo.backends.qulacs import QulacsBackend  # pylint: disable=C0415
 
             return QulacsBackend()
-        else:
-            raise_error(
-                ValueError,
-                f"Backend {backend} is not available. The native qibo backends are {QIBO_NATIVE_BACKENDS}.",
-            )
+
+        return NumpyBackend()
```
BrunoLiegiBastonLiegi, line 58: I am not sure I like this new form of the `load`, I think we are giving up a little bit of readability, and I am not sure why we needed to change this.

src/qibo/models/error_mitigation.py
```diff
@@ -345,7 +345,16 @@ def closure():
 
         optimizer.step(closure)
         return params
-    return curve_fit(model, xdata, ydata, p0=params)[0]
+
+    if backend.platform in ["cupy", "cuquantum"]:  # pragma: no cover
+        # Currrently, ``cupy`` does not have compatibility with ``scipy.optimize``.
+        xdata = backend.to_numpy(xdata)
+        ydata = backend.to_numpy(ydata)
+        params = backend.to_numpy(params)
```
BrunoLiegiBastonLiegi, line 353: I think that even for `tensorflow` a conversion to `numpy` is happening inside of `curve_fit` anyway. Therefore, we could even drop this check and just cast to `numpy` in all backends that are not `pytorch`. In any case, I am planning at some point to add a `curve_fit` method to the backends and get rid of this checks, thus you can safely ignore this.

src/qibo/backends/__init__.py
renatomello: The `load` function is not ending in a return, which is a `lint` issue. I don't know how much any else care about this, so if y'all want the other way, at least the lint issue should be manually disabled then.

src/qibo/models/error_mitigation.py
renatomello: This is needed right now because tests are failing on GPU without this.

pyproject.toml
MatteoRobbiati: I am keeping this because many tests are importing tensorflow (just to give an example: [this one](https://github.com/qiboteam/qibo/blob/958b11bc38b800f217ad9c97b950635896935fd4/tests/test_backends_torch_gradients.py#L50)). 

I agree many of the imports can be removed, but probably this is something we can solve after the migration.

BrunoLiegiBastonLiegi: So, now we can get safely rid of this dependency, right?

renatomello (approved):

scarrazza:
@MatteoRobbiati anything else for this PR, can we merge?

doc/source/code-examples/advancedexamples.rst
```diff
@@ -78,19 +78,8 @@ or programmatically, during runtime, as follows:
...
-
-    import tensorflow as tf
-    tf.config.threading.set_inter_op_parallelism_threads(1)
-    tf.config.threading.set_intra_op_parallelism_threads(1)
-    import qibo
-
-Note that this should be run during Tensorflow initialization in the beginning
-of the script and before creating the qibo backend.
+For similar wariness when using a machine learning backend (such as TensorFlow or Pytorch)
+please refer to the Qiboml documentation.
```
BrunoLiegiBastonLiegi, line 82: A link here would be probably helpful, even though I don't think we have any right now in qiboml. This is more a reminder to add it, once we have it.

BrunoLiegiBastonLiegi (approved):

doc/source/code-examples/advancedexamples.rst
MatteoRobbiati: I think we can add a simple documentation in Qiboml for now, just with these instructions.

PR #1480, opened 2024-10-08
Add hook to check toml files

alecandido:
The hook is one of the default ones, and essentially for free. It will mainly help keeping `pyproject.toml` formatted.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,6 +7,7 @@ repos:
       - id: trailing-whitespace
       - id: end-of-file-fixer
       - id: check-yaml
+      - id: check-toml
       - id: debug-statements
   - repo: https://github.com/psf/black
     rev: 24.10.0


alecandido:
Ok, I was joking, it's just verifying the syntax, it's not formatting anything.

To get the formatting part, we'd need something like:
https://github.com/toml-lang/toml/blob/dc5af251cd348ee053d9355359743767e3d0b123/.pre-commit-config.yaml#L2-L5
or
https://github.com/ComPWA/taplo-pre-commit
but they are not "for free", so I'm not going to propose them (for the time being).

scarrazza (approved):

stavros11 (approved):
Maybe we could also propagate to other repos (if not already there).

alecandido:
> Maybe we could also propagate to other repos (if not already there).

That's exactly why I appointed you as a reviewer :)

Issue #1403, opened 2024-07-24
Migrating to NumPy 2.0


scarrazza:
At some point, we should also take into account migrating to NumPy 2.0, now that is out.

A much relevant improvement (to us) could be the following:
https://numpy.org/doc/stable/release/2.0.0-notes.html#array-api-compatible-functions-aliases
https://numpy.org/doc/stable/release/2.0.0-notes.html#array-api-compatible-functions-for-numpy-linalg
https://numpy.org/doc/stable/release/2.0.0-notes.html#array-api-standard-support-in-the-main-namespace

It was already documented for v1, and there are compatibility packages available
https://numpy.org/doc/stable/reference/array_api.html
but now it is available by default, and it should improve the compatibility with the other array frameworks (i.e. TensorFlow, PyTorch, JAX, CuPy, ...), since the Python API has been standardized (not all of them adopted it already, but they might do it).

_Originally posted by @alecandido in https://github.com/qiboteam/qibo/issues/1395#issuecomment-2238791623_

glpatcern:
Ciao @MatteoRobbiati, @scarrazza, are you going to look into this migration any time soon?

I've got a request about this from the SWAN service at CERN, as they plan to move to numpy 2.0 and would potentially need to uninstall Qibo.

MatteoRobbiati:
> Ciao [@MatteoRobbiati](https://github.com/MatteoRobbiati), [@scarrazza](https://github.com/scarrazza), are you going to look into this migration any time soon?
> 
> I've got a request about this from the SWAN service at CERN, as they plan to move to numpy 2.0 and would potentially need to uninstall Qibo.

Ciao @glpatcern!
We have to solve some dependencies we still have in the tests with machine learning backends (mostly related to Keras). After this, we plan to support Numpy 2.0 indeed!

glpatcern:
Hey @MatteoRobbiati, I guess this will take some time yet, right? I see there's also #1484 that is kind of related, but did not find any PR specifically addressing numpy 2.

At the moment, the SWAN team is ready to migrate to a new set of packages that support numpy 2 and python 3.12, and unless we (you) provide some context or roadmap, they may drop Qibo to give precedence to all the rest.

alecandido:
[...]

PR #1484, opened 2024-10-10
Support Python 3.13

alecandido:
Since py3.13 has been finally published (on Monday), I tried out of curiosity whether it was possible to upgrade our version constraint to use that.

We do not directly rely on Python internals, and py3.x versions are backward compatible (in principle). So, there is no reason for Qibo not to support it, for as long as our dependencies support that.

So, it was just a matter of running `poetry lock`, and... it seems working.

Now, we still need to solve the Qibojit problem (that in the meanwhile became also a Qibotn one, and soon Qiboml as well).

I expected this to be blocked by TensorFlow once again. And it should, since they are not distributing wheels for py3.13 yet (and they apparently won't before next year https://github.com/tensorflow/tensorflow/issues/62003#issuecomment-2385696013), but apparently they didn't care declaring it
https://github.com/tensorflow/tensorflow/blob/a2fadc33162b16989eb10a541d112e6b9fffa0e1/tensorflow/tools/pip_package/setup.py#L412
and Poetry is not examining the wheel compatibility, so we already got a valid lockfile anyhow.

While going through Qibojit and the rest of it (following the much beloved procedure in https://github.com/qiboteam/qibo/pull/1231#issuecomment-1968362299) further problems may arise. But, at this point, I will have a try soon, also to track the present blockers.


alecandido:
The CI is of course failing, since I had to disable `qibojit` and `qibotn` to lock. Re-enabling them is the missing part to complete this PR.

cosenal:
Ciao @alecandido, any update on this? Qibo is one dependency that is blocking [Mitiq](https://github.com/unitaryfund/mitiq) to support Python 3.13.

alecandido:
@cosenal thanks for reaching about this, we're now in a better situation than last time (i.e. #1231), but still not optimal. Let me explain.

Before Qibo was locked behind TensorFlow, which was still the case in October, when I opened this PR.
While NumPy and some other packages have now synced their release cycle with Python's one (which I guess was the main reason for Python itself to establish a well-predictable schedule), so they are already upgrading compatibility based on beta releases, making them ready by October. However, TensorFlow is much slower (certainly not the only major one, but essentially the only one among Qibo's dependencies).

What happened in the last few months is the final migration of the differentiable computation support of Qibo to Qiboml https://github.com/qiboteam/qiboml.
This makes Qibo (almost) independent of TensorFlow, thus I was able to lock.

Unfortunately, we're still in a situation in which part of the backends are tested in Qibo's CI, thus we should still follow the update procedure outlined in https://github.com/qiboteam/qibo/pull/1231#issuecomment-1968362299
We already had plans to solve this situation, by either deeply reworking the Qibo internals layout (https://github.com/qiboteam/qibo-core) or moving/duplicating part of the tests. But it hasn't happened yet...

Thus, the current issue is that Qiboml will be needed to test the TensorFlow backend in the Qibo CI, right now failing for its absence (actually, first encountering Qibojit's absence, but that can be solved with the above procedure). If the dependency were reinstated, then it would fail because of Qiboml not supporting Python 3.13 - eventually coming back to TensorFlow once more.

So, the original issue is not completely solved, but we're now much closer, since we only have to choose how to handle the test issue, introducing the dependency, and then TensorFlow would be out of the main loop (unless Mitiq explicitly targets the TensorFlow backend - at which point Qibo can't solve the issue, until TensorFlow itself will).

@scarrazza

renatomello (approved):

renatomello (approved):
@alecandido maybe it is worth adding a `no cover` tag to the `_sgd_tf` function

alecandido:
Bad news: NumPy v1 is not supporting py3.13, since Numpy v2 transitioned already happened before that.

So, now this PR requires #1403 & #1582 to complete.

alecandido:
Cirq is introducing a transitive dependency on Rigetti's `qcs-sdk-python`, which is obtained from the Rust interface https://github.com/rigetti/qcs-sdk-rust/, and it currently suffers an issue related to the wheel generation, https://github.com/rigetti/qcs-sdk-rust/issues/531 (it appears to come from the [`zip` crate](https://github.com/PyO3/maturin/issues/2499#issuecomment-2710388691) itself...).

However, since continuously testing with Cirq does not seem an actual priority, I temporarily silenced the related tests, and dropped the dependency. It will be reinstated as soon as they will solve the problem on their side (we could limit the transitive dependency on our side to some past version that is actually working - but then we may lose py3.13 again, so it's really not worth it...).

alecandido:
Ok, now it is really failing just because of NumPy (https://github.com/qiboteam/qibo/pull/1484#issuecomment-2709993727), thus there is nothing that can be done just on this PR.

scarrazza (approved):
Thanks @alecandido.

PR #1485, opened 2024-10-11
RGD optimization method for quantum state tomography

HuberyMing:
Checklist:
- [ ] Tests are passing.

To include the optimization method of Riemannian Gradient Descent (RGD) algorithm to do the tomography problem. The implementation follows from the arXiv:2210.04717, which was published in Phys. Rev. Lett. 132, 240804


renatomello:
Not entering the other merits of the PR, this would have to be inside the `tomography` module instead of being a separate `tomography_RGB` module.

src/qibo/tomography_RGD/BasicTools.py
```diff
@@ -0,0 +1,76 @@
...
+        print("   the {}-th qubit".format(i))
+
+        sym_Generated = []
+        for symNow in symList:
+            sym_Generated = sym_Generated + ["".join([symNow, s]) for s in symbols]
+            # print(sym_Generated)
+        symList = sym_Generated
+    print("  totol number of labels {}".format(len(symList)))
+
+    return symList
```
renatomello, line 36:
```suggestion
from itertools import product

all_labels = ["".join(elem) for elem in product(symbols, repeat=nqubits)]
```

src/qibo/tomography_RGD/Run_Tomo.py
```diff
@@ -0,0 +1,196 @@
...
+import methodsRGD_core
+import numpy as np
+import projectors
+import qutip as qu
+from BasicTools import Generate_All_labels, Plt_Err_Time
+
+# from states import GHZState, HadamardState, RandomState
+from qibo_states import GHZState, HadamardState, RandomState
+
+from qibo import quantum_info
```
renatomello, line 18:
```suggestion
from qibo.quantum_info import random_density_matrix

```

src/qibo/tomography_RGD/Run_Tomo.py
```diff
@@ -0,0 +1,196 @@
...
+        )
+
+    elif circuit_Choice == 2:  # directly generate density matrix via qutip
+        Nr = 1
+        rho = qu.rand_dm_ginibre(2**n, dims=[[2] * n, [2] * n], rank=Nr)
+        target_density_matrix = rho.full()
+
+    elif circuit_Choice == 3:  # directly generate density matrix via qibo
+        Nr = 3
+        target_density_matrix = quantum_info.random_density_matrix(2**n, Nr)
```
renatomello, line 77:
```suggestion
        target_density_matrix = random_density_matrix(2**n, Nr)

```

src/qibo/tomography_RGD/qibo_states.py
```diff
@@ -0,0 +1,257 @@
...
+    state.create_circuit()
+    data_dict_list = state.execute_measurement_circuits(labels)
+    # print(data_dict_list)
+
+    target_density_matrix = state.get_state_matrix()
+    target_state = state.get_state_vector()
+    # print(state.get_state_vector())
+
+    Nr = 3
+    random_DM = quantum_info.random_density_matrix(2**n, Nr)
```
renatomello, line 257:
```suggestion
    random_DM = random_density_matrix(2**n, Nr)

```

alecandido:
@renatomello to follow up what I told you about remotes during the Qibo meeting, you could do the following:
```sh
git remote add hubery-ming git@github.com:HuberyMing/qibo.git
git pull hubery-ming
git switch -c QST -t hubery-ming/QST
```
in this way you would set up a remote named `hubery-ming`, tracking @HuberyMing's repo, and then create a local branch in your local repo named `QST`, tracking the https://github.com/HuberyMing/qibo/tree/QST branch.

Of course, you won't have pushing rights (unless @HuberyMing will grant you), but that's in case you want to test it locally.

Maybe you were already familiar with this, but this may also be useful for everyone else who'd like to have a look (and it could be a reference to deal for PRs from forks in general).

src/qibo/tomography_RGD/qibo_states.py
```diff
@@ -0,0 +1,257 @@
...
+            else:
+                op_ind = random.randint(0, 1)
+            if op_ind == 0:  # U3
+                qind = random.randint(0, self.n - 1)
+                circuit.add(
+                    gates.U3(
+                        qind,
+                        2 * np.pi * random.random(),
+                        2 * np.pi * random.random(),
+                        2 * np.pi * random.random(),
```
renatomello, line 220: A tip for the future: This choice of phases does not sample random states uniformly. Please see https://github.com/qiboteam/qibo/blob/958b11bc38b800f217ad9c97b950635896935fd4/src/qibo/quantum_info/random_ensembles.py#L48 and https://pennylane.ai/qml/demos/tutorial_haar_measure

src/qibo/tomography_RGD/qibo_states.py
```diff
@@ -0,0 +1,257 @@
+import random
+
+import numpy as np
+
+import qibo
+from qibo import gates, quantum_info
+
+
+class State:
```
renatomello, line 9: Correct me if I'm wrong, but it seems like this base class is unnecessary since every method is dependent on a `qibo.models.Circuit` method. I'd say that if these state classes are needed at all (which I'm not convinced of), you could use `Circuit` as the base class directly.

src/qibo/tomography_RGD/qibo_states.py
```diff
@@ -0,0 +1,257 @@
...
+        State.__init__(self, n)
+        self.circuit_name = "Hadamard"
+
+    def create_circuit(self):
+        circuit = qibo.Circuit(self.n)
+
+        for i in range(self.n):
+            circuit.add(gates.H(i))
+
+        self.circuit = circuit
```
renatomello, line 189: Since this is just a layer of Hadamards, this is a Hadamard transform of the initial state. A more concise way to implement this would be to modify the following function: https://github.com/qiboteam/qibo/blob/958b11bc38b800f217ad9c97b950635896935fd4/src/qibo/quantum_info/utils.py#L118
For instance, if the input is a circuit, then a layer of Hadamard is added. That would perform the Hadamard of whatever state one wants, including the zero state in your case.

src/qibo/tomography_RGD/qibo_states.py
```diff
@@ -0,0 +1,257 @@
...
+        self.circuit_name = "GHZ"
+
+    def create_circuit(self):
+        circuit = qibo.Circuit(self.n)
+
+        circuit.add(gates.H(0))
+        for i in range(1, self.n):
+            circuit.add(gates.CNOT(0, i))
+
+        self.circuit = circuit
```
renatomello, line 171: This fits as a function in `qibo.models.encodings` receiving `nqubits` as input and returning the corresponding circuit as output.

src/qibo/tomography_RGD/projectors.py
```diff
@@ -0,0 +1,978 @@
...
+        for y in [x for x in product(*[values_dict[letter] for letter in label])]
+    ]
+    ijv = list(map(lambda x: (x[0][0], x[0][1], x[1]), zip(ij, values)))
+
+    i_coords, j_coords, entries = zip(*ijv)
+
+    projector = sparse.coo_matrix(
+        (entries, (i_coords, j_coords)), shape=(d, d), dtype=complex
+    )
+    return projector
```
renatomello, line 168: @BrunoLiegiBastonLiegi this is a case where it would be convenient to have the `quantum_info.basis.pauli_basis` function return a generator because then one could use that to directly access a specific element of the basis.

mho291:
Just curious, would it be helpful to have a call to understand how the algorithm works? It might make it easier to review the code.

mho291:
@HuberyMing If you're doing the changes locally, you could do `pip install pre-commit` followed by `pre-commit install` inside your environment. Each time you do `git commit -m "message"`, the pre-commit will run. If anything fails, you have to do `git commit` again before `git push`. Then we we will avoid having extra commits in this PR.

alecandido:
> Each time you do `git commit -m "message"`, the pre-commit will run. If anything fails, you have to do `git commit` again before `git push`.

Specifically, if it fails the involved hook is "declaring" a failure, that may imply that you have to fix something manually. However, most of the hooks we have registered are also auto-fixing the issues they find (e.g. the formatter fails if the result differs from the original - but if it fails, it also reformats the fails).
The auto-fixes are not staged, so you also need to `git add` the fixed files.

image

However, in these cases, I'm sure that @HuberyMing just applied the suggestions from the review from the PR web interface. In which case there is no chance of having `pre-commit` running anywhere else...

(as you also prepended
> @HuberyMing If you're doing the changes locally,

)

renatomello:
> Just curious, would it be helpful to have a call to understand how the algorithm works? It might make it easier to review the code.

It might me needed, yes. Let me read the paper first though, so I have a better understanding.

HuberyMing:
> > Just curious, would it be helpful to have a call to understand how the algorithm works? It might make it easier to review the code.
> 
> It might me needed, yes. Let me read the paper first though, so I have a better understanding.

Hi @renatomello Thank you for the review. No problem. How do we discuss?

renatomello:
> > > Just curious, would it be helpful to have a call to understand how the algorithm works? It might make it easier to review the code.
> > 
> > 
> > It might me needed, yes. Let me read the paper first though, so I have a better understanding.
> 
> Hi @renatomello Thank you for the review. No problem. How do we discuss?

We can try to arrange a chat via Zoom next week. Right now, I'd like to see a separate PR adding the GHZ state to `qibo.models.encodings`. I think it will be a good lesson on a small scale on how this PR should move forward in terms of code implementation.
You can follow the style of the functions inside `qibo.models.encodings` as well as the test stile inside `tests/test_models_encodings.py`. 

It's a good first issue.

src/qibo/tomography_RGD/qibo_states.py
renatomello: A new branch starting from `master`, completely unrelated to this one. You can name whatever, but probably `ghz` is simpler and better. Then in that branch, you only implement the GHZ circuit as a function in `qibo.models.encodings` and the corresponding tests. Nothing else. After that is merged into `master`, we change the code here accordingly so it can communicate with the newly existing function.

src/qibo/tomography_RGD/Run_Tomo.py
```diff
@@ -0,0 +1,196 @@
...
+            label = data_dict_list[ii]["label"]
+            data_dict[label] = data_dict_list[ii]["count_dict"]
+
+        measurement_list = measurements.MeasurementStore.calc_measurement_list(
+            labels, data_dict
+        )
+
+    elif circuit_Choice == 2:  # directly generate density matrix via qutip
+        Nr = 1
+        rho = qu.rand_dm_ginibre(2**n, dims=[[2] * n, [2] * n], rank=Nr)
```
renatomello, line 72:
```suggestion
        rho = random_density_matrix(2**n, rank=Nr, metric="ginibre")

```

src/qibo/tomography_RGD/Run_Tomo.py
```diff
@@ -0,0 +1,196 @@
...
+#   and directly do the tomography
+#
+#
+
+import measurements
+import methodsMiFGD_core
+import methodsRGD_core
+import numpy as np
+import projectors
+import qutip as qu
```
renatomello, line 12:
```suggestion
```

src/qibo/tomography_RGD/qibo_states.py
HuberyMing: Thank you for the instruction. I have created a new branch called ghz and push it with sync fork. I have tested some functions inside the directory src/qibo/test. However, I do not understand well the qibo.models.encodings you mentioned. There are some encoders in the qibo.models.encodings but probably we don't need them at the moment. I am not sure if we can add some user-defined properties in the circuit. The way I did it now is to create a new class. Probably some properties do not fit into qibo's standards. I am curious if creating some list as a recorded processed result is allowed. Or do you prefer on-the-fly calculation during processing? Thanks.

mho291: I think what @renatomello is that the GHZ state should be added as a function to `qibo.models.encodings`. It will probably look something like this:

``` python
def GHZ_circuit(nqubits):
    c = QiboCircuit(nqubits)
    c.add(gates.H(0))
    for _i in range(nqubits-1):
        c.add(gates.CNOT(_i, _i+1))
    return c
```

I don't think it will take so much time to add it in, I can do it after dinner.
Anyway, would `qibo.models.utils` be a more suitable location than  `qibo.models.encodings`?

HuberyMing:
> Do you plan to continue with this PR @HuberyMing? :)

Sorry for not pushing forward much. I will continue the PR. Hope to finish it soon.

MatteoRobbiati:
> Sorry for not pushing forward much. I will continue the PR. Hope to finish it soon.

No problem! Take your time :) I just wanted to check the status!

HuberyMing:
To have better compatibility with qibo, now I have changed the expectation value calculation of each Pauli string from the shot measurements by using the package expectation_from_samples. Here is the imported package:

from qibochem.measurement import expectation, expectation_from_samples

However, not all Pauli string can be correctly calculated for its expectation value from the shot measurement from this package. I have raised an issue in #1677

renatomello:
> To have better compatibility with qibo, now I have changed the expectation value calculation of each Pauli string from the shot measurements by using the package expectation_from_samples. Here is the imported package:
> 
> from qibochem.measurement import expectation, expectation_from_samples
> 
> However, not all Pauli string can be correctly calculated for its expectation value from the shot measurement from this package. I have raised an issue in #1677

This creates a circular dependency (`qibochem` depends on `qibo` that depends on `qibochem`) that cannot happen. If the function is general and important enough, it should be implemented directly in `qibo` and imported by `qibochem`.

On the other hand, maybe this is not true for the `SymbolicHamiltonian` class that I see you are using, but the `Hamiltonian` class already has two methods called `expectation_from_samples` and `expectation_from_circuit`. Couldn't they be used instead? If not, couldn't they be generalized to fit your needs?

HuberyMing:
> > To have better compatibility with qibo, now I have changed the expectation value calculation of each Pauli string from the shot measurements by using the package expectation_from_samples. Here is the imported package:
> > from qibochem.measurement import expectation, expectation_from_samples
> > However, not all Pauli string can be correctly calculated for its expectation value from the shot measurement from this package. I have raised an issue in #1677
> 
> This creates a circular dependency (`qibochem` depends on `qibo` that depends on `qibochem`) that cannot happen. If the function is general and important enough, it should be implemented directly in `qibo` and imported by `qibochem`.
> 
> On the other hand, maybe this is not true for the `SymbolicHamiltonian` class that I see you are using, but the `Hamiltonian` class already has two methods called `expectation_from_samples` and `expectation_from_circuit`. Couldn't they be used instead? If not, couldn't they be generalized to fit your needs?

Thanks for the comment. Inspired by your comment, I tried something like
```python
from qibo.hamiltonians.hamiltonians import AbstractHamiltonian
from qibo.hamiltonians import Hamiltonian

coef_Pauli_exact = AbstractHamiltonian.expectation(stateGHZ, symbolPauli)    # NotImplementedError: None
coef_Pauli_exact = Hamiltonian.expectation(stateGHZ, symbolPauli)           
coef_Pauli = symbolPauli.expectation_from_samples(frequencies, qubit_map=range(n_qubits)) 
```
I found none of them work. Basically they are not implemented. However, I figured out there is no quick way to do this. We can get the correct results by using pauli_term_measurement_expectation such as
```python
from qibochem.measurement.result import pauli_term_measurement_expectation

circuit.add(gates.M(0, basis=type(X(0).gate))) # H gate     # for 'X' Pauli term
circuit.add(gates.M(1, basis=type(Y(1).gate))) # RX(0.5*pi) gate
circuit.add(gates.M(2, basis=type(Z(2).gate))) # Computational basis remains unchanged

# Now run the circuit to get the circuit measurements
result = circuit(nshots=10000)
frequencies = result.frequencies(binary=True)
# pauli_term_measurement_expectation is a Qibochem function for calculating the expectation value of Hamiltonians with non-Z terms
shots_term1 = pauli_term_measurement_expectation(term1.terms[0], frequencies, qubit_map=range(n_qubits))
```
The conclusion is that pauli_term_measurement_expectation works but expectation_from_samples does not. At least this can fit into my need. Thanks.

HuberyMing:
To have some updates. The issue #1677 has already solved this problem of getting the correct expectation values by using the `SymbolicHamiltonian.expectation_from_circuit` function.

scarrazza:
@HuberyMing what are your plans concerning this PR?

HuberyMing:
> @HuberyMing what are your plans concerning this PR?

@scarrazza  Thank you for your patience regarding this PR. I am currently refactoring the code to remove unnecessary parts and better align it with the qibo framework. After the Lunar New Year, I will devote more time to finalizing it, with the goal of completing the PR within one to two months.

HuberyMing:
Since 'NumpyBackend' object has no attribute 'np', do we use 'np' (numpy) directly instead?

alecandido:
Hi @HuberyMing: in #1711 the backends' mechanism was quite modified, which triggered on its own the release of Qibo v0.3.0.

To a first approximation, the `self.np` attribute has been replaced by the `self.engine`.
https://github.com/qiboteam/qibo/blob/e8bca307f88a54fb77339d27a48c52d74112a89b/src/qibo/backends/abstract.py#L25
Though that is meant to be used mainly internally to the backend. If you are a backend user, then you should directly use the functions exposed by the backend
https://github.com/qiboteam/qibo/blob/e8bca307f88a54fb77339d27a48c52d74112a89b/src/qibo/backends/abstract.py#L332-L335
In this case it will be `backend.einsum(...)`.

HuberyMing:
@scarrazza @mho291  I am ready for the PR. Now files are merged to one file src/qibo/tomography_RGD/RGD_Optimized.py. This is in a temporary working directory where I also put two test files. To finish the PR, in which place do you suggest to put RGD_Optimized.py? It may be renamed as state_tomography_RGD.py if needed.

HuberyMing:
The quantum state tomography via RGD algorithm is now implemented as one file at src/qibo/tomography/state_tomography.py with two classes PauliMap and RGDOptimizer.
The test files are tests/test_tomography_pauli_map.py and tests/test_tomography_rgd_optimizer.py. 
Ready for PR. Thanks.

PR #1489, opened 2024-10-15
Expectation value of observables non diagonal in the computational basis

BrunoLiegiBastonLiegi:
Currently, our `SymbolicHamiltonian.expectation_from_samples` only works for observables that are composition of `Z` operators. To compute the expectation value of some non-`Z` operator, for instance `XZ + YX`, one has to manually rotate the measurement basis accordingly, collect the frequencies and then calculate the expectation value of a `ZZ` operator starting from those frequencies. Therefore, in this example, one has to measure `M(0,1, basis=["X", "Z"])` first, collect the frequencies and calculate the expectation of `ZZ`, then re-execute a second time measuring `M(0,1, basis=["Y", "X"])`, get the frequencies and again calculate the expectation of `ZZ`. The sum of the two expectations is then the global expected value.

This PR automatizes this process, by allowing circuits to be passed as inputs to `SymbolicHamiltonian.expectation_from_samples`.

Checklist:
- [ ] Tests are passing.


chmwzc:
Resolves #852, happy to see that this functionality is getting added in :)
Also, just FYI, I implemented something like this in Qibochem [here](https://github.com/qiboteam/qibochem/blob/main/src/qibochem/measurement/result.py), together with the use of qubit-wise commuting measurements to reduce measurement costs (Ref: [https://arxiv.org/abs/1907.03358](https://arxiv.org/abs/1907.03358)).

marekgluza:
That is very cool and will be very useful for my work with @Sam-XiaoyueLi 

@BrunoLiegiBastonLiegi you can request review from @Sam-XiaoyueLi and she can help add further docs or examples if needed

BrunoLiegiBastonLiegi:
@stavros11 may I trouble you with the review of this PR? I changed some things in the `SymbolicHamiltonian` and related objects, which I presume you first implemented.

@Sam-XiaoyueLi feel free to suggest any meaningfull example for the docs, for the moment I just made up a random observable.

marekgluza:
@Sam-XiaoyueLi maybe TLFIM or XXZ?

doc/source/code-examples/advancedexamples.rst
```diff
@@ -2015,7 +2015,7 @@ to allow calculation of expectation values directly from such samples:
 .. testcode::
 
     from qibo import Circuit, gates
-    from qibo.hamiltonians import Z
+    from qibo.hamiltonians import Z as HZ
```
stavros11, line 2018: Is this causing some issue in the new test? Another option is 
```suggestion
    from qibo import hamiltonians
```
and then use `hamiltonian = hamiltonians.Z(4)`. It is a bit longer but probably less weird than the `HZ` symbol.

src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -552,41 +556,141 @@ def calculate_dense(self):
...
+        circuit.
+
+        Args:
+            circuit (Circuit): input frequencies.
+            qubit_map (dict): qubit map.
+
+        Returns:
+            (float): the calculated expectation value.
+        """
+        from qibo import Circuit, gates
```
stavros11, line 575: Is this causing a circular import if top level?

src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -552,41 +556,141 @@ def calculate_dense(self):
...
+
+        if nshots is None:  # pragma: no cover
+            nshots = 1000
```
stavros11, line 578: Why not use `nshots=1000` directly as the default? I think we are doing the same in `Circuit.execute`.

I understand that this value would be ignored if frequencies are passed in the `expectation_from_samples` function, which probably motivates the use of `None`, but ignoring `1000` should also be fine (up to debate).

src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -552,41 +556,141 @@ def calculate_dense(self):
...
+                    backend=self.backend,
+                )
+            )
+            # prepare the measurement basis and append it to the circuit
+            measurements = [
+                gates.M(factor.target_qubit, basis=factor.gate.__class__)
+                for factor in set(term.factors)
+            ]
+            circ_copy = circuit.copy(True)
+            [circ_copy.add(m) for m in measurements]
```
stavros11, line 600:
```suggestion
            circ_copy.add(measurements)
```
should also work.

src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -552,41 +556,141 @@ def calculate_dense(self):
...
+                dict(
+                    zip(
+                        [qubit_map[q] for q in term.target_qubits],
+                        range(len(term.target_qubits)),
+                    )
+                )
+            )
+        frequencies = [
+            result.frequencies()
+            for result in self.backend.execute_circuits(rotated_circuits, nshots=nshots)
```
stavros11, line 613: I understand this is useful only for hardware execution, right? Since in simulation you could just simulate the circuit and use `expectation` directly, which should also be more efficient (and exact).

src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -552,41 +556,141 @@ def calculate_dense(self):
...
+                    for state in keys
+                ]
+            )
+        expvals = self.backend.cast(expvals, dtype=counts.dtype).reshape(
+            len(self.terms), len(freq)
+        )
+        return self.backend.np.sum(expvals @ counts.T) + self.constant.real
+
+    def expectation_from_samples(
+        self, data: Union[dict, "Circuit"], qubit_map: dict = None, nshots: int = None
```
stavros11, line 664: This is also up for discussion, however it may even be cleaner to expose two separate functions: `expectation_from_samples` and `expectation_from_circuit`, instead of accepting two different types.

doc/source/code-examples/advancedexamples.rst
alecandido: You could even do
```suggestion
    from qibo import hamiltonians as hs
```
that's half-way (but maybe still a bit weird).

src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -552,41 +556,141 @@ def calculate_dense(self):
...
+        """
+        from qibo import Circuit, gates
+
+        if nshots is None:  # pragma: no cover
+            nshots = 1000
+        rotated_circuits = []
+        coefficients = []
+        Z_observables = []
+        q_maps = []
+        for term in self.terms:
```
Edoardo-Pedicillo, line 583: Now this function is evaluating the Hamiltonian expectation value term by term. The next step (not for this PR) should be to reduce the number of circuit executions merging terms acting on non-intersecting qubit subsets.

doc/source/code-examples/advancedexamples.rst
BrunoLiegiBastonLiegi: Yeah no, it was just to avoid the confusion with the `Z` symbol that I use below. Any of the two suggestions should be fine, thanks.

src/qibo/hamiltonians/hamiltonians.py
BrunoLiegiBastonLiegi: Unfortunately, it is.

src/qibo/hamiltonians/hamiltonians.py
BrunoLiegiBastonLiegi: Specifically here, it is None because it depends on `expectation_from_samples` which defaults to None as well. But yes, we could just directly set it to 1000, no problem, I was not just sure whether this would cause any confusion, even though in the docstring is explicitely mentioned.

In any case, probably the best solution would be to completely separate the two functions and expose them directly to the user, as you suggested below.

src/qibo/hamiltonians/hamiltonians.py
BrunoLiegiBastonLiegi: Yes I had exactly in mind hardware execution for this. But even for simulation, you wouldn't want the exact expectation from the final state (is this what you are suggesting? not sure) as this is supposed to be sample dependent.

stavros11: > you wouldn't want the exact expectation from the final state (is this what you are suggesting? not sure) as this is supposed to be sample dependent.

Yes, above I meant to get the exact expectation from the final state, however you can still get a sample dependent estimate by first simulating the circuit with shots, getting the samples and using `expectation_from_samples`. I would expect this to be faster than using `expectation_from_circuit` for simulation, since in the first case the "state preparation" circuit is only simulated once, while `expectation_from_circuit` will repeat the simulation of the same circuit for every basis.

On hardware, or if you are simulating with noise, it is probably a different story.

src/qibo/hamiltonians/hamiltonians.py
stavros11: Thanks for confirming. We could investigate if it is easy to fix this, for sure not in this PR, but maybe open an issue.

stavros11 (approved):
Thanks for the updates @BrunoLiegiBastonLiegi.

Edoardo-Pedicillo (approved):

src/qibo/hamiltonians/hamiltonians.py
BrunoLiegiBastonLiegi: Ah yeah, I see, I've been thinking about doing the state preparation first and then just measuring for simulation. To do that though, we should probably move (at least some part of) the `expectation_from_circuit` inside the backend, which I am not against by the way, but for the moment I decided to keep it simple. In any case, I was assuming that since `execute_circuits` launches several parallel jobs, if I am not mistaken, the overhead of running several copies of the same simulation would be reduced, assuming you have many cores available.

stavros11: I also think we don't need to do anything about this, since it is already available by executing the circuit manually and then calling `expectation_from_samples`. This is a bit more manual for the user, but it is still fairly simple and it is not the best idea to provide multiple ways (almost equally simple) to do the same thing as it would increase maintenance from our side.

Issue #1493, opened 2024-10-16
Transpiler refactoring


csookim:
`wire_names` stores the qubit names, and `Gate.qubits` contains the index of the target qubit names in `wire_names`. However, `wire_names` is currently ignored during the transpilation process. The transpilation needs to be updated to ensure consistency across Qibo.

This update should include:

1. Integration of `wire_names` with `initial_layout` and `final_layout`.
2. Consistency between `wire_names` and `connectivity`.
3. Modifications to the transpilation pass.
4. Updating default transpiler.
5. Updating test files.
6. Usage of `wire_names` in the `QibolabBackend` (in Qibolab's repo).

##### Related Issues

https://github.com/qiboteam/qibo/pull/1440#discussion_r1800997851

https://github.com/qiboteam/qibo/pull/1440#discussion_r1801000856

#1429 

##### References

https://github.com/qiboteam/qibo/pull/1440#issuecomment-2402489708 https://github.com/qiboteam/qibo/pull/1440#discussion_r1795647536

alecandido:
@csookim I just edited your comment to replace the numbered list with checkboxes (at least we could track the upgrades).

Concerning the Qibolab part, open an issue directly in Qibolab (you could reference in here just by pasting the link).

Issue #1429, opened 2024-08-28
Transpiler padding


alecandido:
Currently, the transpiler is already providing some padding
https://github.com/qiboteam/qibo/blob/ecbd16d7d8f2b6780bacbb7b92c9f2b3fc44087b/src/qibo/transpiler/optimizer.py#L19-L33
but just applying the basic approach, i.e. padding at the end.

However, there is not much more that can be done at the moment, since Qibolab does not support using its own qubits' names in circuits.
https://github.com/qiboteam/qibolab/issues/1009

As soon as that will be available, it would be interesting to offer the option to choose which subset of qubits should be targeted. And applying padding only at the end.

This last part could be done even immediately: sometimes it may be interesting to route a circuit on a subset of the whole qubits and connectivity (even though some subsets will make impossible to execute some circuits), so padding as preprocessing and padding as postprocessing are truly two different features, and only the first is currently available (unless I'm missing something).

alecandido:
@Simone-Bordoni can you take a look at the description? I'm not asking you to implement anything of what I described. And maybe it's not even urgent to anything at all. But at least I would have a consistent issue for when someone will be able to take over the project.

csookim:
Similarly, I'd like to open an issue about the naming of qubits and nodes in the connectivity graph. Would it be better to discuss this here, or should I create a separate issue?

alecandido:
> Similarly, I'd like to open an issue about the naming of qubits and nodes in the connectivity graph. Would it be better to discuss this here, or should I create a separate issue?

The two things are related, and they will be most likely solved together. Feel free to discuss it even here :)
(but if you have a different opinion, do what you consider best)

csookim:
Currently, the transpiler is designed to interpret integer connectivity graph nodes `i` as string physical qubit names `"q" + str(i)`.

https://github.com/qiboteam/qibo/blob/21732ad59d31740e8c0641921abe98e49a1ea8c1/src/qibo/transpiler/placer.py#L203-L205

For example, consider the connectivity graph:

![Screenshot 2024-08-29 at 19 56 06](https://github.com/user-attachments/assets/15d28205-e775-4031-b5f7-9de7bd956314)

and the physical-to-logical mapping:

`{'q0': 4, 'q1': 0, 'q2': 2, 'q3': 1, 'q4': 3}`.

If we want to enable custom string names for qubits, the naming scheme will need to be updated. 

Possible options are:

[...]

PR #1500, opened 2024-10-22
Transpiler refactor

csookim:
This covers issue #1493.

This PR updates the qubit representation in `qibo` to support hardware execution after transpilation. `Circuit.wire_names` and `Gate.qubits` will be used on the hardware side.

### Terms

- physical qubit = qubit name = `Circuit.wire_names: list[str | int]`
- logical qubit = qubit index = `Gate.qubits: Tuple[int, ...]`
- connectivity = `nx.Graph`
- mapping = qubit layout

### Updates

- Removed `initial_layout`, which was used to represent the mapping between `{physical qubit: logical qubit}`.
- `wire_names` itself serves as the mapping. Qubit `i` corresponds to the physical qubit named `wire_names[i]`.

#### Transpiler

- To transpile a circuit, the user must provide `connectivity` representing the hardware.
- Node names in `connectivity` must match the `wire_names` in the circuit.
- `Placer` calculates the placement based on `connectivity` and `circuit`. It updates the `wire_names` directly with no return value.
- `Router` does routing based on `connectivity` and `circuit`, returning a `new_circuit` and `final_mapping`.

Checklist:
- [ ] Tests are passing.


alecandido:
> `wire_names` can now contain both `str` and `int`. If None, default is `["q0", "q1", ...]`.

I would suggest this is the best place to give up on this practice, and start using `int` by default (i.e. the equivalent of `list(range(nqubits))`, but keeping `wire_names` to `None`).

It was introduced for the sake of drawing, which was even the original purpose of `.wire_names`. Now that the attribute has a functional role, let's trim it to minimum required.

alecandido:
> #### 1
> 
> How should we handle cases where a user adds multiple `Placer`s? This doesn’t impact the transpiler. In the previous approach, `Passes` generated an error.

I wonder whether we should allow that at all. This we can discuss.

Placing should be an operation that is required only once, though there is nothing bad in placing again a circuit that has already been placed (at most, it's just useless). So, allowing more composition (even useless one) and lifting restrictions, should simplify the handling, and in general give the user more flexibility (possibly useless).

If it results being more complicated (which I'd be surprised of), then we can keep the error.

alecandido:
> #### 2
> 
> If #qubits in the circuit is different from #nodes in the connectivity, users have two explicit options for transpilation:
> 
>   * Use the `Preprocessor` to automatically add qubits.
>   * Use `on_qubits` to restrict the `connectivity`.


Yes, the two options you mentioned it is what I described in #1429 (approximately) as "padding before" transpilation, or "padding after" (which corresponds to slicing the platform).
I would try to engineer a consistent interface for the user. Even just a flag, or a three-state variable, as `Optional[bool]`, which is setting the padding before, after, or no padding at all - resulting in an error if the number of qubits is not matched (which could be the default, but maybe set differently in the default global transpiler).

> Then, Is it acceptable for all `Placer` and `Router` components to enforce that the input #qubits and #nodes must be the same? Previously, some components had this restriction while others did not.

I would say so: this is the part where the error mentioned above should be raised. Since padding/slicing is happening at a different level, for simplicity all the other ones should assume the simplest case. I.e. the number of qubits matches the (possibly sliced) platform.

> [!NOTE]
> By *slicing* I mean that the platform is considered as if the qubits who are not selected were not present, as well as all the pairs in which one of the two components involved is one of the deselected qubits.

alecandido:
> #### 3
> 
> In this implementation, the `Placer` only changes the order of the names.
> Should we also change the wires accordingly?

I would say that's fine as it is. We could change for drawing, but it's not worth to rearrange a whole `Circuit` just to keep the labels sorted in the original way. And that's even meaningful that they are being drawn reshuffled, at least the change is apparent.

In any case, I would consider this only related to drawing, but other than that we may forget about.
Since it has no practical effect, let's keep internally the minimal amount of operations. It will be more efficient (most likely at an irrelevant level), but also simpler to maintain (which instead is relevant).

src/qibo/transpiler/optimizer.py
```diff
@@ -27,7 +27,10 @@ def __call__(self, circuit: Circuit) -> Circuit:
             )
         if logical_qubits == physical_qubits:
             return circuit
-        new_circuit = Circuit(physical_qubits)
+        new_wire_names = circuit.wire_names.copy() + [
+            name for name in self.connectivity.nodes if name not in circuit.wire_names
+        ]
+        new_circuit = Circuit(nqubits=physical_qubits, wire_names=new_wire_names)
```
alecandido, line 33: This could be the inspiration for a non-breaking change.

We could make `nqubits: Optional[int] = None`, and if the `wire_names` is specified, just compute `nqubits=len(wire_names)`.

In principle, we could even interpret the first argument in different ways according to the type, allowing the
```py
Circuit(["A", "D", "B", "C"])
```
syntax.

(There is even dedicated support in standard library for that, as [`@singledispatch`](https://docs.python.org/3/library/functools.html#functools.singledispatch), though we should check that is available also for methods, and in particular for `__init__` - though the second one is more likely if the first is true, since `__init__` is just a normal method, invoked by the class machinery, while the actual constructor is `__new__`)

src/qibo/transpiler/optimizer.py
```diff
@@ -27,7 +27,10 @@ def __call__(self, circuit: Circuit) -> Circuit:
             )
         if logical_qubits == physical_qubits:
             return circuit
-        new_circuit = Circuit(physical_qubits)
+        new_wire_names = circuit.wire_names.copy() + [
+            name for name in self.connectivity.nodes if name not in circuit.wire_names
+        ]
```
alecandido, line 32:
```suggestion
        new_wire_names = circuit.wire_names + list(self.connectivity.nodes - circuit.wire_names)
```

`nx.NodeView` object seems to support set operations (union, intersection, ... and in particular sets' difference).
Moreover, `list.copy()` is a shallow copy. Such a shallow copy is also provided by plain addition.
```py
[ins] In [1]: x = [0, 1, 2]
[ins] In [2]: y = x + [4]
[ins] In [3]: del y[0]
[ins] In [4]: y
Out[4]: [1, 2, 4]                                                                                                                                                                           
[ins] In [5]: x
Out[5]: [0, 1, 2]
```
(of course, it's shallow, so deep modifications within the list elements are shared, but that's the same with `.copy()`)

src/qibo/transpiler/placer.py
```diff
@@ -133,118 +117,68 @@ def __call__(self, circuit: Circuit):
...
-        else:
-            trivial_layout = dict(
-                zip(
-                    ["q" + str(i) for i in range(circuit.nqubits)],
-                    range(circuit.nqubits),
-                )
-            )
-        return trivial_layout
+        if circuit is None:
+            raise_error(ValueError, "Circuit must be provided.")
```
alecandido, line 151: This is pretty much equivalent to just drop the default value (you're just converting a `TypeError` into a `ValueError`). Moreover, the other placers have no default value.

tests/conftest.py
```diff
@@ -81,6 +82,54 @@ def clear():
...
+        ]
+        chip.add_edges_from(graph_list)
+        return chip
+
+    return _grid_connectivity
+
+
+@pytest.fixture
+def line_connectivity():
+    def _line_connectivity(n, names=None):
```
alecandido, line 121: I'd say that is not worth making it a fixture, to return a function.

However, what maybe you don't know is that you can natively parametrize the fixtures during their invocation :)
```py
@pytest.fixture
def fixt(request):
    marker = request.node.get_closest_marker("fixt_data")
    return None if marker is None else marker.args[0]


@pytest.mark.fixt_data(42)
def test_fixt(fixt):
    assert fixt == 42
```
https://docs.pytest.org/en/stable/how-to/fixtures.html#using-markers-to-pass-data-to-fixtures

alecandido (commented):
This is a preliminary review with some ides, but not a complete one.

src/qibo/backends/__init__.py
```diff
@@ -144,12 +144,8 @@ def _default_transpiler(cls):
...
             and connectivity_edges is not None
         ):
-            # only for q{i} naming
-            node_mapping = {q: i for i, q in enumerate(qubits)}
-            edges = [
-                (node_mapping[e[0]], node_mapping[e[1]]) for e in connectivity_edges
-            ]
-            connectivity = nx.Graph(edges)
+            connectivity = nx.Graph(connectivity_edges)
+            connectivity.add_nodes_from(qubits)
```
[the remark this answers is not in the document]
csookim: I put this to insert nodes when #qubits is 1 or if there are remaining unconnected nodes. Do these cases happen?

src/qibo/transpiler/optimizer.py
csookim: https://stackoverflow.com/a/34484742

src/qibo/transpiler/placer.py
csookim: [f19dc6e](https://github.com/qiboteam/qibo/pull/1500/commits/f19dc6ed1ecb1e624ba6c4bbc63f72c3d3f56f20)

src/qibo/transpiler/optimizer.py
csookim: [f19dc6e](https://github.com/qiboteam/qibo/pull/1500/commits/f19dc6ed1ecb1e624ba6c4bbc63f72c3d3f56f20)

src/qibo/models/circuit.py
```diff
@@ -406,8 +407,9 @@ def light_cone(self, *qubits):
         qubit_map = {q: i for i, q in enumerate(sorted(qubits))}
         kwargs = dict(self.init_kwargs)
         kwargs["nqubits"] = len(qubits)
+        new_wire_names = [self.wire_names[q] for q in list(sorted(qubits))]
+        kwargs["wire_names"] = new_wire_names
```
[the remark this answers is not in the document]
csookim: [f19dc6e](https://github.com/qiboteam/qibo/pull/1500/commits/f19dc6ed1ecb1e624ba6c4bbc63f72c3d3f56f20)

src/qibo/backends/__init__.py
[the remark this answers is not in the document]
alecandido: Unconnected nodes are not relevant for the various processes involving the connectivity (i.e. routing and placing), because they do not require nor can be involved in swaps.

So, in those cases, the best is just that you work on top of the given connectivity, and you work out the best permutation/swaps additions.
If you are just playing with swaps, adding them on the involved qubits should leave the unconnected ones unaltered (at most, if a new circuit is being created, you may iterate on them at the end, appending all the gates for the qubits in the original circuits, which are not nodes in the connectivity).
Concerning the permutations, instead, you could compute them restricted to the given connectivity, and then apply them to the original list, instead of creating a new list (which could miss the unconnected elements). E.g. if the final permutation is something like:
```py
pi = {0: 4, 1: 1, 4: 0}
```
on a 5-qubit chip with arbitrary names, you could apply it as follows:
```py
new_circ.wire_names = old_circ.wire_names.copy()
for i, o in pi.items():
    new_circ.wire_names[i] = old_circ.wire_names[o]
```

---

`i` and `o` may be swapped, according to the convention used on the permutation. Moreover, the minimal representation of a permutation would be:
```py
[4, 1, 0]
```
i.e. just the values. And you would obtain back the case above by `zip`ping them with the keys obtained from the (sorted) graph nodes.

src/qibo/models/circuit.py
```diff
@@ -300,32 +300,33 @@ def wire_names(self, wire_names: Union[list, dict]):
...
                 )
 
             self._wire_names = [
-                wire_names.get(f"q{i}", f"q{i}") for i in range(self.nqubits)
+                k for k, _ in sorted(wire_names.items(), key=lambda x: x[1])
             ]
         else:
             self._wire_names = [f"q{i}" for i in range(self.nqubits)]
 
+        self.init_kwargs["wire_names"] = self._wire_names
```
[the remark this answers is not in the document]
alecandido: The default could be moved to the getter (the `@property` decorated one), and we could drop the `"q{}"` mangling.

src/qibo/transpiler/optimizer.py
alecandido: Actually, it's working:
```py
[ins] In [1]: from functools import singledispatchmethod
[nav] In [2]: class Ciao:
         ...:     @singledispatchmethod
         ...:     def __init__(self, arg):
         ...:         raise NotImplementedError
         ...:
         ...:     @__init__.register
         ...:     def _(self, arg: int):
         ...:         print(-arg)
         ...:
         ...:     @__init__.register
         ...:     def _(self, arg: bool):
         ...:         print(not arg)
[ins] In [3]: Ciao(30)
-30
Out[3]: <__main__.Ciao at 0x7ff57f710510>
[ins] In [4]: Ciao(False)
True
Out[4]: <__main__.Ciao at 0x7ff58418a3d0>
```
https://docs.python.org/3/library/functools.html#functools.singledispatchmethod

src/qibo/models/circuit.py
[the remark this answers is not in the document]
csookim: When we draw the circuit, it will draw like this when `wire_names` is not set `(circuit._wire_names is None)`.

```py
"0: ─H─U1─U1─U1─U1───────────────────────────x───\n"
"1: ───o──|──|──|──H─U1─U1─U1────────────────|─x─\n"
"2: ──────o──|──|────o──|──|──H─U1─U1────────|─|─\n"
"3: ─────────o──|───────o──|────o──|──H─U1───|─x─\n"
"4: ────────────o──────────o───────o────o──H─x───"
```

Should add an additional q in that cases `(circuit._wire_names is None)`?

```py
"q0: ─H─U1─U1─U1─U1───────────────────────────x───\n"
"q1: ───o──|──|──|──H─U1─U1─U1────────────────|─x─\n"
"q2: ──────o──|──|────o──|──|──H─U1─U1────────|─|─\n"
"q3: ─────────o──|───────o──|────o──|──H─U1───|─x─\n"
"q4: ────────────o──────────o───────o────o──H─x───"
```

src/qibo/transpiler/optimizer.py
csookim: > note that the dispatch happens on the type of the first non-self or non-cls argument:

I tried using @singledispatchmethod, but it requires the specific first argument to find dispatch, which conflicts with initializing circuit using keyword args. We need to add preprocessing before every ``circuit = Circuit(**init_kwargs)``.

The code was:

```py
class Circuit:
...
    @singledispatchmethod
    def __init__(
        self,
        nqubits: Optional[int],
        wire_names: Optional[list],
        accelerators = None,
        density_matrix: bool = False,
    ):
        NotImplementedError("This method should not be called directly.")
                            
    @__init__.register
    def _(self, nqubits: int, wire_names=None, accelerators=None, density_matrix=False):
        self._init(nqubits, wire_names, accelerators, density_matrix)
    
    @__init__.register
    def _(self, wire_names: list, nqubits=None, accelerators=None, density_matrix=False):
        self._init(len(wire_names), wire_names, accelerators, density_matrix)

    def _init(self, nqubits, wire_names, accelerators, density_matrix):
        if nqubits < 1:
            raise_error(
                ValueError,
                f"Number of qubits must be positive but is {nqubits}.",
            )
        ...
```

src/qibo/models/circuit.py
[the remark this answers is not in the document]
alecandido: I'm not sure. To be fair, to me, it's fine both ways: one will be more familiar (with `"q"`s), the other one more accurate (reflecting the actual labels used).

It is mostly aesthetics, but let's ask for @stavros11 opinion.

src/qibo/transpiler/optimizer.py
alecandido: Yes, the idea was indeed to have `nqubits: int`, and `nqubits: list` and interpret the `nqubits: list` as `wire_names` (or choose a more agnostic name to span the two of them).

In any case, the idea is to treat them as mutually exclusive arguments, since one can be inferred from the other. `@singledispatchmethod` was just an option, but you may also do it manually, applying a `None` default and checking that one of the two *has* to be `None`, and it will be inferred from the other.

> which conflicts with initializing circuit using keyword args.

In any case, this was exactly to allow not using only keyword args, but to allow passing both `nqubits` and `wire_names` as positional (since they are mutually alternative).
You can just keep use the signature:
```py
def __init__(self, nqubits: Optional[int, list] = None, ..., wire_names: Optional[list] = None):
    ...
```
and just parse the options by yourself:
- exactly one has to be different from `None`
- the other one is inferred
- if `nqubits` is a `list`, is interpreted as `wire_names`

The only pain point is that this will allow you to also do:
```py
Circuit(nqubits=["A", "B", "C", "D"])
```
which is a bit confusing, and it would have been ruled out by `single_dispatch` (otherwise you can not easily distinguish if an argument has been passed as positional or keyword).

src/qibo/transpiler/placer.py
csookim, line 1: Sorry, the commit is mixed. It includes the new `assert_qubit_match` function, which checks that #qubits matches #nodes in the connectivity in every `Placer`.

src/qibo/transpiler/router.py
csookim, line 1: `assert_qubit_match` in every `Router`.

src/qibo/transpiler/optimizer.py
csookim: [93ac772](https://github.com/qiboteam/qibo/pull/1500/commits/93ac772c579f932e9d70fe7c79f4862b81da829f)

I think `qubits_spec` or `qubit` could work as agnostic names. Let me know if you have other names.

src/qibo/models/circuit.py
```diff
@@ -203,6 +204,30 @@ def __init__(
                 )
             self._distributed_init(nqubits, accelerators)
 
+    def _parse(self, qubit_spec, wire_names):
```
alecandido, line 207: I'd put this out of the class (it does not depend on self), and give it a more telling name (something like `_extract_qubits` or `_determine_qubits`, or similar)

src/qibo/transpiler/optimizer.py
csookim: I think it is better to handle in another PR, if needed.

src/qibo/models/circuit.py
csookim: [41f5eff](https://github.com/qiboteam/qibo/pull/1500/commits/41f5eff1cc44ad8df1c6c9485eda96d638b4f557)

src/qibo/models/circuit.py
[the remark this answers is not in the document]
stavros11: > It is mostly aesthetics, but let's ask for @stavros11 opinion.

We discussed this yesterday with @csookim and I believe we will go without `"q"`s for simplicity. Looking at some changes, I think this has also been implemented.

src/qibo/models/circuit.py
```diff
@@ -120,7 +120,7 @@ class Circuit:
     A specific backend has to be used for performing calculations.
 
     Args:
-        nqubits (int): Total number of qubits in the circuit.
+        nqubits (int, list): Number of qubits in the circuit or a list of wire names.
```
alecandido, line 123: It could also be `None`. Though I'm not sure I want to advertise it...

This argument listing is unsuitable to multiple signatures, the free form would have better served the purpose. Maybe, the best thing we could is providing a couple of examples on top (with `nqubits` and with `wire_names`).

src/qibo/models/circuit.py
```diff
@@ -141,11 +141,9 @@ class Circuit:
             Defaults to ``False``.
         accelerators (dict, optional): Dictionary that maps device names to the number of times each
             device will be used. Defaults to ``None``.
-        wire_names (list or dict, optional): Names for qubit wires.
-            If ``None``, defaults to (``q0``, ``q1``... ``qn``).
+        wire_names (list, optional): Names for qubit wire names.
+            If ``None``, defaults to [``0``, ``1``, ..., ``nqubits - 1``].
             If ``list`` is passed, length of ``list`` must match ``nqubits``.
```
alecandido, line 146: This is not enough now, since `nqubits` became optional (so, if `None` you don't have anything to match).

src/qibo/models/circuit.py
```diff
@@ -406,8 +385,8 @@ def light_cone(self, *qubits):
         qubit_map = {q: i for i, q in enumerate(sorted(qubits))}
         kwargs = dict(self.init_kwargs)
         kwargs["nqubits"] = len(qubits)
+        kwargs["wire_names"] = [self.wire_names[q] for q in list(sorted(qubits))]
```
alecandido, line 388:
```suggestion
        kwargs["wire_names"] = [self.wire_names[q] for q in sorted(qubits)]
```

alecandido (commented):
Some minor comments, but I still need to review the actual transpiler (the diff there is huge, even if some pieces just migrated, sorry)

src/qibo/models/circuit.py
csookim: [a4f2092](https://github.com/qiboteam/qibo/pull/1500/commits/a4f2092708a9f15e066e28fdeb6da18eaaaa302f)

doc/source/code-examples/advancedexamples.rst
```diff
@@ -2094,22 +2094,24 @@ Multiple transpilation steps can be implemented using the :class:`qibo.transpile
...
+    from qibo.transpiler.utils import assert_transpiling
 
     # Define connectivity as nx.Graph
     def star_connectivity():
         chip = nx.Graph()
-        chip.add_nodes_from(list(range(5)))
-        graph_list = [(i, 2) for i in range(5) if i != 2]
-        chip.add_edges_from(graph_list)
+        chip.add_nodes_from(["q0", "q1", "q2", "q3", "q4"])
+        chip.add_edges_from([("q0", "q2"), ("q1", "q2"), ("q2", "q3"), ("q2", "q4")])
```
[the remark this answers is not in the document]
csookim: [a4f2092](https://github.com/qiboteam/qibo/pull/1500/commits/a4f2092708a9f15e066e28fdeb6da18eaaaa302f)

src/qibo/models/circuit.py
csookim: [a4f2092](https://github.com/qiboteam/qibo/pull/1500/commits/a4f2092708a9f15e066e28fdeb6da18eaaaa302f)

src/qibo/models/circuit.py
csookim: [a4f2092](https://github.com/qiboteam/qibo/pull/1500/commits/a4f2092708a9f15e066e28fdeb6da18eaaaa302f)

Please review the updated docstring, and if everything ok, please close this conversation.

csookim:
@alecandido There is an issue in `Circuit.__add__`. Currently, circuits with different `kwargs` cannot be added. `wire_names` is also in `kwargs`.

https://github.com/qiboteam/qibo/blob/a074d176779c15bac7838679aa4a87fcfcdbe597/src/qibo/models/circuit.py#L265-L271

This should be updated, because there are cases where an initial state (wire_names is None) is added to a circuit (wire_names is not None).

https://github.com/qiboteam/qibolab/blob/f74c334d278f9c45de84f38e0f719a25c0dc1fc5/src/qibolab/_core/backends.py#L90-L94

What I have in mind is:
```py
# circuit1 + circuit2
wire_names1 = circuit1.init_kwargs.get("wire_names")
wire_names2 = circuit2.init_kwargs.get("wire_names")

if wire_names1 is None or wire_names2 is None:
    new_name = wire_names1 or wire_names2
elif wire_names1 == wire_names2:
    new_name = wire_names1
raise error
```

stavros11:
> @alecandido There is an issue in `Circuit.__add__`. Currently, circuits with different `kwargs` cannot be added. `wire_names` is also in `kwargs`.

I would say it is fine to raise an error in this case since circuits with different wire names cannot be trivially added, at least using the current implementation of `__add__`.

> This should be updated, because there are cases where an initial state (wire_names is None) is added to a circuit (wire_names is not None).

We were discussing this with @alecandido yesterday, and the circuit initial state is a not so much used feature. Since it's already there, it would be good to have, but I would not consider a priority. Also, if I understand correctly, if the `initial_state` circuit has the same wire names as the `circuit`, `__add__` will work. This is just forcing the user of this feature to use consistent wire names, which is reasonable.

> What I have in mind is:
> 
> ```python
> # circuit1 + circuit2
> wire_names1 = circuit1.init_kwargs.get("wire_names")
> wire_names2 = circuit2.init_kwargs.get("wire_names")
> 
> if wire_names1 is None or wire_names2 is None:
>     new_name = wire_names1 or wire_names2
> elif wire_names1 == wire_names2:
>     new_name = wire_names1
> else:
>     raise error
> ```

If I understand this, here you are basically treating the case where one of the circuits has wire names and the other does not. This is an enhancement on the above and is certainly sufficient for now.

alecandido:
I generally agree with @stavros11, but I'm a bit more inclined in keeping the current behavior as it is.

What I would suggest, is to deprecate the usage of `initial_state` in `QibolabBackend`: the argument has to be there, for compatibility with the backend interface, but if `initial_state is not None`, I'd just raise an error in the backend immediately, not only in the case in which is not a `Circuit`, or a `Circuit` that does not add properly.

The rationale is that there are alternative ways of obtaining the same result (concatenating the initial state on the user side, passing a single circuit to the backend - the one to be executed), and this is even much less confusing.
Indeed, the circuit to be executed in general will also be transpiled, but the initial state may bypass transpilation, because of taking a different route. Not only: even if it's transpiled, transpiling them separately and then concatenating will not be equivalent to first concatenate and then transpiling. Other than the performances (which in general are expected to improve if you can take into account the entire execution as whole), you may even end up in an incorrect result even transpiling both, because in general you will need to introduce a permutation at the end in order to counteract the SWAPs that you introduced during routing (which at the moment we counteract only at the level of measurements, if I understood correctly).

All in all, if your execution requires some kind of preparation, better to prepend it to the `Circuit` to be executed, and **then** transpile and execute.

For this reason, I'd stop supporting `initial_state` argument in `QibolabBackend` immediately (for a transition period it can be a "soft" deprecation, not doing anything, but eventually dropping the concatenation code and just raising an error).

csookim:
> For this reason, I'd stop supporting `initial_state` argument in `QibolabBackend` immediately (for a transition period it can be a "soft" deprecation, not doing anything, but eventually dropping the concatenation code and just raising an error).

Should we keep the `initial_state` variable but update it as deprecated (should not be used when using the hardware backend)? Then, I will open an issue on the qibolab side.

csookim:
Currently, `Optimizer`, `Placer`, and `Router` have both `__init__()` and `__call__()`.

Since `connectivity` is checked in `__init__()`, the connectivity is duplicated when defining the transpiler. And `Passes` already overwrites the specified `connectivity` for all passes.

What if we move the `connectivity` check to `__call__()`?

```py
p = Passes(
    connectivity=connectivity,
    passes=[
        Preprocessing(connectivity),
        Trivial(connectivity),
        Sabre(connectivity),
        Unroller(NativeGates[natives]),
    ],
)
```
Then above code will be

```py
p = Passes(
    connectivity=connectivity,
    passes=[Preprocessing(), Trivial(), Sabre(), Unroller(NativeGates[natives])],
)
```

csookim:
#### 1
Do we need a default transpiler? This is not be a default backend transpiler but a default one for star connectivity.

https://github.com/qiboteam/qibo/blob/ff89c58f93215fbec091b84eb231a747a805dfb9/src/qibo/transpiler/pipeline.py#L71-L88

#### 2

Should we remove the `Custom` placer? It could be replaced with the `wire_names` setter.

alecandido:
> Do we need a default transpiler? This is not be a default backend transpiler but a default one for star connectivity.

Not at all

> Should we remove the `Custom` placer? It could be replaced with the `wire_names` setter.

That would definitely make sense :)

src/qibo/transpiler/utils.py
alecandido, line 1: In principle, since this file is fully dedicated to asserts, you could even name it `asserts.py` :P

alecandido (approved):
Thanks @csookim, to me, it seems mostly alright.

Obviously, this PR is large enough that it's hard to catch all the minor details and consequences, but since(/assuming) you even managed to run on hardware, I see no reason to keep it stuck any further.

I just have a few minor comments, but other than that, feel free to keep going.

(as discussed offline, some guide about the role of `.wire_names` would be certainly appreciated by our users, but I would open a separate PR for it)

src/qibo/models/circuit.py
```diff
@@ -155,29 +165,20 @@ class Circuit:
 
     def __init__(
         self,
-        nqubits: int,
+        nqubits: Optional[Union[str, list]] = None,
```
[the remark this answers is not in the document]
csookim: [797331e](https://github.com/qiboteam/qibo/pull/1500/commits/797331ed774f374481bac37984522a94d9e66cf7)

src/qibo/transpiler/optimizer.py
```diff
@@ -13,10 +13,18 @@ class Preprocessing(Optimizer):
         connectivity (:class:`networkx.Graph`): hardware chip connectivity.
     """
 
-    def __init__(self, connectivity: nx.Graph):
+    def __init__(self, connectivity: nx.Graph = None):
```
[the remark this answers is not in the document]
csookim: [797331e](https://github.com/qiboteam/qibo/pull/1500/commits/797331ed774f374481bac37984522a94d9e66cf7)

src/qibo/transpiler/optimizer.py
```diff
@@ -13,10 +13,18 @@ class Preprocessing(Optimizer):
...
         self.connectivity = connectivity
 
     def __call__(self, circuit: Circuit) -> Circuit:
+        if self.connectivity is None or not all(
```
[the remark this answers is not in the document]
csookim: [797331e](https://github.com/qiboteam/qibo/pull/1500/commits/797331ed774f374481bac37984522a94d9e66cf7)

alecandido (approved):

stavros11 (approved):

PR #1504, opened 2024-10-29
Unsupported native gates

csookim:
`Unroller` generates an error when initialized with unsupported gates.

https://github.com/qiboteam/qibolab/actions/runs/11569426877/job/32203123767?pr=1083#step:10:259

In this PR, when qibo receives native gates, unsupported gates will be ignored. Should we require backends to match qibo’s native gate set?

Checklist:
- [ ] Tests are passing.

https://github.com/qiboteam/qibolab/pull/1083

### DIFF
--- a/src/qibo/transpiler/unroller.py
+++ b/src/qibo/transpiler/unroller.py
@@ -4,7 +4,7 @@ from operator import or_
 
 from qibo import gates
 from qibo.backends import _check_backend
-from qibo.config import raise_error
+from qibo.config import log, raise_error
 from qibo.models import Circuit
 from qibo.transpiler._exceptions import DecompositionError
 from qibo.transpiler.decompositions import (
@@ -22,7 +22,11 @@ class FlagMeta(EnumMeta):
 
     def __getitem__(cls, keys):
         if isinstance(keys, str):
-            return super().__getitem__(keys)
+            try:
+                return super().__getitem__(keys)
+            except KeyError:
+                log.info(f"Native gate {keys} is not supported by qibo.")
+                return super().__getitem__("NONE")
         return reduce(or_, [cls[key] for key in keys])  # pylint: disable=E1136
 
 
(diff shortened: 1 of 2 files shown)


alecandido:
Ok, it should be fine to ignore the native gates which are unsupported, and restrict to the known ones. The alternative is just failing, but this is what it would happen anyhow if the restricted set of natives won't allow you to transpile (to save computation, we may introduce ahead of time checks, but if there is no way to tell it in advance, attempting and failing is just the best we could do).

I'm not sure which is the practical motivation in this case (I guess coming from Qibolab), which instead may be worth to address even in a separate way (though this is still fine to me).
@stavros11 do you see any major drawback in this approach?

src/qibo/transpiler/unroller.py
```diff
@@ -22,7 +22,11 @@ class FlagMeta(EnumMeta):
 
     def __getitem__(cls, keys):
         if isinstance(keys, str):
-            return super().__getitem__(keys)
+            try:
+                return super().__getitem__(keys)
+            except KeyError:
+                log.info(f"Native gate {keys} is not supported by qibo.")
```
alecandido, line 28: I know we're inconsistent in logging, but in this respect I became strongly opinionated.

https://github.com/qiboteam/qibo-cloud-backends/pull/30#discussion_r1820379326

csookim: Should we remove the log?

csookim: [23423a9](https://github.com/qiboteam/qibo/pull/1504/commits/23423a9b0b45d5acddd839eb19ceac3197c82954)

stavros11 (approved):
If I understand this correctly, the motivation is to avoid failure when the qibolab platform/compiler provides more native gates than the transpiler "knows". In that case, transpilation should still be possible using the intersection of gates supported by the transpiler and compiler/platform. 

Other than the fact that it may be a bit suboptimal, as I would expect the gate limitation to come primarily from the hardware, not the transpiler, I do not see any major drawback so I am approving.

alecandido (approved):

PR #1510, opened 2024-10-30
Migrating Pytorch backend to Qiboml

MatteoRobbiati:
Checklist:
- [x] Tests are passing.

### Some things to be discussed and still not touched in this PR

1. I think we should fix the procedure based on [this function](https://github.com/qiboteam/qibo/blob/f37c024189e701e82a045a8e8829b9cfe7e917e5/src/qibo/gates/gates.py#L2638), otherwise we will always need to depend on Torch;
2. In examples like [this one](https://github.com/qiboteam/qibo/blob/f37c024189e701e82a045a8e8829b9cfe7e917e5/doc/source/code-examples/advancedexamples.rst?plain=1#L786) we define a torch tensor which is provided to our circuit. I would like to drop any dependence on pytorch in Qibo, thus I propose to delete these kind of advanced examples, pointing to Qiboml documentation (where we will rewrite dedicated examples). Otherwise, we should build the torch tensor doing something like the following (which is very confusing and we must avoid):

```python
backend = construct_backend(backend="qiboml", platform="pytorch")

# create a torch tensor from backend using torch which is renamed np
tensor = backend.np.ones( .... )
```


doc/source/code-examples/advancedexamples.rst
```diff
@@ -543,7 +543,9 @@ refer to the :ref:`Optimizers <Optimizers>` section of the documentation.
 Note that if the Stochastic Gradient Descent optimizer is used then the user
 has to use a backend based on tensorflow or pytorch primitives and not the default custom
 backend, as custom operators currently do not support automatic differentiation.
-To switch the backend one can do ``qibo.set_backend("tensorflow")`` or ``qibo.set_backend("pytorch")``.
+To switch the backend one can do ``qibo.set_backend(backend="qiboml", platform="tensorflow")``
+or ``qibo.set_backend(backend="qiboml", platform="pytorch")``, after ensuring the
+Qiboml package has been installed.
```
renatomello, line 548:
```suggestion
``qiboml`` package has been installed.
```

doc/source/getting-started/installation.rst
```diff
@@ -7,15 +7,15 @@ Operating systems support
...
-+------------------+------+---------+------------+
++------------------+------+---------+-----------+---------+
+| Operating System | qibo | qibojit |Tensorflow | Pytorch |
++==================+======+=========+===========+=========+
+| Linux x86        | Yes  | Yes     | Yes       | Yes     |
++------------------+------+---------+-----------+---------+
+| MacOS >= 10.15   | Yes  | Yes     | Yes       | Yes     |
++------------------+------+---------+-----------+---------+
+| Windows          | Yes  | Yes     | Yes       | Yes     |
++------------------+------+---------+-----------+---------+
```
renatomello, line 18: Should this be updated to contemplate the CPU / GPU subdivision? Both `qibojit` and `qiboml` have backends that work on both.

renatomello:
>     1. I think we should fix the procedure based on [this function](https://github.com/qiboteam/qibo/blob/f37c024189e701e82a045a8e8829b9cfe7e917e5/src/qibo/gates/gates.py#L2638), otherwise we will always need to depend on Torch;

That function is needed to properly have the `gates.Unitary.dagger` working for all possible input types the matrix can be (_i.e._ `np.ndarray`, `tf.Tensor`, `torch.Tensor`, `cp.ndarray`). I'm not sure there is a way out of this. One immediate thing we could do is to rename `np` as `platform` in all backends since that's the name we give in `set_backend` anyways.

doc/source/getting-started/installation.rst
```diff
@@ -197,26 +168,45 @@ please do:
...
+into these well-known machine learning frameworks. Quantum layers can be constructed
+and added to your Pytorch or Tensorflow models, and trained using any supported
+optimization routine.
 
-In order to switch to the ``pytorch`` backend please do:
+In order to use these quantum machine learning backends please make sure the
+preferred package is installed following `Tensorflow's <https://www.tensorflow.org/install>`_
+or `Pytorch's <https://pytorch.org/get-started/locally/>`_ installation instructions.
+
+To switch to Tensorflow or Pytorch backend please do:
```
BrunoLiegiBastonLiegi, line 191: Not sure whether it is worth to have details about the ml backends. I would probably place this in the `qiboml` docs and link them here.

tests/test_hamiltonians.py
```diff
@@ -69,10 +69,10 @@ def transformation_d(a, b, use_eye=False):
...
         mH1, mH2 = backend.to_numpy(H1.matrix), backend.to_numpy(H2.matrix)
     else:
-        if backend.platform == "tensorflow":
-            pytest.skip("Tensorflow does not support operations with sparse matrices.")
-        elif backend.name == "pytorch":
-            pytest.skip("Pytorch does not support operations with sparse matrices.")
+        if backend.platform in ["tensorflow", "pytorch"]:
+            pytest.skip(
+                "Tensorflow and Pytorch do not support operations with sparse matrices."
+            )
```
BrunoLiegiBastonLiegi, line 75: This is not really for this PR, but I think that both of them support sparse matrices to some extent [https://pytorch.org/docs/stable/generated/torch.sparse.mm.html](https://pytorch.org/docs/stable/generated/torch.sparse.mm.html)
[https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/sparse-mat-mul](https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/sparse-mat-mul)

I am not sure what was causing problems here, but I guess it is more a matter of how the operation is called rather than the impossibility to perform it. Anyway, this is probably out of the scope of this PR.

tests/test_models_variational.py
```diff
@@ -102,13 +102,11 @@ def myloss(parameters, circuit, target):
 def test_vqe(backend, method, options, compile, filename):
     """Performs a VQE circuit minimization test."""
     if (method == "sgd" or compile) and (
-        (backend.name != "pytorch") or (backend.platform != "tensorflow")
+        backend.platform not in ["tensorflow", "pytorch"]
```
BrunoLiegiBastonLiegi, line 105: This could become a property of the backend, e.g.:
``` python
@property
def support_differentiation(self,):
    return False
```
to avoid checking precisely for which backend this works, but again might be out of scope here.

tests/test_hamiltonians.py
MatteoRobbiati: I just opened an issue with this comment.

tests/test_models_variational.py
MatteoRobbiati: I think the best thing for Qibo will be to remove all these tests about variational models.

doc/source/getting-started/installation.rst
MatteoRobbiati: I don't know, since this is already shown in the backends diagram.

MatteoRobbiati:
By my side, I think we can stop modifications here for this PR. Still we need to remove a lot of examples and subsections from the docs, but I would prefer to do that after we populate the Qiboml documentation. Otherwise we will have vacations in the documentation. What you think @renatomello, @BrunoLiegiBastonLiegi?

doc/source/getting-started/installation.rst
MatteoRobbiati: I agree, I would remove these lines. Probably it is better to wait the moment the Qiboml docs are ready so that we can just link to the proper section.

renatomello (approved):

MatteoRobbiati:
> We need to update this https://github.com/qiboteam/qiboml/blob/800e800d54f4962f4890716c4b1539014c69c1b5/src/qiboml/backends/__init__.py#L3

Done in [qiboml](https://github.com/qiboteam/qiboml/pull/52), thanks!

> BTW, is the `PytorchBackend` of `qiboml` updated or we are still waiting for [qiboteam/qiboml#30](https://github.com/qiboteam/qiboml/pull/30) ?

That was outdated. Just closed!

renatomello:
`qibo` is importing the `torch` backend from `qiboml` and `qiboml` is importing the `torch` backend from `qibo`

https://github.com/qiboteam/qibo/actions/runs/11888475398/job/33209011064?pr=1510

MatteoRobbiati:
> `qibo` is importing the `torch` backend from `qiboml` and `qiboml` is importing the `torch` backend from `qibo`

This should have been solved merging https://github.com/qiboteam/qiboml/pull/52#issue-2671210521 yesterday. Rerunning the tests to double check :)

renatomello:
> I just removed a gradient test from here and opened a dedicated issue in qiboml. Also, the `test_list_available_backends` should be fixed here, so after merging the master into your branch, @renatomello, the problem mentioned in #1520 should be solved.

You mean after this PR is merged, right?

MatteoRobbiati:
> You mean after this PR is merged, right?

Yes, I mean this PR should be solving that problem too.

BrunoLiegiBastonLiegi (approved):

Issue #1524, opened 2024-11-15
Copying circuit `wire_names`


csookim:
When copying the circuit, `wire_names is not transferred to the new circuit.

Minimal update to enable hw execution test.

csookim:
@scarrazza Could you give permission to create a branch?

scarrazza:
Retry now please.

csookim:
Enabling int wire_names is necessary.

Instead of merging this, I will create a branch that includes both this update and the int `wire_names` update and test the branch on qrccluster.

Close this issue.

PR #1525, opened 2024-11-15
Copying circuit `wire_names`

csookim:
This covers issue #1524. 

Checklist:
- [ ] Tests are passing.


csookim:
This does not fully resolve the issue and enabling int `wire_names` is necessary.

Instead of merging this, I will create a branch that includes both this update and the int `wire_names` update and test the branch on qrccluster.

Close this PR.

Issue #1502, opened 2024-10-25
Parameterized gates for GST


mho291:
In this current version of gate set tomography, it seems like only non-parameterized gates can be passed into GST.

There are two ways that I generate non-parameterized gates:
``` python
gate_set = [gates.I, gates.X, gates.H, gates.S]
```
or
``` python
target_gates = [gates.SX(0), gates.Z(0), gates.RX(0, np.pi/2)]
gate_set = [g.__class__ for g in target_gates]
```

Then we pass them into `GST`:
``` python
from qibo.tomography import gate_set_tomography 
gate_set_tomography.GST(gate_set=gate_set, 
                        nshots=int(1e3), 
                        noise_model=None, 
                        include_empty=True, 
                        pauli_liouville=False, 
                        gauge_matrix=None, 
                        backend=NumpyBackend())
```

In lines 288-300 of `gate_set_tomography.py`, which is this:
``` python
for gate in gate_set:
        if gate is not None:
            init_args = signature(gate).parameters
            if "q" in init_args:
                nqubits = 1
            elif "q0" in init_args and "q1" in init_args and "q2" not in init_args:
                nqubits = 2
            else:
                raise_error(
                    RuntimeError,
                    f"Gate {gate} is not supported for `GST`, only 1- and 2-qubits gates are supported.",
                )
            gate = gate(*range(nqubits))
```
it only extracts which qubits the gate acts on. I've tried to extract the `parameters` from the gates in `gate_set` but it always returns something like this `<property object at 0x123456789>, i.e.
``` python
target_gates = [gates.SX(0), gates.Z(0), gates.RX(0, np.pi/2)]
gate_set = [g.__class__ for g in target_gates]

print(gate_set[2].parameters)

>> <property object at 0x123456789> 
```
Looking back at our input for `gate_set`, it turns out that it's difficult to extract `parameters` from each of the gates in `gate_set`.

I realise that if we can extract the parameters from the `target_gates`.
``` python
target_gates = [gates.SX(0), gates.Z(0), gates.RX(0, np.pi/2)]

print(target_gates[2].parameters)

> (1.5707963267948966,)
```

Hence, for GST, is there another way of inputting the gates instead of doing `gate_set = [g.__class__ for g in target_gates]`?

alecandido:
`.parameters` is a property bound to an instance. If you take the class, there is no way you can access parameters

```py
[ins] In [1]: from qibo.gates import RX
[...]

PR #1534, opened 2024-12-03
Updating GST to accept parameterised gates

mho291:
Checklist:
- [ ] Tests are passing.

### DIFF
--- a/doc/source/code-examples/advancedexamples.rst
+++ b/doc/source/code-examples/advancedexamples.rst
@@ -2242,8 +2242,7 @@ Let's first define the set of gates we want to estimate:
 
    from qibo import gates
 
-   target_gates = [gates.RX(0, np.pi/3), gates.Z(0), gates.PRX(0, np.pi/2, np.pi/3), gates.GPI(0, np.pi/7), gates.CNOT(0,1)]
-   gate_set = [(g.__class__, [g.parameters[i] for i in range(len(g.parameters))]) if g.parameters else (g.__class__, []) for g in target_gates]
+   gate_set = [(gates.RX, [np.pi/3]), gates.Z, (gates.PRX, [np.pi/2, np.pi/3]), (gates.GPI, [np.pi/7]), gates.CNOT]
 
 For simulation purposes we can define a noise model. Naturally this is not needed when running on real quantum hardware, which is intrinsically noisy. For example, we can suppose that the three gates we want to estimate are going to be noisy:
 
--- a/doc/source/code-examples/advancedexamples.rst
+++ b/doc/source/code-examples/advancedexamples.rst
@@ -2242,7 +2242,8 @@ Let's first define the set of gates we want to estimate:
 
    from qibo import gates
 
-   gate_set = {gates.X(0), gates.H(0), gates.CZ(0, 1)}
+   target_gates = [gates.RX(0, np.pi/3), gates.Z(0), gates.PRX(0, np.pi/2, np.pi/3), gates.GPI(0, np.pi/7), gates.CNOT(0,1)]
+   gate_set = [(g.__class__, [g.parameters[i] for i in range(len(g.parameters))]) if g.parameters else (g.__class__, []) for g in target_gates]
 
 For simulation purposes we can define a noise model. Naturally this is not needed when running on real quantum hardware, which is intrinsically noisy. For example, we can suppose that the three gates we want to estimate are going to be noisy:
 
--- a/doc/source/code-examples/advancedexamples.rst
+++ b/doc/source/code-examples/advancedexamples.rst
@@ -2240,7 +2240,7 @@ Let's first define the set of gates we want to estimate:
 
    from qibo import gates
 
-   gate_set = {gates.X, gates.H, gates.CZ}
+   gate_set = {gates.X(0), gates.H(0), gates.CZ(0, 1)}
 
 For simulation purposes we can define a noise model. Naturally this is not needed when running on real quantum hardware, which is intrinsically noisy. For example, we can suppose that the three gates we want to estimate are going to be noisy:
 
--- a/src/qibo/tomography/gate_set_tomography.py
+++ b/src/qibo/tomography/gate_set_tomography.py
@@ -1,4 +1,5 @@
 from functools import cache
+from inspect import signature
 from itertools import product
 from random import Random
 from typing import List, Union
@@ -227,7 +228,12 @@ def GST(
     """Runs Gate Set Tomography on the input ``gate_set``.
 
     Args:
-        gate_set (tuple or set or list): set of :class:`qibo.gates.Gate` to run GST on.
+        gate_set (tuple or set or list): set of :class:`qibo.gates.Gate` and parameters to run
+            GST on.
+            E.g. target_gates = [gates.RX(0, np.pi/3), gates.Z(0), gates.PRX(0, np.pi/2, np.pi/3),
+                                 gates.GPI(0, np.pi/7), gates.CNOT(0,1)]
+                 gate_set = [(g.__class__, [g.parameters[i] for i in range(len(g.parameters))])
+                             if g.parameters else (g.__class__, []) for g in target_gates]
         nshots (int, optional): number of shots used in Gate Set Tomography per gate.
             Defaults to :math:`10^{4}`.
         noise_model (:class:`qibo.noise.NoiseModel`, optional): noise model applied to simulate
@@ -286,12 +292,25 @@ def GST(
 
     for gate in gate_set:
         if gate is not None:
-            nqubits = len(gate.qubits)
-            if nqubits not in (1, 2):
+            init_args = signature(gate[0]).parameters
+            params = gate[1]
+
+            angle_names = [name for name in init_args if name in {"theta", "phi"}]
+
+            angle_values = {}
+            for name, value in zip(angle_names, params):  # Zip ensures correct order
+                angle_values[name] = value
+
+            if "q" in init_args:
+                nqubits = 1
+            elif "q0" in init_args and "q1" in init_args and "q2" not in init_args:
+                nqubits = 2
+            else:
                 raise_error(
                     RuntimeError,
                     f"Gate {gate} is not supported for `GST`, only 1- and 2-qubits gates are supported.",
                 )
+            gate = gate[0](*range(nqubits), **angle_values)
 
         matrices.append(
             _gate_tomography(
--- a/src/qibo/tomography/gate_set_tomography.py
+++ b/src/qibo/tomography/gate_set_tomography.py
@@ -307,7 +307,7 @@ def GST(
             else:
                 raise_error(
                     RuntimeError,
-                    f"Gate {gate} is not supported for `GST`, only 1- and 2-qubits gates are supported.",
+                    f"Gate {gate} is not supported for `GST`, only 1- and 2-qubit gates are supported.",
                 )
             gate = gate(*range(nqubits), **angle_values)
 
--- a/src/qibo/tomography/gate_set_tomography.py
+++ b/src/qibo/tomography/gate_set_tomography.py
@@ -293,10 +293,12 @@ def GST(
             if isinstance(gate, tuple):
                 angles = ["theta", "phi", "lam"]
                 gate, params = gate
-                angle_values = dict(zip(angles[: len(params)], params))
+                init_args = signature(gate).parameters
+                valid_angles = [arg for arg in init_args if arg in angles]
+                angle_values = dict(zip(valid_angles, params))
             else:
                 angle_values = {}
-            init_args = signature(gate).parameters
+                init_args = signature(gate).parameters
 
             if "q" in init_args:
                 nqubits = 1
--- a/src/qibo/tomography/gate_set_tomography.py
+++ b/src/qibo/tomography/gate_set_tomography.py
@@ -1,13 +1,13 @@
 from functools import cache
-from inspect import signature
 from itertools import product
+from random import Random
 from typing import List, Union
 
 import numpy as np
 from sympy import S
 
 from qibo import Circuit, gates, symbols
-from qibo.backends import _check_backend, get_transpiler
+from qibo.backends import _check_backend
 from qibo.config import raise_error
 from qibo.hamiltonians import SymbolicHamiltonian
 from qibo.transpiler.optimizer import Preprocessing
@@ -260,7 +260,15 @@ def GST(
     backend = _check_backend(backend)
 
     if backend.name == "qibolab" and transpiler is None:  # pragma: no cover
-        transpiler = get_transpiler()
+        transpiler = Passes(
+            connectivity=backend.platform.topology,
+            passes=[
+                Preprocessing(backend.platform.topology),
+                Random(backend.platform.topology),
+                Sabre(backend.platform.topology),
+                Unroller(NativeGates.default()),
+            ],
+        )
 
     matrices = []
     empty_matrices = []
@@ -278,17 +286,12 @@ def GST(
 
     for gate in gate_set:
         if gate is not None:
-            init_args = signature(gate).parameters
-            if "q" in init_args:
-                nqubits = 1
-            elif "q0" in init_args and "q1" in init_args and "q2" not in init_args:
-                nqubits = 2
-            else:
+            nqubits = len(gate.qubits)
+            if nqubits not in (1, 2):
                 raise_error(
                     RuntimeError,
                     f"Gate {gate} is not supported for `GST`, only 1- and 2-qubits gates are supported.",
                 )
-            gate = gate(*range(nqubits))
 
         matrices.append(
             _gate_tomography(
--- a/src/qibo/tomography/gate_set_tomography.py
+++ b/src/qibo/tomography/gate_set_tomography.py
@@ -1,7 +1,6 @@
 from functools import cache
 from inspect import signature
 from itertools import product
-from random import Random
 from typing import List, Union
 
 import numpy as np
@@ -230,10 +229,8 @@ def GST(
     Args:
         gate_set (tuple or set or list): set of :class:`qibo.gates.Gate` and parameters to run
             GST on.
-            E.g. target_gates = [gates.RX(0, np.pi/3), gates.Z(0), gates.PRX(0, np.pi/2, np.pi/3),
-                                 gates.GPI(0, np.pi/7), gates.CNOT(0,1)]
-                 gate_set = [(g.__class__, [g.parameters[i] for i in range(len(g.parameters))])
-                             if g.parameters else (g.__class__, []) for g in target_gates]
+            E.g. gate_set = [(gates.RX, [np.pi/3]), gates.Z, (gates.PRX, [np.pi/2, np.pi/3]),
+                             (gates.GPI, [np.pi/7]), gates.CNOT]
         nshots (int, optional): number of shots used in Gate Set Tomography per gate.
             Defaults to :math:`10^{4}`.
         noise_model (:class:`qibo.noise.NoiseModel`, optional): noise model applied to simulate
@@ -292,21 +289,14 @@ def GST(
 
     for gate in gate_set:
         if gate is not None:
-            if isinstance(gate, tuple):
-                init_args = signature(gate[0]).parameters
-                params = gate[1]
-                gate_class = gate[0]
-            else:
-                init_args = signature(gate).parameters
-                params = []
-                gate_class = gate
 
-            angle_names = [
-                name for name in init_args if name in {"theta", "phi", "lam"}
-            ]
-            angle_values = {}
-            for name, value in zip(angle_names, params):
-                angle_values[name] = value
+            if isinstance(gate, tuple):
+                angles = ["theta", "phi", "lam"]
+                gate, params = gate
+                angle_values = dict(zip(angles[: len(params)], params))
+            else:
+                angle_values = {}
+            init_args = signature(gate).parameters
 
             if "q" in init_args:
                 nqubits = 1
@@ -317,12 +307,12 @@ def GST(
                     RuntimeError,
                     f"Gate {gate} is not supported for `GST`, only 1- and 2-qubits gates are supported.",
                 )
-            what_gate = gate_class(*range(nqubits), **angle_values)
+            gate = gate(*range(nqubits), **angle_values)
 
         matrices.append(
             _gate_tomography(
                 nqubits=nqubits,
-                gate=what_gate,
+                gate=gate,
                 nshots=nshots,
                 noise_model=noise_model,
                 backend=backend,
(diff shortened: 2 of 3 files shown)


BrunoLiegiBastonLiegi:
I am no expert of tomography, but exactly what do you mean by "tomography of a parametrized gate"? I mean, I presume you are thinking of: fixing the set of parameters and performing the tomography for that specific parameter choice. So it's not really a parametrized gate anymore.

alecandido:
@BrunoLiegiBastonLiegi the problem I believe @mho291 is trying to address should not be theoretical, but rather technical. Cf. https://github.com/qiboteam/qibo/issues/1502

So, you're right that once the parameters are fixed, the gate is completely analogous to any other fixed one. But from the point of view of Qibo is still holding some parameters in an attribute. Thus, it's parameterized in that sense...

mho291:
Thanks @BrunoLiegiBastonLiegi and @alecandido. I guess the most straightforward reason to passing in a gate with parameters is so that the user does not need to transpile it to the standard gates. Will it be odd if we passed the entire gate but not use the qubit parameter, for e.g. `gates.RX(qubit=15, theta=np.pi/4)`? Seems less complicated than having the gate + gate parameters. But otherwise, I like the suggestion to use separate `gate_parameters` and we can think about it.

Following the GST wishlist [here](https://github.com/qiboteam/qibo/issues/1377#issue-2387354902), we might want to consider adding a feature that allows the user to specify which qubit to use and gates to use for state preparation & measurements, so that GST can be used on a device that supports verbatim compilation, like the Amazon Braket devices. It's more complicated than adding `gate_parameters`, but I feel that adding `gate_parameters` now is one step towards full customisability of GST. 

I've been doing GST with verbatim compilation on IQM Garnet recently so I have some gates that we can use for state preparation and measurements. There, I extracted the qubit information and used it for state preparation and measurements. But maybe let's leave that for a later task?

alecandido:
> Will it be odd if we passed the entire gate but not use the qubit parameter, for e.g. `gates.RX(qubit=15, theta=np.pi/4)`? Seems less complicated than having the gate + gate parameters. But otherwise, I like the suggestion to use separate `gate_parameters` and we can think about it.

In case, I would explore the option of using some clearly invalid value as a placeholder. More in details, not to check for this value (and just ignore it, as you would do anyhow), but rather to systematically use that to document the GST usage.

The best one would be `None`, i.e. `gates.RX(None, theta=np.pi/4)`. If the type is checked anywhere, we could try with `-1`.

This is just to make it clear that this value is completely irrelevant (which makes explicit that the interface would be better without, but hiding it is just worse).

> Following the GST wishlist [here](https://github.com/qiboteam/qibo/issues/1377#issue-2387354902), we might want to consider adding a feature that allows the user to specify which qubit to use and gates to use for state preparation & measurements, so that GST can be used on a device that supports verbatim compilation, like the Amazon Braket devices. It's more complicated than adding `gate_parameters`, but I feel that adding `gate_parameters` now is one step towards full customisability of GST.
> 
> I've been doing GST with verbatim compilation on IQM Garnet recently so I have some gates that we can use for state preparation and measurements. There, I extracted the qubit information and used it for state preparation and measurements. But maybe let's leave that for a later task?

Definitely, let's keep for a separate task.

The one in this PR is a patch, to allow working with something that before was not accepted (unexpectedly, perhaps).
If you really want to add the qubit specification as a feature, better to plan it and designing it properly. So, a dedicated PR would be certainly recommended.

mho291:
Thanks @alecandido . To carry on the discussion further, what might be the best way forward? I see two possible routes we can take:

1. To modify this PR to add gate parameters, qubit specification & gates for state prep and measurement.
2. To add only the gate parameters. (I feel that this will be sufficient for the general end-user like myself)

Thanks a lot! Enjoy the weekends!

alecandido:
@mho291 in case of doubt, do less (per PR). So, 2. is definitely better than 1.. @BrunoLiegiBastonLiegi's proposal was to do even less, not more (i.e. to avoid instances, in order to avoid the qubit specification).

Concerning the alternative of doing more or less, the answer is less - and then move to further PRs.
Regarding @BrunoLiegiBastonLiegi suggestion, that's a relevant UI design question. But since it's about a UI which has a limited number of users anyhow (at least for the time being), I'd suggest you to limit the design time as well, and find a solution which is suitable to you. Whatever we do, it will never be perfect at first shot, and we'll refactor later on ^^

> Thanks a lot! Enjoy the weekends!

Sorry for the late reply. I actually enjoyed it :P

mho291:
Thanks @alecandido , let me do it like how @BrunoLiegiBastonLiegi suggested, to incorporate the gate parameters:
`GST(gate_set=[(gates.RX, rx_param), gates.Y, (gates.U3, u3_params)], ...)`
We can think about adding qubit specification later on.

Glad you enjoyed the weekends! ;)

tests/test_tomography_gate_set_tomography.py
```diff
@@ -218,11 +218,18 @@ def test_GST(backend, target_gates, pauli_liouville):
...
     ]
-    # gate_set = [g.__class__ for g in target_gates]
+    gate_set = [
+        (
+            (g.__class__, [g.parameters[i] for i in range(len(g.parameters))])
+            if g.parameters
+            else (g.__class__, [])
+        )
+        for g in target_gates
+    ]
```
BrunoLiegiBastonLiegi, line 228:
```suggestion
    gate_set = [
        (
            (g.__class__, g.parameters)
            if g.parameters
            else g.__class__
        )
        for g in target_gates
    ]
```

isn't this working?

tests/test_tomography_gate_set_tomography.py
```diff
@@ -268,10 +275,17 @@ def test_GST_with_transpiler(backend, star_connectivity):
...
     target_gates = [gates.SX(0), gates.Z(0), gates.CNOT(0, 1)]
-    # gate_set = [g.__class__ for g in target_gates]
+    gate_set = [
+        (
+            (g.__class__, [g.parameters[i] for i in range(len(g.parameters))])
+            if g.parameters
+            else (g.__class__, [])
+        )
+        for g in target_gates
+    ]
```
BrunoLiegiBastonLiegi, line 285: same as above

BrunoLiegiBastonLiegi (commented):
In general I still struggle to find a strong motivation for this, as it would make more sense to me to decompose your parametrized (fixed) gate in actual non parametrized ones, which will compose your gate set for GST then. 
In any case, even though the `(gate, params)` approach is probably not the cleanest, it should be fine for the moment.

src/qibo/tomography/gate_set_tomography.py
```diff
@@ -286,12 +292,25 @@ def GST(
...
-            nqubits = len(gate.qubits)
-            if nqubits not in (1, 2):
+            init_args = signature(gate[0]).parameters
+            params = gate[1]
+
+            angle_names = [name for name in init_args if name in {"theta", "phi"}]
+
+            angle_values = {}
+            for name, value in zip(angle_names, params):  # Zip ensures correct order
+                angle_values[name] = value
```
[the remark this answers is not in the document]
mho291: Thanks Andrea! I realise that if we do
``` python
for angle, param in zip(["theta", "phi"][:len(gate[1])], gate[1])
```
it won't work for gates like `gates.GPI(0, np.pi)` because `GPI` gate doesn't have `theta` parameter. I run into this error message:
```
---> 46 gate = gate(*range(nqubits), **params)

TypeError: GPI.__init__() got an unexpected keyword argument 'theta'
```
Hence my train of thought is that if we search for `theta` or `phi` or both, then we just extract that angle_name out and assign the parameter directly (Lines 295 to 302)
``` python
angle_names = [name for name in init_args if name in {"theta", "phi"}]

angle_values = {}
for name, value in zip(angle_names, params):  # Zip ensures correct order
    angle_values[name] = value
```

mho291:
Thanks @BrunoLiegiBastonLiegi for the comments. 

> In general I still struggle to find a strong motivation for this, as it would make more sense to me to decompose your parametrized (fixed) gate in actual non parametrized ones, which will compose your gate set for GST then.

True, I do see your point of view. However, if I wanted to quickly do GST for parameterised gates without the decomposition, then we'll probably need this small functionality. Correct me if I'm wrong: I have the impression that since `qw11q` device uses `RX` gate as its native gate, then we'll need to pass in a parameter regardless, right?

Anyway, back to the PR, when I was constructing the `gate_set`, I thought that maybe it'd be better to standardise the input. Instead of 
`gate_set = [gates.X, (gates.RX, [np.pi/3]), gates.Z, (gates.PRX, [np.pi/2, np.pi/3]), (gates.GPI, [np.pi/7])]`,
we can leave a blank `[]` for gates that don't have any parameters. So the `gate_set` will look like this:
`gate_set = [(gates.X, []), (gates.RX, [np.pi/3]), (gates.Z, []), (gates.PRX, [np.pi/2, np.pi/3]), (gates.GPI, [np.pi/7])]`

From an end-user perspective, the most convenient way is to dump the entire gate and its parameters, including qubit specifications, into the `gate_set`. However, since we're doing it incrementally, I find it easier to configure the `gate_set` when all gates are within the parentheses `(gates.X, [])` rather than needing to remember which gates to avoid wrapping with parentheses. I'd like to hear your opinion whether my thought process makes sense!

BrunoLiegiBastonLiegi:
> True, I do see your point of view. However, if I wanted to quickly do GST for parameterised gates without the decomposition, then we'll probably need this small functionality. 

Yeah, fair enough, it's surely faster to just allow for passing the parametrized gate here, rather than manually decomposing

> Correct me if I'm wrong: I have the impression that since `qw11q` device uses `RX` gate as its native gate, then we'll need to pass in a parameter regardless, right?

Yes and now, because if I remember correctly our `RX` is implemented as `GPI2(pi/2)` pulses. In any case, the decomposition happens internally anyway when you execute on hardware, the `transpiler` argument is indeed there for that.

> Anyway, back to the PR, when I was constructing the `gate_set`, I thought that maybe it'd be better to standardise the input. Instead of `gate_set = [gates.X, (gates.RX, [np.pi/3]), gates.Z, (gates.PRX, [np.pi/2, np.pi/3]), (gates.GPI, [np.pi/7])]`, we can leave a blank `[]` for gates that don't have any parameters. So the `gate_set` will look like this: `gate_set = [(gates.X, []), (gates.RX, [np.pi/3]), (gates.Z, []), (gates.PRX, [np.pi/2, np.pi/3]), (gates.GPI, [np.pi/7])]`
> From an end-user perspective, the most convenient way is to dump the entire gate and its parameters, including qubit specifications, into the `gate_set`. However, since we're doing it incrementally, I find it easier to configure the `gate_set` when all gates are within the parentheses `(gates.X, [])` rather than needing to remember which gates to avoid wrapping with parentheses. I'd like to hear your opinion whether my thought process makes sense!

I understand your wish of standardizing the inputs, but personally I am not a fan of making the user specify a non parametrized gate as (gate, []) everytime. Say for instance, that I want to run GST on `X`, `Y` and `Z` then I have to call
``` python
GST(gate_set=[(X, []), (Y, []), (Z, [])])
```
not the best in my opinion. We could actually leverage the asimettry, instead, and infer from the presence of a tuple the presence of a parametrized gate.

src/qibo/tomography/gate_set_tomography.py
```diff
@@ -278,7 +292,15 @@ def GST(
 
     for gate in gate_set:
         if gate is not None:
-            init_args = signature(gate).parameters
+            init_args = signature(gate[0]).parameters
+            params = gate[1]
+
+            angle_names = [name for name in init_args if name in {"theta", "phi"}]
```
MatteoRobbiati, line 298:
```suggestion
            angle_names = [name for name in init_args if name in {"theta", "phi", "lam"}]
```

This should cover any case, including cases like `gates.U2` and `gates.U3`.

src/qibo/tomography/gate_set_tomography.py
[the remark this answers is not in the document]
MatteoRobbiati: As suggested by @BrunoLiegiBastonLiegi, we need here to firstly check whether the gate in the list is a tuple, and then access to first or second element of the tuple.

MatteoRobbiati (commented):
Firstly, thanks @mho291 for the work! I sounds in general good to me.
About input parameters, I also think as @BrunoLiegiBastonLiegi a good solution for now would be to pass parametrised gates as tuples and non-parametrised gates as single elements in the `gate_set` list. Another possibility would be to re-discuss how a gate is defined, and if we really need it to depend on a specific qubit number. 
Anyway, this would be a way bigger discussion and I don't think this PR is the right place where to do it. 

I left a couple of comments more and I would reccomend addressing the last comments from @BrunoLiegiBastonLiegi.

Thanks again :)

mho291:
> Firstly, thanks @mho291 for the work! It sounds in general good to me. About input parameters, I also think as @BrunoLiegiBastonLiegi a good solution for now would be to pass parametrised gates as tuples and non-parametrised gates as single elements in the `gate_set` list. Another possibility would be to re-discuss how a gate is defined, and if we really need it to depend on a specific qubit number. Anyway, this would be a way bigger discussion and I don't think this PR is the right place where to do it.
> 
> I left a couple of comments more and I would reccomend addressing the last comments from @BrunoLiegiBastonLiegi.
> 
> Thanks again :)

Thanks @MatteoRobbiati and @BrunoLiegiBastonLiegi . Yes, I have coded it to go with @BrunoLiegiBastonLiegi 's suggestion: If it's a parameterised gate, put it in a tuple. Then inside GST, check if it's a tuple, then access the parameters.

I've done this:
```python
for gate in gate_set:
    if gate is not None:
        if isinstance(gate, tuple):
            init_args = signature(gate[0]).parameters
            params = gate[1]
            gate_class = gate[0]
        else:
            init_args = signature(gate).parameters
            params = []
            gate_class = gate
            
        angle_names = [name for name in init_args if name in {"theta", "phi", "lam"}]
        angle_values = {}
        for name, value in zip(angle_names, params):
            angle_values[name] = value

        if "q" in init_args:
            nqubits = 1
        elif "q0" in init_args and "q1" in init_args and "q2" not in init_args:
            nqubits = 2
        else:
            raise_error(
                RuntimeError,
                f"Gate {gate} is not supported for `GST`, only 1- and 2-qubits gates are supported.",
            )
        gate = gate_class(*range(nqubits), **angle_values)
```
It worked locally but the Pytests seemed to fail with error message
`target_gates = [<qibo.gates.gates.SX object at 0x17748bc50>, <qibo.gates.gates.RX object at 0x17748bd10>, <qibo.gates.gates.PRX object at 0x17748be50>, <qibo.gates.gates.CY object at 0x17748bf50>]`
and traced back to
`/qibo/tomography/gate_set_tomography.py:281: in GST
    init_args = signature(gate).parameters
`
where it was complaining about the tuple input.
`E           TypeError: (<class 'qibo.gates.gates.RX'>, [0.7853981633974483]) is not a callable object`

These TypeErrors occur for `numpy` and `qibojit-numba`:
``` python
=========================================================================================== short test summary info ===========================================================================================
FAILED test_tomography_gate_set_tomography.py::test_GST[numpy-False-target_gates0] - TypeError: (<class 'qibo.gates.gates.RX'>, [0.7853981633974483]) is not a callable object
FAILED test_tomography_gate_set_tomography.py::test_GST[numpy-True-target_gates0] - TypeError: (<class 'qibo.gates.gates.RX'>, [0.7853981633974483]) is not a callable object
FAILED test_tomography_gate_set_tomography.py::test_GST[qibojit-numba-False-target_gates0] - TypeError: (<class 'qibo.gates.gates.RX'>, [0.7853981633974483]) is not a callable object
FAILED test_tomography_gate_set_tomography.py::test_GST[qibojit-numba-True-target_gates0] - TypeError: (<class 'qibo.gates.gates.RX'>, [0.7853981633974483]) is not a callable object
```

Later I checked with [this code](https://github.com/qiboteam/qibo/pull/1534#discussion_r1880450888) which also seems to fail due to the TypeError. 

Oddly, the tests here are passing. Should I be concerned that tests on my local machine are failing? 
I've been staring at this TypeError for a few hours now but can't find a solution. I'd really appreciate some feedback. 

Thank you!

BrunoLiegiBastonLiegi:
Not sure, can you post the complete traceback of your local error? Silly question, you have installed this branch and not the qibo main, right?

doc/source/code-examples/advancedexamples.rst
```diff
@@ -2242,7 +2242,8 @@ Let's first define the set of gates we want to estimate:
 
    from qibo import gates
 
-   gate_set = {gates.X, gates.H, gates.CZ}
+   target_gates = [gates.RX(0, np.pi/3), gates.Z(0), gates.PRX(0, np.pi/2, np.pi/3), gates.GPI(0, np.pi/7), gates.CNOT(0,1)]
+   gate_set = [(g.__class__, [g.parameters[i] for i in range(len(g.parameters))]) if g.parameters else (g.__class__, []) for g in target_gates]
```
BrunoLiegiBastonLiegi, line 2246:
```suggestion
   gate_set = [(gates.RX, [np.pi/3]), gates.Z, (gates.PRX, [np.pi/2, np.pi/3]), (gates.GPI, [np.pi/7]), gates.CNOT]
```

I would directly build this as we intend the user to do it.

src/qibo/tomography/gate_set_tomography.py
```diff
@@ -227,7 +228,12 @@ def GST(
     """Runs Gate Set Tomography on the input ``gate_set``.
 
     Args:
-        gate_set (tuple or set or list): set of :class:`qibo.gates.Gate` to run GST on.
+        gate_set (tuple or set or list): set of :class:`qibo.gates.Gate` and parameters to run
+            GST on.
+            E.g. target_gates = [gates.RX(0, np.pi/3), gates.Z(0), gates.PRX(0, np.pi/2, np.pi/3),
+                                 gates.GPI(0, np.pi/7), gates.CNOT(0,1)]
+                 gate_set = [(g.__class__, [g.parameters[i] for i in range(len(g.parameters))])
+                             if g.parameters else (g.__class__, []) for g in target_gates]
```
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: what about this one?

src/qibo/tomography/gate_set_tomography.py
```diff
@@ -1,13 +1,14 @@
 from functools import cache
 from inspect import signature
 from itertools import product
+from random import Random
```
BrunoLiegiBastonLiegi, line 4: is this the random placer of the transpiler?

src/qibo/tomography/gate_set_tomography.py
```diff
@@ -278,7 +292,22 @@ def GST(
...
+                init_args = signature(gate).parameters
+                params = []
+                gate_class = gate
+
+            angle_names = [
+                name for name in init_args if name in {"theta", "phi", "lam"}
+            ]
+            angle_values = {}
+            for name, value in zip(angle_names, params):
+                angle_values[name] = value
```
BrunoLiegiBastonLiegi, line 309:
```suggestion
            if isinstance(gate, tuple):
                angles = ["theta", "phi", "lam"]
                gate, params = gate
                angle_values = dict(zip(angles[:len(params)], params))
            else:
                angle_values = {}
            init_args = signature(gate).parameters
```

this should do the same I believe

mho291:
> Silly question, you have installed this branch and not the qibo main, right?

Not silly at all! It my oversight that I did `pip install qibo` in a fresh environment instead of the branch. I just installed the branch and there are no errors. Thank you for guiding, I will write it down in my notes and not repeat this again. :)

tests/test_tomography_gate_set_tomography.py
```diff
@@ -204,7 +204,16 @@ def test_gate_tomography_noise_model(backend):
 
 @pytest.mark.parametrize(
     "target_gates",
-    [[gates.SX(0), gates.Z(0), gates.CY(0, 1)], [gates.TOFFOLI(0, 1, 2)]],
+    [
+        # [gates.SX(0), gates.X(0), gates.Z(0), gates.CY(0, 1)],
```
BrunoLiegiBastonLiegi, line 208:
```suggestion
```
remove comment

tests/test_tomography_gate_set_tomography.py
```diff
@@ -215,17 +224,21 @@ def test_GST(backend, target_gates, pauli_liouville):
...
+    if len(target_gates) == 4:
         empty_1q, empty_2q, *approx_gates = GST(
             gate_set=gate_set,
             nshots=int(1e4),
             include_empty=True,
             pauli_liouville=pauli_liouville,
             backend=backend,
         )
-        print(type(empty_1q), type(empty_2q))
+        # print(type(empty_1q), type(empty_2q))
```
BrunoLiegiBastonLiegi, line 241:
```suggestion
```
remove

tests/test_tomography_gate_set_tomography.py
```diff
@@ -265,7 +278,18 @@ def test_GST_with_transpiler(backend, star_connectivity):
...
     target_gates = [gates.SX(0), gates.Z(0), gates.CNOT(0, 1)]
-    gate_set = [g.__class__ for g in target_gates]
+    # gate_set = [
+    #     (
+    #         (g.__class__, [g.parameters[i] for i in range(len(g.parameters))])
+    #         if g.parameters
+    #         else (g.__class__, [])
+    #     )
+    #     for g in target_gates
+    # ]
```
BrunoLiegiBastonLiegi, line 288:
```suggestion
```
remove

BrunoLiegiBastonLiegi (commented):
Thanks @mho291, looks good, just some minor suggestions.

src/qibo/tomography/gate_set_tomography.py
mho291: I think so, I think qibo.transpiler.placer already imports Random so it's double imports. I have removed line 4 and it works.

src/qibo/tomography/gate_set_tomography.py
```diff
@@ -227,7 +227,10 @@ def GST(
     """Runs Gate Set Tomography on the input ``gate_set``.
```
MatteoRobbiati, line 227:
```suggestion
    """Run Gate Set Tomography on the input ``gate_set``.
```

src/qibo/tomography/gate_set_tomography.py
```diff
@@ -288,7 +309,7 @@ def GST(
                     RuntimeError,
                     f"Gate {gate} is not supported for `GST`, only 1- and 2-qubits gates are supported.",
```
MatteoRobbiati, line 310:
```suggestion
                    f"Gate {gate} is not supported for `GST`, only 1- and 2-qubit gates are supported.",
```

MatteoRobbiati (approved):
Thanks! I would ask you to implement two more minor suggestions, but LGTM.

BrunoLiegiBastonLiegi (approved):

PR #1544, opened 2024-12-16
Fix documentation of `models.circuit.Circuit`

renatomello:
Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/models/circuit.py
+++ b/src/qibo/models/circuit.py
@@ -133,6 +133,7 @@ class Circuit:
 
             from qibo import Circuit
 
+            # Every circuit initialization below is valid
             circuit = Circuit(5)  # Default wire names are [0, 1, 2, 3, 4]
             circuit = Circuit(["A", "B", "C", "D", "E"])
             circuit = Circuit(5, wire_names=["A", "B", "C", "D", "E"])
--- a/src/qibo/models/circuit.py
+++ b/src/qibo/models/circuit.py
@@ -129,7 +129,7 @@ class Circuit:
             circuit = Circuit(5)  # Default wire names are [0, 1, 2, 3, 4]
             circuit = Circuit(["A", "B", "C", "D", "E"])
             circuit = Circuit(5, wire_names=["A", "B", "C", "D", "E"])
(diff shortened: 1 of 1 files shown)


src/qibo/models/circuit.py
```diff
@@ -120,9 +120,12 @@ class Circuit:
     A specific backend has to be used for performing calculations.
 
     Circuits can be created with a specific number of qubits and wire names.
+
     Example:
-        .. testcode::
+        .. code-block:: python
```
alecandido, line 125: Just to understand: was there any issue with the `testcode` directive?

renatomello: `testcode` does not display the code in the compiled documentation, leaving the example blank in the doc https://qibo.science/qibo/stable/api-reference/qibo.html#qibo.models.circuit.Circuit

src/qibo/models/circuit.py
```diff
@@ -120,9 +120,12 @@ class Circuit:
...
+
             from qibo import Circuit
+
             c = Circuit(5)  # Default wire names are [0, 1, 2, 3, 4]
             c = Circuit(["A", "B", "C", "D", "E"])
             c = Circuit(5, wire_names=["A", "B", "C", "D", "E"])
```
renatomello, line 131:
```suggestion
            circuit = Circuit(5)  # Default wire names are [0, 1, 2, 3, 4]
            circuit = Circuit(["A", "B", "C", "D", "E"])
            circuit = Circuit(5, wire_names=["A", "B", "C", "D", "E"])
```

src/qibo/models/circuit.py
alecandido: Ok. I'm not sure why it is happening, since in principle even `testcode` blocks should be shown, according to Sphinx
https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#directive-testcode
> By default it is shown as a highlighted code block.

(there is an explicit attribute `:hide:` to avoid showing it).

I wonder why... but a workaround is still better than an empty block. Thanks for the fix!

alecandido (approved):

Edoardo-Pedicillo (approved):

PR #1547, opened 2024-12-17
Add decomposition for `gates.CRY`

renatomello:
Checklist:
- [x] Tests are passing.

### DIFF
--- a/tests/test_gates_gates.py
+++ b/tests/test_gates_gates.py
@@ -729,7 +729,7 @@ def test_cun(backend, name, params):
         matrix = Circuit(2)
         matrix.add(decomposition)
         matrix = matrix.unitary(backend=backend)
-        backend.assert_allclose(matrix, _matrix)
+        backend.assert_allclose(matrix, _matrix, atol=1e-10)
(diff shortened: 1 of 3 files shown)


Edoardo-Pedicillo (approved):

tests/test_gates_gates.py
```diff
@@ -722,6 +725,12 @@ def test_cun(backend, name, params):
 
     backend.assert_allclose(final_state, target_state, atol=1e-6)
 
+    if name == "CRY":
+        matrix = Circuit(2)
+        matrix.add(decomposition)
```
stavros11, line 730: Maybe `decomposition` can be defined inside this if, or even do directly
```suggestion
        matrix.add(gate.decompose())
```
to avoid having another if above (line 701).

stavros11 (approved):

tests/test_gates_gates.py
renatomello: I tried that but gate is being redefined as an ndarray before this line, this is why I added line 701.

Issue #1494, opened 2024-10-16
`SymbolicHamiltonian.terms` setter invalidates the object


BrunoLiegiBastonLiegi:
`SymbolicHamiltonian` provides a setter for `terms` that do not take care of updating the form. 
This means that by doing the following for instance:

``` python
from qibo.symbols import Z, X
from qibo.hamiltonians import SymbolicHamiltonian

h0 = SymbolicHamiltonian(Z(0)*Z(1) + Z(2)*Z(1), nqubits = 3)
h1 = SymbolicHamiltonian(X(0)*X(1) + X(2)*X(1), nqubits = 3)
print(h0.form)
h0.terms = h1.terms
print(h0.form)
```

you make its `form` and `terms` inconsistent.
I would say, in general, that `terms` should not have a setter since it is computed starting from the `form` and I would rather make it a `@cached_property` probably. At the moment the main reason it is there, I believe, is beacuse it makes it easier to compose hamiltonians through the `__add__` operator for instance, without the need to recompute the terms.

PR #1548, opened 2024-12-18
Remove `SymbolicHamiltonian` `terms` and `dense` setters

BrunoLiegiBastonLiegi:
This should address #1494 

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/hamiltonians/hamiltonians.py
+++ b/src/qibo/hamiltonians/hamiltonians.py
@@ -359,6 +359,9 @@ class SymbolicHamiltonian(AbstractHamiltonian):
         return self._form
 
     def _calculate_nqubits_from_form(self, form):
+        """Calculate number of qubits in the system described by the given
+        Hamiltonian formula
+        """
         nqubits = 0
         for symbol in form.free_symbols:
             if isinstance(symbol, self._qiboSymbol):
@@ -383,25 +386,8 @@ class SymbolicHamiltonian(AbstractHamiltonian):
                 TypeError,
                 f"Symbolic Hamiltonian should be a ``sympy`` expression but is {type(form)}.",
             )
-        # Calculate number of qubits in the system described by the given
-        # Hamiltonian formula
-        nqubits = 0
-        for symbol in form.free_symbols:
-            if isinstance(symbol, self._qiboSymbol):
-                q = symbol.target_qubit
-            elif isinstance(symbol, sympy.Expr):
-                if symbol not in self.symbol_map:
-                    raise_error(ValueError, f"Symbol {symbol} is not in symbol map.")
-                q, matrix = self.symbol_map.get(symbol)
-                if not isinstance(matrix, self.backend.tensor_types):
-                    # ignore symbols that do not correspond to quantum operators
-                    # for example parameters in the MaxCut Hamiltonian
-                    q = 0
-            if q > nqubits:
-                nqubits = q
-
         self._form = form
-        self.nqubits = nqubits + 1
+        self.nqubits = self._calculate_nqubits_from_form(form)
 
     @cached_property
     def terms(self):
@@ -409,22 +395,21 @@ class SymbolicHamiltonian(AbstractHamiltonian):
 
         Terms will be objects of type :class:`qibo.core.terms.HamiltonianTerm`.
         """
-        if self._terms is None:
-            # Calculate terms based on ``self.form``
-            from qibo.hamiltonians.terms import (  # pylint: disable=import-outside-toplevel
-                SymbolicTerm,
-            )
+        # Calculate terms based on ``self.form``
+        from qibo.hamiltonians.terms import (  # pylint: disable=import-outside-toplevel
+            SymbolicTerm,
+        )
 
-            form = sympy.expand(self.form)
-            terms = []
-            for f, c in form.as_coefficients_dict().items():
-                term = SymbolicTerm(c, f, self.symbol_map)
-                if term.target_qubits:
-                    terms.append(term)
-                else:
-                    self.constant += term.coefficient
-            self._terms = terms
-        return self._terms
+        self.constant = 0.0
+
+        form = sympy.expand(self.form)
+        terms = []
+        for f, c in form.as_coefficients_dict().items():
+            term = SymbolicTerm(c, f, self.symbol_map)
+            if term.target_qubits:
+                terms.append(term)
+            else:
+                self.constant += term.coefficient
 
     @property
     def matrix(self):
@@ -548,10 +533,10 @@ class SymbolicHamiltonian(AbstractHamiltonian):
         return Hamiltonian(self.nqubits, matrix, backend=self.backend) + self.constant
 
     def calculate_dense(self):
-        if self._terms is None:
-            # calculate dense matrix directly using the form to avoid the
-            # costly ``sympy.expand`` call
-            return self._calculate_dense_from_form()
+        # if self._terms is None:
+        # calculate dense matrix directly using the form to avoid the
+        # costly ``sympy.expand`` call
+        #    return self._calculate_dense_from_form()
         return self._calculate_dense_from_terms()
 
     def expectation(self, state, normalize=False):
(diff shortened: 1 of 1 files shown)


BrunoLiegiBastonLiegi:
I am still unsure whether it is worth keeping the two different ways for computing the dense form of the `SymbolicHamiltonian`, namely `from_terms` and `from_form`.
Therefore I've done a little benchmark on 4 different hamiltonians `("TFIM", "XXZ", "TFIM**2", "XXZ**2")` with the numpy backend for qubits from 2 to 11. In particular the number of terms (after expansion) for a given number of qubits seems to be equal for all the 4 hamiltonians (weirdly, I was expecting the squared ones to have more terms):
``` python
{
    "2q": 36,
    "3q": 81,
    "4q": 144,
    "5q": 225,
    "6q": 324,
    "7q": 441,
    "8q": 576,
    "9q": 729,
    "10q": 900,
    "11q": 1089
}
```
the average runtimes across 10 runs are as follows:

image

BrunoLiegiBastonLiegi:
I marked the PR as ready for review, but keep in mind that this
> Note: tests on cupy are still failing due to this https://github.com/qiboteam/qibojit/issues/196 .

is still up. I am testing right now whether there I can fix them in qibojit directly (qiboteam/qibojit#199), otherwise I will have to distribute some casts here.

@stavros11 @renatomello may I ask you guys to have a look a this if you have time, I presume you two were the main contributors to the hamiltonian module. There are still some parts of the code that I left in commented because they are up for discussion and I would like to conduct some more tests on a GPU.

stavros11:
Thanks @BrunoLiegiBastonLiegi for the refactor. This part of is admittedly quite messy. I did not have a detailed look in the code (will do soon), just some quick feedback on the first comment.

> EDIT 2: I went ahead and provided both `Symbols` and `SymbolicTerm` with a `backend` attribute which is useful whenever you move to the dense form representation. Initially, I thought about making `backend` an additional input argument to the functions that needed it (similarly to what we do with gates), but this would have meant breaking the api (e.g. `symbol.matrix` -> `symbol.matrix(backend)`). 

Personally, I don't see this as a big issue, particularly because the old `gate.matrix` was also converted to `gate.matrix(backend)` at some point. I guess you could also have a default value so that `symbol.matrix()` is sufficient. It would still be breaking and of course up to discussion, maybe to be postponed to a major version release (if we follow that). In any case, if not done here, I would open an issue to do it in the next major release,

> What it means now though, is that there are cases where the backends of the different pieces may mismatch, thus causing problems. To mitigate this, I decided to forcely set the backend of the symbols here https://github.com/qiboteam/qibo/blob/4d8c9c1211c850a5e740607de943da75fc31f8b2/src/qibo/hamiltonians/terms.py#L176C1-L179C50 which allows for the end user to not care about which backend to construct a symbol with, maintaining in practice the usual API. 

particularly to avoid the need for such workarounds.

> As part of the refactor, I removed the `symbol_map` argument of the hamiltonians, 

I don't fully remember what's that for, but based on the docstring it was planned to be removed anyway, so it should be fine.

> the `from_symbolic` method and the `TrotterHamiltonian`.

In principle, these are also kind of "breaking" so we could even just do a breaking release with this PR.

Edoardo-Pedicillo (commented):
The refactoring is really nice. Thanks @BrunoLiegiBastonLiegi !

Edoardo-Pedicillo (approved):

src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -472,42 +424,34 @@ def _get_symbol_matrix(self, term):
             # note that we need to use matrix multiplication even though
             # we use scalar symbols for convenience
             factors = term.as_ordered_factors()
-            result = self._get_symbol_matrix(factors[0])
-            for subterm in factors[1:]:
-                result = result @ self._get_symbol_matrix(subterm)
+            result = reduce(
+                self.backend.np.matmul,
+                [self._get_symbol_matrix(subterm) for subterm in factors],
+            )
```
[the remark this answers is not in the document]
alecandido: Just because it happened to me to see this comment
```suggestion
            result = reduce(
                self.backend.np.matmul,
                (self._get_symbol_matrix(subterm) for subterm in factors),
            )
```

This is an incredibly marginal comment, but Python has comprehension for both structures (lists `[x for x in ...]` and set/dictionaries `{x/x: y for x/x, y in ...}`) and generators (with `(x for x in ...)`).
The advantage of generators is that they are iterators, and not structures. So, the iteration itself would happen lazily, and you would not need the intermediate structure. By contrast, if you use list comprehension it will
1. consume the whole iteration ahead of time and
2. collect all the elements in an intermediate list

However, sometimes you really need a list, because alternatives are not feasible, or exceedingly cumbersome (e.g. when you need multiple iterations, that you could achieve with `itertools.tee` - but it would not be thread safe, and potentially consume a lot of resources as well).

Now, we do not care much/at all about performances at this level (essentially, whenever we care we do not use Python and its iterations, including `reduce()`). But better being aware :)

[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: A side comment about this comment xD: I was expecting as well to get better performance by doing the product of all the matrices with a single einsum instead of a ``reduce/loop``, however, I found this to not always be true, on CPU at least. It may be that with GPUs the einsum might become preferable but I have to check in practice.

BrunoLiegiBastonLiegi:
> There are still some parts of the code that I left in commented because they are up for discussion and I would like to conduct some more tests on a GPU.

Just a small update on the GPU scaling, I tested the ``multikron`` function with the two approaches ``reduce(kron)`` and pure ``einsum``:
``` python
def multikron_reduce(matrix_list, kron):
    return reduce(kron, matrix_list)
    
def multikron_einsum(matrix_list, einsum):
    indices = list(range(2 * len(matrix_list)))
    even, odd = indices[::2], indices[1::2]
    lhs = zip(even, odd)
    rhs = even + odd
    einsum_args = [item for pair in zip(matrix_list, lhs) for item in pair]
    N = 2 ** len(matrix_list)
    h = einsum(*einsum_args, rhs)
    return np.sum(np.reshape(h, (-1, N, N)), axis=0)
```
the results with numpy and cupy with 10 2x2 matrices are:
```
## numpy:
reduce --> 0.8061706170010439
einsum --> 4.193189857000107
## cupy:
reduce --> 0.7108887559988943
einsum --> 0.056594685000163736
```
the GPU scales wonderfully as expected and I suspect this would transfer to the other similar functions that I encountered here. However, the drawback for the CPU seems too big to consider einsum a viable option, and differentiating among the backends would mean to move these functions to the backend directly. Thus I am not sure what to do.

alecandido:
Just as an attempt, could you try to use
```py
einsum(..., optimize=True)
```
or even
```py
einsum(..., optimize="optimal")
```
?

BrunoLiegiBastonLiegi:
I tested ``optimize=True`` with no difference already, I am testing now ``optimize="optimal"``, which I was not aware of btw, no difference for numpy as well, cupy instead gets misteriously utterly slow, in the sense that it is still running while I am typing.

alecandido:
Well, if NumPy is getting no benefit, than it's not worth in any case.

However, this would be interesting to investigate: `"optimal"` attempts (at least to a certain extent) to find the best contraction, possibly wasting a lot of time in computing it.
Now, I expect this not to be that hard for the problem you're laying out. And after you know the contraction path, it's just a matter of executing it.
So, I'm now wondering: how is it possible that `reduce()` ends up in a better contraction than the optimal one?

Just to exclude that the contraction's computation is actually taking a huge amount of time, I'd propose to compute the contraction executed by, and pass it explicitly to `optimize=...` (which also accepts a contraction path as argument).
> Controls if intermediate optimization should occur. No optimization will occur if `False` and `True` will default to the ‘greedy’ algorithm. Also accepts an explicit contraction list from the np.einsum_path function. See `np.einsum_path` for more details. Defaults to `False`.
>
> https://numpy.org/doc/stable/reference/generated/numpy.einsum.html

BrunoLiegiBastonLiegi:
I tried calculating the path separately, but no difference at all, it looks like the contraction is taking that time and that's it.

``` python
def get_path(matrix_list):
    indices = list(range(2 * len(matrix_list)))
    even, odd = indices[::2], indices[1::2]
    lhs = zip(even, odd)
    rhs = even + odd
    einsum_args = [item for pair in zip(matrix_list, lhs) for item in pair]
    return np.einsum_path(*einsum_args, rhs, optimize="optimal")
    
def multikron_einsum(matrix_list, einsum, path):
    indices = list(range(2 * len(matrix_list)))
    even, odd = indices[::2], indices[1::2]
    lhs = zip(even, odd)
    rhs = even + odd
    einsum_args = [item for pair in zip(matrix_list, lhs) for item in pair]
    N = 2 ** len(matrix_list)
    h = einsum(*einsum_args, rhs, optimize=path[0])
    return np.sum(np.reshape(h, (-1, N, N)), axis=0)
```

tests/test_hamiltonians_from_symbols.py
```diff
@@ -19,63 +19,38 @@ def test_symbols_pickling(symbol):
...
     final_matrix = ham.matrix
     target_matrix = hamiltonians.TFIM(nqubits, h=h, backend=backend).matrix
     backend.assert_allclose(final_matrix, target_matrix)
 
 
-@pytest.mark.parametrize("hamtype", ["normal", "symbolic"])
-@pytest.mark.parametrize("calcterms", [False, True])
-def test_from_symbolic_with_power(backend, hamtype, calcterms):
+@pytest.mark.parametrize("hamtype", ["symbolic"])
+def test_from_symbolic_with_power(backend, hamtype):
```
stavros11, line 42: Do we still need `hamtype`? I think it is not used anywhere.

tests/test_hamiltonians_from_symbols.py
```diff
@@ -88,160 +63,100 @@ def test_from_symbolic_with_power(backend, hamtype, calcterms):
     backend.assert_allclose(final_matrix, target_matrix)
 
 
-@pytest.mark.parametrize("hamtype", ["normal", "symbolic"])
-@pytest.mark.parametrize("calcterms", [False, True])
-def test_from_symbolic_with_complex_numbers(backend, hamtype, calcterms):
+@pytest.mark.parametrize("hamtype", ["symbolic"])
+def test_from_symbolic_with_complex_numbers(backend, hamtype):
```
stavros11, line 67: Same here.

src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -860,19 +724,3 @@ def circuit(self, dt, accelerators=None):
         )
 
         return circuit
-
-
-class TrotterHamiltonian:
```
stavros11, line 865: Note that there are a few remaining references to this object, for example here https://github.com/qiboteam/qibo/blob/312efb9d6934e85cfdaafcdb7027cdfe46bf87c4/src/qibo/solvers.py#L47 and some other places in the docs (easily found using search).

src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -663,120 +621,40 @@ def expectation_from_samples(self, freq: dict, qubit_map: list = None) -> float:
...
-            )
-            if self._form is not None:
-                new_ham.form = self.form - o
-            if self._terms is not None:
-                new_ham.terms = self.terms
-                new_ham.constant = self.constant - o
-            if self._dense is not None:
-                new_ham.dense = self.dense - o
+    def __add__(self, o):
+        return self._compose(o, lambda x, y: x + y)
```
stavros11, line 648: Here (and in the other similar function) you can also use standard library's [`operator`](https://docs.python.org/3/library/operator.html)
```suggestion
        import operator
        
        return self._compose(o, operator.add)
```

pyproject.toml
```diff
@@ -56,7 +56,7 @@ ipython = "^8.10.0"
 qulacs = { version = "^0.6.4", markers = "(sys_platform == 'darwin' and python_version > '3.9') or sys_platform != 'darwin'" }
 seaborn = "^0.13.2"
 ipykernel = "^6.29.4"
-qibojit = { git = "https://github.com/qiboteam/qibojit.git"}
+qibojit = { git = "https://github.com/qiboteam/qibojit.git", branch = "cupy_identity_matrix"}
```
stavros11, line 59: The qibojit PR is approved, so we can already merge and revert this to main branch.

src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -525,32 +469,46 @@ def _calculate_dense_from_form(self) -> Hamiltonian:
         matrix = self._get_symbol_matrix(self.form)
         return Hamiltonian(self.nqubits, matrix, backend=self.backend)
 
+    """
     def _calculate_dense_from_terms(self) -> Hamiltonian:
```
stavros11, line 473: If I understand correctly, this is unlikely to be relevant anymore, so we could directly drop.

src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -376,53 +349,31 @@ def form(self, form):
...
         """
-        if self._terms is None:
-            # Calculate terms based on ``self.form``
-            from qibo.hamiltonians.terms import (  # pylint: disable=import-outside-toplevel
-                SymbolicTerm,
-            )
+        # Calculate terms based on ``self.form``
+        from qibo.hamiltonians.terms import (  # pylint: disable=import-outside-toplevel
+            SymbolicTerm,
+        )
```
stavros11, line 364: I think we could try to move this top level as well. `terms` does not seem to import anything from `hamiltonians`, so if there are circular import issues we may be able to resolve them relatively easily.

src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -206,7 +180,9 @@ def __add__(self, o):
                     "Only hamiltonians with the same number of qubits can be added.",
                 )
             new_matrix = self.matrix + o.matrix
-        elif isinstance(o, self.backend.numeric_types):
+        elif isinstance(o, self.backend.numeric_types) or isinstance(
```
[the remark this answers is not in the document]
stavros11: `other` is a potentially more readable name that could work here (to be fair `o` was not introduced in that PR, but we could still take it as an opportunity to fix).

pyproject.toml
BrunoLiegiBastonLiegi: It depends on whether we wish to merge that one, or implement what @andrea-pasquale proposed there in this PR. Both solutions would work for me.

andrea-pasquale: I think we can merge the PR as it is @BrunoLiegiBastonLiegi.

BrunoLiegiBastonLiegi:
Thanks to all of you for the comments, I should have addressed everything already. 
The are just two last things that I would like to discuss with you:
- What to do with the whole ``einsum`` vs ``reduce`` argument we discussed above together with @alecandido. At the moment, it looks like it would be better to just stick with ``reduce``. Another option would be to make the ``multikron`` and similar functions part of the backends, if it's worth it.
- In the ``hamiltonian.models`` module most of the hamiltonians implement a ``dense`` construction separately from the ``symbolic`` one:
https://github.com/qiboteam/qibo/blob/617788e0c0cd5f8997fa7fbb5b382bf87442fa9f/src/qibo/hamiltonians/models.py#L84C1-L93C58
and I wonder whether this is still useful when one could just do ``SymbolicHamiltonian.dense()``

alecandido:
> - What to do with the whole `einsum` vs `reduce` argument we discussed above together with @alecandido. At the moment, it looks like it would be better to just stick with `reduce`. Another option would be to make the `multikron` and similar functions part of the backends, if it's worth it.

I would go for the plain `reduce`. It is good as anything else in performances, and the simplest.
Even making part of the backends would gain at most marginal performance for some backends (for the reason mentioned above). So, I'd not worry that much :)

stavros11 (approved):
Thanks @BrunoLiegiBastonLiegi.

> * What to do with the whole `einsum` vs `reduce` argument we discussed above together with @alecandido. At the moment, it looks like it would be better to just stick with `reduce`. Another option would be to make the `multikron` and similar functions part of the backends, if it's worth it.

From my side, I don't think I ever tried to benchmark these methods, so I do not have any feedback in terms of performance. I would also go with reduce, as it looks simpler in terms of code. If there is any chance that a different implementation would provide a significant advantage in performance we could do it in a different PR after verifying such advantage (you could even open an issue if you think that is worth exploring).

> * In the `hamiltonian.models` module most of the hamiltonians implement a `dense` construction separately from the `symbolic` one:
>   https://github.com/qiboteam/qibo/blob/617788e0c0cd5f8997fa7fbb5b382bf87442fa9f/src/qibo/hamiltonians/models.py#L84C1-L93C58
>   and I wonder whether this is still useful when one could just do `SymbolicHamiltonian.dense()`

I am also not a big fan of providing multiple, equally simple, ways to do the same thing, so I would agree with you that it is not very useful. However, one thing to point out is that removing the `dense` option would be breaking for anyone using that interface and worse, even for people not using it, as it would change the current default (`dense=True`). It could be fine, as there are a few other (potentially less) breaking changes implemented here and I am not sure if we are following a strict version convention for such changes in qibo.

> Run on QPU `sim` completed! :atom:

This is an interesting QPU partition...

BrunoLiegiBastonLiegi:
Thanks for the feedback, I can then maybe open two issues:
- one where I report the different implementations of reduce vs einsum to be tested and, in case, introduced the fastest in a future PR
- one reminding us about the option of dropping one of the two hamiltonian implementations

After that, I think this can be safely merge.

> This is an interesting QPU partition...

that's the best ever, it works perfectly everytime as long as you don't ask for more than 30 qubits

BrunoLiegiBastonLiegi:
Ok, after the tests passed this can be merged.

PR #1551, opened 2024-12-23
Add `models.encodings.binary_encoder` for real-valued data

renatomello:
Will generalize to complex-valued data in a separate PR soon.

Checklist:
- [x] Tests are passing.


src/qibo/models/encodings.py
```diff
@@ -116,8 +116,69 @@ def phase_encoder(data, rotation: str = "RY", **kwargs):
     return circuit
 
 
+def binary_encoder(data, **kwargs):
```
MatteoRobbiati, line 119: Could you share the reference/the idea you followed here to implement the binary encoding on this way?
Is that an alternative way to implement the "standard" binary encoding by means of controlled rotations?

src/qibo/models/encodings.py
```diff
@@ -116,8 +116,69 @@ def phase_encoder(data, rotation: str = "RY", **kwargs):
...
+    if not nqubits.is_integer():
+        raise_error(ValueError, "`data` size must be a power of 2.")
+    nqubits = int(nqubits)
+
+    base_strings = [f"{elem:0{nqubits}b}" for elem in range(dims)]
+    base_strings = np.reshape(base_strings, (-1, 2))
+    strings = [base_strings]
+    for _ in range(nqubits - 1):
+        # new_row = [row[0] for row in base_strings]
+        # base_strings = np.reshape(new_row, (-1, 2))
```
MatteoRobbiati, line 144:
```suggestion
```

In case it is not needed.

src/qibo/models/encodings.py
renatomello: I'm not sure what you mean by "standard", since there are many ways to do binary encoding. This specific one is something that @GoGoKo699 and I have been working on, but it is not published yet. The only difference here is that the amplitudes are parameterized in a specific coordinate system in a $d$-dimensional hypersphere (in this case, called Hopf coordinates).

PR #1559, opened 2025-01-22
Add nontrivial entangling layers to `models.encodings.entangling_layer`

renatomello:
[Reference](https://arxiv.org/abs/2209.08167).

Due to time, I will probably leave the addition of images exemplifying the new layers for a separate PR.

Checklist:
- [x] Tests are passing.


renatomello:
@MatteoRobbiati here I'm officially implementing the Hamming-weight-preserving layers that I once shared with you for the double-bracket project.

src/qibo/models/encodings.py
```diff
@@ -657,11 +683,117 @@ def _parametrized_two_qubit_gate(gate, q0, q1, params=None):
     return gate(q0, q1)
 
 
+def _non_trivial_layers(
```
BrunoLiegiBastonLiegi, line 686: I think it would be more readable if the different non trivial layers had their own separate functions. Hence, something like this:
``` python
if architecture == "next_nearest":
    _next_nearest(...)
elif ...:
   ...
.
.
.

```

doc/source/api-reference/qibo.rst
```diff
@@ -377,9 +377,9 @@ any of the two-qubit gates implemented in ``qibo`` can be selected to customize
 If the chosen gate is parametrized, all phases are set to :math:`0.0`.
 Note that these phases can be updated a posterior by using
 :meth:`qibo.models.Circuit.set_parameters`.
-There are four possible choices of layer ``architecture``:
-``diagonal``, ``shifted``, ``even-layer``, and ``odd-layer``.
-For instance, we show below an example of each architecture for ``nqubits = 6``.
+The possible choices of layer ``architecture`` are the following, in alphabetical order:
```
MatteoRobbiati, line 380: I would probably add an example for each of layer type (not urgent, but could be useful).

src/qibo/models/encodings.py
```diff
@@ -427,20 +427,23 @@ def entangling_layer(
...
+            ``"v"``, and ``"x"``. The ``"x"`` architecture is only defined for an even number
+            of qubits. Defaults to ``"diagonal"``.
         entangling_gate (str or :class:`qibo.gates.Gate`, optional): Two-qubit gate to be used
             in the entangling layer. If ``entangling_gate`` is a parametrized gate,
             all phases are initialized as :math:`0.0`. Defaults to  ``"CNOT"``.
-        closed_boundary (bool, optional): If ``True`` adds a closed-boundary condition
-            to the entangling layer. Defaults to ``False``.
+        closed_boundary (bool, optional): If ``True`` and ``architecture not in
+            ["pyramid", "v", "x"]``, adds a closed-boundary condition to the entangling layer.
+            Defaults to ``False``.
```
MatteoRobbiati, line 446: Why is the default to `False`?  (curiosity)

src/qibo/models/encodings.py
```diff
@@ -488,46 +505,51 @@ def entangling_layer(
...
+    if architecture in ["next_nearest", "pyramid", "v", "x"]:
+        circuit = _non_trivial_layers(
+            nqubits,
+            architecture=architecture,
+            entangling_gate=entangling_gate,
+            closed_boundary=closed_boundary,
+            **kwargs,
         )
+    else:
+        # Finds the number of correct number of parameters to initialize the gate class.
```
MatteoRobbiati, line 517:
```suggestion
        # Finds the correct number of parameters to initialize the gate class.
```

src/qibo/models/encodings.py
```diff
@@ -657,11 +679,140 @@ def _parametrized_two_qubit_gate(gate, q0, q1, params=None):
...
+
+
+def _non_trivial_layers(
+    nqubits: int,
+    architecture: str = "pyramid",
+    entangling_gate: Union[str, gates.Gate] = "RBS",
+    closed_boundary: bool = False,
+    **kwargs,
+):
+    """Create
```
MatteoRobbiati, line 777: I think here we miss part of the docstrings.

src/qibo/models/encodings.py
MatteoRobbiati: In terms of readability I agree with @BrunoLiegiBastonLiegi. It is also true those function are not supposed to be exposed, and anyway I would implement the common part in a separate function.

src/qibo/models/encodings.py
```diff
@@ -716,6 +868,22 @@ def _get_next_bistring(bitstring, markers, hamming_weight):
 
 
 def _ehrlich_algorithm(initial_string, return_indices: bool = True):
+    """Return list of bitstrings with mininal Hamming distance between consecutive strings.
```
MatteoRobbiati, line 871: I think also this docstring should be completed (?)

MatteoRobbiati (commented):
Thanks @renatomello for this! Some comments follow.

src/qibo/models/encodings.py
```diff
@@ -528,28 +528,24 @@ def entangling_layer(
...
+            qubits = range(0, nqubits - 1, 2)
         elif architecture == "odd_layer":
-            circuit.add(
-                _parametrized_two_qubit_gate(gate, qubit, qubit + 1, parameters)
-                for qubit in range(1, nqubits - 1, 2)
-            )
+            qubits = range(1, nqubits - 1, 2)
         else:
+            qubits = range(0, nqubits - 1, 2)
+            qubits_2 = range(1, nqubits - 1, 2)
```
BrunoLiegiBastonLiegi, line 538:
```suggestion
            qubits = tuple(range(0, nqubits - 1, 2)) + tuple(range(1, nqubits - 1, 2))
```
this should lift the need of the following ``architecture == "shifted"`` check I believe

doc/source/api-reference/qibo.rst
renatomello: I mentioned above that would include images with examples in a separate PR due to lack of time right now.

src/qibo/models/encodings.py
renatomello: No particular reason

src/qibo/models/encodings.py
renatomello: Itś already resolved

renatomello: > It's already resolved

src/qibo/models/encodings.py
```diff
@@ -716,6 +874,22 @@ def _get_next_bistring(bitstring, markers, hamming_weight):
...
+
+    Based on the Gray code called Ehrlich algorithm.
+
+    Args:
+        initial_string (_type_): _description_
+        return_indices (bool, optional): _description_. Defaults to True.
+
+    Returns:
+        _type_: _description_
+
```
MatteoRobbiati, line 887: This is what I meant her with uncomplete @renatomello! I would address these placeholders before merging! :)

MatteoRobbiati (approved):
Thanks!

PR #1560, opened 2025-01-22
Add `hamming_weight_encoder` to `models.encodings`

renatomello:
[Reference](https://arxiv.org/abs/2405.20408)

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/models/encodings.py
+++ b/src/qibo/models/encodings.py
@@ -390,7 +390,6 @@ def hamming_weight_encoder(
         )
 
     if optimize_controls:
-        controls_to_remove = list(range(last_qubit, last_qubit - weight, -1))
         indices = [
             int(binom(nqubits - j, weight - j)) - 1 for j in range(weight - 1, 0, -1)
         ]
@@ -680,7 +679,7 @@ def _get_markers(bitstring, last_run: bool = False):
 
 
 def _get_next_bistring(bitstring, markers, hamming_weight):
-    if len(markers) == 0:
+    if len(markers) == 0:  # pragma: no cover
         return bitstring
 
     new_bitstring = np.copy(bitstring)
@@ -718,7 +717,7 @@ def _get_next_bistring(bitstring, markers, hamming_weight):
 
 def _ehrlich_algorithm(initial_string, return_indices: bool = True):
     k = np.unique(initial_string, return_counts=True)
-    if len(k[1]) == 1:
+    if len(k[1]) == 1:  # pragma: no cover
         return ["".join([str(item) for item in initial_string])]
 
     k = k[1][1]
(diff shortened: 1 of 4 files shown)


Edoardo-Pedicillo (approved):
Looks good to me

src/qibo/models/encodings.py
```diff
@@ -323,6 +324,106 @@ def unary_encoder_random_gaussian(
...
+    ):
+        targets = list(last_qubit - np.asarray(targets))
+        controls = list(last_qubit - np.asarray(controls))
+        controls.sort()
+
+        if optimize_controls:
+            comparison = k < np.asarray(indices)
+            controls_to_remove = list(np.argwhere(comparison == True).flatten())
+            for control_index in controls_to_remove[::-1]:
+                controls.pop(control_index)
```
BrunoLiegiBastonLiegi, line 409: I am not completely sure about this, but can't you simplify this by just taking
```suggestion
            controls = controls[k > np.asarray(indices)]
```
or something along these lines?

src/qibo/models/encodings.py
```diff
@@ -558,3 +659,166 @@ def _parametrized_two_qubit_gate(gate, q0, q1, params=None):
...
+
+
+def _get_markers(bitstring, last_run: bool = False):
+    nqubits = len(bitstring)
+    markers = [len(bitstring) - 1]
+    for ind, value in zip(range(nqubits - 2, -1, -1), bitstring[::-1][1:]):
+        if value == bitstring[-1]:
+            markers.append(ind)
+        else:
+            break
```
BrunoLiegiBastonLiegi, line 675: Also here not 100% sure is equivalent, but if I got it right you are checking whether there is another element (except the last one bitstring[-1]) in bitstring with the same value as bitstring[-1], right?
```suggestion
    ind = bitstring[::-1][1:].index(bitstring[-1])
    markers.append(
        list(range(nqubits - 2, -1, -1))[ind]
    )
```

tests/test_models_encodings.py
```diff
@@ -213,6 +217,57 @@ def test_unary_encoder_random_gaussian(backend, nqubits, seed):
...
+        circuit.queue = [
+            gates.X(nqubits - 1 - qubit) for qubit in range(weight)
+        ] + circuit.queue
+    state = backend.execute_circuit(circuit).state()
+
+    backend.assert_allclose(state, target, atol=1e-7)
+
+    # edge case for coverage
+    _ehrlich_algorithm(np.array([0] * nqubits))
+    _get_next_bistring(np.array([0] * nqubits), [], 2)
```
BrunoLiegiBastonLiegi, line 268: Do these need to be tested as well?

BrunoLiegiBastonLiegi (approved):

PR #1558, opened 2025-01-21
Cupy matrices

BrunoLiegiBastonLiegi:
This adapts `qibo` for qiboteam/qibojit#199, which is addressing the `cupy` failures related to #1548.

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.

PR #1561, opened 2025-01-22
Selfhosted GPU workflow and GPU Tests fixes

BrunoLiegiBastonLiegi:
<del>Playing around with the selfhosted workflow.</del>

Edit: now that the workflow is working, this PR will also take care of fixing the GPU tests that are currently failing in main. @andrea-pasquale I will leave the ``cuquantum`` tests to you, I will take care of ``cupy``. This PR is coupled with qiboteam/qibojit#200 for the eventual fixes needed there.

Checklist:
- [ ] Tests are passing.


scarrazza:
@alecandido @BrunoLiegiBastonLiegi I have reinstalled the runner, the ssl problem is gone.

BrunoLiegiBastonLiegi:
Thanks indeed, now the tests are just failing. There is a particular failure related to `numba`:
``` python
RuntimeError: Cannot set NUMBA_NUM_THREADS to a different value once the threads have been launched (currently have 1, trying to set 2)
```
which I encountered locally sometimes as well. 
I have the feeling that this may be caused by this part of the `NumbaBackend.__init__`

https://github.com/qiboteam/qibojit/blob/b7828052c4cc847d44af6159b01508ceb8f6c17b/src/qibojit/backends/cpu.py#L67C1-L70C67

which is called even when the `CupyBackend` and `CuQuantumBackend` are initialized.

alecandido:
@BrunoLiegiBastonLiegi [`NUMBA_NUM_THREADS`](https://numba.pydata.org/numba-doc/dev/reference/envvars.html#envvar-NUMBA_NUM_THREADS) is only the maximum amount of threads allowed. So, setting it, you're explicitly limiting them in Numba, that leads to the generated error.

I believe the problem is originated by the following:
```py
     self.set_threads(len(psutil.Process().cpu_affinity()))
```
but that's a guess.
Specifically, [`.cpu_affinity()`](https://psutil.readthedocs.io/en/latest/index.html#psutil.Process.cpu_affinity), which somehow corresponds to more than one element (which means that the process is actually capable of running on multiple processors...), while the number of threads is being limited by `NUMBA_NUM_THREADS`.

So, the constructor is attempting to maximize the number of threads, and the environment variable is limiting it at the same time...

BrunoLiegiBastonLiegi:
Ok I was able to reproduce this locally on a fresh new environment, I am currently investigating where the max number of threads might be changed and why not always.

BrunoLiegiBastonLiegi:
One weird thing that I noticed, for instance, concerns the order you execute the tests with. Namely, I was expecting that by executing the tests isolated to a single module, say for example `test_backends_clifford.py`, I would not have seen the same error, as the failure is most likely due to some global variable being overwritten, but the weird thing is that this isolated execution seems to "cure" the specific tests. Take for instance three groups of tests that we see failing here: `test_backends_clifford.py`, `test_callbacks.py` and `test_cirq.py`, if I run globally:
``` python
pytests tests/
```
all three of them fail, but if I run just one of them, say `test_cirq.py`:
``` python
pytest test_cirq.py
```
 it does not fail and it won't fail anymore when I run globally ``pytest tests/``

alecandido:
If `pytest tests/` is not behaving twice in the same way, despite being invoked in exactly the same manner, then it means that is sharing some kind of persisting state. We just have to find where (most likely, an environment variable).

BrunoLiegiBastonLiegi:
Ok, finally the tests are failing but not due to numba, just because they are.

src/qibo/backends/clifford.py
```diff
@@ -48,6 +48,10 @@ def __init__(self, engine=None):
         if engine == "numpy":
             pass
         elif engine == "numba":
+            from numba import set_num_threads
+
+            set_num_threads(1)
```
alecandido, line 53: In the end, what is this needed for?

BrunoLiegiBastonLiegi: I realized that it is probably a better approach to set ``parallel=True`` in the operations 

https://github.com/qiboteam/qibojit/blob/dd5bfc0280c2efc4fb9134dfecf1ff29bfc3406e/src/qibojit/backends/clifford_operations_cpu.py#L6

and then restrict the number of threads here, instead of having ``parallel=False`` at all times. The effect should be the same from what I understood. This would also allow for dynamically switching between the parallel and single threaded execution of the custom operators I believe.

alecandido: What I may be missing is why do you actually need either one thing or the other. Can't you have `parallel=True` and more than a single a thread?

In case you can allow more than a single thread, I'm not sure why you're setting it to one during the backend creation...

BrunoLiegiBastonLiegi: Simply because in my testing back at the time single threaded used to be faster. But I actually want to come back at this and double check at some point...

tests/test_backends_qibotn.py
```diff
@@ -1,9 +1,9 @@
-import os
+from numba import set_num_threads
 
 from qibo import get_backend, get_threads, set_backend
 
 # Force quimb to use qibojit default number of threads.
-os.environ["NUMBA_NUM_THREADS"] = f"{get_threads()}"
+set_num_threads(get_threads())
```
alecandido, line 6: About this, it was kind of a workaround for Quimb, it should be more or less discussed somewhere (PR/issue in here or Qibotn).

In any case, scoping it in a fixture, and possibly resetting/unsetting it (if allowed) may be generally a better solution, not to leak state on the other tests.

BrunoLiegiBastonLiegi: Ah yeah I can just back up the original number of threads and restore it after the ``qibotn`` test.

BrunoLiegiBastonLiegi:
While addressing the problems in main, I realized that it may be better to rebase this to #1558. The reason being that making patches to the tests (with ``cupy`` matrices being ``np.arrays``) that will most likely be reversed when we move to ``cupy`` arrays, is a waste of time. What do you think @andrea-pasquale @alecandido ?

andrea-pasquale:
> While addressing the problems in main, I realized that it may be better to rebase this to #1558. The reason being that making patches to the tests (with `cupy` matrices being `np.arrays`) that will most likely be reversed when we move to `cupy` arrays, is a waste of time. What do you think @andrea-pasquale @alecandido ?

I agree. Feel free to go ahead with the rebase.

BrunoLiegiBastonLiegi:
Ok done, thanks. Now, before merging, I would address this [comment](https://github.com/qiboteam/qibo/pull/1561/#discussion_r1936008349) above, rebase the ``qibojit`` PR as well and double check that the ``pyproject`` and lock make sense.

tests/test_backends_qibotn.py
BrunoLiegiBastonLiegi: In the end I created a fixture that creates a context manager which sets the number of threads to a desired value and then set it back to the original number. That might be useful more in general if it works as I intended, but please double check because I am not totally sure.

pyproject.toml
```diff
@@ -56,7 +56,7 @@ ipython = "^8.10.0"
 qulacs = { version = "^0.6.4", markers = "(sys_platform == 'darwin' and python_version > '3.9') or sys_platform != 'darwin'" }
 seaborn = "^0.13.2"
 ipykernel = "^6.29.4"
-qibojit = { git = "https://github.com/qiboteam/qibojit.git" }
+qibojit = { git = "https://github.com/qiboteam/qibojit.git", branch="numba_threads" }
```
alecandido: Since here we do not have the same problem of tests (there isn't anything circular, since the changes in Qibojit do not depend on changes here), I'd suggest first merging on the other side, and switch back to the default branch before merging this to `master`.

tests/conftest.py
```diff
@@ -136,3 +137,20 @@ def pytest_generate_tests(metafunc):
...
+
+    @contextmanager
+    def nthreads(n: int):
+        set_num_threads(n)
+        try:
+            yield
+        finally:
+            set_num_threads(original_threads)
+
+    yield nthreads
```
alecandido, line 156: Since `get_threads()` takes no input even on the other side, you may simplify this to
```suggestion
    set_num_threads(get_threads())
    yield
    set_num_threads(original_threads)
```

To be fair, the code above is even doing something more than preventing the context manager in the unit test. The context manager you used is resetting the original number of threads in case an error happens, and only in that case. However, I would always reset the state. And, with fixtures, you do not need to explicitly handle exceptions, since the Pytest machinery will capture the error anyhow, and always execute the finalization block (i.e. resume the execution after the `yield`).

alecandido: Moreover, this hack was strictly related to Quimb, and I have no evidence that the current numba fixture may be useful for anything else beyond that.

So, you could even use something like:
```py
from numba import get_num_threads, set_num_threads

@pytest.fixture
def quimb_backend():
    original_threads = get_num_threads()
    set_num_threads(get_threads())
    from qibotn.backends.quimb import QuimbBackend
    yield QuimbBackend
    set_num_threads(original_threads)
```
and put it directly in the `test_backends_qibotn.py` file.
Fixtures are scoped, you can make them available for all tests, but it is not worth to expose them in the most global scope, if they are needed only in a very specific place.

Btw, to the best of my knowledge is the Quimb import to be troublesome, not the Numba one - so that one can stay top-level.

alecandido (commented):
Just a note on the merge arrangement, and some comments about the fixture. Btw, thanks for that!

tests/conftest.py
BrunoLiegiBastonLiegi: > The context manager you used is resetting the original number of threads in case an error happens

shouldn't the ``finally`` always happen regardless of the ``try`` outcome (which anyway should never fail since is just yielding none)? 
In general I liked more the idea to call ``with numba_threads(n):`` as a context manager. it conveys more the idea of some scoped resources setting, rather than plainly calling  ``numba_threads(n)``. Furthermore, the resetting of the threads is happening only at the end of the test in the second case, which is less modular (not necessarily worse, just a choice).

In any case, having the fixture to return the ``QuimbBackend`` works as well, but I don't find it very useful since it's used only there. I mean, at that point we could just manually set the threads in that specific test without involving a fixture.

alecandido: > shouldn't the `finally` always happen regardless of the `try` outcome (which anyway should never fail since is just yielding none)?

Whoops, you're right: I saw a `try:` and inferred an `except:`. My bad.

alecandido: > In general I liked more the idea to call `with numba_threads(n):` as a context manager. it conveys more the idea of some scoped resources setting, rather than plainly calling `numba_threads(n)`. Furthermore, the resetting of the threads is happening only at the end of the test in the second case, which is less modular (not necessarily worse, just a choice).

Well, but then you put the whole test in the context manager...
In the end, you're creating an extra scope, which is identical to the one you already had. And the `fixture` is already doing the job of the context manager, handling the finalization.

So, you may argue to keep the context manager instead of the fixture. Which is fair - I just find the fixture easier/more suitable for a test, but they will perform the exact same task.

> In any case, having the fixture to return the `QuimbBackend` works as well, but I don't find it very useful since it's used only there.

Well, it is used only there in any case.

> I mean, at that point we could just manually set the threads in that specific test without involving a fixture.

This is just an argument against modularity: they would be performing two different tasks, so it is relevant to split into different functions.
I do not consider a great practice having long functions, just because the other ones you may write would be used in a single place...

BrunoLiegiBastonLiegi: > Well, but then you put the whole test in the context manager...

yes, in this case, but nothing prevents you from having a tests with different sections using a different number of threads, which is not possible with the other approach.

> Well, it is used only there in any case.

True, but more likely to be reused in other places in the future, rather than the ``quimb_backend``, at least this is my feeling.

alecandido: > True, but more likely to be reused in other places in the future, rather than the `quimb_backend`, at least this is my feeling.

Ok, I agree that is more likely to be reused than the other. And we may just have a different perception.

My point is that it is just unlikely, in absolute terms, that even the Numba one will be reused.
The need for setting the number of threads in an unscoped manner, outside a backend or anything else, is not something that I expect we're going to test for.

In particular, we would not be using that even here, it was just a workaround to avoid Quimb failing (if you remove that, it seemed to fail - but it looked like a Quimb/Autoray bug, that may even be gone by now...).

BrunoLiegiBastonLiegi: So, the test used to fail here, in qibo, and that is the reason why we were manually setting the threads, right? Because I tried right now without any manual threads setting and it passes.

alecandido: I think so - either that, or it was failing in Qibotn, and then it got copy-pasted here.

However, I believe it is possible it was failing even before the test, directly during the import.

If it's not failing (or not failing any longer), better not doing anything, and keep all the imports top-level. Mostly as you did in 445013e (but you can lift the `QuimbBackend` import as well).

Issue #1562, opened 2025-01-23
Improvements to final `logical-physical` qubit mapping, documentation


GuillermoAbadLopez:
Hello, I'm a developer for Qilimanjaro, and we are currently implementing your placing/routing into our stack,
here I leave some small improvements that would have helped me, hopefully, they are useful to you.

- version: `qibo 0.2.15`

## Small problems:

1) All the docstrings for the call of the `Router`/`Placer` passes, say something similar to: `"final physical-logical qubits mapping."` for explaining the final mapping/layout, without explicitly stating what the keys and values are. And for an external user, although I can guess the format is: `keys-values`, I think stating it explicitly would be better.

2) Assuming the order of: `physical-logical` in the dosctring corresponds to `keys-values`, then there might be a contradiction. In the `Sabre` router call it says `logical-physical`, but then the function it ends up calling, has the opposite `physical-logical`:
<img width="359" alt="Screenshot 2025-01-22 at 10 45 50" src="https://github.com/user-attachments/assets/f7c719a1-10cb-4ccb-ac13-1bf83c4690be" />
<img width="359" alt="Screenshot 2025-01-22 at 10 45 50" src="https://github.com/user-attachments/assets/b8aff3d7-4fd7-4547-be87-db313f2d7b16" />

## Possible solutions:
1) State explicitly what are the keys and values for the final mapping, in the `Router`/`Placer` passes.
2) Ensure consistency in the docstrings of the methods that call each other.

GuillermoAbadLopez:
Tagging possible relevant people following PR https://github.com/qiboteam/qibo/pull/1500:
@stavros11 @alecandido @csookim

GuillermoAbadLopez:
Opened a PR with very small changes, in case you think it's enough: https://github.com/qiboteam/qibo/pull/1563

alecandido:
@GuillermoAbadLopez thanks for your interest and the issue!

In a sense, the current transpiler is more or less in an intermediate stage. It is usable, but not incredibly optimized and deeply reviewed. That's just for you to gauge the depth you want to reach in there.

Moreover, Qibo is suffering from a common Python shortcoming, which is exposing essentially everything from a package, and relying on conventions for internals recognition (which are clear up to a point, and need to be supplemented with project-wide policies - ttbomu).
In particular, Qibo itself is not making (yet) any effort to clarify which is the public interface.

[...]

PR #1563, opened 2025-01-23
Improving docstrings in final qubit mapping, w/ `{logical: physical}`

GuillermoAbadLopez:
## PR addressing Issue: https://github.com/qiboteam/qibo/issues/1562
- [x] Changes incorrect `CircuitMap.final_layout()` docstring from `physical-logical` to `logical-physical`
- [x] Changes all `logical-physical` by `{logical: physical}`, for showing more explicitly which are the keys/values

## Extra for reviewers:
- [ ] Please check the first change from above is correct, since I didn't went too deeply to check it. Instead I assumed it has to be that way based on the contradiction there was, and checking the next method names it calls 👍 

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/transpiler/abstract.py
+++ b/src/qibo/transpiler/abstract.py
@@ -47,7 +47,7 @@ class Router(ABC):
             circuit (:class:`qibo.models.circuit.Circuit`): Circuit to be routed.
 
         Returns:
-            (:class:`qibo.models.circuit.Circuit`, dict): Routed circuit and final logical-physical qubit mapping.
+            (:class:`qibo.models.circuit.Circuit`, dict): Routed circuit and final {logical: physical} qubit mapping.
         """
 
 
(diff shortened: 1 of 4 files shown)


GuillermoAbadLopez:
### PR itself:
> Nothing to amend, we can merge.

Cool to hear 🙌 

---
### Beyond the PR:

> Just notice that the logical and physical adjectives were also a poor choice, since most often they are just intermediate steps, so it's just a way of handling a permutation.

Yes, and qiskit has changed all these names recently, so the industry hasn't yet settled on anything, it can be tricky to keep up, with what each permutation represents for each different SDK.. 😅 

> In principle, we could return just a permutation as a sequence of permuted indices.

Precisely!


---

> So, I would discourage any further focus on permutations. Including the current final_layout, which is only sometimes present, and leading to patterns like the following
>  `[source_code]`
> Just relying the Circuit.wire_names will embed the optional permutation in the resulting Circuit.

That makes sense, I'll try it, thanks!

renatomello (approved):

PR #1565, opened 2025-01-25
New backend specialised for Hamming-weight-preserving circuits

renatomello:
Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/backends/hamming_weight.py
+++ b/src/qibo/backends/hamming_weight.py
@@ -1,0 +1,94 @@
+import numpy as np
+from scipy.special import binom
+
+from qibo.backends.numpy import NumpyBackend
+
+
+class HammingWeightBackend(NumpyBackend):
+    def __init__(self):
+        super().__init__()
+        self._dict_cached_strings = {}
+
+    def _get_cached_strings(self, nqubits: int, weight: int, ncontrols: int = 0):
+        initial_string = np.array(
+            [1] * (weight - 1 - ncontrols)
+            + [0] * ((nqubits - 2 - ncontrols) - (weight - 1 - ncontrols))
+        )
+        strings =_ ehrlich_algorithm(initial_string, False)
+        strings = [list(string) for string in strings]
+        strings = np.asarray(strings, dtype=int)
+
+        return strings
+
+    def apply_gate(self, gate, state, nqubits, weight):
+        # Right now, it works only with two-qubit Givens rotations,
+        # e.g. gates.RBS, gates.GIVENS, gates.SWAP, gates.iSWAP,
+        # gates.SiSWAP, and gates.RZZ (up to global phase).
+        qubits = list(gate.target_qubits)
+        controls = list(gate.control_qubits)
+        ncontrols = len(controls)
+        other_qubits = list(set(list(range(nqubits))) ^ set(qubits + controls))
+
+        key = f"{ncontrols}"
+        if key not in self._dict_cached_strings:
+            self._dict_cached_strings[key] = self._get_cached_strings(
+                nqubits, weight, ncontrols
+            )
+
+        strings = self._dict_cached_strings[key]
+
+        matrix = gate.matrix().real
+        matrix_0101 = matrix[1, 1]
+        matrix_0110 = matrix[1, 2]
+        matrix_1001 = matrix[2, 1]
+        matrix_1010 = matrix[2, 2]
+
+        indexes_in = np.zeros((len(strings), nqubits), dtype=str)
+        indexes_in[:, other_qubits] = strings
+        if len(controls) > 0:
+            indexes_in[:, controls] = "1"
+        indexes_in[:, qubits] = ["1", "0"]
+        indexes_out = np.copy(indexes_in)
+        indexes_out[:, qubits] = ["0", "1"]
+
+        indexes_in = np.array(
+            [self._dict_indexes["".join(elem)] for elem in indexes_in]
+        )
+        indexes_out = np.array(
+            [self._dict_indexes["".join(elem)] for elem in indexes_out]
+        )
+
+        old_in, old_out = state[indexes_in], state[indexes_out]
+
+        new_amplitudes_in = matrix_1010 * old_in + matrix_1001 * old_out
+        new_amplitudes_out = matrix_0101 * old_out + matrix_0110 * old_in
+
+        state[indexes_in] = new_amplitudes_in
+        state[indexes_out] = new_amplitudes_out
+
+        return state
+
+    def execute_circuit(self, circuit, weight: int, initial_state=None, nshots=1000):
+        # Right now, it works only for ``weight'>=3`` and for gates with
+        # number of controls < weight.
+        nqubits = circuit.nqubits
+        n_choose_k = int(binom(nqubits, weight))
+        indexes = list(range(n_choose_k))
+
+        lexicographical_order = self._get_cached_strings(nqubits + 2, weight + 1)
+        lexicographical_order = [
+            "".join(item.astype(str)) for item in lexicographical_order
+        ]
+        lexicographical_order.sort()
+        self._dict_indexes = dict(zip(lexicographical_order, indexes))
+        del lexicographical_order, indexes
+
+        if initial_state is None:
+            initial_state = self.np.zeros(n_choose_k, dtype=float)
+            initial_state[0] = 1
+
+        state = initial_state
+        for gate in circuit.queue:
+            state = self.apply_gate(gate, state, nqubits, weight)
+
+        return state
--- a/src/qibo/backends/hamming_weight.py
+++ b/src/qibo/backends/hamming_weight.py
@@ -2,19 +2,35 @@ import numpy as np
 from scipy.special import binom
 
 from qibo.backends.numpy import NumpyBackend
+from qibo.models.encodings import _ehrlich_algorithm
 
 
 class HammingWeightBackend(NumpyBackend):
-    def __init__(self):
+    def __init__(self, engine=None):
         super().__init__()
+
+        if engine is None:
+            from qibo.backends import (
+                _check_backend,
+                _get_engine_name,
+            )  # pylint: disable=C0415
+
+            engine = _get_engine_name(_check_backend(engine))
+
+        self.platform = engine
+
+        self.np = self.engine.np
+        
         self._dict_cached_strings = {}
+
+        self.name = "hamming_weight"
 
     def _get_cached_strings(self, nqubits: int, weight: int, ncontrols: int = 0):
         initial_string = np.array(
             [1] * (weight - 1 - ncontrols)
             + [0] * ((nqubits - 2 - ncontrols) - (weight - 1 - ncontrols))
         )
-        strings =_ ehrlich_algorithm(initial_string, False)
+        strings = _ehrlich_algorithm(initial_string, False)
         strings = [list(string) for string in strings]
         strings = np.asarray(strings, dtype=int)
 
@@ -37,7 +53,7 @@ class HammingWeightBackend(NumpyBackend):
 
         strings = self._dict_cached_strings[key]
 
-        matrix = gate.matrix().real
+        matrix = gate.matrix(backend=self.engine).real
         matrix_0101 = matrix[1, 1]
         matrix_0110 = matrix[1, 2]
         matrix_1001 = matrix[2, 1]
--- a/src/qibo/backends/hamming_weight.py
+++ b/src/qibo/backends/hamming_weight.py
@@ -1,3 +1,5 @@
+"""Module defining the Hamming-weight-preserving backend."""
+
 import numpy as np
 from scipy.special import binom
 
@@ -85,8 +87,6 @@ class HammingWeightBackend(NumpyBackend):
         return state
 
     def execute_circuit(self, circuit, weight: int, initial_state=None, nshots=1000):
-        # Right now, it works only for ``weight'>=3`` and for gates with
-        # number of controls < weight.
         nqubits = circuit.nqubits
         n_choose_k = int(binom(nqubits, weight))
         indexes = list(range(n_choose_k))
--- a/src/qibo/backends/hamming_weight.py
+++ b/src/qibo/backends/hamming_weight.py
@@ -12,17 +12,17 @@ class HammingWeightBackend(NumpyBackend):
         super().__init__()
 
         if engine is None:
-            from qibo.backends import (
+            from qibo.backends import (  # pylint: disable=C0415
                 _check_backend,
                 _get_engine_name,
-            )  # pylint: disable=C0415
+            )
 
             engine = _get_engine_name(_check_backend(engine))
 
         self.platform = engine
 
         self.np = self.engine.np
-        
+
         self._dict_cached_strings = {}
 
         self.name = "hamming_weight"
--- a/src/qibo/backends/numpy.py
+++ b/src/qibo/backends/numpy.py
@@ -1,3 +1,5 @@
+"""Module defining the Numpy backend."""
+
 import collections
 import math
 from typing import Union
--- a/src/qibo/backends/__init__.py
+++ b/src/qibo/backends/__init__.py
@@ -286,6 +286,10 @@ def _check_backend(backend):
     return backend
 
 
+def _get_engine_name(backend):
+    return backend.platform if backend.platform is not None else backend.name
+
+
 def list_available_backends(*providers: str) -> dict:
     """Lists all the backends that are available."""
     available_backends = MetaBackend().list_available()
--- a/src/qibo/backends/clifford.py
+++ b/src/qibo/backends/clifford.py
@@ -12,10 +12,6 @@ from qibo.backends.numpy import NumpyBackend
 from qibo.config import raise_error
 
 
-def _get_engine_name(backend):
-    return backend.platform if backend.platform is not None else backend.name
-
-
 class CliffordBackend(NumpyBackend):
     """Backend for the simulation of Clifford circuits following
     `Aaronson & Gottesman (2004) <https://arxiv.org/abs/quant-ph/0406196>`_.
@@ -35,7 +31,10 @@ class CliffordBackend(NumpyBackend):
             self._stim = stim
         else:
             if engine is None:
-                from qibo.backends import _check_backend  # pylint: disable=C0415
+                from qibo.backends import (
+                    _check_backend,
+                    _get_engine_name,
+                )  # pylint: disable=C0415
 
                 engine = _get_engine_name(_check_backend(engine))
 
--- a/src/qibo/backends/clifford.py
+++ b/src/qibo/backends/clifford.py
@@ -31,10 +31,10 @@ class CliffordBackend(NumpyBackend):
             self._stim = stim
         else:
             if engine is None:
-                from qibo.backends import (
+                from qibo.backends import (  # pylint: disable=C0415
                     _check_backend,
                     _get_engine_name,
-                )  # pylint: disable=C0415
+                )
 
                 engine = _get_engine_name(_check_backend(engine))
 
--- a/src/qibo/backends/qulacs.py
+++ b/src/qibo/backends/qulacs.py
@@ -16,7 +16,7 @@ from qibo.result import CircuitResult, QuantumState
 
 
 def circuit_to_qulacs(
-    circuit: "qibo.Circuit", # type: ignore
+    circuit: "qibo.Circuit",  # type: ignore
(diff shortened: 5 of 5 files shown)


renatomello:
@BrunoLiegiBastonLiegi could you help me out with setting the correct engine from the name of the platform? I didn't understand what you did for the Clifford backend.

renatomello:
This was closed by mistake, so I'm reopening it.

BrunoLiegiBastonLiegi:
What is the problem with the engine here? I quickly looked into the implementation but everything seems to be fine, except for the fact that the ``.platform`` attribute is a backend once again, which is the thing we decided to avoid for the clifford backend in the end.
General question: which are the custom operations each engine should provide to power this specific backend? None, right? Everything plays around the ``self.np = engine.np`` and the redefinition of ``apply_gate`` and ``execute_circuit``, right?

renatomello:
>General question: which are the custom operations each engine should provide to power this specific backend? None, right? Everything plays around the `self.np = engine.np` and the redefinition of `apply_gate` and `execute_circuit`, right?

Yes.

src/qibo/backends/hamming_weight.py
```diff
@@ -64,6 +60,15 @@ def execute_circuit(self, circuit, weight: int, initial_state=None, nshots=None)
         for gate in circuit.queue:
             if isinstance(gate, gates.Channel):
                 raise_error(RuntimeError, "Circuit must not contain channels.")
+            if (
+                not gate.hamming_weight
+                and not gate.__class__.__name__ == "M"
+                and not gate.__class__.__name__ == "Unitary"
```
renatomello, line 66:
```suggestion
                and not gate.__class__.__name__ not in ["M", "Unitary"]
```

src/qibo/gates/abstract.py
```diff
@@ -74,6 +74,11 @@ def clifford(self):
         """Return boolean value representing if a Gate is Clifford or not."""
         return False
 
+    @property
+    def hamming_weight(self):
+        """Return boolean value representing if a Gate is Hamming weight preserving or not."""
```
renatomello, line 79:
```suggestion
        """Return boolean value representing if a Gate is Hamming-weight-preserving or not."""
```

src/qibo/gates/gates.py
```diff
@@ -2545,6 +2698,10 @@ def __init__(
         self.init_args = [qubits_in, qubits_out]
         self.init_kwargs = {"theta": theta, "phi": phi, "trainable": trainable}
 
+    @property
+    def hamming_weight(self):
+        return len(self.init_args[0]) == len(self.init_args[1])
```
renatomello, line 2703: Nice

src/qibo/quantum_info/hamming_weight.py
```diff
@@ -54,7 +55,7 @@ def full_state(self):
         return state
 
     def probabilities(self, qubits: Optional[Union[list, set]] = None):
-        if self.nshots is None:
+        if self.nshots is None or len(self.measurements) == 0:
             return self.exact_probabilities(qubits)
         else:
             return self.probabilities_from_samples(qubits)
```
renatomello:
```suggestion
        if self.nshots is None or len(self.measurements) == 0:
            return self.exact_probabilities(qubits)

        return self.probabilities_from_samples(qubits)
```

doc/source/api-reference/qibo.rst
```diff
@@ -2697,6 +2697,55 @@ Alternatively, a Clifford circuit can also be executed starting from the :class:
...
+by specifying the engine used for calculation, if not provided the current backend is used
+
+.. testcode::  python
+
+    import qibo
+
+    # setting numpy as the global backend
+    qibo.set_backend("numpy")
+    # the clifford backend will use the numpy backend as engine
+    backend = qibo.backends.CliffordBackend()
```
renatomello, line 2739:
```suggestion
    from qibo import set_backend
    from qibo.backends import HammingWeightBackend

    # setting numpy as the global backend
    set_backend("numpy")
    # the clifford backend will use the numpy backend as engine
    backend = HammingWeightBackend()
```

doc/source/api-reference/qibo.rst
```diff
@@ -2697,6 +2697,56 @@ Alternatively, a Clifford circuit can also be executed starting from the :class:
...
+by specifying the engine used for calculation, if not provided the current backend is used
+
+.. testcode::  python
+
+    from qibo import set_backend
+    from qibo.backends import HammingWeightBackend
+
+    # setting numpy as the global backend
+    set_backend("numpy")
+    # the clifford backend will use the numpy backend as engine
```
renatomello, line 2739:
```suggestion
    from qibo import set_backend
    from qibo.backends import HammingWeightBackend

    # setting numpy as the global backend
    set_backend("numpy")
    # the Hamming-weight backend will use the numpy backend as engine
```

doc/source/api-reference/qibo.rst
```diff
@@ -2697,6 +2697,56 @@ Alternatively, a Clifford circuit can also be executed starting from the :class:
...
+:meth:`qibo.quantum_info.hamming_weight.HammingWeightResult.full_state` method.
+Note, however, that this process is inefficient as it involves the construction of
+a statevector with size exponential in the number of qubits.
+
+As for the other backends, the Hamming-weight backend can be set with
+
+.. testcode::  python
+
+    import qibo
+    qibo.set_backend("hamming_weight", platform="numpy")
```
renatomello, line 2728:
```suggestion
    from qibo import set_backend

    set_backend("hamming_weight", platform="numpy")
```

renatomello:
@BrunoLiegiBastonLiegi @stavros11 sans a handful of docstrings that @AlejandroSopena will update today, this is ready for review.

src/qibo/backends/hamming_weight.py
```diff
@@ -505,6 +521,24 @@ def _apply_gate_CCZ(self, gate, state, nqubits: int, weight: int):
         return state
 
     def _apply_gate_n_qubit(self, gate, state, nqubits, weight):
+        """Custom ``apply_gate`` method for Hamming-weight-preserving n-qubit gates.
+
+        This method performs matrix multiplication directly in the subspace with Hamming-weight
+        ``weight`` without the need to calculate the full matrix representation of the gate.
+
+        .. note::
+            The attribute ``gate.hamming_weight`` must be set to ``True`` for this method to work.
```
renatomello, line 530:
```suggestion
        .. note::
            The attribute ``gate.hamming_weight`` must be manually set to ``True`` for this method to work.
```

src/qibo/backends/hamming_weight.py
```diff
@@ -505,6 +521,24 @@ def _apply_gate_CCZ(self, gate, state, nqubits: int, weight: int):
...
+
+        Args:
+            gate (:class:`qibo.gates.abstract.Gate`): Hamming-weight-preserving
+                n-qubit gate to be applied to ``state``
```
renatomello, line 534:
```suggestion
                :math:`n`-qubit gate to be applied to ``state``
```

src/qibo/backends/hamming_weight.py
```diff
@@ -0,0 +1,716 @@
...
+
+            engine = _get_engine_name(_check_backend(engine))
+
+        self.platform = engine
+        if engine == "numpy":
+            self.engine = construct_backend(self.platform)
+        elif engine in ["numba", "cupy", "cuquantum"]:
+            self.engine = construct_backend("qibojit", platform=self.platform)
+        elif engine in ["tensorflow", "pytorch"]:  # pragma: no cover
+            self.engine = construct_backend("qiboml", platform=self.platform)
```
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: Ok I double checked, the backend is used mostly for the ``cast`` method and retrieving the gate matrices, which cannot be circumvented unfortunately. What I was wondering is whether we could dynamically inherit from the correct backend, and make this something similar to our ``MetaBackend``:
``` python
from qibo.backends import NumpyBackend

def some_method(self, x):
    return self.np.log(x)

def create_HW_backend(backend):

    hw_backend = type("HWBackend", (backend,), {"some_method": some_method,})()
    hw_backend.name = "hamming_weight"
    return hw_backend

b = create_HW_backend(NumpyBackend)
print(b.np) # this returns numpy
print(b.some_method(2)) # this returns np.log(2)

from qibo import gates
print(gates.x(0).matrix(b)) # this returns the numpy matrix of X
```

BrunoLiegiBastonLiegi (commented):
Thanks, I started looking into this, but before going ahead with the review I wished to discuss about the possibility of changing the structure of the backend, as I detail in one of the comments below.

BrunoLiegiBastonLiegi (commented):
Ok I dug a little deeper, here some more comments. Broadly speaking I think you might benefit from the use of ``functools.cache`` instead of manually caching (assuming the number of different combinations is indeed restricted).

src/qibo/backends/hamming_weight.py
```diff
@@ -0,0 +1,716 @@
...
+        Uses the ``ehrlich_algorithm`` with an ``initial_string``.
+
+        Args:
+            initial_string (ndarray): Array of bits representing the input
+                of the Ehrlich algorithm.
+
+        Returns:
+            ndarray: All bitstrings with the same Hamming weight as ``initial_string``.
+        """
+        from qibo.models.encodings import _ehrlich_algorithm  # pylint: disable=C0415
```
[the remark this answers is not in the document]
renatomello: Yes

src/qibo/backends/numpy.py
```diff
@@ -69,17 +71,20 @@ def set_threads(self, nthreads):
...
-        elif self.is_sparse(x):
+
+        if self.is_sparse(x):
             return x.astype(dtype, copy=copy)
+
         return np.asarray(x, dtype=dtype, copy=copy if copy else None)
 
     def is_sparse(self, x):
-        from scipy import sparse
+        from scipy import sparse  # pylint: disable=C0415
```
renatomello, line 87:
```suggestion
```

AlejandroSopena:
This is ready for review @BrunoLiegiBastonLiegi

src/qibo/backends/_hamming_weight_operations.py
```diff
@@ -0,0 +1,651 @@
...
+    return self._apply_gate_n_qubit(gate, state, nqubits, weight)
+
+
+def execute_circuit(self, circuit, weight: int, initial_state=None, nshots: int = 1000):
+    """Execute ``circuit`` by applying the queue of gates to the ``initial_state``.
+
+    Args:
+        circuit (:class:`qibo.models.circuit.Circuit`): Hamming-weight-preserving circuit
+            to be executed.
+        weight (int): fixed Hamming weight of the ``initial_state``.
```
BrunoLiegiBastonLiegi, line 48: is there any way to make this optional? I don't know maybe there's a way to extract this from the ``initial_state`` or use a default value. I am saying this because otherwise this backend would fail with many of the functions we have in qibo, the first one that came to my mind: expectation value calculation of non diagonal observables.

src/qibo/backends/hamming_weight.py
```diff
@@ -1,716 +1,80 @@
...
-        except self.oom_error:  # pragma: no cover
-            raise_error(
-                RuntimeError,
-                f"State does not fit in {self.device} memory."
-                "Please switch the execution device to a "
-                "different one using ``qibo.set_device``.",
-            )
+    backend = None  # needed for pylint
+    if engine == "numpy":
+        from qibo.backends import NumpyBackend  # pylint: disable=C0415
```
BrunoLiegiBastonLiegi, line 23: This chain of ``if/elif`` can probably be made slightly more compact using ``qibo``'s ``construct_backend`` https://github.com/qiboteam/qibo/blob/9af40ab360b2d86760f4cd5f35e6101fff428c72/src/qibo/backends/__init__.py#L312
Namely something like:
``` python
if engine in ("numba", "cupy", "cuquantum"):
    backend = construct_backend("qibojit", platform=engine)
elif ...
```

src/qibo/backends/hamming_weight.py
```diff
@@ -1,716 +1,80 @@
...
-
-        measured_strings = {}
-        for string, index in zip(strings, indexes):
-            measured_string = "".join(string[q] for q in qubits)
-            measured_strings[measured_string] = measured_strings.get(
-                measured_string, 0
-            ) + self.np.real(self.np.abs(state[index]) ** 2)
+    spec = find_spec("qibo.backends._hamming_weight_operations")
+    module = module_from_spec(spec)
+    spec.loader.exec_module(module)
```
BrunoLiegiBastonLiegi, line 55: can't you just
```suggestion
    import qibo.backends._hamming_weight_operations as module
```
?
since you always import the methods from the same file and it's not runtime dependent

BrunoLiegiBastonLiegi (commented):
Thanks, the dynamic inheritance looks great to me! Just a handful of final small comments.

src/qibo/backends/_hamming_weight_operations.py
renatomello: To be fair, given that `nqubits` can be extracted from `circuit`, one could think of solving the equation $\binom{n}{k} = \\text{len}(\\text{initial\\_state})$. However, this involves finding the roots of a polynomial of degree $n^{k}$. It is much simpler to just pass it as an argument.

BrunoLiegiBastonLiegi (approved):

src/qibo/quantum_info/hamming_weight.py
```diff
@@ -0,0 +1,260 @@
...
+        Args:
+            numpy (bool, optional): If ``True`` the returned tensor will be a ``numpy`` array,
+                otherwise it will follow the backend tensor type.
+                Defaults to ``False``.
+
+        Returns:
+            ndarray: State in the computational basis of the subspace
+            with fixed Hamming weight ordered in lexicographical order.
+        """
+        return super().state(numpy=numpy)
```
stavros11, line 94: I believe these can just be inherited from `QuantumState`.

src/qibo/quantum_info/hamming_weight.py
```diff
@@ -0,0 +1,260 @@
...
+        self.weight = weight
+        self.n_choose_k = int(binom(self.nqubits, self.weight))
+
+        self._state = state
+        self._samples = None
+        self._probs = None
+        self._frequencies = None
+        self._measurement_gate = None
+        self._repeated_execution_frequencies = None
+        self._measurement_gate = self.measurement_gate
```
stavros11, line 60: These are also not needed as they are already assigned by the `MeasurementOutcomes.__init__(self, measurements, backend)`.

src/qibo/quantum_info/hamming_weight.py
```diff
@@ -0,0 +1,260 @@
...
+
+        backend = HammingWeightBackend(engine)
+        QuantumState.__init__(self, state, backend)
+        MeasurementOutcomes.__init__(self, measurements, backend)
+
+        self.measurements = measurements
+        self.nqubits = nqubits
+        self.nshots = nshots
+        self.weight = weight
+        self.n_choose_k = int(binom(self.nqubits, self.weight))
```
stavros11, line 52: Is this used anywhere?

stavros11 (commented):
Thanks @renatomello. I tried running:
```py
from qibo import gates, Circuit, set_backend

c = Circuit(2)
c.add(gates.S(0))
c.add(gates.T(1))

set_backend("hamming_weight")
result = c()
```
and I got the following error:
```py
  File "/home/stavros/qiboteam/qibo/src/qibo/backends/_hamming_weight_operations.py", line 197, in _get_lexicographical_order
    n_choose_k = int(binom(nqubits, weight))
                     ^^^^^^^^^^^^^^^^^^^^^^
TypeError: ufunc 'binom' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```
Is this expected?

src/qibo/quantum_info/hamming_weight.py
renatomello: @AlejandroSopena

renatomello:
> Thanks @renatomello. I tried running:
> 
> ```python
> from qibo import gates, Circuit, set_backend
> 
> c = Circuit(2)
> c.add(gates.S(0))
> c.add(gates.T(1))
> 
> set_backend("hamming_weight")
> result = c()
> ```
> 
> and I got the following error:
> 
> ```python
>   File "/home/stavros/qiboteam/qibo/src/qibo/backends/_hamming_weight_operations.py", line 197, in _get_lexicographical_order
>     n_choose_k = int(binom(nqubits, weight))
>                      ^^^^^^^^^^^^^^^^^^^^^^
> TypeError: ufunc 'binom' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
> ```
> 
> Is this expected?

The circuit execution depends on the Hamming weight $k$, which in your example was never set. This is why it is failing. I would say that it should fail, but this probably is not the best error message to raise.

stavros11:
> The circuit execution depends on the Hamming weight k , which in your example was never set. This is why it is failing. I would say that it should fail, but this probably is not the best error message to raise.

Then, if I understand correctly, the only way to execute circuits with that backend is to use `backend.execute_circuit` and pass the weight. Can the usual `GlobalBackend` execution of qibo, like in my example, still work? If no, regardless of the error, it is probably worth to mention in the docs that a `weight` should be set somehow, or even better include an example of circuit execution with this backend, because from a quick look I could not find this information anywhere in the current docs, so I am not sure how an external user could use that.

renatomello:
> > The circuit execution depends on the Hamming weight k , which in your example was never set. This is why it is failing. I would say that it should fail, but this probably is not the best error message to raise.
> 
> Then, if I understand correctly, the only way to execute circuits with that backend is to use `backend.execute_circuit` and pass the weight. Can the usual `GlobalBackend` execution of qibo, like in my example, still work? If no, regardless of the error, it is probably worth to mention in the docs that a `weight` should be set somehow, or even better include an example of circuit execution with this backend, because from a quick look I could not find this information anywhere in the current docs, so I am not sure how an external user could use that.

I added kwargs to `Circuit.execute`, allowing the user to set the Hamming weight in a circuit `__call__` _e.g._ `circuit(weight=k)`. This should fix the issue you expertly found. @AlejandroSopena could you add the notes in the documentation like @stavros11 suggested? Thanks.

src/qibo/quantum_info/hamming_weight.py
AlejandroSopena: We need `self.nqubits`, `self.nshots` and `self.weight`. I realized that `self.nshots`  can be assigned by `MeasurementOutcomes.__init__()`

src/qibo/quantum_info/hamming_weight.py
AlejandroSopena: The `symbolic` method is a custom implementation to work specifically with this backend. It requires passing the `weight` to `backend.calculate_symbolic`. As for `state`, it differs only in the docstring. I have removed it from this class and added a note in the `QuantumState` method to clarify this.

AlejandroSopena:
> > > The circuit execution depends on the Hamming weight k , which in your example was never set. This is why it is failing. I would say that it should fail, but this probably is not the best error message to raise.
> > 
> > 
> > Then, if I understand correctly, the only way to execute circuits with that backend is to use `backend.execute_circuit` and pass the weight. Can the usual `GlobalBackend` execution of qibo, like in my example, still work? If no, regardless of the error, it is probably worth to mention in the docs that a `weight` should be set somehow, or even better include an example of circuit execution with this backend, because from a quick look I could not find this information anywhere in the current docs, so I am not sure how an external user could use that.
> 
> I added kwargs to `Circuit.execute`, allowing the user to set the Hamming weight in a circuit `__call__` _e.g._ `circuit(weight=k)`. This should fix the issue you expertly found. @AlejandroSopena could you add the notes in the documentation like @stavros11 suggested? Thanks.

I updated the documentation

doc/source/api-reference/qibo.rst
```diff
@@ -2709,6 +2709,96 @@ Alternatively, a Clifford circuit can also be executed starting from the :class:
...
+    # the above execution is equivalent to the following
+    # that uses the global backend and the ``weight`` kwarg
+
+    set_backend("hamming_weight", platform="numpy")
+
+    result = circuit(weight=2)
+
+.. testsetup::  python
+
+    set_backend("numpy")
```
renatomello, line 2771:
```suggestion
.. testsetup::  python
    from qibo import set_backend

    set_backend("numpy")
```

renatomello:
@stavros11 apart from whatever this CI error is, this should be ready

stavros11 (approved):

PR #1571, opened 2025-01-31
Fix docstring of `gates.PRX`

renatomello:
Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/gates/gates.py
+++ b/src/qibo/gates/gates.py
@@ -694,11 +694,11 @@ class PRX(ParametrizedGate):
 
     Corresponds to the following unitary matrix
 
-        .. math::
-            \\begin{pmatrix}
-                \\cos{(\\theta / 2)} & -i e^{-i \\phi} \\sin{(\\theta / 2)} \\
-                -i e^{i \\phi} \\sin{(\\theta / 2)} & \\cos{(\\theta / 2)}
-            \\end{pmatrix}
+    .. math::
+        \\begin{pmatrix}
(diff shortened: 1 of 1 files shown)


src/qibo/gates/gates.py
```diff
@@ -694,11 +694,11 @@ class PRX(ParametrizedGate):
...
-        .. math::
-            \\begin{pmatrix}
-                \\cos{(\\theta / 2)} & -i e^{-i \\phi} \\sin{(\\theta / 2)} \\
-                -i e^{i \\phi} \\sin{(\\theta / 2)} & \\cos{(\\theta / 2)}
-            \\end{pmatrix}
+    .. math::
+        \\begin{pmatrix}
+            \\cos{(\\theta / 2)} & -i e^{-i \\phi} \\sin{(\\theta / 2)} \\\\
+            -i e^{i \\phi} \\sin{(\\theta / 2)} & \\cos{(\\theta / 2)}
+        \\end{pmatrix}
```
alecandido, line 701: To save a few `\`, you can put an `r` in front of the docstring's opening quotes, i.e. `r"""` (`r` for raw). This will deactivate escape codes, so `\n` won't be any longer a newline, and `\t` is not a tab (but they will be literally interpreted as the two characters you typed). However, that's usually what you want in a docstring.
https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals

You may have been already aware. In case you chose explicitly, sorry for the extra noise.

alecandido (approved):

AlejandroSopena (approved):

PR #1572, opened 2025-02-03
Modify `models.encodings.hamming_weight_encoder` to upload data in the lexicographical order

renatomello:
Data is amplitude-encoded in the lexicographical order instead of the order of the original Gray code.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/models/encodings.py
+++ b/src/qibo/models/encodings.py
@@ -373,12 +373,6 @@ def hamming_weight_encoder(
 
     # sort data such that the encoding is performed in lexicographical order
     lex_order = [int(string, 2) for string in bitstrings]
-    # bitstrings_lexicographical = list(
-    #     zip(bitstrings_lexicographical, range(len(bitstrings_lexicographical)))
-    # )
-    # bitstrings_lexicographical.sort()
-    # bitstrings_lexicographical = np.asarray(bitstrings_lexicographical)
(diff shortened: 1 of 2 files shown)


src/qibo/models/encodings.py
```diff
@@ -371,6 +371,15 @@ def hamming_weight_encoder(
     initial_string = np.array([1] * weight + [0] * (nqubits - weight))
     bitstrings, targets_and_controls = _ehrlich_algorithm(initial_string)
 
+    # sort data such that the encoding is performed in lexicographical order
+    bitstrings_lexicographical = list(zip(bitstrings, range(len(bitstrings))))
```
BrunoLiegiBastonLiegi, line 375: I am not sure what `bitstrings` contains exactly, but if you can sort with `numpy` why not to use [argsort](https://numpy.org/doc/stable/reference/generated/numpy.argsort.html) instead of manually building the indices?
``` python
data = data[np.argsort(bitstrings)]
```

renatomello: I need the indexes of the sorted list but the array of indexes itself needs to be ordered in the same order as the original array. Moreover, there was a bug in the code actually so I pushed a fix now.

BrunoLiegiBastonLiegi (approved):

PR #1576, opened 2025-02-05
Adding ``wire_names`` to ``circuit.raw`` and ``from_dict``

BrunoLiegiBastonLiegi:
As per title this fixes the serialization of circuit by including the ``wire_names``, should fix the problems encountered in qiboteam/qibo-client#99.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/tests/test_models_circuit.py
+++ b/tests/test_models_circuit.py
@@ -4,10 +4,13 @@ from collections import Counter
 
 import numpy as np
 import pytest
+from networkx import Graph
 
-from qibo import Circuit, gates
+from qibo import Circuit, gates, get_backend, get_transpiler
 from qibo.models.circuit import _resolve_qubits
 from qibo.models.utils import initialize
+from qibo.transpiler import Sabre
+from qibo.transpiler._exceptions import PlacementError
 
 
 def test_parametrizedgates_class():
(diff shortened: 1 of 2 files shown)


scarrazza (approved):

alecandido (approved):
If you could add a test it'd be even better :)

(you could actually test the entire use case, just by dumping, loading, and executing the result - it should fail on `master`, and work with your fix)

BrunoLiegiBastonLiegi:
I added a small test specifically checking that the wire names are serialized, however, triggering the transpiler error is not straightforward, as we would need to test on a backend that has a topology, like the ``qibolab`` one.

alecandido:
> I added a small test specifically checking that the wire names are serialized, however, triggering the transpiler error is not straightforward, as we would need to test on a backend that has a topology, like the `qibolab` one.

In principle, you could make a mock backend with some non-trivial topology, just for the sake of testing:
```py
class MockBackend(NumpyBackend): # <- maybe you could even inherit from AbstractBackend, not sure
    @property
    def topology(self):
        return # whatever you need
```

However, I will let you decide what is best :)

tests/test_models_circuit.py
```diff
@@ -375,6 +378,32 @@ def test_circuit_serialization():
...
+    # otherwise it crashes
+    transpiler = get_transpiler()
+    g = Graph()
+    g.add_edge(*wire_names)
+    transpiler.connectivity = g
+    transpiler.passes = [Sabre()]
+
+    c, _ = transpiler(c)
+    new_c, _ = transpiler(new_c)
+    assert new_c.wire_names == c.wire_names
```
alecandido, line 400: > In the end I just directly triggered the transpiler error as globally setting a mock backend is not straightforward.

That is perfectly fine, but at this point you do not need to pass through a backend at all
```suggestion
    transpiler = Passes(passes=[Sabre()], connectivity=Graph([wire_names]))

    c, _ = transpiler(c)
    new_c, _ = transpiler(new_c)
    assert new_c.wire_names == c.wire_names
```

BrunoLiegiBastonLiegi: ah ok, so without getting the transpiler with ``get_transpiler()``, it makes sense.

PR #1579, opened 2025-02-07
Selfhosted GPU tests restricted to GPU backends only

BrunoLiegiBastonLiegi:
Same as qiboteam/qibojit#206. A ``--gpu-only`` option is added to pytest to run tests on GPU backends only.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -82,7 +82,7 @@ def pytest_configure(config):
 
 @pytest.fixture
 def backend(backend_name, request):
-    if request.config.getoption("--gpu_only"):  # pragma: no cover
+    if request.config.getoption("--gpu-only"):  # pragma: no cover
         if backend_name not in ("cupy", "cuquantum"):
             pytest.skip("Skipping non-gpu backend.")
     yield get_backend(backend_name)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -81,7 +81,7 @@ def pytest_configure(config):
 
 
 @pytest.fixture
-def backend(backend_name):
+def backend(backend_name, request):
     if request.config.getoption("--gpu_only"):  # pragma: no cover
         if backend_name not in ("cupy", "cuquantum"):
             pytest.skip("Skipping non-gpu backend.")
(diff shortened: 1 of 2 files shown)


BrunoLiegiBastonLiegi:
Ok now the gpu tests take ~6 minutes to complete. Should we include even the pytorch and tensorflow (I don't think we have gpu support for jax yet) gpu tests?

tests/conftest.py
```diff
@@ -10,6 +10,16 @@
 
 from qibo.backends import _Global, construct_backend
 
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--gpu_only",
```
alecandido, line 16:
```suggestion
        "--gpu-only",
```

Sorry, it was my fault...

alecandido (approved):

andrea-pasquale:
> Ok now the gpu tests take ~6 minutes to complete. Should we include even the pytorch and tensorflow (I don't think we have gpu support for jax yet) gpu tests?

If you want feel free to do so, however if we see that tests are not passing for PyTorch and tensorflow we can address them in another PR.

BrunoLiegiBastonLiegi:
The thing is that right now I am not even sure how well is integrated the use of a device in the torch and tf backend. For example:

``` python
from qibo import set_backend
from qibo.models import QFT

set_backend(backend="qiboml", platform="pytorch")
c = QFT(5)
r = c()
r.probabilities().device
# device(type='cpu')
r.state().device
# device(type='cpu')
```
and the same happens for tensorflow.
Hence it's not worth it enabling them now as they would probably just run on CPU anyway. A revision of the device attribute has to be done for both of them first.

andrea-pasquale:
> The thing is that right now I am not even sure how well is integrated the use of a device in the torch and tf backend. For example:
> 
> ```python
> from qibo import set_backend
> from qibo.models import QFT
> 
> set_backend(backend="qiboml", platform="pytorch")
> c = QFT(5)
> r = c()
> r.probabilities().device
> # device(type='cpu')
> r.state().device
> # device(type='cpu')
> ```
> 
> and the same happens for tensorflow. Hence it's not worth it enabling them now as they would probably just run on CPU anyway. A revision of the device attribute has to be done for both of them first.

At this point I agree we can skip it, just open an issue so that we don't forget.

andrea-pasquale (approved):

PR #1582, opened 2025-02-10
Migrating to Numpy 2.0

BrunoLiegiBastonLiegi:
This attempts to migrate ``qibo`` to ``numpy 2.0``. Due to the lack of compatibility between ``cirq`` and ``numpy 2.0`` I had to drop the ``cirq`` dep and hide the tests. This is anyway still very much a work in progress, thus any decision in those regards can be postponed to when the migration is fully functional on our side.

Checklist:
- [x] Tests are passing.


alecandido:
I dropped support for py3.9, since required.

However, this would have happened anyhow in #1484, so it's only anticipated a little bit.

alecandido (approved):

alecandido:
@renatomello as noted in the related PR, this is required for py3.13 (cf. https://pypi.org/project/numpy/1.26.4/#files)

renatomello (approved):

Issue #1136, opened 2023-12-13
Improving integration of `quantum_info` module and `qibojit`


renatomello:
Right now, even though the `quantum_info` module is compatible to all backends, it is not making use of any of the `jit` properties from `numba`. Therefore, `numpy` and `numba` performances are the same across the board. It would be good to profile the module and see where it is worth to invest time to code custom operations in `qibojit`.
```[tasklist]
### Tasks
- [x] https://github.com/qiboteam/qibojit/pull/161
- [x] https://github.com/qiboteam/qibo/pull/1150
- [x] https://github.com/qiboteam/qibo/pull/1459
```

renatomello:
@AlejandroSopena @BrunoLiegiBastonLiegi 

This is list of the functions that have priority in being optimized:

Basis:
- [ ] `comp_basis_to_pauli`
- [ ] `pauli_to_comp_basis`

Random ensembles:
- [ ] `random_statevector`
- [ ] `random_density_matrix`
- [ ] `random_unitary`
- [ ] `random_quantum_channel`
- [ ] `random_clifford`

Superoperator transformations:
- [ ] `vectorization`
- [ ] `unvectorization`
- [ ] `_reshuffliing`
- [ ] `to_pauli_liouville`
- [ ] `choi_to_kraus`
- [ ] `kraus_to_choi`
- [ ] `kraus_to_stinespring`
- [ ] `stinespring_to_kraus`

BrunoLiegiBastonLiegi:
@renatomello should we create a separate module in `quantum_info`, something like `_utils.py` and create the relative modules in `qibojit` for `numba` and `cupy` for this?

renatomello:
> @renatomello should we create a separate module in `quantum_info`, something like `_utils.py` and create the relative modules in `qibojit` for `numba` and `cupy` for this? Then I thought we could delegate to the backend the loading of this utils through a method. For example in `NumbaBackend`:
> 
> ```
> @staticmethod
> def _load_quantum_info_utils():
>     from qibojit.quantum_info import _utils_cpu
>     
>     return _utils_cpu
> ```
> 
> and then the functions inside the `quantum_info` module will use the provided backend to load the utils and call them.

Which routines would be there?

BrunoLiegiBastonLiegi:
> Which routines would be there?

All the ones we want to improve with `numba` and `cupy`

renatomello:
> > Which routines would be there?
> 
[...]

PR #1583, opened 2025-02-10
``quantum_info`` module refactor and customization

BrunoLiegiBastonLiegi:
This tries to address #1136, but a refactor of the whole module is needed.

Checklist:
- [ ] Tests are passing.


BrunoLiegiBastonLiegi:
@renatomello is this check 

https://github.com/qiboteam/qibo/blob/312efb9d6934e85cfdaafcdb7027cdfe46bf87c4/tests/test_quantum_info_utils.py#L213

a known property of ``hellinger_shot_error/fidelity`` or just something that happened to be true for that particular seed? Because, this is failing after some of these changes.

renatomello:
> @renatomello is this check
> 
> https://github.com/qiboteam/qibo/blob/312efb9d6934e85cfdaafcdb7027cdfe46bf87c4/tests/test_quantum_info_utils.py#L213
> 
> a known property of `hellinger_shot_error/fidelity` or just something that happened to be true for that particular seed? Because, this is the only one failing in the `quantum_info` module after the changes, which might mean I messed up something (most likely `random_clifford`).

@AlejandroSopena

AlejandroSopena:
> > @renatomello is this check
> > https://github.com/qiboteam/qibo/blob/312efb9d6934e85cfdaafcdb7027cdfe46bf87c4/tests/test_quantum_info_utils.py#L213
> > 
> > a known property of `hellinger_shot_error/fidelity` or just something that happened to be true for that particular seed? Because, this is the only one failing in the `quantum_info` module after the changes, which might mean I messed up something (most likely `random_clifford`).
> 
> @AlejandroSopena

This works for a particular seed. A stronger test can be
``` python
assert hellinger_error < np.sqrt(2 * hellinger_fid / nshots) * 2**nqubits
```
This is an upper bound for the error.

BrunoLiegiBastonLiegi:
@renatomello I am basically done with the ``numpy`` low level implementation, I will now move to ``numba`` and ``cupy`` customization. However, there are two remaining parts in the ``numpy`` implementation which I am not very satisfied of:

- in the ``_sample_from_quantum_mallows_distribution`` function there is a loop where an array is gradually deleted inside a loop while we also access its elements, this is working but seems a little risky not easily readable. I tried to reverse engineer it and convert it to just simple array and indices operations, but I couldn't figure out a way. Do you have a clearer picture about this, how should you generate the permutation in practice in more simple terms?

https://github.com/qiboteam/qibo/blob/94ac52dc1bc60e1d07009e84968bd397640ffce4/src/qibo/quantum_info/quantum_info.py#L286C1-L290C30

- in the ``random_clifford`` function, when you build the gamma and delta matrices, there was a very convoluted loop with crossed checks, that I tried converting to array operations. 

https://github.com/qiboteam/qibo/blob/725e6460be5865f48158959e566119aa577fa825/src/qibo/quantum_info/random_ensembles.py#L650C1-L691C39 

I think I managed, but the problem is that the test is not very robust, in the sense that we just generated a fixed result for a precise seed and used that as a target. The problem is that whenever you change the random generation, the old target is no longer valid and you have to update it. Basically, we are building a target assuming that our code is correct to check whether our code is correct, which does not make any sense at all. Since this random clifford generation supposedly satisfies some properties of the clifford space (i.e. uniformity if I remember correctly), it may be better in the test to generate a bunch of cliffords and measure how uniform they are.

renatomello:
@BrunoLiegiBastonLiegi can you add to this PR the fact that the Pauli-Liouville matrices are always real? I think right now the return is a complex matrix. Thanks.

tests/test_backends_clifford.py
```diff
@@ -131,7 +131,12 @@ def test_two_qubits_gates(backend, gate):
         [2],
     ],
 )
-@pytest.mark.parametrize("seed", [15, 19, 25])
+@pytest.mark.parametrize(
+    "seed",
+    [
+        42,
```
renatomello, line 137:
```suggestion
        42
```

BrunoLiegiBastonLiegi:
> @BrunoLiegiBastonLiegi can you add to this PR the fact that the Pauli-Liouville matrices are always real? I think right now the return is a complex matrix. Thanks.

It depends, are they multiplied to complex arrays or matrix at some point? If yes, I would keep them as ``complex128`` arrays, otherwise errors will be raised if we try to multiply ``flota64`` and ``complex128``.

BrunoLiegiBastonLiegi:
The GPUs on the cluster are busy, but while we wait for the GPU tests to run, I think it is safe to start reviewing this.

BrunoLiegiBastonLiegi:
The ``qcs-sdk-python`` package is causing problems with the installation, this comes as ``cirq`` dependency as far as I understood. Bumping up ``cirq``'s minimum version solved the problem locally for me but not in the CI, completely removing it solved installation problems in the CI but naturally makes the ``cirq`` test fail.

src/qibo/backends/_clifford_operations.py
```diff
@@ -404,10 +404,9 @@ def _determined_outcome(state, q, nqubits):
 
 def _random_outcome(state, p, q, nqubits):
     p = p[0] + nqubits
-    tmp = state[p, q].copy()
     state[p, q] = 0
     h = state[:-1, q].nonzero()[0]
-    state[p, q] = tmp
+    state[p, q] = 1
```
renatomello, line 409: just an aside: could you add a docstring to this function?

src/qibo/backends/_clifford_operations.py
```diff
@@ -463,14 +462,14 @@ def _pack_for_measurements(state, nqubits):
...
+    z = _unpackbits(state[:, _packed_size(nqubits) : -1], axis=1, count=nqubits)
     return np.hstack((x, z, state[:, -1][:, None]))
 
 
 def _init_state_for_measurements(state, nqubits, collapse):
     if collapse:
-        return _unpackbits(state, axis=0, count=_dim_xz(nqubits))[: _dim(nqubits)]
+        return _unpackbits(state, axis=0, count=_dim(nqubits))
     else:
         return state.copy()
```
renatomello, line 474:
```suggestion
        return _unpackbits(state, axis=0, count=_dim(nqubits))

    return state.copy()
```

src/qibo/models/error_mitigation.py
```diff
@@ -204,7 +205,7 @@ def ZNE(
             nshots,
             readout,
             qubit_map,
-            seed=local_state,
+            seed=seed,
```
renatomello, line 208: Will this reset the rng object?

src/qibo/quantum_info/basis.py
```diff
@@ -89,43 +75,20 @@ def pauli_basis(
...
+        return getattr(backend.qinfo, f"_vectorize_sparse{fname}")(
+            nqubits, *_get_paulis(pauli_order, backend), normalization=normalization
+        )
+    elif vectorize:
+        return getattr(backend.qinfo, f"_vectorize{fname}")(
+            nqubits, *_get_paulis(pauli_order, backend), normalization=normalization
+        )
+    return backend.qinfo._pauli_basis(
+        nqubits, *_get_paulis(pauli_order, backend), normalization=normalization
+    )
```
renatomello, line 91:
```suggestion
        func = getattr(backend.qinfo, f"_vectorize_sparse{fname}")
    elif vectorize:
        func = getattr(backend.qinfo, f"_vectorize{fname}")
    else:
        func = backend.qinfo._pauli_basis
    
    return func(nqubits, *_get_paulis(pauli_order, backend), normalization=normalization)
```

src/qibo/quantum_info/utils.py
```diff
@@ -449,15 +449,30 @@ def haar_integral(
...
+        rho = backend.np.einsum(
+            "ijk,ijl->ikl", random_states, backend.np.conj(random_states)
+        )
+
+        for state in rho:
+            # haar_state = backend.np.reshape(
+            #    random_statevector(dim, backend=backend), (-1, 1)
+            # )
+
+            # rho = haar_state @ backend.np.conj(haar_state).T
```
renatomello, line 472:
```suggestion
```

tests/test_backends_clifford.py
```diff
@@ -121,73 +121,61 @@ def test_two_qubits_gates(backend, gate):
...
+    clifford_bkd.set_seed(seed)
 
-    c = random_clifford(nqubits, seed=1, backend=backend)
-    c.density_matrix = True
+    c = random_clifford(nqubits, seed=seed, density_matrix=True, backend=backend)
     c_copy = c.copy()
+    MEASURED_QUBITS = np.random.choice(list(range(nqubits)), size=2, replace=False)
+    prob_qubits = np.random.choice(MEASURED_QUBITS, size=prob_qubits, replace=False)
     c.add(gates.M(*MEASURED_QUBITS))
     c_copy.add(gates.M(*MEASURED_QUBITS))
```
renatomello, line 148:
```suggestion
    circuit = random_clifford(nqubits, seed=seed, density_matrix=True, backend=backend)
    circuit_copy = circuit.copy()
    MEASURED_QUBITS = np.random.choice(list(range(nqubits)), size=2, replace=False)
    prob_qubits = np.random.choice(MEASURED_QUBITS, size=prob_qubits, replace=False)
    circuit.add(gates.M(*MEASURED_QUBITS))
    circuit_copy.add(gates.M(*MEASURED_QUBITS))
```

tests/test_backends_clifford.py
```diff
@@ -121,73 +121,61 @@ def test_two_qubits_gates(backend, gate):
...
-
-    numpy_bkd.set_seed(2024)
     numpy_result = numpy_bkd.execute_circuit(c, nshots=nshots)
```
renatomello, line 149:
```suggestion
    numpy_result = numpy_bkd.execute_circuit(circuit, nshots=nshots)
```

tests/test_backends_clifford.py
```diff
@@ -121,73 +121,61 @@ def test_two_qubits_gates(backend, gate):
...
-
-    clifford_bkd.set_seed(2024)
     clifford_result = clifford_bkd.execute_circuit(c_copy, nshots=nshots)
```
renatomello, line 150:
```suggestion
    clifford_result = clifford_bkd.execute_circuit(circuit_copy, nshots=nshots)
```

tests/test_backends_clifford.py
```diff
@@ -240,11 +228,11 @@ def test_initial_state(backend):
 
 @pytest.mark.parametrize("seed", [10])
 def test_bitflip_noise(backend, seed):
-    rng = np.random.default_rng(seed)
+    backend.set_seed(seed)
     clifford_bkd = construct_clifford_backend(backend)
-    c = random_clifford(5, seed=rng, backend=backend)
+    c = random_clifford(5, seed=seed, backend=backend)
     c_copy = c.copy()
```
renatomello, line 234:
```suggestion
    circuit = random_clifford(5, seed=seed, backend=backend)
    circuit_copy = circuit.copy()
```

tests/test_backends_clifford.py
```diff
@@ -240,11 +228,11 @@ def test_initial_state(backend):
...
-    qubits = rng.choice(range(3), size=2, replace=False)
+    qubits = backend.np.random.choice(range(3), size=2, replace=False)
     c.add(gates.M(*qubits, p0=0.1, p1=0.5))
     c_copy.add(gates.M(*qubits, p0=0.1, p1=0.5))
     numpy_res = numpy_bkd.execute_circuit(c_copy)
```
renatomello, line 238:
```suggestion
    circuit.add(gates.M(*qubits, p0=0.1, p1=0.5))
    circuit_copy.add(gates.M(*qubits, p0=0.1, p1=0.5))
    numpy_res = numpy_bkd.execute_circuit(circuit_copy)
```

tests/test_backends_clifford.py
```diff
@@ -256,30 +244,32 @@ def test_bitflip_noise(backend, seed):
...
+    numpy_bkd.set_seed(seed)
     backend.set_seed(seed)
 
     clifford_bkd = construct_clifford_backend(backend)
     clifford_bkd.set_seed(seed)
 
     nqubits = 3
 
-    c = random_clifford(nqubits, density_matrix=True, seed=seed, backend=backend)
+    c = random_clifford(nqubits, density_matrix=False, seed=seed, backend=backend)
```
renatomello, line 255:
```suggestion
    circuit = random_clifford(nqubits, density_matrix=False, seed=seed, backend=backend)
```

tests/test_backends_clifford.py
```diff
@@ -256,30 +244,32 @@ def test_bitflip_noise(backend, seed):
...
 
     noise = NoiseModel()
-    noisy_gates = np.random.choice(c.queue, size=1, replace=False)
+    noisy_gates = [c.queue[0]]
```
renatomello, line 258:
```suggestion
    noisy_gates = [circuit.queue[0]]
```

tests/test_backends_clifford.py
```diff
@@ -256,30 +244,32 @@ def test_bitflip_noise(backend, seed):
...
 
-    c.add(gates.M(*range(nqubits)))
     c_copy = c.copy()
+    non_noisy = c.copy()
+    non_noisy.add(gates.M(*range(nqubits)))
+    c.add(gates.M(*range(nqubits)))
+    c_copy.add(gates.M(*range(nqubits)))
 
     c = noise.apply(c)
     c_copy = noise.apply(c_copy)
```
renatomello, line 269:
```suggestion
    circuit_copy = circuit.copy()
    non_noisy = circuit.copy()
    non_noisy.add(gates.M(*range(nqubits)))
    circuit.add(gates.M(*range(nqubits)))
    circuit_copy.add(gates.M(*range(nqubits)))

    circuit = noise.apply(circuit)
    circuit_copy = noise.apply(circuit_copy)
```

tests/test_backends_clifford.py
```diff
@@ -256,30 +244,32 @@ def test_bitflip_noise(backend, seed):
...
 
-    numpy_bkd.set_seed(2024)
     numpy_result = numpy_bkd.execute_circuit(c)
     clifford_result = clifford_bkd.execute_circuit(c_copy)
```
renatomello, line 272:
```suggestion
    numpy_result = numpy_bkd.execute_circuit(circuit)
    clifford_result = clifford_bkd.execute_circuit(circuit_copy)
```

tests/test_quantum_info_clifford.py
```diff
@@ -289,7 +289,7 @@ def test_clifford_samples_frequencies(backend, binary):
     clifford_backend = construct_clifford_backend(backend)
     if not clifford_backend:
         return
-    c = random_clifford(5)
+    c = random_clifford(5, backend=backend)
     c.add(gates.M(3, register_name="3"))
     c.add(gates.M(0, 1, register_name="01"))
     obj = Clifford.from_circuit(c, nshots=50, engine=_get_engine_name(backend))
```
renatomello, line 295:
```suggestion
    circuit = random_clifford(5, backend=backend)
    circuit.add(gates.M(3, register_name="3"))
    circuit.add(gates.M(0, 1, register_name="01"))
    obj = Clifford.from_circuit(circuit, nshots=50, engine=_get_engine_name(backend))
```

tests/test_quantum_info_random.py
```diff
@@ -62,51 +62,39 @@ def test_uniform_sampling_U3(backend, seed):
...
-        dims = 0
-        random_hermitian(dims, backend=backend)
-    with pytest.raises(TypeError):
-        dims = 2
-        random_hermitian(dims, seed=0.1, backend=backend)
+
+    if backend.platform != "tensorflow":
+        global PRECISION_TOL
+    else:
+        PRECISION_TOL = 4e-7
```
renatomello, line 97: I hate `tensorflow`so much

src/qibo/quantum_info/random_ensembles.py
```diff
@@ -215,35 +198,15 @@ def random_unitary(dims: int, measure: Optional[str] = None, seed=None, backend=
...
-        from scipy.linalg import expm
-
-        H = random_hermitian(dims, seed=seed, backend=NumpyBackend())
-        unitary = expm(-1.0j * H / 2)
-        unitary = backend.cast(unitary, dtype=unitary.dtype)
-
-    return unitary
+        return backend.qinfo._random_unitary(dims)
+    else:
+        raise_error(ValueError, f"measure {measure} not implemented.")
```
renatomello, line 209: Function is not ending in a return, which is not recommended

src/qibo/models/error_mitigation.py
renatomello: Unless I~m missing something, I do not think that the seed of a specific function should alter any type of global seed.

BrunoLiegiBastonLiegi: Both ``numba`` and ``cupy`` were having problems, ``pytorch`` I am not even sure has something like that. Thus either we give up on specifying the seed of the random ensemble functions, or we drop the customization for those functions.

src/qibo/quantum_info/random_ensembles.py
BrunoLiegiBastonLiegi: What do you mean exactly? It is raising an error I can't return anything after the error is raised.

src/qibo/models/error_mitigation.py
renatomello: `cupy` has an equivalent implementation of the [`numpy.random.default_rng`](https://docs.cupy.dev/en/stable/reference/generated/cupy.random.default_rng.html#cupy.random.default_rng) class. Idk what errors were occurring but in principle, it should work. `torch` has implementations of [rng Generators](https://pytorch.org/docs/stable/generated/torch.Generator.html). I remember you said `numba` didn't handle well rngs with parallelization. If it can't be done, it is best to just use the `numpy` implementation of the rng and cast to the backend than to reset the global seed. The functions not having a seed argument is not an option.

BrunoLiegiBastonLiegi: Yes ``cupy`` has it, but the support is limited, if you compare to ``numpy``'s [generator](https://numpy.org/doc/stable/reference/random/generator.html) you quickly notice that only a fraction of the methods is present. The ``pytorch`` generator object you linked, first does not support a seed to be passed, but rather I presume it makes use of the global seed, second doesn't seem to provide sampling functions as for ``numpy`` (``Generator.distribution()``). Casting doesn't solve the problem for ``numba`` (which btw still uses ``numpy`` arrays), you precisely need to wrap the ``numpy`` random sampling in a ``@numba.njit`` decorated function otherwise you won't see any speedup.
If the backend seed to be reset by a call of a function is a problem, we could always back up the original seed and reset it afterwards. Even though to me it doesn't seem like a huge problem to reset the seed globally, as I don't see particular reasons to use different seeds for different function calls inside the same execution.

renatomello: It's not about the seed of the backend, but rather about the current state of the rng object. Backing up the original seed and resetting the rng object brings it back to the beginning of the sequence of random numbers instead of continuing from where it had stopped. Maybe it is possible to copy the state of the rng object instead of just the seed?

BrunoLiegiBastonLiegi: But there's no rng object anymore, apart from the global internal rng I presume each engine uses to draw samples upon a ``engine.random.distribution()`` call. Indeed ``seed`` is now intended to only be an integer, which will be set as the internal seed of the backend, and thus, of its engine (e.g. via ``np.random.seed(seed)``).

src/qibo/quantum_info/basis.py
```diff
@@ -1,12 +1,25 @@
...
+    pauli_labels = {"I": backend.matrices.I()}
+    pauli_labels.update(
+        {label: getattr(backend.matrices, label) for label in ("X", "Y", "Z")}
+    )
+    return [pauli_labels[label] for label in order]
+
+
+@cache
+def _normalization(nqubits: int):
+    return np.sqrt(2**nqubits)
```
renatomello, line 22:
```suggestion
    return float(np.sqrt(2**nqubits))
```

Maybe it's best to force the `python` `float` type (or doing that by directly using `math.sqrt`) than using `numpy.float64`.

src/qibo/quantum_info/basis.py
```diff
@@ -259,27 +221,18 @@ def pauli_to_comp_basis(
     """
     backend = _check_backend(backend)
 
-    unitary = pauli_basis(
+    if sparse:
+        normalization = _normalization(nqubits) if normalize else 1.0
+        return getattr(backend.qinfo, f"_pauli_to_comp_basis_sparse_{order}")(
+            nqubits, *_get_paulis(pauli_order, backend), normalization=normalization
+        )
```
renatomello, line 228:
```suggestion
        func = getattr(backend.qinfo, f"_pauli_to_comp_basis_sparse_{order}")
        return func(nqubits, *_get_paulis(pauli_order, backend), normalization=normalization)
```

tests/utils.py
```diff
@@ -11,9 +11,19 @@ def random_sparse_matrix(backend, n, sparse_type=None):
...
+        im = sparse.rand(n, n, format=sparse_type, dtype=complex)
+        try:
+            matrix = re + 1j * im
+            break
+        except ValueError:
+            n_tries += 1
+            continue
+    if n_tries == 100:
+        raise RuntimeError("Maximum number of tries reached for random sparse matrix.")
+    return matrix
```
renatomello, line 26: Why do we need this?

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -125,40 +112,26 @@ def unvectorization(state, order: str = "row", backend=None):
...
-        nqubits = int(np.log2(dim))
-        axes_old = list(np.arange(0, 2 * nqubits))
-        state = backend.np.reshape(state, [2] * 2 * nqubits)
-        state = backend.np.transpose(state, axes_old[1::2] + axes_old[0::2])
-        state = backend.np.reshape(state, [2**nqubits] * 2)
-
+    dim = int(np.sqrt(state.shape[-1]))
+    if len(state.shape) == 1:
+        state = state.reshape(1, -1)
+    state = getattr(backend.qinfo, f"_unvectorization_{order}")(state, dim)
```
renatomello, line 132:
```suggestion
    func = getattr(backend.qinfo, f"_unvectorization_{order}")
    state = func(state, dim)
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -11,6 +11,7 @@
 from qibo.gates.abstract import Gate
 from qibo.gates.gates import Unitary
 from qibo.gates.special import FusedGate
+from qibo.quantum_info.basis import _get_paulis, _normalization
 from qibo.quantum_info.linalg_operations import singular_value_decomposition
```
renatomello, line 15: Maybe these two functions could be moved to `quantum_info.quantum_info.py`.

```suggestion
from qibo.quantum_info.basis import _get_paulis, _normalization
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -125,40 +112,26 @@ def unvectorization(state, order: str = "row", backend=None):
...
         )
-    else:
-        if order not in ["row", "column", "system"]:
-            raise_error(
-                ValueError,
-                f"order must be either 'row' or 'column' or 'system', but it is {order}.",
-            )
-
     backend = _check_backend(backend)
     state = backend.cast(state)
```
renatomello, line 127:
```suggestion
    state = backend.cast(state, dtype=state.dtype)
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -186,10 +159,7 @@ def to_choi(channel, order: str = "row", backend=None):
     """
     backend = _check_backend(backend)
 
-    channel = vectorization(channel, order=order, backend=backend)
-    channel = backend.np.outer(channel, backend.np.conj(channel))
-
-    return channel
+    return getattr(backend.qinfo, f"_to_choi_{order}")(channel)
```
renatomello, line 162:
```suggestion
    func_order = getattr(backend.qinfo, f"_to_choi_{order}")
    return func_order(channel)
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -212,10 +182,7 @@ def to_liouville(channel, order: str = "row", backend=None):
     """
     backend = _check_backend(backend)
 
-    channel = to_choi(channel, order=order, backend=backend)
-    channel = _reshuffling(channel, order=order, backend=backend)
-
-    return channel
+    return getattr(backend.qinfo, f"_to_liouville_{order}")(channel)
```
renatomello, line 185:
```suggestion
    func_order = getattr(backend.qinfo, f"_to_liouville_{order}")
    return func_order(channel)
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -247,22 +214,13 @@ def to_pauli_liouville(
...
 
-    nqubits = int(np.log2(channel.shape[0]))
-
-    channel = to_liouville(channel, order=order, backend=backend)
-
-    unitary = comp_basis_to_pauli(
-        nqubits, normalize, pauli_order=pauli_order, backend=backend
+    normalization = _normalization(int(np.log2(channel.shape[0]))) if normalize else 1.0
+    return getattr(backend.qinfo, f"_to_pauli_liouville_{order}")(
+        channel, *_get_paulis(pauli_order, backend), normalization=normalization
```
renatomello, line 221:
```suggestion
    normalization = _normalization(int(np.log2(channel.shape[0]))) if normalize else 1.0
    func_order = getattr(backend.qinfo, f"_to_pauli_liouville_{order}")
    return func_order(
        channel, *_get_paulis(pauli_order, backend), normalization=normalization
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -490,28 +448,9 @@ def choi_to_kraus(
...
         precision_tol = PRECISION_TOL
 
-    if not isinstance(validate_cp, bool):
-        raise_error(
-            TypeError,
-            f"validate_cp must be type bool, but it is type {type(validate_cp)}.",
-        )
-
     backend = _check_backend(backend)
     choi_super_op = backend.cast(choi_super_op)
```
renatomello, line 455:
```suggestion
    choi_super_op = backend.cast(choi_super_op, dtype=choi_super_op.dtype)
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -541,37 +480,14 @@ def choi_to_kraus(
...
-                kraus_ops.append(
-                    eig * unvectorization(kraus, order=order, backend=backend)
-                )
-                coefficients.append(eig)
-
-    kraus_ops = backend.cast(kraus_ops)
-    coefficients = backend.cast(coefficients)
+        kraus_ops, coefficients = getattr(backend.qinfo, f"_choi_to_kraus_cp_{order}")(
+            eigenvalues, eigenvectors, precision_tol
+        )
```
renatomello, line 490:
```suggestion
        func_order =  getattr(backend.qinfo, f"_choi_to_kraus_cp_{order}")
        kraus_ops, coefficients = func_order(eigenvalues, eigenvectors, precision_tol)
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -727,19 +643,15 @@ def kraus_to_choi(kraus_ops, order: str = "row", backend=None):
...
         kraus_op.append(gate)
-        kraus_op = kraus_op.matrix(backend)
-        kraus_op = vectorization(kraus_op, order=order, backend=backend)
-        super_op = super_op + backend.np.outer(kraus_op, backend.np.conj(kraus_op))
+        kraus_ops.append(kraus_op.matrix(backend)[None, :])
         del kraus_op
-
-    return super_op
+    kraus_ops = backend.np.vstack(kraus_ops)
+    return getattr(backend.qinfo, f"_kraus_to_choi_{order}")(kraus_ops)
```
renatomello, line 654:
```suggestion
    kraus_ops = backend.np.vstack(kraus_ops)
    func_order = getattr(backend.qinfo, f"_kraus_to_choi_{order}")
    return func_order(kraus_ops)
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -541,37 +480,14 @@ def choi_to_kraus(
         warnings.warn("Input choi_super_op is a non-completely positive map.")
 
         # using singular value decomposition because choi_super_op is non-CP
-        U, coefficients, V = singular_value_decomposition(
-            choi_super_op, backend=backend
+        kraus_ops, coefficients = getattr(backend.qinfo, f"_choi_to_kraus_{order}")(
+            choi_super_op
         )
```
renatomello, line 485:
```suggestion
        func_order = getattr(backend.qinfo, f"_choi_to_kraus_{order}")
        kraus_ops, coefficients = func_order(choi_super_op)
```

src/qibo/quantum_info/random_ensembles.py
```diff
@@ -6,10 +6,16 @@
 from typing import Optional, Union
 
 import numpy as np
+from numpy.random import permutation
 from scipy.stats import rv_continuous
 
 from qibo import Circuit, gates, matrices
```
renatomello, line 12:
```suggestion
from qibo import Circuit, gates
```

src/qibo/quantum_info/random_ensembles.py
```diff
@@ -316,42 +280,34 @@ def random_quantum_channel(
...
         super_op = _super_op_from_bcsz_measure(
-            dims=dims, rank=rank, order=order, seed=local_state, backend=backend
+            dims=dims, rank=rank, order=order, seed=seed, backend=backend
+        )
+    elif measure == "haar":
+        super_op = getattr(backend.qinfo, f"_super_op_from_haar_measure_{order}")(dims)
+    elif measure is None:
+        super_op = getattr(backend.qinfo, f"_super_op_from_hermitian_measure_{order}")(
+            dims
         )
```
renatomello, line 310:
```suggestion
    elif measure == "haar":
        func_order = getattr(backend.qinfo, f"_super_op_from_haar_measure_{order}")
        super_op = func_order(dims)
    elif measure is None:
        func_order = getattr(backend.qinfo, f"_super_op_from_hermitian_measure_{order}")
        super_op = func_order(dims)
```

src/qibo/quantum_info/quantum_info.py
```diff
@@ -0,0 +1,393 @@
+import numpy as np
+from numpy import ndarray
+from scipy.linalg import expm
+
+ENGINE = np
```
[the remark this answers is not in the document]
renatomello: @BrunoLiegiBastonLiegi

tests/utils.py
BrunoLiegiBastonLiegi: Sometimes this was failing randomly, thus I decided to try several times.

src/qibo/quantum_info/superoperator_transformations.py
BrunoLiegiBastonLiegi: They are shared by all the backends, in the sense that we do not need to customize them for each backend. Thus, I would prefer to leave them out of ``quantum_info.quantum_info`` which contains the backend specific customizations. Maybe I can move them to utils, in case.

renatomello: I'm fine with `utils`

src/qibo/models/error_mitigation.py
BrunoLiegiBastonLiegi: I was thinking again about this one. Maybe, I missed the point the other time, the problem is that every time you pass the seed, it is reset? Thus for instance if you do:
```
res1 = some_function(..., seed=1)
res2 = some_function(..., seed=1)
```
you obtain the same results ``res1==res2`` whereas with the local state object you would have obtained two different results with subsequent calls. One thing that we maybe can do here to replicate the same behaviour, is to do nothing if the ``seed=None`` or it is equal to the current seed.

renatomello: In this example, the results have to be the same. The problem I was raising was more in another direction.

For instance, a backend is created. This sets a seed to start. Then, under the hood, some stochastic processes happen (e.g. circuit measurement / sampling) and consequently random numbers are generated. Then, outside of the backend we call `some_function(..., seed=1). The implementation in this branch will reset the backend seed to `1`. This breaks the sequence of random numbers that were being generated before. Even if you re-reset the backend with the original seed, now the sequence of random numbers start from the beginning instead of starting from they previously were.

This may affect reproducibility.

BrunoLiegiBastonLiegi: Mmmmh okay, but then why would you want to use two different seeds: one for the backend operations (shots sampling, ecc...) and one for the single function? Can't you just set the seed globally at the beginning or later when you run the function?

tests/utils.py
renatomello: But where are random sparse matrices being used and why?

src/qibo/quantum_info/superoperator_transformations.py
renatomello: Now that we will have `float64` and `float32` dtypes, not explicitly specifying the dtype may lead to different issues e.g. trying to cast a complex tensor as a float dtype.

src/qibo/quantum_info/superoperator_transformations.py
renatomello: same answer as above

tests/utils.py
BrunoLiegiBastonLiegi: This was used by some tests in ``test_hamiltonians.py``, concerning the why, I guess they needed some random sparse matrix...

src/qibo/quantum_info/superoperator_transformations.py
BrunoLiegiBastonLiegi: But it is still not merged, is it changing the default behaviour ``dtype=self.dtype`` with ``self.dtype`` always being ``complex128/64``?

renatomello: Yes, the backend's dtype might be `float32/64`

BrunoLiegiBastonLiegi: Mmmh, okay but does ``backend.cast`` default to? It would make sense to default to the complex dtype as we are dealing with quantum objects most of the time. Regardless, the proposed solution is not viable, as I pointed out above, it will fail for sure with ``pytorch``, but in general is going to probably fail everytime the input ``state`` and the backend don't match.

renatomello: > Mmmh, okay but what does `backend.cast` default to? It would make sense to default to the complex dtype as we are dealing with quantum objects most of the time. Regardless, the proposed solution is not viable, as I pointed out above, it will fail for sure with `pytorch`, but in general is going to probably fail everytime the input `state` and the backend don't match.

Right now, `cast` defaults to the current backend's `dtype`. The default backend `dtype` is `complex128`. However, the following
```python
from qibo import set_backend

set_backend("numpy", dtype="float64")

x = ...
x = backend.cast(x)
```
would try to cast to ``float64``.

AlejandroSopena (commented):
Thanks @BrunoLiegiBastonLiegi for all the work. It looks great.

The tests `test_with_unitaries` and `test_with_comb` in [test_quantum_info_quantum_networks.py](https://github.com/qiboteam/qibo/blob/qinfo_jit/tests/test_quantum_info_quantum_networks.py) are failing sometimes with `cupy` and `cuquantum` locally.

Some more comments below:

renatomello (approved):

BrunoLiegiBastonLiegi:
Ok this is now synced with main and all the tests are passing. The only tricky point is how to handle the seed, which I am open to discuss.

src/qibo/backends/_clifford_operations.py
```diff
@@ -357,6 +357,9 @@ def _exponent(
     Returns:
         (np.array): The calculated exponents.
```
renatomello:
```suggestion
        ndarray: The calculated exponents.
```

src/qibo/backends/_clifford_operations.py
```diff
@@ -372,6 +375,7 @@ def _rowsum(symplectic_matrix, h, i, nqubits, determined=False):
     Returns:
         (np.array): The updated symplectic matrix.
```
renatomello:
```suggestion
        ndarray: The updated symplectic matrix.
```

src/qibo/quantum_info/utils.py
```diff
@@ -435,15 +450,25 @@ def haar_integral(
...
         )
-        for _ in range(samples):
-            haar_state = backend.np.reshape(
-                random_statevector(dim, backend=backend), (-1, 1)
-            )
 
-            rho = haar_state @ backend.np.conj(haar_state).T
+        random_states = backend.qinfo.ENGINE.random.standard_normal(
+            (samples, dim)
+        ).astype(rand_unit_density.dtype)
```
renatomello:
```suggestion
        random_states = backend.qinfo.ENGINE.random.standard_normal(
            (samples, dim)
        )
        random_states = backend.cast(random_states, dtype=rand_unit_density.dtype)
```

Otherwise, this will create issues for some backends, especially the `qiboml` ones.

src/qibo/quantum_info/utils.py
```diff
@@ -1,18 +1,31 @@
...
 import numpy as np
 
-from qibo import matrices
-from qibo.backends import _check_backend
+from qibo.backends import Backend, _check_backend
 from qibo.config import PRECISION_TOL, raise_error
 
 
+@cache
+def _get_paulis(order: str, backend: Backend):
```
renatomello:
```suggestion
def _get_single_paulis(order: str, backend: Backend):
```

Or maybe `_get_paulis_single`. One or the other. Just to be very clear that it's single-qubit Paulis.

src/qibo/quantum_info/utils.py
```diff
@@ -1,18 +1,31 @@
...
+    pauli_labels = {"I": backend.matrices.I()}
+    pauli_labels.update(
+        {label: getattr(backend.matrices, label) for label in ("X", "Y", "Z")}
+    )
+    return [pauli_labels[label] for label in order]
+
+
+@cache
+def _normalization(nqubits: int):
```
renatomello:
```suggestion
def _pauli_basis_normalization(nqubits: int):
```
Or `_pauli_norm` or any variant that makes explicit that it is the normalization of the Pauli basis.

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -12,6 +12,7 @@
 from qibo.gates.gates import Unitary
 from qibo.gates.special import FusedGate
 from qibo.quantum_info.linalg_operations import singular_value_decomposition
+from qibo.quantum_info.utils import _get_paulis, _normalization
```
renatomello: propagate changes

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -125,40 +112,27 @@ def unvectorization(state, order: str = "row", backend=None):
...
-        )
-    else:
-        nqubits = int(np.log2(dim))
-        axes_old = list(np.arange(0, 2 * nqubits))
-        state = backend.np.reshape(state, [2] * 2 * nqubits)
-        state = backend.np.transpose(state, axes_old[1::2] + axes_old[0::2])
-        state = backend.np.reshape(state, [2**nqubits] * 2)
+    state = backend.cast(state, dtype=state.dtype)
 
+    dim = int(np.sqrt(state.shape[-1]))
```
renatomello, line 129: This should be the same as the now cached normalization

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -490,28 +452,9 @@ def choi_to_kraus(
...
         precision_tol = PRECISION_TOL
 
-    if not isinstance(validate_cp, bool):
-        raise_error(
-            TypeError,
-            f"validate_cp must be type bool, but it is type {type(validate_cp)}.",
-        )
-
     backend = _check_backend(backend)
     choi_super_op = backend.cast(choi_super_op)
```
renatomello:
```suggestion
    choi_super_op = backend.cast(choi_super_op, dtype=choi_super_op.dtype)
```

src/qibo/quantum_info/random_ensembles.py
```diff
@@ -5,10 +5,14 @@
 from typing import Optional, Union
 
 import numpy as np
+from numpy.random import permutation
```
renatomello:
```suggestion
```
Not being used.

src/qibo/quantum_info/random_ensembles.py
```diff
@@ -518,71 +455,37 @@ def random_density_matrix(
...
-        raise_error(
-            TypeError, f"metric must be type str, but it is type {type(metric)}."
-        )
     if metric not in ["hilbert-schmidt", "ginibre", "bures"]:
         raise_error(ValueError, f"metric {metric} not implemented.")
 
-    if basis is not None and not isinstance(basis, str):
-        raise_error(TypeError, f"basis must be type str, but it is type {type(basis)}.")
-    elif basis is not None and basis not in ["pauli"]:
+    if basis is not None and basis not in ["pauli"]:
```
renatomello:
```suggestion
    if basis is not None and basis != "pauli":
```

src/qibo/quantum_info/random_ensembles.py
```diff
@@ -518,71 +455,37 @@ def random_density_matrix(
...
             or set(basis.split("-")[1]) != {"I", "X", "Y", "Z"}
         ):
             raise_error(ValueError, f"basis {basis} nor recognized.")
 
-    if not isinstance(normalize, bool):
-        raise_error(
-            TypeError, f"normalize must be type bool, but it is type {type(normalize)}."
-        )
-    elif normalize is True and basis is None:
+    if normalize is True and basis is None:
```
renatomello:
```suggestion
    if normalize and basis is None:
```

src/qibo/quantum_info/random_ensembles.py
```diff
@@ -518,71 +455,37 @@ def random_density_matrix(
...
-            state += random_unitary(dims, seed=local_state, backend=backend)
-            state = state @ random_gaussian_matrix(
-                dims, rank, seed=local_state, backend=backend
-            )
-            state = state @ backend.np.transpose(backend.np.conj(state), (1, 0))
-            state /= backend.np.trace(state)
+            state = backend.qinfo._random_density_matrix_bures(dims, dims, 0.0, 1.0)
 
-    state = backend.cast(state, dtype=state.dtype)
+    # state = backend.cast(state, dtype=state.dtype)
```
renatomello:
```suggestion
```

src/qibo/quantum_info/basis.py
```diff
@@ -1,12 +1,8 @@
...
-
-from qibo import matrices
-from qibo.backends import _check_backend
+from qibo.backends import Backend, _check_backend
 from qibo.config import raise_error
-from qibo.quantum_info.superoperator_transformations import vectorization
+from qibo.quantum_info.utils import _get_paulis, _normalization
 
 
 def pauli_basis(
```
renatomello, line 8: I've been suggested more than once by people who use these functions to set the normalized Pauli basis as True by default. Could you take advantage of this opportunity and do that on this PR? Also in the other functions that deal with Pauli basis, _e.g._ in `superoperator_tranformations.py`.

src/qibo/quantum_info/_quantum_info.py
```diff
@@ -0,0 +1,459 @@
...
+def _pauli_basis(
+    nqubits: int,
+    pauli_0: ndarray,
+    pauli_1: ndarray,
+    pauli_2: ndarray,
+    pauli_3: ndarray,
+    normalization: float = 1.0,
+) -> ndarray:
+    basis_single = ENGINE.vstack((pauli_0, pauli_1, pauli_2, pauli_3)).reshape(4, 2, 2)
+    dim = 2**nqubits
```
renatomello, line 17: cached normalization squared?

src/qibo/quantum_info/_quantum_info.py
```diff
@@ -0,0 +1,459 @@
...
+    axes_old = list(range(1, 2 * nqubits + 1))
+    state = ENGINE.reshape(state, (state.shape[0],) + (2,) * 2 * nqubits)
+    state = ENGINE.transpose(
+        state,
+        [
+            0,
+        ]
+        + axes_old[1::2]
+        + axes_old[0::2],
+    )
```
renatomello:
```suggestion
    state = ENGINE.transpose(state, [0] + axes_old[1::2] + axes_old[0::2])
```

src/qibo/quantum_info/superoperator_transformations.py
BrunoLiegiBastonLiegi: the problem is that I don't have (direct) access to the nqubits here, I would need to first extrapolate it through a log and then call the `_normalization`, at that point it's probably better to just stick with the sqrt.

src/qibo/quantum_info/_quantum_info.py
BrunoLiegiBastonLiegi: well yes I could I guess, not sure whether it is worth it though, honestly I don't even remember why I created the `_normalization` function in the first place...

src/qibo/models/error_mitigation.py
```diff
@@ -1015,16 +1026,15 @@ def error_sensitive_circuit(circuit, observable, seed=None, backend=None):
     """
     from qibo import gates
 
-    backend_temp, local_state = _check_backend_and_local_state(seed, backend)
+    backend_temp = _check_backend(backend)
     backend = (
-        CliffordBackend(engine=_get_engine_name(backend_temp))
+        CLIFFORD_BACKEND()  # CliffordBackend(engine=_get_engine_name(backend_temp))
```
AlejandroSopena:
```suggestion
        CLIFFORD_BACKEND()
```

AlejandroSopena (approved):

PR #1591, opened 2025-02-17
Add `sparse_encoder` to `models.encodings`

renatomello:
This is based on Section IV of https://arxiv.org/abs/2405.20408

Depends on https://github.com/qiboteam/qibojit/pull/220

Checklist:
- [x] Tests are passing.


src/qibo/gates/gates.py
```diff
@@ -2538,6 +2538,7 @@ def __init__(
         self.target_qubits = tuple(qubits_in) + tuple(qubits_out)
         self.unitary = True
 
+        self.parameter_names = "theta"
```
AlejandroSopena, line 2541:
```suggestion
```

src/qibo/models/encodings.py
```diff
@@ -364,7 +506,8 @@ def hamming_weight_encoder(
         *Quantum encoder for fixed Hamming-weight subspaces*
         `arXiv:2405.20408 [quant-ph] <https://arxiv.org/abs/2405.20408>`_.
     """
-    complex_data = bool(data.dtype in [complex, np.dtype("complex128")])
+    # complex_data = bool(data.dtype in [complex, np.dtype("complex128")])
```
AlejandroSopena, line 509:
```suggestion
```

AlejandroSopena (approved):
GPU tests are not passing but they will be fixed in #1622

src/qibo/models/encodings.py
```diff
@@ -1186,3 +1330,103 @@ def _intermediate_gate(
...
+            bitstring = f"{bitstring:0{nqubits}b}"
+        bitstrings.append(bitstring)
+        _data.append(elem)
+    _data = list(zip(bitstrings, _data))
+    _data = sorted(_data, key=lambda x: hamming_weight(x[0]))
+
+    data_sorted, bitstrings_sorted = [], []
+    for string, elem in _data:
+        bitstrings_sorted.append(np.array(list(string)).astype(int))
+        data_sorted.append(elem)
```
BrunoLiegiBastonLiegi, line 1352:
```suggestion
    bitstring_sorted, data_sorted = zip(*_data)
    bitstring_sorted = np.array(list(bistring_sorted)).astype(int)
```

src/qibo/models/encodings.py
```diff
@@ -1186,3 +1330,103 @@ def _intermediate_gate(
...
+        gate = (
+            gates.U3(qubit, 2 * theta, 2 * phi, 0.0).controlled_by(*controls)
+            if complex_data
+            else gates.RY(qubit, 2 * theta).controlled_by(*controls)
+        )
+    elif distance == 2 and hw_0 == hw_1:
+        qubits = [np.where(difference == -1)[0][0], np.where(difference == 1)[0][0]]
+        for qubit in qubits:
+            if qubit not in touched_qubits:
+                touched_qubits.append(qubit)
```
BrunoLiegiBastonLiegi, line 1373:
```suggestion
        touched_qubits += list(set(qubits) - set(touched_qubits))
```

src/qibo/models/encodings.py
```diff
@@ -1186,3 +1330,103 @@ def _intermediate_gate(
...
+        qubits = [np.where(difference == -1)[0], np.where(difference == 1)[0]]
+        for row in qubits:
+            for qubit in row:
+                if qubit not in touched_qubits:
+                    touched_qubits.append(qubit)
+        gate = gates.GeneralizedRBS(qubits[0], qubits[1], theta, -phi).controlled_by(
+            *controls
+        )
+
+    return gate, touched_qubits
```
BrunoLiegiBastonLiegi, line 1392:
```suggestion
    return gate
```
In principle you wouldn't need to return ``touched_qubits`` as it is a list passed as input that is grown by the function. I would either copy the input list and modify and return the copy, or grow directly the input list and not return it for better clarity.

renatomello:
@BrunoLiegiBastonLiegi in #1622 I'm making all the functions in `models.encodings` backend-aware, which, among other things, fixes the tests on GPU that are failing here.

BrunoLiegiBastonLiegi:
Okay, let's wait for that one to be merged then.

renatomello:
> Okay, let's wait for that one to be merged then. Or rebase this one.

That branch was forked from this branch, so that one already has the `sparse_encoder` there.

BrunoLiegiBastonLiegi:
Ah ok, so that one should be rebased to this one, I guess. I would prefer not to merge something that is failing in main, this will make all the other PR fail.

renatomello:
> @BrunoLiegiBastonLiegi @AlejandroSopena all tests in `models.encodings` are passing for GPUs now. I'm not going to fix all the other tests unrelated to that module in this PR to not make it even more bloated than it already is.

I lied. It's fixed now.

src/qibo/models/encodings.py
```diff
@@ -184,38 +186,53 @@ def sparse_encoder(data, nqubits: int = None, **kwargs):
...
+    data_sorted, bitstrings_sorted = _sort_data_sparse(data, nqubits, backend)
     # calculate phases
-    _data_sorted = np.abs(data_sorted) if complex_data else data_sorted
-    thetas = _generate_rbs_angles(_data_sorted, architecture="diagonal")
-    phis = np.zeros(len(thetas) + 1)
+    _data_sorted = backend.np.abs(data_sorted) if complex_data else data_sorted
+    thetas = _generate_rbs_angles(
+        _data_sorted, architecture="diagonal", backend=backend
+    )
+    phis = backend.np.zeros(len(thetas) + 1)
```
BrunoLiegiBastonLiegi, line 203: This may be failing with ``CupyBackend`` as ``backend.np`` is ``numpy`` and not ``cupy``. Or better, it may not fail but it most likely won't use ``cupy`` arrays, and thus GPU.

src/qibo/models/encodings.py
```diff
@@ -184,38 +186,53 @@ def sparse_encoder(data, nqubits: int = None, **kwargs):
...
     # marking qubits that have suffered the action of a gate
     touched_qubits = []
-    circuit = comp_basis_encoder(list(bitstrings_sorted[0]), nqubits=nqubits, **kwargs)
+    circuit = comp_basis_encoder(
+        [int(bit) for bit in bitstrings_sorted[0]], nqubits=nqubits, **kwargs
+    )
     for qubit, bit in enumerate(bitstrings_sorted[0]):
         if bit == 1:
-            touched_qubits.append(qubit)
+            touched_qubits.append(int(qubit))
```
BrunoLiegiBastonLiegi, line 219: Do you actually need the explicit cast to int, enumerate should produce an int in any case I believe. In any case, if ``bitstring_sorted`` is a ``int8`` np array you could probably just do
```python
touched_qubits = (bitstrings_sorted[0] == 1).nonzero()[0]
```

src/qibo/models/encodings.py
```diff
@@ -184,38 +186,53 @@ def sparse_encoder(data, nqubits: int = None, **kwargs):
...
-        distance = hamming_distance("".join(b_1.astype(str)), "".join(b_0.astype(str)))
+        distance = hamming_distance(b_1, b_0)
         difference = b_1 - b_0
 
-        ones, new_ones = np.argsort(b_0)[-hw_0:], np.argsort(b_1)[-hw_1:]
+        ones, new_ones = (
+            list(backend.np.argsort(b_0)[-hw_0:]),
+            list(backend.np.argsort(b_1)[-hw_1:]),
+        )
+        ones, new_ones = {int(elem) for elem in ones}, {int(elem) for elem in new_ones}
```
BrunoLiegiBastonLiegi, line 232: Can't you just do
```suggestion
        ones, new_ones = set(ones), set(new_ones)
```
?

src/qibo/models/encodings.py
```diff
@@ -1261,37 +1288,40 @@ def _binary_encoder_hyperspherical(data, nqubits, complex_data: bool, **kwargs):
...
+            )
+    phis = backend.cast(phis, dtype=phis[0].dtype)
 
     angles = []
     for k, (theta, phi) in enumerate(zip(thetas, phis)):
         if k in indexes_to_double:
-            angle = [2 * theta, 2 * phi, 0.0] if complex_data else [2 * theta]
+            zero_casted = backend.cast(
+                0.0, dtype=backend.np.float64
+            )  # because of GPU backends
```
BrunoLiegiBastonLiegi, line 1309: This can probably be done outside of the loop though, right?

src/qibo/models/encodings.py
```diff
@@ -1261,37 +1288,40 @@ def _binary_encoder_hyperspherical(data, nqubits, complex_data: bool, **kwargs):
...
         angles.extend(angle)
 
     if complex_data:
         angles[-2] = 2 * _angle_mod_two_pi(
-            (np.angle(data[-1]) - np.angle(data[-2])) / 2
+            (backend.np.angle(data[-1]) - backend.np.angle(data[-2])) / 2
         )
         angles[-1] = 2 * _angle_mod_two_pi(
-            (-1 / 2) * (np.angle(data[-2]) + np.angle(data[-1])) + np.sum(phis[:-2])
+            (-1 / 2) * (backend.np.angle(data[-2]) + backend.np.angle(data[-1]))
```
BrunoLiegiBastonLiegi, line 1321:
```suggestion
            (-0.5) * (backend.np.angle(data[-2]) + backend.np.angle(data[-1]))
```

src/qibo/models/encodings.py
```diff
@@ -1332,48 +1362,67 @@ def _intermediate_gate(
...
+    phi,
+    backend=None,
 ):
+    backend = _check_backend(backend)
     if distance == 1:
-        qubit = np.where(difference == 1)[0][0]
+        qubit = int(backend.np.where(difference == 1)[0][0])
         if qubit not in touched_qubits:
-            touched_qubits.append(qubit)
+            touched_qubits.append(int(qubit))
```
BrunoLiegiBastonLiegi, line 1407:
```suggestion
            touched_qubits.append(qubit)
```
qubit should already be an int, as you just casted it

src/qibo/models/encodings.py
renatomello: This fix, in my opinion, is to be done outside of this PR. I have been thinking of refactoring this part of the backends as something like `backend.engine` or `backend.platform` (or even `backend.math`) and drop `backend.np` entirely from every backend.

src/qibo/models/encodings.py
renatomello: Yes, the cast to int needs to explicit because of the GPU backends.

src/qibo/models/encodings.py
renatomello: This doesn't work again because of GPU backends, without the explicit casting to `int`, the elements are unhashable.

src/qibo/models/encodings.py
BrunoLiegiBastonLiegi: Yeah yeah I agree, just pointing out here, if you want to simply patch it here you could just do a ``backend.cast`` even though it's not the cleanest solution...

src/qibo/models/encodings.py
renatomello: They need to be actual `int`s instead of `cp.int8` otherwise some other errors are raised down the line.

src/qibo/models/encodings.py
renatomello: True.

src/qibo/models/encodings.py
```diff
@@ -91,14 +91,8 @@ def phase_encoder(data, rotation: str = "RY", **kwargs):
...
     """
-    if isinstance(data, list):
-        data = np.array(data)
 
-    if len(data.shape) != 1:
-        raise_error(
-            TypeError,
-            f"``data`` must be a 1-dimensional array, but it has dimensions {data.shape}.",
-        )
+    backend = _check_backend(backend)
```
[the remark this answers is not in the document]
renatomello: Thanks for spotting it.

src/qibo/models/encodings.py
```diff
@@ -184,38 +180,53 @@ def sparse_encoder(data, nqubits: int = None, **kwargs):
     if isinstance(data[0][0], str) and nqubits is None:
         nqubits = len(data[0][0])
 
-    complex_data = bool("complex" in str(type(data[0][1])))
+    _data_test = data[0][1]
+    _data_test = (
+        _data_test.dtype if "array" in str(type(_data_test)) else type(_data_test)
+    )
```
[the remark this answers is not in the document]
renatomello: I didn't want to do that to not call `data[0][1]` three times in the same line

src/qibo/models/encodings.py
```diff
@@ -171,10 +165,18 @@ def sparse_encoder(data, nqubits: int = None, **kwargs):
...
+
     if isinstance(data, zip):
         data = list(data)
 
-    if "int" in str(type(data[0][0])) and nqubits is None:
+    # TODO: Fix this mess with qibo native dtypes
+    try:
+        type_test = bool("int" in str(data[0][0].dtype))
+    except AttributeError:
+        type_test = bool("int" in str(type(data[0][0])))
```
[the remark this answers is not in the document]
renatomello: Maybe that is a good issue for the `unitaryHACK` @alecandido @AlejandroSopena

src/qibo/models/encodings.py
BrunoLiegiBastonLiegi: but ``qubit`` should be an int already regardless of what ``bitstring_sorted`` is, as it is produced by ``enumerate``

src/qibo/models/encodings.py
BrunoLiegiBastonLiegi: but you already did the explicit cast here https://github.com/qiboteam/qibo/blob/db5c529106537cc5ce944de344134b0d4f64fe3c/src/qibo/models/encodings.py#L1405

you're doing:
``` python
qubit = int(...)
if qubit not in touched_qubits:
    # cast again, but qubit is already an int
    touched_qubits.append(int(qubit))
```

BrunoLiegiBastonLiegi (approved):
Thanks, just double check the two redundant casts to int, because they don't seem to be necessary to me.

PR #1594, opened 2025-02-26
Fix `gates.GeneralizedRBS.apply` method

renatomello:
This is needed in #1591 

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/gates/gates.py
+++ b/src/qibo/gates/gates.py
@@ -2533,8 +2533,11 @@ class GeneralizedRBS(ParametrizedGate):
         super().__init__(trainable)
         self.name = "grbs"
         self.draw_label = "gRBS"
-        self.target_qubits = tuple(qubits_in) + tuple(qubits_out)
         self.unitary = True
+
+        target_qubits = list(qubits_in) + list(qubits_out)
+        target_qubits.sort()
+        self.target_qubits = tuple(target_qubits)
 
         self.parameter_names = "theta"
         self.parameters = theta, phi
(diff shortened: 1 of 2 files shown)


src/qibo/gates/gates.py
```diff
@@ -2535,9 +2535,12 @@ def __init__(
         super().__init__(trainable)
         self.name = "grbs"
         self.draw_label = "gRBS"
-        self.target_qubits = tuple(qubits_in) + tuple(qubits_out)
         self.unitary = True
 
+        target_qubits = list(qubits_in) + list(qubits_out)
+        target_qubits.sort()
+        self.target_qubits = tuple(target_qubits)
```
stavros11, line 2542: Looking at this and following our previous discussion, I initially struggled to understand how it works, because here it seems like you are losing all information about the user-passed qubit order, for example `GeneralizedRBS([0, 1], [2])` will end up having the same `target_qubits` with `GeneralizedRBS([0, 2], [1])` which normally shouldn't be the case. However, after doing some doing some quick tests locally, it seemed that this fix works well and the "cheat" actually happens in https://github.com/qiboteam/qibo/blob/f1a7574577647a0e683b44372f6a501351b8e38e/src/qibo/backends/numpy.py#L132-L138
where you are retaining which are the `qubits_in` and `qubits_out`.

I am not sure if there is a cleaner solution, but I guess that should be fine for now.

stavros11 (approved):
Thanks @renatomello.

AlejandroSopena (approved):

PR #1599, opened 2025-03-03
Matrix representation of `gates.GeneralizedRBS` independent of qubit order

AlejandroSopena:
The  `gates.GeneralizedRBS` gate must be Hamming weight preserving when the number of input and output qubits is the same, but here we can see that it is not: 

```Python
from qibo import gates
from qibo.backends import NumpyBackend

backend = NumpyBackend()

gate = gates.GeneralizedRBS([1,2],[3,4],theta=0.123)

print(gate.matrix(backend))
```
It is mixing states $|0001\rangle$ and $|0110\rangle$ .

This PR address this issue. I construct the RBS gate without considering the order of the qubits passed as input. This is the same approach we follow for other multi-qubit gates. 

This simplifies the code and makes the cheat introduced in PR #1594 unnecessary.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/backends/npmatrices.py
+++ b/src/qibo/backends/npmatrices.py
@@ -491,21 +491,17 @@ class NumpyMatrices:
         )
 
     def GeneralizedRBS(self, qubits_in, qubits_out, theta, phi):
-        bitstring_length = len(qubits_in) + len(qubits_out)
-        integer_in = "".join(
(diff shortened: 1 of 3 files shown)


renatomello (approved):

stavros11 (approved):
Thanks @AlejandroSopena.

AlejandroSopena:
The GPU is fully occupied and the tests are cancelled but I have checked that they pass locally.

PR #1605, opened 2025-03-12
Implementing function to visualize states in UI module

MatteoRobbiati:
This PR introduces a plotscript in the UI module. 

---

The function is `qibo.ui.result_visualization.visualize_state`, and can be used to visualize circuit result's properties according to the chosen mode in the form of an histogram. 
It accepts `mode` values in `["amplitudes", "probabilities", "frequencies"]`.
Since the xticks population can become easily very crowded with more than 7/8 qubits, I thought it can be good to be able to restrict the number of ticks to the "most relevant components" of the state. This can be set by the user doing e.g.:

```python
import numpy as np

from qibo.ui.result_visualization import visualize_state

nqubits = 5
mode = "frequencies"

c = Circuit(nqubits)
for q in range(nqubits):
    c.add(gates.RY(q, theta=np.random.randn()))
    c.add(gates.RZ(q, theta=np.random.randn()))

# Execute your circuit
outcome = c(nshots=1000)

# Plot results
_, fig = visualize_state(
    execution_outcome=outcome,
    mode=mode,
    n_most_relevant_components=8,
)
```

And it will return the following result.

image
---

Regarding the amplitudes visualization, I opted for showing both Imag and Real parts with their signs (an example follows). 

image

Checklist:
- [ ] Tests are passing.


MatteoRobbiati:
> @MatteoRobbiati, I would like to include this feature in the next release planned this week, could you please merge main here and let me know when it is ready?

It is ready to be reviewed. I am fixing now the coverage, but apart from this we can proceed with the process.

scarrazza (approved):
Thanks @MatteoRobbiati looks pretty good to me!

PR #1615, opened 2025-03-20
Make ``backend.samples_to_binary`` use the backend's engine

BrunoLiegiBastonLiegi:
As per title, this makes this backend method use the backend engine rather than always ``numpy``. In particular, this is needed by qiboteam/qiboml#72, as running on GPU would fail otherwise.

Checklist:
- [ ] Tests are passing.


renatomello (approved):

MatteoRobbiati (approved):
LGTM.

PR #1616, opened 2025-03-20
``pytest --gpu-only`` fix + GPU backends shots sampling

BrunoLiegiBastonLiegi:
In #1604 it was found that some GPU tests are incorrectly skipped when running ``pytest --gpu-only``. This fixes that.

EDIT: on top of this, I found out that shots generation was done by numpy even with GPU backends, such as cupy, I am fixing that here as well.

Checklist:
- [x] Tests are passing.


BrunoLiegiBastonLiegi:
I also noticed that there is a backend method:

https://github.com/qiboteam/qibo/blob/d10caa050fa11ffb59fe34b813ecc99160df54fa/src/qibo/backends/numpy.py#L846C1-L863C14

that is there just to provide a fixed target for some tests in ``test_measurement_probabilstic.py``. I am skipping those tests with gpus, because that would require to provide the cupy backend with a similar method, but I am not sure it's worth spending time on this. Actually, I would advocate for removing this method from the backend completely, and in case we still wish to keep such a test, to have the fixed targets defined in the tests directly as we do in many other locations. What do you think @alecandido?

alecandido:
> Actually, I would advocate for removing this method from the backend completely, and in case we still wish to keep such a test, to have the fixed targets defined in the tests directly as we do in many other locations. What do you think @alecandido?

The function has been introduced 3y ago, and it seems fully dedicated to tests, used only in `test_measurements_probability.py`.

I believe the intention was to differentiate the values per backend. But there is no reason to distribute it as part of the package, since it is fully intended for testing.
The same can be obtained just defining these functions as part of the tests, and construct a mapping (easily implemented as a `dict`) from the relevant backends (which are to be tested in Qibo's CI) to the respective functions, all defined within the `tests/` folder itself.
If, at any time, a backend will be added, we can either require the respective function to be provided (failing the test using it), or instead we could make it non-mandatory, just skipping the test if the function is missing.

In any case, I fully agree that the function has to be dropped.

src/qibo/backends/numpy.py
```diff
@@ -91,6 +91,14 @@ def to_numpy(self, x):
     def compile(self, func):
         return func
 
+    def zeros(self, shape, dtype=None):
+        if dtype is None:
+            dtype = self.dtype
+        return self.np.zeros(shape, dtype=dtype)
+
+    def random_choice(self, a, **kwargs):
+        return self.np.random.choice(a, **kwargs)
```
renatomello, line 100: They have to be added to the abstract backend object too, right?

renatomello (approved):

BrunoLiegiBastonLiegi:
@alecandido  @AlejandroSopena @MatteoRobbiati can you quickly look into this, I would like to merge it soon, because qibojit is depending on this branch right now

https://github.com/qiboteam/qibojit/blob/63c700d1732030c7647f653a9c76f0f52f871a91/pyproject.toml#L26

which I'd like to fix asap.

MatteoRobbiati (approved):
LGTM!

PR #1618, opened 2025-03-20
Enhanced `gates.Unitary` draw with `matplotlib`

sergiomtzlosa:
Checklist:
- [ ] Tests are passing.

Now for the circuit below:

```
c = models.Circuit(6)

c.add(gates.Unitary(np.random.random((8, 8)), 0, 2, 3))
c.add(gates.Unitary(np.random.random((2, 2)), 2))
c.add(gates.Unitary(np.random.random((2, 2)), 4))

plot_circuit(c)
```
The output image is:

imagen


renatomello (approved):

renatomello:
Is there any way of distinguishing these unitaries from each other? For instance, the second and third unitaries in your example are different matrices but have the same symbol, so it might lead someone to think they are the same.

sergiomtzlosa:
> Is there any way of distinguishing these unitaries from each other? For instance, the second and third unitaries in your example are different matrices but have the same symbol, so it might lead someone to think they are the same.

This is the properties from the U gates you mention:

```
{'name': 'Unitary',
 'draw_label': 'U',
 'is_controlled_by': False,
 'init_args': [array([[0.59587452, 0.61818819, 0.41597643, 0.90393119, 0.06621149,
          0.72290538, 0.77908855, 0.83662002],
         [0.42430393, 0.61242686, 0.88895057, 0.88425884, 0.52700795,
          0.90525265, 0.32402481, 0.41618417],
         [0.45058784, 0.06722899, 0.80816688, 0.66207815, 0.52801049,
          0.68084687, 0.42146987, 0.54617903],
         [0.86071525, 0.80881038, 0.15703967, 0.45291621, 0.04587542,
          0.81400596, 0.54256902, 0.32047022],
         [0.20801591, 0.69555816, 0.57814184, 0.21689202, 0.85507719,
          0.99327673, 0.84065296, 0.67474639],
         [0.83281928, 0.30620483, 0.06071928, 0.83475402, 0.37167302,
          0.43024746, 0.5071005 , 0.47252334],
         [0.27345507, 0.89303954, 0.68043481, 0.03191491, 0.68112454,
          0.52691193, 0.50706933, 0.19426751],
         [0.30004403, 0.1706441 , 0.8723928 , 0.47353485, 0.96016128,
          0.45373964, 0.59661599, 0.29120244]]),
  0,
  2,
  3],
 'init_kwargs': {'name': None, 'check_unitary': True, 'trainable': True},
 'unitary': False,
 '_target_qubits': (0, 2, 3),
 '_control_qubits': (),
 '_parameters': (array([[0.59587452, 0.61818819, 0.41597643, 0.90393119, 0.06621149,
          0.72290538, 0.77908855, 0.83662002],
         [0.42430393, 0.61242686, 0.88895057, 0.88425884, 0.52700795,
          0.90525265, 0.32402481, 0.41618417],
         [0.45058784, 0.06722899, 0.80816688, 0.66207815, 0.52801049,
          0.68084687, 0.42146987, 0.54617903],
         [0.86071525, 0.80881038, 0.15703967, 0.45291621, 0.04587542,
          0.81400596, 0.54256902, 0.32047022],
         [0.20801591, 0.69555816, 0.57814184, 0.21689202, 0.85507719,
          0.99327673, 0.84065296, 0.67474639],
         [0.83281928, 0.30620483, 0.06071928, 0.83475402, 0.37167302,
          0.43024746, 0.5071005 , 0.47252334],
         [0.27345507, 0.89303954, 0.68043481, 0.03191491, 0.68112454,
          0.52691193, 0.50706933, 0.19426751],
         [0.30004403, 0.1706441 , 0.8723928 , 0.47353485, 0.96016128,
          0.45373964, 0.59661599, 0.29120244]]),),
 'symbolic_parameters': {},
 'device_gates': set(),
 'original_gate': None,
 'parameter_names': 'u',
 'nparams': 64,
 'trainable': True,
 '_clifford': False}
```

I think that qibo does not provide enough information to distisguish the behaviour from each U-gate. I thinks that this related to this very old issue: https://github.com/qiboteam/qibo/pull/478#issuecomment-933478007

IMHO, this is not possible yet...or am I wrong?

alecandido:
Not that is simple, but in principle you could hash the matrix values (`hash(array)` won't work, but you can hash its data buffer), and keep a global register as a dictionary (i.e. global to the single drawing, but local to the drawing function).

Every time you need to grab a symbol, you could just query the register. If it's in there, you take the index, if it's not, you add a new entry, with the `d[hash] = len(d)`. Then, you could use the index as $U_{42}$ (for the hypothetical 42nd independent unitary in the drawing)

sergiomtzlosa:
> Not that is simple, but in principle you could hash the matrix values in the gate parameters (`hash(array)` won't work, but you can hash its data buffer), and keep a global register as a dictionary (i.e. global to the single drawing, but local to the drawing function).
> 
> Every time you need to grab a symbol, you could just query the register. If it's in there, you take the index, if it's not, you add a new entry, with the `d[hash] = len(d)`. Then, you could use the index as U 42 (for the hypothetical 42nd independent unitary in the drawing)

Thanks @alecandido , but I do not understand how to pick up the data buffer and check on every U-gate, do you mean to hash the full U-gate? I do not catch how to pick up each single U-gate, can you point me a little example please?

renatomello:
@sergiomtzlosa the `gates,Unitary` class has a `name` attribute. I guess we could think of a mechanism that uses `name` to distinguish between unitaries.

sergiomtzlosa:
thank @alecandido , the property `name` holds the same name for the three single unitaries on three-unitary gate:

`gates.Unitary(np.random.random((8, 8)), 0, 2, 3)`

Should we think about add another property on the Unitary class to set a name to each single Unitary gate? What do you think?

renatomello:
> Should we think about add another property on the Unitary class to set a name to each single Unitary gate? What do you think?

My first inclination is to say that the change needs to be in the `Circuit.add` method. For instance, there could be a hidden integer `number` in the `Circuit` object that keeps track of how many `gates.Unitary` were added to the circuit and if one is added without a custom name, then the gate name becomes something like `$U_{number}$ ` and then `number += 1`.

alecandido:
@sergiomtzlosa 

```py
[ins] In [3]: u = gates.Unitary(np.ones((2,2)), 0)
[ins] In [4]: hash(u.parameters[0].data.tobytes())
Out[4]: 6949360754640926774
```

No need to add any further info.

renatomello:
> @sergiomtzlosa
> 
> ```python
> [ins] In [3]: u = gates.Unitary(np.ones((2,2)), 0)
> [ins] In [4]: hash(u.parameters[0].data.tobytes())
> Out[4]: 6949360754640926774
> ```
> 
> No need to add any further info (if you accept to name them { U i } i ).

A `hash` is a unique identifier but it isn't user-friendly, especially for plotting.

alecandido:
> A `hash` is a unique identifier but it isn't user-friendly, especially for plotting.

I was not proposing to use a hash as the matrix index, but to use the hash to build a register:
```py
def unitary_id(h: int, register: dict) -> int:
    id_ = register.get(h)
    if id_ is not None:
    	return id_
    id_ = register[h] = len(register)
    return id_


register: dict[int, int] = {}
h = hash(u.parameters[0].data.tobytes())
id_ = unitary_id(h, register)
```

sergiomtzlosa:
thanks @renatomello and @alecandido !! both solutions are good. I think that we can simplify the @alecandido solution by getting a global hash for the Unitary gate to identy it and then label each single U-gate with an index according to the number of parameters, from 0 to len(parameters)....I keep on thinking a solution.

alecandido:
Sorry @sergiomtzlosa, but anything "global" is only simple for relative short time span. As soon as you will have to maintain (change, debug it, or debug something related) it would get much more complicate. Also because it will be more complicate to reproduce.

I do not want to argue that the solution I proposed is the unique and best one. Just be careful about global objects (including having classes so huge which will act almost as a global context for most executions, as it could be a `Circuit` itself, if the whole application consists in its evaluation).

sergiomtzlosa:
thanks @alecandido , sorry for that, I will focus on a more flexible solution like yours😉

sergiomtzlosa:
great @alecandido ! I did not feel criticised at any time. I truly understand your point of view and as you know more than me qibo from the inside I feel very happy to be part of the discussion, and as you have more expertise than me on qibo framework and I will follow your ideas to do a better framework.

Thanks a lot!

alecandido:
I understand that you're not drawing unique rectangles, because it would be otherwise impossible to act on non-adjacent qubits (the alternative would be to mark differently connected and "jumping through" wires, e.g. making them dashed just before - but I'm not sure how simple it is, and not even sure whether it would be clear).

However, concerning the numbering, maybe it'd be worth to associate the ID to the whole unitary, and mark all elements the same (e.g. repeating $U_0$ for all the three components). Alternatively, you may use something like $\sideset{_2\ }{_0}{U}$, to clearly mark which is the element involved for each block (in this case, the third qubit for the gate $U_0$), conveying more information (not all operators are symmetric under swap of the two qubits, e.g. the CNOT is not).

sergiomtzlosa:
thanks @alecandido ! I will have a look about how to do that

alecandido:
Just in case: the source for ${}_2U_0$ is just `${}_2U_0$` :)

sergiomtzlosa:
before pushing changes, what about this?

imagen

src/qibo/ui/mpldrawer.py
```diff
@@ -257,13 +261,21 @@ def _draw_controls(ax, i, gate, labels, gate_grid, wire_grid, plot_params, measu
...
 
                 symbol = SYMBOLS.get(name, name)
 
                 if is_dagger:
                     symbol += r"$\rm{^{\dagger}}$"
 
+                if name == "UNITARY" and target.count("_") > 1:
+                    u_gate_single_hash = controls[cci].split("_")[2]
+                    subindex = plot_params["hash_unitary_gates"][u_gate_single_hash]
+                    symbol = r"$\rm_{}$".format(subindex) + symbol
```
alecandido, line 275: Here, and in every other place handling subscripts, I'd suggest you to keep a literal pair of braces (that you need to escape as double braces `{{}}`, for a total of 3 pairs...).
```suggestion
                    symbol = r"$\rm_{{{}}}$".format(subindex) + symbol
```

The reason is that your format may be otherwise screwed for multiple digits (or characters, in general)
- `$U_{42}$`: $U_{42}$
- `$U_42$`: $U_42$

tests/test_ui_mpldrawer.py
```diff
@@ -245,6 +245,18 @@ def test_fuse_cluster():
...
+def test_plot_unitaries():
+    """Test for plotting unitaries"""
+    c = Circuit(6)
+    c.add(gates.Unitary(np.random.random((8, 8)), 1, 2, 3))
+    c.add(gates.Unitary(np.random.random((8, 8)), 0, 2, 4))
+    c.add(gates.Unitary(np.random.random((2, 2)), 2))
+    c.add(gates.Unitary(np.random.random((2, 2)), 4))
+    c.add(gates.Unitary(np.random.random((8, 8)), 0, 2, 5))
+    _, fig = plot_circuit(c)
+    assert match_figure_image(fig, BASEPATH + "/test_plot_unitaries.npy") == True
```
renatomello, line 257:
```suggestion
def test_plot_unitaries(backend):
    """Test for plotting unitaries"""
    from qibo.quantum_info import random_unitary

    circuit = Circuit(6)
    circuit.add(gates.Unitary(random_unitary(8, backend=backend), 1, 2, 3))
    circuit.add(gates.Unitary(random_unitary(8, backend=backend), 0, 2, 4))
    circuit.add(gates.Unitary(random_unitary(2, backend=backend), 2))
    circuit.add(gates.Unitary(random_unitary(2, backend=backend), 4))
    circuit.add(gates.Unitary(random_unitary(8, backend=backend), 0, 2, 5))

    _, fig = plot_circuit(c)

    assert match_figure_image(fig, BASEPATH + "/test_plot_unitaries.npy")
```

tests/test_ui_mpldrawer.py
```diff
@@ -224,36 +225,41 @@ def test_layered_circuit():
...
-    c.add(gates.Unitary(np.random.random((2, 2)), 4))
-    c.add(gates.Unitary(np.random.random((8, 8)), 0, 2, 5))
-    _, fig = plot_circuit(c)
+
+    backend = construct_backend("numpy")
+    backend.set_seed(42)
+    np.random.seed(42)
+
+    circuit = Circuit(6)
+    circuit.add(gates.Unitary(random_unitary(8, backend=backend), 1, 2, 3))
```
renatomello, line 257: If you want, the `random_unitary` function also accepts `seed` as an argument.

sergiomtzlosa: good! I set it there too!!

renatomello:
@sergiomtzlosa what does the left subscript mean?

sergiomtzlosa:
hi @renatomello ! I set the left subscript just to enumerate the gates, that'a all! they do nothing

renatomello:
> hi @renatomello ! I set the left subscript just to enumerate the gates, that'a all! they do nothing

I thought the right subscripts were enumerating the unitaries...

sergiomtzlosa:
as they do nothing on both sides, then I should remove then...

alecandido:
@renatomello I actually suggested it above:
- the right ones are enumerating the gates, as you originally proposed
- the left ones are enumerating the qubits involved in the gates

This is to distinguish `CNOT(1,3)` from `CNOT(3,1)`, when `CNOT` is represented as a unitary (i.e. to explicitly identify the action of unitary which is not symmetric under qubits swaps).
In any case, it is just a refinement for unitaries used multiple times on different sets of qubits.

sergiomtzlosa:
I do not know how to do that because this Unitary gates representation do not distinguish between target and control qubits...can you point me how to achive this please?🙏

renatomello:
> @renatomello I actually suggested it above:
> 
>     * the right ones are enumerating the gates, as you originally proposed
> 
>     * the left ones are enumerating the qubits involved in the gates
> 
> 
> This is to distinguish `CNOT(1,3)` from `CNOT(3,1)`, when `CNOT` is represented as a unitary (i.e. to explicitly identify the action of unitary which is not symmetric under qubits swaps). In any case, it is just a refinement for unitaries used multiple times on different sets of qubits.

For that, I would suggest a superscript like $U_{1}^{(1,3)}$ since it is much more common and natural than a left subscript.

alecandido:
> I do not know how to do that because this Unitary gates representation do not distinguish between target and control qubits...can you point me how to achive this please?🙏

I expected it to be exactly what you did. In the plot in https://github.com/qiboteam/qibo/pull/1618#issuecomment-2744136005, I expected that, if you reused twice $U_0$, the same $U_0$ symbol would be used.

So, if a circuit contains
```py
c.add(Unitary(array, 0, 1, 3))
c.add(Unitary(array, 2, 1, 3))
```
both gates would be represented with 3 boxes using ${}_0 U_0$, ${}_0 U_1$, and ${}_0 U_2$. But they will be sorted accordingly to the qubit order, with ${}_0 U_0$ placed on the first wire in the former case, and on the third wire on the latter (while the other two boxes will be on the same wires for both gates).

sergiomtzlosa:
> > I do not know how to do that because this Unitary gates representation do not distinguish between target and control qubits...can you point me how to achive this please?🙏
> 
> I expected it to be exactly what you did. In the plot in [#1618 (comment)](https://github.com/qiboteam/qibo/pull/1618#issuecomment-2744136005), I expected that, if you reused twice U 0 , the same U 0 symbol would be used.
> 
> So, if a circuit contains
> 
> ```python
> c.add(Unitary(array, 0, 1, 3))
> c.add(Unitary(array, 2, 1, 3))
> ```
> 
> both gates would be represented with 3 boxes using 0 U 0 , 0 U 1 , and 0 U 2 . But they will be sorted accordingly to the qubit order, with 0 U 0 placed on the first wire in the former case, and on the third wire on the latter (while the other two boxes will be on the same wires for both gates).

I have a question, the `array` is the samefor both gates?

alecandido:
Yes, that's correct.

But apparently the right subscript it is not: if the `array` is the same (what I implied using twice the `array` identifier in the example above), the right subscript should have been the same as well.

sergiomtzlosa:
Given the following circuit as example:

```python
import numpy as np
from qibo import Circuit, gates, models
from qibo.ui import plot_circuit

c = models.Circuit(4)

array1 = np.random.random((8, 8))
array2 = np.random.random((8, 8))
array3 = np.random.random((8, 8))

c.add(gates.Unitary(array1, 0, 1, 3))
c.add(gates.X(0))
c.add(gates.Unitary(array1, 3, 1, 2))
c.add(gates.CNOT(0,2))
c.add(gates.Unitary(array2, 1, 0, 3))
c.add(gates.Z(1))
c.add(gates.Unitary(array2, 1, 0, 3))
c.add(gates.Y(2))
c.add(gates.Unitary(array3, 1, 0, 3))

plot_circuit(c);
```

The circuit os drawn as:

imagen

sergiomtzlosa:
I reviewed the merge conflicts with the master branch, you can review it for merging.

Issue #1497, opened 2024-10-17
The max index of qubits in gates.CU1 is out ranged  in qibo/src/qibo/models /hep.py


Landau1908:
**Describe the bug**
A minimun example: if nqubits =even number, it works, otherwise, it doesn't work.
```python
from qibo.ui import plot_circuit
from qibo.models.hep import qPDF

ansatz = 'Weighted'
multi_output = True
layers = 3
nqubits = 7

mypdf = qPDF(ansatz, layers, nqubits, multi_output=multi_output)
plot_circuit(mypdf.circuit, scale = 0.8, cluster_gates = True, style="quantumspain")
```
Ouputs:
```python
Traceback (most recent call last):

  Cell In[7], line 1
    plot_circuit(mypdf.circuit, scale = 0.8, cluster_gates = True, style="quantumspain")

  File E:\ProgramData\Lib\site-packages\qibo\ui\mpldrawer.py:729 in plot_circuit
    ax = _plot_quantum_schedule(gates_cluster, inits, params, labels, scale=scale)

  File E:\ProgramData\Lib\site-packages\qibo\ui\mpldrawer.py:59 in _plot_quantum_schedule
    return _plot_quantum_circuit(

  File E:\ProgramData\Lib\site-packages\qibo\ui\mpldrawer.py:120 in _plot_quantum_circuit
    _draw_gates(

  File E:\ProgramData\Lib\site-packages\qibo\ui\mpldrawer.py:180 in _draw_gates
    _draw_controls(

  File E:\ProgramData\Lib\site-packages\qibo\ui\mpldrawer.py:198 in _draw_controls
    control_indices = _get_flipped_indices(controls, labels)

  File E:\ProgramData\Lib\site-packages\qibo\ui\mpldrawer.py:472 in _get_flipped_indices
    return [_get_flipped_index(t, labels) for t in targets]

  File E:\ProgramData\Lib\site-packages\qibo\ui\mpldrawer.py:445 in _get_flipped_index
    i = labels.index(target)

ValueError: 'q_7' is not in list
```

**Expected behavior**
After reading the code (qibo/src/qibo/models /hep.py), this error is induced by this line
```python
for q in range(1, qubits + 1, 2):
```
which should be corrected to
```python
for q in range(1, qubits + 0, 2):
```
There are total 4 lines should be corrected: Line163，193，237，260

**Desktop (please complete the following information):**
 - OS: Win10, 64

**Additional context**

[...]

PR #1623, opened 2025-03-28
review plot out of bounds qubit number - bug #1497

sergiomtzlosa:
Checklist:
- [x] Tests are passing.

The circuit does not print with odd number of qubits, like 5, 7 and so on, because the circuit is using out of bounds number of qubits

```python
from qibo.ui import plot_circuit
from qibo.models.hep import qPDF

ansatz = 'Weighted'
multi_output = True
layers = 3
nqubits = 7

mypdf = qPDF(ansatz, layers, nqubits, multi_output=multi_output)
plot_circuit(mypdf.circuit, scale = 0.8, cluster_gates = True, style="quantumspain")
```

### DIFF
--- a/src/qibo/ui/mpldrawer.py
+++ b/src/qibo/ui/mpldrawer.py
@@ -577,12 +577,14 @@ def _process_gates(array_gates, nqubits):
         ]:
             for qbit in gate._target_qubits:
                 item = (init_label,)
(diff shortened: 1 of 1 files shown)


renatomello (approved):

MatteoRobbiati (approved):
Thanks!

PR #1628, opened 2025-04-16
Downgrade `sympy` dependency

renatomello:
As per discussed in the `qibo` meeting, this is necessary because of the `torch` dependencies.

Checklist:
- [x] Tests are passing.


pyproject.toml
```diff
@@ -24,7 +24,7 @@ include = ["*.out", "*.yml"]
 python = ">=3.10,<3.14"
 numpy = "^2.0.0"
 scipy = "^1.15.2"
-sympy = "^1.13.3"
+sympy = ">=1.13.1"
```
alecandido, line 27: The `^` is just more restrictive than `>=`, because the meaning is exactly "greater equal, but less than next major".

In this case, it would be equivalent to `>=1.13.1,<2.0.0`.

So, I would keep the `^`

alecandido (approved):

pyproject.toml
renatomello, line 27:
```suggestion
sympy = "^1.13.1"
```

AlejandroSopena (approved):

PR #1227, opened 2024-02-27
Creates `IBMQNoiseModel`

renatomello:
Closes #1214 .  

New class `IBMQNoiseModel` inherits `NoiseModel`, but allows for the creating of a noise model involving `ThermalRelaxationError`, `DepolarizingError`, and `ReadoutError`.

Noise model initialisation is done using a dictionary of parameters.

This class is meant to replace `CompositeNoiseModel` defined in `noise_model.py`.

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [x] Tests are passing.
- [x] Coverage does not decrease.
- [x] Documentation is updated.

PR #1218, opened 2024-02-14
Move `hellinger_shot_error` from `noise_model` to `quantum_info.utils`

renatomello:
Closes #1213 

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [x] Tests are passing.
- [x] Coverage does not decrease.
- [x] Documentation is updated.

PR #1629, opened 2025-04-17
Drop `noise_model.py`

renatomello:
Most of the functionalities in this file were transferred to the regular `noise.py` in #1218 and #1227 but this code was never removed.

Checklist:
- [x] Tests are passing.


alecandido (approved):
I only made a shallow check that the features exposed in the removed module have been reintroduced elsewhere (e.g. by the mentioned PRs), but it seems actually the case.

Thus, I perfectly agree to reduce redundancy, and favor the better integration with the quantum info module.

AlejandroSopena (approved):

PR #1639, opened 2025-05-05
Clifford measurement fix

### LINKED ISSUE #1598: Measurements with the clifford backend failing for some specific seeds
The ``test_backends_clifford::test_random_clifford_circuit`` test is failing for some specific seeds (for instance 15 and 25 locally for me). I am sure that the problem is the measurement because the final state is correct and matches the one obtained with the numpy backend. It's not even a post processing problem, e.g. probabilities calculation, because starting from the same collection of samples the exact same probabilities to the numpy backend are recovered. It is indeed a problem of incorrect samples extracted.

I double checked the implementation against the original proposal by [Aaronson and Gottesman](https://arxiv.org/abs/quant-ph/0406196) but I could not find any problem.

I am still trying to figure out what specifically is changing with those particular seeds to track where the problem might originate from.

BrunoLiegiBastonLiegi:
This closes #1598. The solution consists in moving the packing/unpacking of the bits inside of ``_rowsum`` after exponents calculation.

Checklist:
- [ ] Tests are passing.


renatomello:
Since you're going to have to do custom (un)packing, is it worth to think of using more than 8 bits, `stim` style?

PR #1641, opened 2025-05-13
Making Fused gate matrix backend dependent

BrunoLiegiBastonLiegi:
This just makes the fused gate to be constructed through the backend and avoid converting to numpy. The reason for this is that in some cases conversion to numpy can be tricky: with gpu backends you have to copy data to cpu, whereas for ml backends conversion to numpy is not always possible and will fail (e.g. pytorch with gradients attached, symbolic tensorflow or traced jax arrays).

Checklist:
- [ ] Tests are passing.


BrunoLiegiBastonLiegi:
This is no longer needed, I ended up directly overwriting the method in the qiboml backends qiboteam/qiboml#89.

PR #1651, opened 2025-05-28
[unitaryHack 2025] Support component-wise decompositions

### LINKED ISSUE #1585: Bug when using `Gate.decompose` on controlled gates
Most `qibo` gate classes have a custom `decompose` method that returns the equivalent gate decomposed in a "standard" gate set. This gate set is usually `CNOT` + single-qubit gates. For instance, the `RBS` gate admits the following decomposition:

```python
In [6]: from qibo import Circuit, gates
   ...: 
   ...: nqubits = 3
   ...: 
   ...: gate = gates.RBS(1, 2, 0.1)
   ...: 
   ...: circuit = Circuit(nqubits)
   ...: circuit.add(gate.decompose())
   ...: circuit.draw()
0: ────────────
1: ─H─o─RY─o─H─
2: ───X─RY─X───
```

The gates in the decomposition above correctly act on qubits $1$ and $2$, as the original gate does. However, there is a bug when the decomposed gate is defined with control qubits. For instance:

```python
In [7]: from qibo import Circuit, gates
   ...: 
   ...: nqubits = 3
   ...: 
   ...: gate = gates.RBS(1, 2, 0.1).controlled_by(0)
   ...: 
   ...: circuit = Circuit(nqubits)
   ...: circuit.add(gate)
   ...: circuit.draw()
   ...: print()
   ...: 
   ...: circuit = Circuit(nqubits)
   ...: circuit.add(gate.decompose())
   ...: circuit.draw()
0: ─o───
1: ─RBS─
2: ─RBS─

0: ─H─o─RY─o─H─
1: ───X─RY─X───
2: ────────────
```
As one can see the in the example above, not only the gates act on different qubits than the original gate but the decomposition completely lost the information that the gate was a controlled gate.

What an acceptable solution looks like for the example above:
```python
0: ─o─o─o──o──o─o─
1: ─H─o─RY─|──o─H─
2: ───X────RY─X───
```
For all gates, controlling every gate in the decomposition by the correct qubits is an acceptable solution. However, a lot of the gate decompositions only subset of the gates in the decomposition actually need to be controlled because the rest of the circuit amounts to an identity. For instance, in the example above the following is what an ideal solution looks like:
```python
0: ─────o──o──────
1: ─H─o─RY─|──o─H─
2: ───X────RY─X───
```

y-richie-y:
Closes #1585 

I implemented the "ideal" solution by leaving conjugation gates (when it is repeated on both sides) uncontrolled.

### DIFF
--- a/src/qibo/gates/gates.py
+++ b/src/qibo/gates/gates.py
@@ -2821,7 +2821,7 @@ def _check_engine(array):
 
 
 def _controlled_decompose(self, *free, use_toffolis: bool = True) -> List[Gate]:
-    """Decompose non-conjugation gates component-wise. """
+    """Decompose non-conjugation gates component-wise."""
     from qibo.transpiler.decompositions import (  # pylint: disable=C0415
         standard_decompositions,
     )
@@ -2847,10 +2847,7 @@ def _controlled_decompose(self, *free, use_toffolis: bool = True) -> List[Gate]:
             gate_control_qubits = gate.control_qubits
             gate.is_controlled_by = False
             gate.control_qubits = ()
-            c_gate = gate.controlled_by(
-                *control_qubits,
-                *gate_control_qubits
-            )
+            c_gate = gate.controlled_by(*control_qubits, *gate_control_qubits)
             decomp[i] = c_gate
         self.is_controlled_by = True
         decomp = [*c_decomps1, *decomp, *c_decomps2[::-1]]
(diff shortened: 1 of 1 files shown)


y-richie-y:
@renatomello do you plan on reviewing this PR?

renatomello:
> @renatomello do you plan on reviewing this PR?

I'm sorry for the delay. We have three PRs open on the same issue and need to decide the "winner". I'm sorry I have been neglecting this PR but I'll be coming around to it soon.

One thing that I can say right now is that even though this solution works for the gates that were mentioned in the PR, it is not being applied to all the other native gates in the module.

renatomello:
@y-richie-y Thanks for opening this PR. Even though your solution works, we decided to move forward with the solution implemented in #1653 due to it being a bit more complete. We thank you for your input though.

PR #1606, opened 2025-03-14
New transpiler techniques (gate optmization):  U3GateFusion, RotationGateFusion and InverseCancellation

carlos-luque:
We have worked on issue #1588 and added new techniques in transpiler optimization:

- U3GateFusion: Merges pairs of U3 gates.
- RotationGateFusion: Merges pairs of rotation gates along the same axis.
- InverseCancellation: Eliminates pairs of adjacent gates (gates.H, gates.Y, gates.Z, gates.X, gates.CNOT, gates.CZ, gates.SWAP).

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.

PR #1653, opened 2025-05-29
[unitaryHack 2025] Refactor gate decomposition to support controlled gates

GiacomoFrn:
## Description

This PR addresses the first step of [issue #1585](https://github.com/qiboteam/qibo/issues/1585), which aims to generalize the `decompose()` method for all gates, including controlled gates.

## Summary of changes

- Refactored all custom gates in `gates.py` to implement a new internal method `_base_decompose()`, which defines their standard decomposition.
- Modified the `decompose()` method in abstract.py to support controlled gates by:
  - Calling `_base_decompose()` on the base gate (i.e., the gate with no control qubits),
  - Using `control_mask_after_stripping()` to determine which gates in the decomposition require controls,
  - Adding controls only to the necessary gates, avoiding redundant controlled gates,
  - Preserving the correct qubit ordering and control structure.

## Notes

- The new `_base_decompose()` method is for internal use only and separates decomposition logic from control logic.
- The behavior for non-controlled gates remains unchanged.
- The new logic improves circuit efficiency by not applying controls to gates that cancel out.

### DIFF
--- a/tests/test_gates_abstract.py
+++ b/tests/test_gates_abstract.py
@@ -480,6 +480,7 @@ def test_gates_cancel(g1, g2, expected):
     result = Gate._gates_cancel(g1, g2)
     assert result is expected
 
+
 def test_toffoli_congruent(backend):
     congruent = gates.TOFFOLI(0, 1, 2)
 
@@ -491,6 +492,8 @@ def test_toffoli_congruent(backend):
     target[4, 4] = -1
 
     assert backend.calculate_matrix_norm(congruent - target) < 1e-8
+
+
 def test_special_gate():
     from qibo.gates.abstract import SpecialGate
 


GiacomoFrn:
Hi @renatomello,
I'm reaching out with a couple of doubts about how I could improve my PR. There are two main issues I’d like to clarify:

1. **Controlled gate behavior**: I noticed that when a gate has a `_controlled_gate` defined (like `U3`), the expected behavior — adding a control to each gate in the decomposition — doesn't work as intended. Instead, I only get a single `CU3` gate, and the decomposition is skipped. Also, when I inspect the resulting gate, the `is_controlled_by` flag is not set to `True`, which makes me think that the control is handled differently in this case (perhaps by directly creating the controlled version, like `CU3`). I'm unsure how to intercept or modify this behavior — do you have any suggestions?

2. **Control optimization**: I’ve implemented a simple version of the control optimization that handles basic sandwich-like identity cases. However, I'm not sure how to scale this to more general decompositions, where the gate set might be arbitrary (e.g., universal gates). A numerical approach to identify identity-like patterns seems too costly, while hardcoding all possible identities feels inelegant. Do you have any thoughts on a better strategy?

Thanks in advance for your time and help!

src/qibo/gates/abstract.py
```diff
@@ -3,6 +3,7 @@
 from typing import List, Sequence, Tuple
 
 import sympy
+from numpy import isclose, pi
```
renatomello, line 6:
```suggestion
from math import pi
```

renatomello:
@GiacomoFrn I'm sorry for the delay, I'll revisit this PR tomorrow.

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,95 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
+        if g1.target_qubits != g2.target_qubits:
+            return False
+        if getattr(g1, "is_controlled_by", True):
+            if g1.control_qubits != g2.control_qubits:
+                return False
+        # Identity conditions
+        name = g1.name
+        if name in ("h", "cx"):
+            return True
+        if name == "ry":
```
renatomello, line 391:
```suggestion
        if g1.__class__ != g2.__class__:
            return False

        if g1.target_qubits != g2.target_qubits:
            return False

        if getattr(g1, "is_controlled_by", True):
            if g1.control_qubits != g2.control_qubits:
                return False

        # Identity conditions
        name = g1.name
        if name in ("h", "cx"):
            return True

        if name == "ry":
```

renatomello:
@GiacomoFrn I like your solution the best. Can you comment on the issue so I can assign it to you, please? Could you also star our repo?

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,95 @@ def controlled_by(self, *qubits: int) -> "Gate":
             self.control_qubits = qubits
         return self
 
-    def decompose(self, *free) -> List["Gate"]:
+    def _base_decompose(self, *free, use_toffolis=True) -> List["Gate"]:
+        """Base decomposition for gates.
+
+        Returns a list containing the gate itself. Should be overridden by subclasses
+        that support decomposition to simpler gates.
```
renatomello:
```suggestion
        Returns a list containing the gate itself. Should be overridden by
        subclasses that support decomposition to simpler gates.
```

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,95 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
+
+        Args:
+            free: Ids of free qubits to use for the gate decomposition.
```
renatomello, line 370: Missing `use_toffolis` in the Args

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,95 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
+
+        Returns:
+            list: gates that have the same effect as applying the original gate.
```
renatomello:
```suggestion
            list: Synthesis of the original gate in another gate set.
```

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,95 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
+        """
+        return [self.__class__(*self.init_args, **self.init_kwargs)]
+
+    @staticmethod
+    def gates_cancel(g1, g2):
+        """Check if same gate type, same targets, same controls"""
+        if g1.__class__ != g2.__class__:
+            return False
```
renatomello, line 400:
```suggestion
        if g1.__class__ != g2.__class__:
            return False

```

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,95 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
+        if g1.target_qubits != g2.target_qubits:
+            return False
```
renatomello, line 403:
```suggestion
        if g1.target_qubits != g2.target_qubits:
            return False

```

I can see a problem here with symmetric gates like `CZ`. For instance, `CZ(0, 1)` is the same as `CZ(1, 0)`, but this current test would find the two target qubits to be different.

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,95 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
+        if getattr(g1, "is_controlled_by", True):
+            if g1.control_qubits != g2.control_qubits:
+                return False
```
renatomello:
```suggestion
        if getattr(g1, "is_controlled_by", True) and g1.control_qubits != g2.control_qubits:
            return False

```

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,95 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
+            return True
+        if name == "ry":
+            theta1 = g1.parameters[0]
+            theta2 = g2.parameters[0]
+            # Check if theta1 + theta2 is a multiple of 2π
+            return [
+                bool((first + second) % (2 * pi) < 1e-8)
+                for first, second in zip(theta1, theta2)
+            ]
+        return False
```
renatomello:
```suggestion
        name = g1.name
        if name in ("h", "cx"):
            return True

        if name == "ry":
            theta1 = g1.parameters[0]
            theta2 = g2.parameters[0]
            # Check if theta1 + theta2 is a multiple of 2π
            return [
                bool((first + second) % (2 * pi) < 1e-8)
                for first, second in zip(theta1, theta2)
            ]

        return False
```

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,95 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
+        Args:
+            free: Ids of free qubits to use for the gate decomposition.
+
+        Returns:
+            list: gates that have the same effect as applying the original gate.
+        """
+        return [self.__class__(*self.init_args, **self.init_kwargs)]
+
+    @staticmethod
+    def gates_cancel(g1, g2):
```
renatomello:
```suggestion
    @staticmethod
    def _gates_cancel(g1, g2):
```
It seems that this method does not need to be user-facing.

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,95 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
+            theta1 = g1.parameters[0]
+            theta2 = g2.parameters[0]
+            # Check if theta1 + theta2 is a multiple of 2π
+            return [
+                bool((first + second) % (2 * pi) < 1e-8)
+                for first, second in zip(theta1, theta2)
+            ]
+        return False
+
+    def control_mask_after_stripping(self, gates: List["Gate"]) -> List[bool]:
```
renatomello:
```suggestion
    def _control_mask_after_stripping(self, gates: List["Gate"]) -> List[bool]:
```

Same here

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,95 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
+                right -= 1
+            else:
+                break
+        return mask
+
+    def decompose(self, *free, use_toffolis=True) -> List["Gate"]:
         """Decomposes multi-control gates to gates supported by OpenQASM.
 
         Decompositions are based on `arXiv:9503016 <https://arxiv.org/abs/quant-ph/9503016>`_.
+        If the gate is already controlled, it recursively decomposes the base gate and updates the control qubits accordingly.
```
renatomello:
```suggestion
        If the gate is already controlled, it recursively decomposes the base gate and updates
        the control qubits accordingly.
```

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,95 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
 
         Args:
             free: Ids of free qubits to use for the gate decomposition.
```
renatomello, line 444: Missing `use_toffolis`

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,95 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
+        if self.is_controlled_by:
+            # Step 1: Error check with all controls/targets
+            error_check_gate = self.__class__(*self.init_args, **self.init_kwargs)
+            error_check_gate.target_qubits = self.target_qubits
+            error_check_gate.control_qubits = self.control_qubits
+            if set(free) & set(error_check_gate.qubits):
+                raise ValueError(
+                    "Cannot decompose multi-control X gate if free "
+                    "qubits coincide with target or controls."
+                )
```
renatomello:
```suggestion
                raise_error(ValueError, 
                    "Cannot decompose multi-controlled ``X`` gate if free "
                    "qubits coincide with target or controls."
                )
```

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,95 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
+            decomposed = base_gate._base_decompose(*free, use_toffolis=use_toffolis)
+            mask = self.control_mask_after_stripping(decomposed)
+            for i, g in enumerate(decomposed):
+                if not mask[i]:
+                    continue
+                if not g.is_controlled_by:
+                    g.is_controlled_by = True
+                    g.control_qubits = self.control_qubits
+                else:
+                    g.control_qubits += self.control_qubits
```
renatomello, line 449:
```suggestion
                if mask[i]:
                    if not g.is_controlled_by:
                        g.is_controlled_by = True
                        g.control_qubits = self.control_qubits
                    else:
                        g.control_qubits += self.control_qubits
```

renatomello:
```suggestion
            for k, g in enumerate(decomposed):
                if mask[k]:
                    if not g.is_controlled_by:
                        g.is_controlled_by = True
                        g.control_qubits = self.control_qubits
                    else:
                        g.control_qubits += self.control_qubits
```

renatomello:
@GiacomoFrn can you create a test for the new feature that fixes the coverage?

GiacomoFrn:
@renatomello yes, I'll do that

src/qibo/gates/abstract.py
GiacomoFrn: I see your point, in general, I was mentioning this in a comment before this conversation. While the proposed solution works correctly when the decomposition uses the set (cx, h, ry), the optimization part doesn't currently support different sets of universal gates. I don't really know which is the optimal way of taking this into account for all the possible set in which the gate could be decomposed. Manually checking if their product is identity could be a way, but probably not optimal in term of performance. I'm happy to implement any suggestions to solve this.

renatomello: Maybe the things that are being done in https://github.com/qiboteam/qibo/pull/1606 could help?

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,84 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
+    def control_mask_after_stripping(self, gates: List["Gate"]) -> List[bool]:
+        """Returns a mask indicating which gates should be controlled."""
+        left = 0
+        right = len(gates) - 1
+        mask = [True] * len(gates)
+        while left < right:
+            g1, g2 = gates[left], gates[right]
+            if self.gates_cancel(g1, g2):
+                mask[left] = False
+                mask[right] = False
```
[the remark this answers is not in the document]
GiacomoFrn: @renatomello, here I noticed I get an error since mask is a list, should I import numpy and make it a numpy array or leave the two lines?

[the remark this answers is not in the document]
renatomello: Either way is fine by me

src/qibo/gates/abstract.py
GiacomoFrn: Yes, probably I could use the (InverseCancellation, RotationGateFusion) in the transpiler.optimizer. How should I proceed? Because that PR is not closed yet. I can merge that branch to my branch, but then you have to merge that PR before mine.

renatomello: I think it's easier to leave it as it is right now and revisit it after both PRs are merged. Probably we open an issue as a reminder that this needs to be done at some point.

renatomello:
@GiacomoFrn could you fix the lint issues?

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,109 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
+            return True
+
+        if name == "ry":
+            theta1 = g1.parameters[0]
+            theta2 = g2.parameters[0]
+            # Check if theta1 + theta2 is a multiple of 2π
+            return [
+                bool((first + second) % (2 * pi) < 1e-8)
+                for first, second in zip(theta1, theta2)
+            ]
```
BrunoLiegiBastonLiegi: maybe I am missing something here, are ``theta1`` and ``theta2`` supposed to be lists, tuples or anything longer than a scalar? If yes, then, don't you need to do a ``return all([...])`` or something similar? If not, what is the loop needed for?

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,109 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
         Returns:
             list: gates that have the same effect as applying the original gate.
         """
-        return [self.__class__(*self.init_args, **self.init_kwargs)]
+        if self.is_controlled_by:
+            # Step 1: Error check with all controls/targets
+            error_check_gate = self.__class__(*self.init_args, **self.init_kwargs)
+            error_check_gate.target_qubits = self.target_qubits
+            error_check_gate.control_qubits = self.control_qubits
+            if set(free) & set(error_check_gate.qubits):
```
BrunoLiegiBastonLiegi:
```suggestion
            # Step 1: Error check with all controls/targets
            if set(free) & set(self.qubits):
```
do you actually need to build the gate?

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,109 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
+                    ValueError,
+                    "Cannot decompose multi-controlled ``X`` gate if free "
+                    "qubits coincide with target or controls.",
+                )
+            # Step 2: Decompose base gate without controls
+            base_gate = self.__class__(*self.init_args, **self.init_kwargs)
+            decomposed = base_gate._base_decompose(*free, use_toffolis=use_toffolis)
+            mask = self._control_mask_after_stripping(decomposed)
+            for k, g in enumerate(decomposed):
+                if mask[k]:
```
BrunoLiegiBastonLiegi:
```suggestion
            for m, g in zip(mask, decomposed):
                if m:
```

renatomello:
> @renatomello yes, I'll do that

this is the only think holding approval now.

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,109 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
+                    if not g.is_controlled_by:
+                        g.is_controlled_by = True
+                        g.control_qubits = self.control_qubits
+                    else:
+                        g.control_qubits += self.control_qubits
```
BrunoLiegiBastonLiegi: I am not sure you actually need this ``if/else`` here, because the only gates that have ``is_controlled_by==True`` are those created by a ``.controlled_by(q)`` call (thus not simple two qubits gates such as CNOT), which hardly should be obtained through a ``_base_decompose`` call, I guess (but correct me if I am wrong). 
But rather, what happens if you have indeed that ``g`` is, for instance, a ``CNOT`` that has one control qubit but ``is_controlled_by==False`` and you cannot simply increase its control qubits? In the ``RBS`` case from the issue, luckily the two CNOTs cancel out to the identity and thus do not enter here, but what if you had only one (I'm not sure whether there is any notable gate to test for this @renatomello )?
```python 
                    if len(g.control_qubits) == 0:
                        g.is_controlled_by = True
                        g.control_qubits = self.control_qubits
                    else:
                        # here you should do the control of a controlled gate
```

BrunoLiegiBastonLiegi (commented):
Thanks, this looks good for the most part to me, my only doubt concerns the case where you might have to do the control of a controlled gate as detailed in the comment below, but that might actually not apply here, thus in that case, just ignore it.

src/qibo/gates/abstract.py
GiacomoFrn: That is true, we don't need it

src/qibo/gates/abstract.py
GiacomoFrn:
```suggestion
                    if len(g.control_qubits) == 0:
                        g.is_controlled_by = True
                        g.control_qubits = self.control_qubits
                    else:
                        g.is_controlled_by = True
                        g.control_qubits += self.control_qubits
```

would this work? I did a local test and for RBS gate without adding the optimization it behaves correctly adding the double control to the CNOT

GiacomoFrn:
> > @renatomello yes, I'll do that
> 
> this is the only think holding approval now.

I added a test but apparently the coverage is still not enough, do you have suggestions on a test that I could add?

src/qibo/gates/abstract.py
BrunoLiegiBastonLiegi: Ah yeah, sorry, yesterday while testing it was giving me an error, but probably I did something wrong. Apparently it is enough to extend the control qubits and set ``is_controlled_by=True``. I would just suggest to move the ``g.is_controlled_by = True`` outside of the ``if/else`` since you're gonna do it anyway.

tests/test_gates_abstract.py
```diff
@@ -420,6 +420,22 @@ def test_decompose():
     assert isinstance(decomp_gates[0], gates.H)
 
 
+def test_decompose_controlled():
+    target = gates.H(0).controlled_by(1)
+    decomp = target.decompose()
+    assert len(decomp) == 1
+    assert isinstance(decomp[0], gates.H)
+    assert decomp[0].control_qubits == (1,)
```
BrunoLiegiBastonLiegi:
```suggestion
def test_decompose_controlled():
    backend = NumpyBackend()
    target = gates.H(0).controlled_by(1)
    decomp = target.decompose()
    assert len(decomp) == 1
    assert isinstance(decomp[0], gates.H)
    assert decomp[0].control_qubits == (1,)
    c1 = Circuit(2)
    c2 = c1.copy()
    c1.add(target)
    c2.add(decomp)
    s1 = backend.execute_circuit(c1).state()
    s2 = backend.execute_circuit(c2).state()
    backend.assert_allclose(s1, s2)
```
I would try to verify that you get indeed the same final state, just to double check that nothing breaks in circuit execution.

tests/test_gates_abstract.py
```diff
@@ -420,6 +420,22 @@ def test_decompose():
...
+
+
+def test_decompose_controlled_optimized():
+    target = gates.RBS(1, 2, 0.1).controlled_by(0)
+    decomp = target.decompose()
+    assert len(decomp) == 6
+    controls_on_zero = sum([0 in g.control_qubits for g in decomp])
+    assert controls_on_zero == 2
```
BrunoLiegiBastonLiegi: I would do the same execution test here.

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,103 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
+        """Check if same gate type, same targets, same controls"""
+        if g1.__class__ != g2.__class__:
+            return False
+
+        if g1.target_qubits != g2.target_qubits:
+            return False
+        if (
+            getattr(g1, "is_controlled_by", True)
+            and g1.control_qubits != g2.control_qubits
+        ):
```
BrunoLiegiBastonLiegi:
```suggestion
        if (
            g1.control_qubits != g2.control_qubits
        ):
```
I think you only need to check the control qubits, as we said below a CNOT has ``is_controlled_by==False`` and thus would not enter here.

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,103 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
+            return False
+
+        # Identity conditions
+        name = g1.name
+        if name in ("h", "cx"):
+            return True
```
BrunoLiegiBastonLiegi: I am not sure why you need this very specific name check, this method is supposed to be general right?

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,103 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
+
+        if name == "ry":
+            theta1 = g1.parameters[0]
+            theta2 = g2.parameters[0]
+            # Check if theta1 + theta2 is a multiple of 2π
+            return bool((theta1 + theta2) % (2 * pi) < 1e-8)
```
BrunoLiegiBastonLiegi: even here, why ``ry`` precisely? I guess you would need to have it generally for parametrized gates, right?
Something like this:
```suggestion
        if isinstance(g1, gates.ParametrizedGate):
            # Check if theta1 + theta2 is a multiple of 2π
            same_angle = [bool((theta1 + theta2) % (2 * pi) < 1e-8 for theta1, theta2 in zip(g1.parameters, g2.parameters)]
            return all(same_angle)
```

BrunoLiegiBastonLiegi (commented):
Sorry, just a few further minor things which should also help with coverage.

src/qibo/gates/abstract.py
GiacomoFrn:
```suggestion
                    g.is_controlled_by = True
                    if len(g.control_qubits) == 0:
                        g.control_qubits = self.control_qubits
                    else:
                        g.control_qubits += self.control_qubits
```

BrunoLiegiBastonLiegi: yep, actually, now that I think about it you probably don't even need the ``if/else`` at all, because non controlled gates, such as ``gates.H`` for instance, still have a ``.control_qubits==(,)`` initialized with an empty tuple. Therefore, just a:
```python
                    g.is_controlled_by = True
                    g.control_qubits += self.control_qubits
```
will probably work in all cases.
(So sorry for the constant updates)

src/qibo/gates/abstract.py
GiacomoFrn: I raised this issue before in a comment and asked for directions on how to proceed. With @renatomello, as it my understanding, we discussed that we should leave it like this in a way that is not general and wait for #1606 to be merged such that one can use the methods defined there (InverseCancellation, RotationGateFusion) to make it general.

GiacomoFrn: here ry could be improved by adding rx and rz in a quick edit, before I could add other "inverse gates" like y, z, x, swap

renatomello: IIRC `ry`, `rx`, and `rz` all have the same parent class `_Rn_`, so one could check for that class.

tests/test_gates_abstract.py
GiacomoFrn: ok

renatomello:
@GiacomoFrn @BrunoLiegiBastonLiegi there seems to be a global issue with `numba` and `python 3.13` on `macOS`, *e.g.* tests failing for the same reason in #1667. Since this is unrelated to this PR, I'm gonna move forward and approve it.

renatomello (approved):

src/qibo/gates/abstract.py
```diff
@@ -385,25 +385,19 @@ def _gates_cancel(g1, g2):
...
-            getattr(g1, "is_controlled_by", True)
-            and g1.control_qubits != g2.control_qubits
-        ):
+        if g1.control_qubits != g2.control_qubits:
             return False
 
         # Identity conditions
         name = g1.name
-        if name in ("h", "cx"):
+        if name in ("h", "cx", "x", "y", "z", "swap"):
```
BrunoLiegiBastonLiegi: I am still not understanding why you need this hard-coded name check. I think you could drop this completely and just return ``True`` at the end of the function, after all checks:
https://github.com/qiboteam/qibo/blob/0fe9c4798921ce39525097fb733ddc1294c54db3/src/qibo/gates/abstract.py#L402

If I got it right, you're checking for reasons the gates do not cancel (different class, different controls, different targets), if you don't find any then the gates do cancel out, right? (and indeed that is one of the lines missing in the coverage from what I see)

src/qibo/gates/abstract.py
```diff
@@ -385,25 +385,19 @@ def _gates_cancel(g1, g2):
...
             return True
 
-        if name == "ry":
+        if "_Rn_" in [base.__name__ for base in g1.__class__.__bases__]:
```
BrunoLiegiBastonLiegi, line 414: Better thanks, however, this is still not supporting multi parameters gates, which might be fine  for the moment, maybe just point this out in the docstring.

BrunoLiegiBastonLiegi (commented):
Thank you so much @GiacomoFrn. I have just one final comment, after that one this can go by my side.

tests/test_gates_abstract.py
```diff
@@ -420,6 +423,55 @@ def test_decompose():
...
+@pytest.mark.parametrize(
+    "g1, g2, expected",
+    [
+        (gates.RX(0, pi), gates.RX(0, -pi), True),
+        (gates.RY(1, 0.5), gates.RY(1, -0.5), True),
+        (gates.RZ(2, 2 * pi), gates.RZ(2, -2 * pi), True),
+        (gates.X(0), gates.X(0), True),
+        (gates.Y(1), gates.Y(1), True),
+        (gates.Z(2), gates.Z(2), True),
+        (gates.RX(0, pi / 2), gates.RX(0, pi / 3), False),
```
BrunoLiegiBastonLiegi:
```suggestion
        (gates.X(0), gates.Y(0), False),
        (gates.CNOT(0,1), gates.CNOT(0,1), True),
        (gates.CNOT(0,2), gates.CNOT(1,2), False),
        (gates.RX(0, pi), gates.RX(0, -pi), True),
        (gates.RY(1, 0.5), gates.RY(1, -0.5), True),
        (gates.RZ(2, 2 * pi), gates.RZ(2, -2 * pi), True),
        (gates.X(0), gates.X(0), True),
        (gates.Y(1), gates.Y(1), True),
        (gates.Z(2), gates.Z(2), True),
        (gates.RX(0, pi / 2), gates.RX(0, pi / 3), False),
```
these further cases should help with the coverage

src/qibo/gates/abstract.py
GiacomoFrn: Yes, let me try to explain what my prior was in doing this. I expect that the gates that go into _gates_cancel are the output of a decomposition, standard decomposition with a universal gate set or custom decomposition defined in the gate. Then, two gates of course won't cancel if they have different class/controls or targets, but this is not enough to claim that if these conditions are met, the gates will cancel. 

One case is for rotation gates, which we handle below. But I would also expect an S or T gate to fail this test, for example.

src/qibo/gates/abstract.py
GiacomoFrn: That is true, I thought that in general I could assume that multi parameters gates wouldn't appear at this stage of the decomposition, but I am probably wrong. I will update the docstring.

renatomello:
@GiacomoFrn can you merge the master branch and fix the conflicts?

src/qibo/gates/abstract.py
BrunoLiegiBastonLiegi: @GiacomoFrn this is the only thing blocking my approval right now, could you please look into this?

GiacomoFrn: does the comment above clarify or is there anything else you think would work better?

tests/test_gates_abstract.py
```diff
@@ -422,6 +425,71 @@ def test_decompose():
...
+    assert len(decomp) == 1
+    assert isinstance(decomp[0], gates.H)
+    assert decomp[0].control_qubits == (1,)
+    c1 = Circuit(2)
+    c2 = c1.copy()
+    c1.add(target)
+    c2.add(decomp)
+    s1 = backend.execute_circuit(c1).state()
+    s2 = backend.execute_circuit(c2).state()
+    backend.assert_allclose(s1, s2)
```
renatomello:
```suggestion
def test_decompose_controlled(backend):
    target = gates.H(0).controlled_by(1)
    decomp = target.decompose()

    assert len(decomp) == 1
    assert isinstance(decomp[0], gates.H)
    assert decomp[0].control_qubits == (1,)
    
    circuit_1 = Circuit(2)
    circuit_2 = circuit_1.copy(deep=True)
    circuit_1.add(target)
    circuit_2.add(decomp)

    backend.assert_circuitclose(circuit_1, circuit_2)
```

tests/test_gates_abstract.py
```diff
@@ -2,13 +2,16 @@
 `qibo/gates/gates.py`."""
 
 import json
+from math import pi
 from typing import Optional
 
 import pytest
 
 from qibo import gates, matrices
+from qibo.backends import NumpyBackend
```
renatomello:
```suggestion
```

tests/test_gates_abstract.py
```diff
@@ -422,6 +425,71 @@ def test_decompose():
...
+    assert len(decomp) == 6
+    controls_on_zero = sum([0 in g.control_qubits for g in decomp])
+    assert controls_on_zero == 2
+    c1 = Circuit(3)
+    c2 = c1.copy()
+    c1.add(target)
+    c2.add(decomp)
+    s1 = backend.execute_circuit(c1).state()
+    s2 = backend.execute_circuit(c2).state()
+    backend.assert_allclose(s1, s2)
```
renatomello:
```suggestion
def test_decompose_controlled_optimized(backend):
    target = gates.RBS(1, 2, 0.1).controlled_by(0)
    decomp = target.decompose()

    assert len(decomp) == 6

    controls_on_zero = sum([0 in g.control_qubits for g in decomp])
    assert controls_on_zero == 2

    circuit_1 = Circuit(3)
    circuit_2 = circuit_1.copy(deep=True)
    circuit_1.add(target)
    circuit_2.add(decomp)

    backend.assert_circuitclose(circuit_1, circuit_2)
```

tests/test_gates_abstract.py
```diff
@@ -422,6 +424,72 @@ def test_decompose():
...
+    circuit_2.add(decomp)
+
+    backend.assert_circuitclose(circuit_1, circuit_2)
+
+
+@pytest.mark.parametrize(
+    "g1, g2, expected",
+    [
+        (gates.X(0), gates.Y(0), False),
+        (gates.CNOT(0, 1), gates.CNOT(0, 1), True),
```
renatomello:
```suggestion
        (gates.X(0), gates.Y(0), False),
        (gates.X(0), gates.X(1), False),
        (gates.CNOT(0, 1), gates.CNOT(0, 1), True),
```

tests/test_gates_abstract.py
```diff
@@ -422,6 +424,73 @@ def test_decompose():
...
+        (gates.CNOT(0, 2), gates.CNOT(1, 2), False),
+        (gates.RX(0, pi), gates.RX(0, -pi), True),
+        (gates.RY(1, 0.5), gates.RY(1, -0.5), True),
+        (gates.RZ(2, 2 * pi), gates.RZ(2, -2 * pi), True),
+        (gates.X(0), gates.X(0), True),
+        (gates.Y(1), gates.Y(1), True),
+        (gates.Z(2), gates.Z(2), True),
+        (gates.RX(0, pi / 2), gates.RX(0, pi / 3), False),
```
renatomello:
```suggestion
        (gates.X(0), gates.X(0), True),
        (gates.X(0), gates.X(1), False),
        (gates.X(0), gates.Y(0), False),
        (gates.Y(1), gates.Y(1), True),
        (gates.Z(2), gates.Z(2), True),
        (gates.SX(0), gates.SX(0), False),
        (gates.RX(0, pi), gates.RX(0, -pi), True),
        (gates.RX(0, pi / 2), gates.RX(0, pi / 3), False),
        (gates.RY(1, 0.5), gates.RY(1, -0.5), True),
        (gates.RZ(2, 2 * pi), gates.RZ(2, -2 * pi), True),
        (gates.CNOT(0, 1), gates.CNOT(0, 1), True),
        (gates.CNOT(0, 2), gates.CNOT(1, 2), False),
        (gates.TOFFOLI(0, 1, 2), gates.TOFFOLI(0, 1, 2), True),
        (gates.CCZ(0, 1, 2), gates.CCZ(0, 1, 2), True),
```

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,113 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
+
+        if g1.target_qubits != g2.target_qubits:
+            return False
+
+        if g1.control_qubits != g2.control_qubits:
+            return False
+
+        # Identity conditions for fixed gates
+        name = g1.name
+        if name in ("h", "cx", "x", "y", "z", "swap"):
```
renatomello:
```suggestion
        if name in ("h", "cx", "x", "y", "z", "swap", "ecr", "ccx", "ccz"):
```

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,113 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
 
         Decompositions are based on `arXiv:9503016 <https://arxiv.org/abs/quant-ph/9503016>`_.
+        If the gate is already controlled, it recursively decomposes the base gate and updates
+        the control qubits accordingly.
 
         Args:
             free: Ids of free qubits to use for the gate decomposition.
+            use_toffolis: If ``True`` the decomposition contains only ``TOFFOLI`` gates.
+                If ``False`` a congruent representation is used for ``TOFFOLI`` gates.
+                See :class:`qibo.gates.TOFFOLI` for more details on this representation.
```
renatomello:
```suggestion
    def decompose(self, *free, use_toffolis: bool = True) -> List["Gate"]:
        """Decomposes multi-control gates to gates supported by OpenQASM.

        Decompositions are based on `arXiv:9503016 <https://arxiv.org/abs/quant-ph/9503016>`_.
        If the gate is already controlled, it recursively decomposes the base gate and updates
        the control qubits accordingly.

        Args:
            free: Ids of free qubits to use for the gate decomposition.
            use_toffolis(bool, optional): If ``True``, the decomposition contains only 
                :class:`qibo.gates.TOFFOLI` gates. If ``False``, a congruent 
                representation is used for :class:`qibo.gates.TOFFOLI` gates.
                See :class:`qibo.gates.TOFFOLI` for more details on this representation.
```

src/qibo/gates/abstract.py
```diff
@@ -359,18 +360,114 @@ def controlled_by(self, *qubits: int) -> "Gate":
...
+                    "qubits coincide with target or controls.",
+                )
+            # Step 2: Decompose base gate without controls
+            base_gate = self.__class__(*self.init_args, **self.init_kwargs)
+            decomposed = base_gate._base_decompose(*free, use_toffolis=use_toffolis)
+            mask = self._control_mask_after_stripping(decomposed)
+            for m, g in zip(mask, decomposed):
+                if m:
+                    g.is_controlled_by = True
+                    g.control_qubits += self.control_qubits
```
renatomello:
```suggestion
            for bool_value, gate in zip(mask, decomposed):
                if bool_value:
                    gate.is_controlled_by = True
                    gate.control_qubits += self.control_qubits
```

src/qibo/gates/gates.py
```diff
@@ -2537,7 +2537,7 @@ def congruent(self, use_toffolis: bool = True) -> List[Gate]:
             applying the original ``TOFFOLI`` gate.
         """
         if use_toffolis:
```
renatomello:
```suggestion
```

src/qibo/gates/gates.py
```diff
@@ -2537,7 +2537,7 @@ def congruent(self, use_toffolis: bool = True) -> List[Gate]:
...
-            return self.decompose()
+            return self._base_decompose()
```
renatomello:
```suggestion
```

src/qibo/gates/abstract.py
BrunoLiegiBastonLiegi: Ah yeah, sure, sorry this applies to idempotent gates only naturally

tests/test_gates_abstract.py
```diff
@@ -422,6 +424,63 @@ def test_decompose():
...
+        (gates.CNOT(0, 2), gates.CNOT(1, 2), False),
+        (gates.TOFFOLI(0, 1, 2), gates.TOFFOLI(0, 1, 2), True),
+        (gates.CCZ(0, 1, 2), gates.CCZ(0, 1, 2), True),
+    ],
+)
+def test_gates_cancel(g1, g2, expected):
+    result = Gate._gates_cancel(g1, g2)
+    assert result is expected
+
+
```
renatomello, line 483:
```suggestion
def test_toffoli_congruent(backend):
    congruent = gates.TOFFOLI(0, 1, 2)

    circuit = Circuit(3)
    circuit.add(congruent.congruent())
    congruent = circuit.unitary(backend)

    target = backend.matrices.TOFFOLI
    target[4, 4] = -1

    assert backend.calculate_matrix_norm(congruent - target) < 1e-8
```

renatomello:
BTW @GiacomoFrn: https://github.com/qiboteam/qibo/discussions/1659

PR #1662, opened 2025-06-06
Add FHT-based implementation for Pauli-Liouville transformation

santiagolave:
Checklist:
- [ ] Tests are passing.

This implementation follows the algorithm from the paper: Pauli Decomposition via the Fast Walsh-Hadamard Transform, specifically Theorem 1 + Equation (34). It efficiently computes the Pauli-Liouville representation of an operator A ∈ ℂ^{2ⁿ × 2ⁿ} using the Fast Walsh-Hadamard Transform (FHT).

This function can be registered as the default method inside to_pauli_liouville(...):
`def to_pauli_liouville(channel, method=None, ...):
    if method is None:
        return to_pauli_liouville_fht(channel, ...)
    ...
`

### DIFF
--- a/src/qibo/quantum_info/superoperator_transformations.py
+++ b/src/qibo/quantum_info/superoperator_transformations.py
@@ -217,6 +217,64 @@ def to_liouville(channel, order: str = "row", backend=None):
 
     return channel
 
+def to_pauli_liouville_fht(matrix, normalize=False, backend=None):
+    """
+    Converts a square matrix A ∈ ℂ^{2ⁿ × 2ⁿ} into its Pauli-Liouville representation
+    using the Fast Walsh-Hadamard Transform (FHT), based on:
+    "Pauli Decomposition via the Fast Walsh-Hadamard Transform", Sec. 3.
+
+    Args:
+        matrix (ndarray): Input matrix of shape (2^n, 2^n).
+        normalize (bool): Whether to normalize the Pauli basis.
+        backend (Qibo backend, optional): Computational backend.
+
+    Returns:
+        ndarray: Pauli-Liouville matrix of shape (4^n, 4^n) with coefficients α_{r,s}
+                 such that A = ∑_{r,s} α_{r,s} ⋅ P_r ⋅ P_s†
+    """
+    from qibo.quantum_info import vectorization
+    from qibo.backends import _check_backend
+
+    backend = _check_backend(backend)
+
+    dim = matrix.shape[0]
+    nqubits = int(np.log2(dim))
+    if matrix.shape[1] != dim or 2**nqubits != dim:
+        raise ValueError("Input matrix must be square with dimension 2^n.")
+
+    A = backend.cast(matrix, dtype=complex)
+
+    # Step 1: XOR permutation
+    A_xor = backend.np.empty_like(A)
+    for r in range(dim):
+        for s in range(dim):
+            A_xor[r, s] = A[r ^ s, s]
+
+    # Step 2: Walsh-Hadamard transform over columns
+    H1 = backend.cast([[1, 1], [1, -1]], dtype=complex)
+    Hn = H1
+    for _ in range(nqubits - 1):
+        Hn = backend.np.kron(Hn, H1)
+    A_hat = backend.np.dot(A_xor, Hn)
+
+    # Step 3: Phase factor (-i)^wt(r & s)
+    r = backend.np.arange(dim).reshape(-1, 1)
+    s = backend.np.arange(dim).reshape(1, -1)
+    r_and_s = backend.np.bitwise_and(r, s)
+
+    def hamming_weight(x):
+        return backend.np.array([bin(v).count("1") for v in x.flat]).reshape(x.shape)
+
+    wt = hamming_weight(r_and_s)
+    phase = backend.np.power(-1j, wt)
+
+    coeffs = A_hat * phase / (2 ** nqubits)
+
+    if normalize:
+        coeffs /= backend.np.sqrt(2**nqubits)
+
+    return coeffs
+
 
 def to_pauli_liouville(
     channel,
@@ -262,7 +320,7 @@ def to_pauli_liouville(
     channel = unitary @ channel @ backend.np.conj(unitary).T
 
     return channel
-
+    
 
 def to_chi(
     channel,


src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -218,6 +218,65 @@ def to_liouville(channel, order: str = "row", backend=None):
...
+        matrix (ndarray): Input matrix of shape (2^n, 2^n).
+        normalize (bool): Whether to normalize the Pauli basis.
+        backend (Qibo backend, optional): Computational backend.
+
+    Returns:
+        ndarray: Pauli-Liouville matrix of shape (4^n, 4^n) with coefficients α_{r,s}
+                 such that A = ∑_{r,s} α_{r,s} ⋅ P_r ⋅ P_s†
+    """
+    from qibo.backends import _check_backend
+    from qibo.quantum_info import vectorization
```
renatomello:
```suggestion
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -218,6 +218,65 @@ def to_liouville(channel, order: str = "row", backend=None):
     return channel
 
 
+def to_pauli_liouville_fht(matrix, normalize=False, backend=None):
```
renatomello:
```suggestion
def _to_pauli_liouville_fht(matrix, normalize: bool = False, backend=None):
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -218,6 +218,65 @@ def to_liouville(channel, order: str = "row", backend=None):
...
+    """
+    Converts a square matrix A ∈ ℂ^{2ⁿ × 2ⁿ} into its Pauli-Liouville representation
+    using the Fast Walsh-Hadamard Transform (FHT), based on:
+    "Pauli Decomposition via the Fast Walsh-Hadamard Transform", Sec. 3.
+
+    Args:
+        matrix (ndarray): Input matrix of shape (2^n, 2^n).
+        normalize (bool): Whether to normalize the Pauli basis.
```
renatomello:
```suggestion
        normalize (bool, optional): If ``True``, uses the normalized Pauli basis.
            Defaults to ``False``.
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -218,6 +218,65 @@ def to_liouville(channel, order: str = "row", backend=None):
...
+        backend (Qibo backend, optional): Computational backend.
```
renatomello:
```suggestion
        backend (:class:`qibo.backends.abstract.Backend`, optional): backend
            to be used in the execution. If ``None``, it uses
            the current backend. Defaults to ``None``.
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -218,6 +218,65 @@ def to_liouville(channel, order: str = "row", backend=None):
...
+    if matrix.shape[1] != dim or 2**nqubits != dim:
+        raise ValueError("Input matrix must be square with dimension 2^n.")
+
+    A = backend.cast(matrix, dtype=complex)
+
+    # Step 1: XOR permutation
+    A_xor = backend.np.empty_like(A)
+    for r in range(dim):
+        for s in range(dim):
+            A_xor[r, s] = A[r ^ s, s]
```
renatomello: These loops can probably be vectorized (and therefore sped up). However, probably neither is the best solution (see the next comment below).

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -218,6 +218,65 @@ def to_liouville(channel, order: str = "row", backend=None):
...
+
+    # Step 2: Walsh-Hadamard transform over columns
+    H1 = backend.cast([[1, 1], [1, -1]], dtype=complex)
+    Hn = H1
+    for _ in range(nqubits - 1):
+        Hn = backend.np.kron(Hn, H1)
+    A_hat = backend.np.dot(A_xor, Hn)
```
renatomello:
```suggestion
    from functools import reduce

    H1 = float(np.sqrt(2)) * backend.matrices.H
    Hn = reduce(backend.np.kron, [H1]*nqubits)
    A_hat = A_xor @ Hn
```

However, constructing the entire Hadamard transform matrix at once is very memory and time consuming. I'm quite sure the entire point of applying an FHT is to avoid having all the elements of this matrix in memory at once.  A good solution avoids this step.

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -218,6 +218,65 @@ def to_liouville(channel, order: str = "row", backend=None):
...
+
+    # Step 3: Phase factor (-i)^wt(r & s)
+    r = backend.np.arange(dim).reshape(-1, 1)
+    s = backend.np.arange(dim).reshape(1, -1)
+    r_and_s = backend.np.bitwise_and(r, s)
+
+    def hamming_weight(x):
+        return backend.np.array([bin(v).count("1") for v in x.flat]).reshape(x.shape)
```
renatomello:
```suggestion
from qibo.quantum_info.utils import hamming_weight
```

src/qibo/quantum_info/superoperator_transformations.py
santiagolave: Thanks for your response! I will implement this change.

santiagolave:
Hi! I have implemented the mentioned changes in the code implementation, however I had some problems with hamming-weight Qibo implementation, still figuring out how to applied to this code. I have also created a new function "from_pauli_liouville_fht" which uses the coefficients matrix to return the corresponding channel. However, in order to use this method another functions had to be created, while they appear in the "src/qibo/quantum_info/superoperator_transformations.py" file, I think they should be added to the "src/qibo/quantum_info/utils.py". At this point I'm sure some memorie issues will appear regarding the calculation of the tensor products for the X and Z Pauli matrices, looking forward to seeing these corrections!

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -218,6 +218,121 @@ def to_liouville(channel, order: str = "row", backend=None):
...
+        backend (:class:`qibo.backends.abstract.Backend`, optional): Backend
+            to be used in the execution. If ``None``, it uses the current backend.
+
+    Returns:
+        ndarray: Matrix of shape (4^n, 4^n) with coefficients α_{r,s}
+            such that A = ∑_{r,s} α_{r,s} ⋅ P_r ⋅ P_s†.
+    """
+    from functools import reduce
+
+    from qibo.backends import _check_backend
```
renatomello:
```suggestion
```
Already imported at top level

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -218,6 +218,121 @@ def to_liouville(channel, order: str = "row", backend=None):
...
+        backend (:class:`qibo.backends.abstract.Backend`, optional): Backend
+            to be used in the execution. If ``None``, it uses the current backend.
+
+    Returns:
+        ndarray: Reconstructed matrix A ∈ ℂ^{2^n × 2^n}.
+    """
+    import numpy as np
+    from numpy.linalg import matrix_power
+
+    from qibo.backends import _check_backend
```
renatomello:
```suggestion
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -218,6 +218,121 @@ def to_liouville(channel, order: str = "row", backend=None):
...
+
+    def hamming_weight(a, b):
+        return backend.np.sum(a * b)
+
+    def pauli_operator(r, s):
+        X = np.array([[0, 1], [1, 0]])
+        Z = np.array([[1, 0], [0, -1]])
+        bin_r = int_to_binary_array(r, n)
+        bin_s = int_to_binary_array(s, n)
+        op = matrix_power(X, bin_r[0]) @ matrix_power(Z, bin_s[0])
```
renatomello:
```suggestion
        op = backend.calculate_matrix_power(X, bin_r[0]) @ backend.calculate_matrix_power(Z, bin_s[0])
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -218,6 +218,121 @@ def to_liouville(channel, order: str = "row", backend=None):
...
+        for i in range(1, n):
+            xi = matrix_power(X, bin_r[i])
+            zi = matrix_power(Z, bin_s[i])
```
renatomello:
```suggestion
            xi = backend.calculate_matrix_power(X, bin_r[i])
            zi = backend.calculate_matrix_power(Z, bin_s[i])
```

src/qibo/quantum_info/superoperator_transformations.py
renatomello: `qibo` backends have their own `matrix_power` implementation due to requiring compatibility with other engines than `numpy` (e.g. `cupy`, `torch`).

renatomello:
> Hi! I have implemented the mentioned changes in the code implementation, however I had some problems with hamming-weight Qibo implementation, still figuring out how to applied to this code. I have also created a new function "from_pauli_liouville_fht" which uses the coefficients matrix to return the corresponding channel. However, in order to use this method another functions had to be created, while they appear in the "src/qibo/quantum_info/superoperator_transformations.py" file, I think they should be added to the "src/qibo/quantum_info/utils.py". At this point I'm sure some memorie issues will appear regarding the calculation of the tensor products for the X and Z Pauli matrices, looking forward to seeing these corrections!

Maybe you could take a first attempt that I took at this problem in https://github.com/qiboteam/qibo/pull/1633. I'm curious if you could compare time and memory complexity with and without completely constructing the Hadamard matrix in memory. Hopefully you can find a way to combine the best of both attempted solutions. I'd also highlight the second paragraph after Theorem 1 in the paper, it explains why a naive implementation is not ideal.

renatomello:
@santiagolave any progress?

santiagolave:
Sure it is! Hi again! I wanted to share the results of a preliminary performance analysis of my implementation of the to_pauli_liouville_fht function and highlight some optimization opportunities. Using %lprun, I profiled the function on a random 4×4 density matrix (i.e., 2 qubits). The top time-consuming lines were:

Line | Operation                                                     | Time (μs) | % Total Time
---- | ----------------------------------------------- | -----------| -------------
58   | hamming_weight computation (loop)          | 1808.0     | 33.5%
59   | Phase factor: backend.np.power                   | 1544.8     | 28.6%
46   | XOR permutation: A[idx[:, None] ^ idx]        | 1062.8     | 19.7%
51   | Matrix multiplication: A_hat = A_xor @ Hn   | 492.8      | 9.1%
50   | Kronecker product for Hadamard matrix      | 261.1       | 4.8%

So overall, more than 60% of the total time is spent computing the Hamming weights and the complex phase factors. This confirms that we do not need need to avoid constructing the full Hadamard matrix explicitly, but find a better way to calculating the Hamming weights.

This was executed on a personal laptop with the following specs:

💻 CPU: 12th Gen Intel(R) Core(TM) i5-1235U
🧠 RAM: 8 GB DDR4 @ 3200 MT/s
📦 Disk: 477GB SSD (NVMe)
🎮 GPU: Intel Iris Xe (shared memory 3.8 GB)
OS: Windows 11, Python 3.11 (Conda env)

For the comparation between the code in proposed in problem [#1633](https://github.com/qiboteam/qibo/pull/1633) and my current implementation I have found that it is pretty obvious an implementation must be made. Which is the step I'm currently at.
descarga

renatomello:
> Sure it is! Hi again! I wanted to share the results of a preliminary performance analysis of my implementation of the to_pauli_liouville_fht function and highlight some optimization opportunities. Using %lprun, I profiled the function on a random 4×4 density matrix (i.e., 2 qubits). The top time-consuming lines were:
> Line 	Operation 	Time (μs) 	% Total Time
> 58 	hamming_weight computation (loop) 	1808.0 	33.5%
> 59 	Phase factor: backend.np.power 	1544.8 	28.6%
> 46 	XOR permutation: A[idx[:, None] ^ idx] 	1062.8 	19.7%
> 51 	Matrix multiplication: A_hat = A_xor @ Hn 	492.8 	9.1%
> 50 	Kronecker product for Hadamard matrix 	261.1 	4.8%
> 
> So overall, more than 60% of the total time is spent computing the Hamming weights and the complex phase factors. This confirms that we do not need need to avoid constructing the full Hadamard matrix explicitly, but find a better way to calculating the Hamming weights.
> 
> This was executed on a personal laptop with the following specs:
> 
> 💻 CPU: 12th Gen Intel(R) Core(TM) i5-1235U 🧠 RAM: 8 GB DDR4 @ 3200 MT/s 📦 Disk: 477GB SSD (NVMe) 🎮 GPU: Intel Iris Xe (shared memory 3.8 GB) OS: Windows 11, Python 3.11 (Conda env)
> 
> For the comparation between the code in proposed in problem [#1633](https://github.com/qiboteam/qibo/pull/1633) and my current implementation I have found that it is pretty obvious an implementation must be made. Which is the step I'm currently at. descarga

Thanks. Which color is each implementation here?

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -218,6 +218,121 @@ def to_liouville(channel, order: str = "row", backend=None):
...
+            such that A = ∑_{r,s} α_{r,s} ⋅ P_r ⋅ P_s†.
+    """
+    from functools import reduce
+
+    from qibo.backends import _check_backend
+
+    backend = _check_backend(backend)
+
+    dim = matrix.shape[0]
+    nqubits = int(np.log2(dim))
```
Newtech66: `np.log2` may be unstable: consider using `int(dim).bit_length() - 1` as an alternative which does not require floating point.

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -218,6 +218,121 @@ def to_liouville(channel, order: str = "row", backend=None):
...
+
+    Args:
+        matrix (ndarray): Input matrix of shape (2^n, 2^n).
+        normalize (bool, optional): If ``True``, uses the normalized Pauli basis.
+            Defaults to ``False``.
+        backend (:class:`qibo.backends.abstract.Backend`, optional): Backend
+            to be used in the execution. If ``None``, it uses the current backend.
+
+    Returns:
+        ndarray: Matrix of shape (4^n, 4^n) with coefficients α_{r,s}
```
Newtech66: I want to understand: why return a $4^n \times 4^n$ matrix? That unnecessarily makes your algorithm $O(16^n)$.

Newtech66: I still don't quite understand the problem statement of this issue admittedly, but as far as I understand, the approach here should be to take a $2^n \times 2^n$ matrix to a $4^n$ vector that encodes the coefficients in the Pauli basis. And to implement the reverse direction: from the $4^n$ vector to a $2^n \times 2^n$ matrix. A $4^n \times 4^n$ matrix seems unneccesary.

renatomello:
The representation of states in the Paulo basis yield vectors of dimension $4^n$. The representation of channels yields matrices of dimension $4^n \times 4^n$.

santiagolave:
Hi! I forgot to put labels on the graph, sorry! The code implemented in #1633 corresponds to the blue curve and my implementation corresponds to the orange one. I have optimized the hamming_weight funcion through a vectorization in order to calcule all the exponents, and this lead to a reduction of time complexity in my implementation, however it still fell behind the one in problem #1633, so I also took the divide-and-conquer approach for the FWH transformation. Finally I once again measured time complexity as shown on the graph below 
download

santiagolave:
Hi! I have just pushed a new update to this pull request.

This update includes:

- A vectorized implementation of the hamming_weight function.
- A new divide-and-conquer implementation for the Fast Walsh-Hadamard Transform, following the expression described in the literature.
- Overall improvements in execution time, especially noticeable from 5 qubits onwards (see benchmark below).

This optimized version significantly reduces the time complexity compared to my previous version (orange curve) and now performs comparably to the implementation in #1633 (blue curve).

Looking forward to your feedback!

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -218,6 +220,130 @@ def to_liouville(channel, order: str = "row", backend=None):
...
+        backend: Qibo backend.
+    Returns:
+        Array with the same shape, containing the Hamming weight of each element.
+    """
+    arr = backend.np.array(arr, dtype=backend.np.uint64)
+    weights = backend.np.zeros_like(arr)
+    while backend.np.any(arr):
+        weights += arr & 1
+        arr >>= 1
+    return weights
```
renatomello, line 237: What's exactly the input of this function? It seems an extra complication when we already have an implementation for the Hamming weight function.

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -218,6 +220,129 @@ def to_liouville(channel, order: str = "row", backend=None):
...
+            Defaults to ``False``.
+        backend (:class:`qibo.backends.abstract.Backend`, optional): Backend
+            to be used in the execution. If ``None``, it uses the current backend.
+
+    Returns:
+        ndarray: Matrix of shape (4^n, 4^n) with coefficients α_{r,s}
+            such that A = ∑_{r,s} α_{r,s} ⋅ P_r ⋅ P_s†.
+    """
+
+
```
renatomello:
```suggestion

```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -218,6 +220,127 @@ def to_liouville(channel, order: str = "row", backend=None):
...
+    dim = coeffs.shape[0]
+    n = int(np.log2(dim))
+    d = 2**n
+    A = backend.np.zeros((d, d), dtype=complex)
+
+    def int_to_binary_array(x, bits):
+        return backend.np.array([int(b) for b in format(x, f"0{bits}b")])
+
+    def hamming_weight(a, b):
+        return backend.np.sum(a * b)
```
renatomello: Same here. If what you're looking for is Hamming distance between two bitstrings, `qibo.quantum_info.utils` already has an implementation for it.

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -218,6 +220,127 @@ def to_liouville(channel, order: str = "row", backend=None):
...
+
+    backend = _check_backend(backend)
+
+    dim = coeffs.shape[0]
+    n = int(np.log2(dim))
+    d = 2**n
+    A = backend.np.zeros((d, d), dtype=complex)
+
+    def int_to_binary_array(x, bits):
+        return backend.np.array([int(b) for b in format(x, f"0{bits}b")])
```
renatomello: Besides being probably unnecessary, the name of this function is `int_to_binary` but it seems to be converting binary to ints.

santiagolave:
### ✅ **Summary of changes in this pull request**

This pull request introduces an optimized implementation of the **Pauli-Liouville transformation using the Fast Hadamard Transform (FHT)**, with improvements in both efficiency and clarity:

#### 🔹 Main updates

1. **Implementation of `_to_pauli_fht`**

   * Converts a $2^n \times 2^n$ matrix into its Pauli-Liouville representation.
   * Uses in place computation.
   * Returns a matrix of shape $2^n \times 2^n$  with the coefficients $\alpha_{r,s}$.
   * Includes optional normalization by $\sqrt{2^n}$.

2. **New function `_from_pauli_fht`**

   * Reconstructs the original matrix $A$ from its coefficients using the expansion:

     $$A = \sum_{r,s} \alpha_{r,s} P_r P_s^\dagger$$
   * Each $P_r P_s^\dagger$ operator is built as a Kronecker product of Pauli matrices $X$, $Z$, and identity, according to the binary representations of $r$ and $s$.
   * The implementation avoids unnecessary computation by skipping terms with $\alpha_{r,s} = 0$.

---

### ⚠️ Additional note

I attempted to implement `_from_pauli_fht` using the **Walsh-Hadamard Transform** approach described in Equation (30) of the original paper. However, I encountered persistent issues involving incorrect signs and mismatches in the real and imaginary parts of the reconstructed matrix. To ensure correctness, I replaced it with a direct reconstruction based on the explicit expansion into Pauli operators, which consistently reproduces the input matrices across all tested cases.

---

Knowing that the Hackaton is about to close I will be glad to have a final feedback! Thank you very much!

santiagolave:
Hi! Just wanted to check the revision of this implementation.

renatomello:
@santiagolave could you merge the master branch to this one?

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -218,6 +220,115 @@ def to_liouville(channel, order: str = "row", backend=None):
...
+        for ind_col in q_array:
+            elem = 0
+            for q_ind in q_array:
+                q_xor_r = backend.np.bitwise_xor(q_ind, ind_row)
+                hadamard_exp = hamming_weight(
+                    int(backend.np.bitwise_and(q_ind, ind_col))
+                )
+                elem += operator[q_xor_r, q_ind] * ((-1) ** hadamard_exp)
+            hadamard_exp = hamming_weight(int(backend.np.bitwise_and(ind_row, ind_col)))
+            alphas[ind_row, ind_col] = (elem * (1j**-hadamard_exp)) / dims
```
BrunoLiegiBastonLiegi, line 274: This thrice nested loop is not very readable, nor efficient: I think you are repeating some operations multiple times, e.g. ``q_xor_r = backend.np.bitwise_xor(q_ind, ind_row)`` is repetead for each ``ind_col`` but will yield the same result everytime since ``ind_row`` is fixed. Therefore, I would advocate for leveraging the vectorization capabilities of numpy and other engines like it. This is a quick example of how this loop could be done using vector operations (please double check it! I just quickly sketched it just to give you an overall idea):
```suggestion
    q_xor_r = backend.np.bitwise_xor(q_array, backend.np.reshape(q_array, (-1,1)))
    q_and_r = backend.np.bitwise_and(q_array, backend.np.reshape(q_array, (-1,1)))
    hadamard_exp = backend.cast(
        [hamming_weight(int(i)) for i in q_and_r.ravel()],
    dtype=q_and_r.dtype).reshape(q_and_r.shape)
    elem = backend.sum(operator[q_xor_r, q_array] * ((-1) ** hadamard_exp), axis=-1)
    alphas = (elem * (1j**-hadamard_exp)) / dims
```

src/qibo/quantum_info/superoperator_transformations.py
```diff
@@ -218,6 +220,115 @@ def to_liouville(channel, order: str = "row", backend=None):
...
+                xi = X if bin_r[0] else backend.np.eye(2)
+                zi = Z if bin_s[0] else backend.np.eye(2)
+                op = backend.np.dot(xi, zi)
+
+                for i in range(1, nqubits):
+                    xi = X if bin_r[i] else backend.np.eye(2)
+                    zi = Z if bin_s[i] else backend.np.eye(2)
+                    op = backend.np.kron(op, backend.np.dot(xi, zi))
+                exponent = hamming_weight(int(backend.np.bitwise_and(ind_row, ind_col)))
+                a_matrix += coeff * backend.cast(op, dtype=complex) * (1j**exponent)
```
BrunoLiegiBastonLiegi, line 327: Same as above: try thinking how you could vectorize these loops.

Issue #1643, opened 2025-05-19
Product of two SymbolicHamiltonian's not being reduced completely


chmwzc:
**Describe the bug**
@jf-kong was trying to implement the [folded spectrum Hamiltonian approach](https://github.com/qiboteam/qibochem/pull/136) for excited states: Variationally optimize $`(H - \omega)^2`$ instead of just $`H`$.
In the process, we found that the product of two `SymbolicHamiltonians` wasn't being simplified completely:

**To Reproduce**
```
from qibo.hamiltonians import SymbolicHamiltonian
from qibo.symbols import Z

a1 = SymbolicHamiltonian(0.2*Z(0)*Z(1) - 0.5)
a2 = SymbolicHamiltonian(Z(0), nqubits=a1.nqubits)
a = a1 @ a2
for term in a.terms:
    print(f"{[factor for factor in term.factors]}")
# Output:
# [Z0]
# [Z0, Z1, Z0]
```

**Expected behavior**
It's not wrong, but it would be good if the `[Z0, Z1, Z0]` term was reduced to `[Z1]` directly.

**Additional context**
I have a rough idea of how to do it, but wanted to ask for comments before creating a pull request.
Basically, sympy has a [`evaluate_pauli_product`](https://docs.sympy.org/latest/modules/physics/paulialgebra.html) function to evaluate the product of Pauli operators, so I'm thinking to add a new class method (`reduce`? `expand`? `simplify`? I'm open to suggestions about the method name) to `SymbolicHamilonian` that does something like this:

1. Expand `SymbolicHamiltonian.form`
2. For each term in the expanded form:
    a. Map every `qibo.symbol` in each factor to a sympy `Pauli` object
    b. Use the Sympy `evaluate_pauli_product` function to get the final Pauli operator for each qubit
    b. Map the Pauli object back to `qibo.symbol`'s
3. Update `SymbolicHamiltonian.form` with the simplified terms

There's probably some refactoring that can be done in `SymbolicHamiltonian.terms` and `SymbolicTerm` as well, will look at that too if the above pseudocode looks OK.

If I'm not wrong, this should be under @BrunoLiegiBastonLiegi?

BrunoLiegiBastonLiegi:
Hey, thanks, yeah you're perfectly right, I think when I did the refactor I did not spend much time in optimizing this. There actually is a simplification step in place for powers of paulis here:

[...]

PR #1664, opened 2025-06-10
Merge and simplify common terms in a SymbolicHamiltonian after expanding

chmwzc:
My attempted fix for #1643. The main changes are to the code for `SymbolicTerm`, to merge and combine different factors acting on a qubit if they are Pauli operators.

Unfortunately, the tests for `TSP` (`test_models_tsp.py`) are failing, and I'm not familiar with it, could I get any advice on what went wrong there?

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/hamiltonians/terms.py
+++ b/src/qibo/hamiltonians/terms.py
@@ -192,7 +192,7 @@ class SymbolicTerm(HamiltonianTerm):
 
     @staticmethod
     def _compute_power(factor):
-        """Break down a factor into a base term and its power, e.g. `X**n` returns `X` and `n`. In case the base term is a Pauli also simplifies the power calculation, i.e. return the identity for even `n` and the Pauli itsel for odd `n`."""
+        """Break down a factor into a base term and its power, e.g. `X**n` returns `X` and `n`. In case the base term is a Pauli also simplify the power calculation, i.e. return the identity for even `n` and the Pauli itself for odd `n`."""
         if isinstance(factor, sympy.Pow):
             factor, power = factor.args
             assert isinstance(power, sympy.Integer)


mho291:
Maybe we can ignore the tests for TSP, since we might move it to qibo comb optimisation. Right @shangtai)?

renatomello:
> Maybe we can ignore the tests for TSP, since we might move it to qibo comb optimisation. Right @shangtai)?

We can't ignore the test without at least knowing why it is failing.

src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -357,10 +357,17 @@ def terms(self):
...
             if term.target_qubits:
-                terms.append(term)
+                # Check for any terms with the same factors and add their coefficients together
+                same_terms = [
+                    _term for _term in terms if set(_term.factors) == set(term.factors)
+                ]
+                term.coefficient = term.coefficient + sum(
+                    _term.coefficient for _term in same_terms
+                )
+                terms = [_term for _term in terms if _term not in same_terms] + [term]
```
BrunoLiegiBastonLiegi: Do you actually need to skim the complete list of terms? In principle, if everything goes as expected, you can stop as soon as you find one, that should be the only one (as you're supposed to merge together those sharing the same factors).
```python
found = False
for i, _term in enumerate(terms):
    if set(_term.factors) == set(term.factors):
        found = True
        terms[i].coefficient += term.coefficient
        break
if not found:
    terms.append(term)
```

src/qibo/hamiltonians/terms.py
```diff
@@ -150,44 +150,102 @@ def __init__(self, coefficient, factors=1, backend: Optional[Backend] = None):
         # Dictionary that maps target qubit ids to a list of matrices that act on each qubit
         self.matrix_map = {}
         if factors != 1:
+            # For mapping between qibo symbols and sympy Pauli objects
+            pauli_mapping = {X: Pauli(1), Y: Pauli(2), Z: Pauli(3)}
+            inverse_pauli_mapping = {_v: _k for _k, _v in pauli_mapping.items()}
```
BrunoLiegiBastonLiegi: Mmmh at this point I wonder whether we could make our ``PauliSymbol`` 

https://github.com/qiboteam/qibo/blob/e100c5bd4aef8c12b0250fc1713ecbebd8102e09/src/qibo/symbols.py#L131

inherit from ``sympy.physics.paulialgebra.Pauli`` directly:
```python
class PauliSymbol(Symbol, Pauli)
```

src/qibo/hamiltonians/terms.py
```diff
@@ -150,44 +150,102 @@ def __init__(self, coefficient, factors=1, backend: Optional[Backend] = None):
...
+                                    if (
+                                        q_factor.__class__ in pauli_mapping
+                                        and self.backend.np.allclose(
+                                            q_factor.matrix, self.matrix_map[q][-1]
+                                        )
+                                    ):
+                                        pauli_product = evaluate_pauli_product(
+                                            pauli_mapping[q_factor.__class__]
+                                            * pauli_mapping[factor.__class__]
+                                        )
```
BrunoLiegiBastonLiegi: ok I am trying to follow this, you are basically looking for all the factors you appended to ``self.factors`` already that share current factor's qubit and are paulis. Then, you look for pairs of paulis that sit close to each other and simplify them with sympy. If that's what's going on, I was wondering why don't you instead look forward for any following pauli on the same qubit whenever you find one?
```python
n = 0
while n < len(factors):
    factor = factors[n]
    # ... all the power stuff
    # ... and the other checks
    if isinstance(factor, (X, Y, Z)):
        k = n + 1
        paulis = [factor]
        while k < len(factors):
            next_factor = factors[k]
            if next_factor.target_qubit == factor.target_qubit:
                if next_factor.__class__ in pauli_mapping:
                   paulis.append(next_factor)
                else:
                    break
            k += 1
    n = k
    if len(paulis > 1):
        pauli_product = evaluate_pauli_product(np.prod(paulis))
    # ... then all the coefficient adjustments
    # .... checks and matrix_map/factors updates, etc...
```

BrunoLiegiBastonLiegi:
I did some checks for the test that is failing. The final state is indeed deviating from the one saved in ``regressions/`` more than expected, however I manually checked the parameters learned in training and they are very close to those you obtain in ``master``:
```
# this branch
2 layers: [0.1        0.09820968]
4 layers: [0.1        0.09796367 0.10000245 0.09796367]
# master
2 layers: [0.1        0.09794202]
4 layers: [0.1        0.09812937 0.10000213 0.09812812]
```
Thus I think this is just a problem of training finding a slightly different minimum, that produces a different but compatible final state. We had similar problems with some backends in this test already, and probably fixing a target final state for a variational problem that depends on training was not a good idea in the first place. It would be better in my opinion, for instance to check that a target energy is reached.
Therefore, for now I just increased the tolerance, but we might want to reformulate this (and probably the whole variational module in light of ``qiboml``) in the future.

src/qibo/hamiltonians/terms.py
chmwzc: Yeah, that feels a lot cleaner that what I did here, I'll go try it out

src/qibo/hamiltonians/terms.py
chmwzc: > you are basically looking for all the factors you appended to self.factors already that share current factor's qubit and are paulis. Then, you look for pairs of paulis that sit close to each other and simplify them with sympy

Yup, that's what I was trying to do

> look forward for any following pauli on the same qubit whenever you find one

Initial thoughts: Need to keep track of what factor has been simplified through the `for factor in factors.as_ordered_factors():` loop if we try to look forward instead. Let me try it out👍

src/qibo/hamiltonians/terms.py
chmwzc: - Tried and couldn't figure out how to do the multiclass inheritance; it's beyond my poor Python skills 😢
- Instead, I'm trying to use a function (`_one_qubit_paulis_string_product`) from `qibo.quantum_info._clifford_utils` directly, but ran into some circular import error when I tried to import and use it in `terms.py`
  - Have raised an issue #1683 for that

BrunoLiegiBastonLiegi: Yeah, thanks, I tried as well. I managed to get it to work to some extent but it is complicated and causes problems in several scenarios. We can keep it as is, thanks for trying this out.

src/qibo/hamiltonians/terms.py
chmwzc: Tried it, but it seems like looking forward is a lot more complicated.
The factors being given as powers won't be reduced if you try to look forward, e.g. something like `(X(0)**3) * (Z(0)**3)` would require the code about powers to be repeated

chmwzc:
I have completely no idea why, but the TSP tests are passing again now...🤷‍♂️😅

src/qibo/hamiltonians/terms.py
```diff
@@ -150,44 +150,97 @@ def __init__(self, coefficient, factors=1, backend: Optional[Backend] = None):
         # Dictionary that maps target qubit ids to a list of matrices that act on each qubit
         self.matrix_map = {}
         if factors != 1:
+            # For mapping between qibo symbols and sympy Pauli objects
+            pauli_mapping = {X: Pauli(1), Y: Pauli(2), Z: Pauli(3)}
+            inverse_pauli_mapping = {_v: _k for _k, _v in pauli_mapping.items()}
```
BrunoLiegiBastonLiegi: these could be defined globally without redefining them everytime:
```python
PAULI_MAPPING = {X: Pauli(1), Y: Pauli(2), Z: Pauli(3)}
INVERSE_PAULI_MAPPING = {_v: _k for _k, _v in PAULI_MAPPING.items()}
```

chmwzc: > defining two separate methods

Thanks, just a note, I did the second method slightly differently from what you suggested

src/qibo/hamiltonians/terms.py
```diff
@@ -199,6 +190,66 @@ def __init__(self, coefficient, factors=1, backend: Optional[Backend] = None):
 
         self.target_qubits = tuple(sorted(self.matrix_map.keys()))
 
+    @staticmethod
+    def _compute_power(factor):
+        """Break down a factor into a base term and its power, e.g. `X**n` returns `X` and `n`"""
```
BrunoLiegiBastonLiegi:
```suggestion
        """Break down a factor into a base term and its power, e.g. `X**n` returns `X` and `n`. In case the base term is a Pauli also simplifies the power calculation, i.e. return the identity for even `n` and the Pauli itsel for odd `n`."""
```
just a small addition

BrunoLiegiBastonLiegi (approved):
Perfect, thank you so much!

src/qibo/hamiltonians/terms.py
```diff
@@ -199,6 +190,66 @@ def __init__(self, coefficient, factors=1, backend: Optional[Backend] = None):
...
+                else:
+                    power = 1
+            else:
+                power = int(power)
+        else:
+            power = 1
+        return factor, power
+
+    def _simplify_q_factors(self, factor, q):
+        """Simplifies the factors acting on qubit q using Pauli algebra after factor is added to q"""
```
MatteoRobbiati:
```suggestion
        """Simplify the factors acting on qubit ``q`` using Pauli algebra after factor is added to ``q``."""
```

src/qibo/hamiltonians/terms.py
```diff
@@ -199,6 +190,66 @@ def __init__(self, coefficient, factors=1, backend: Optional[Backend] = None):
...
+        q_factors = [_factor for _factor in self.factors if _factor.target_qubit == q]
+        while q_factors:
+            q_factor = q_factors[-1]
+            # Only simplify if the last term in self.matrix_map[q] is a Pauli matrix
+            if q_factor.__class__ in PAULI_MAPPING:
+                pauli_product = evaluate_pauli_product(
```
MatteoRobbiati, line 221: I don't know how much this could impact on big Hamiltonians, but we could think to precompute the nine combinations and avoid calling the `evaluate_pauli_product` everytime.

MatteoRobbiati (commented):
Thanks a lot! Just a couple of comments, but LGTM.

src/qibo/hamiltonians/terms.py
chmwzc: Yeah, that was why I was looking at the `_one_qubit_paulis_string_product` function from `quantum_info._clifford_utils`; the combinations are hardcoded there.
If that worked (#1683), the code would probably be something like `_one_qubit_paulis_string_product(q_factor.__class__.__name__, factor.__class__.__name__)` instead, and then we could have skipped using `sympy.physics.paulialgebra` as well.

src/qibo/hamiltonians/terms.py
```diff
@@ -199,6 +190,66 @@ def __init__(self, coefficient, factors=1, backend: Optional[Backend] = None):
 
         self.target_qubits = tuple(sorted(self.matrix_map.keys()))
 
+    @staticmethod
+    def _compute_power(factor):
+        """Break down a factor into a base term and its power, e.g. `X**n` returns `X` and `n`. In case the base term is a Pauli also simplifies the power calculation, i.e. return the identity for even `n` and the Pauli itsel for odd `n`."""
```
MatteoRobbiati:
```suggestion
        """Break down a factor into a base term and its power, e.g. `X**n` returns `X` and `n`. In case the base term is a Pauli also simplify the power calculation, i.e. return the identity for even `n` and the Pauli itself for odd `n`."""
```

PR #1668, opened 2025-06-12
Bloch sphere 

niccololaurora:
The pr contains an implementation of the Bloch sphere for the `ui` module.
There are `tests` to check the correct behaviour of the sphere and a `tutorial`.

The sphere relies on 4 main functions: 
1. `bs.add_state()`: to add a state to the sphere;
2. `bs.add_vector()`: to add a vector to the sphere;
3. `bs.plot()`: to plot the sphere or to save in a .pdf file;
4. `bs.clear()`: to reset the sphere;

States and vectors can have two possible representations: `point` or `vector`.
An example is the following:

`
bs = Bloch()
circ = Circuit(1)
circ.add(gates.RX(0, theta=np.random.rand()))
circ.add(gates.RY(0, theta=np.random.rand()))
circ.add(gates.RZ(0, theta=np.random.rand()))
state = circ().state()

bs.add_state(state, color="black", mode="vector")
bs.add_vector(np.array([0,1,0]), color="black", mode="point")
bs.plot()
`

[multi_class_classification.pdf](https://github.com/user-attachments/files/20712227/multi_class_classification.pdf)

Checklist:
- [ ] Tests are passing.


examples/bloch_sphere/first_bloch_sphere.pdf
alecandido, line 1: This file is only an output of the notebook: if it is not going to be displayed anywhere (in the docs, I mean), I would actually gitignore it.

src/qibo/ui/bloch.py
alecandido, line 1: Using the global API of Matplotlib is kind of an antipattern in a package, stemming from the general refrain of use global objects (since you may pollute the global environment and cause unintended interactions).

It is fine to use `plt.figure()` as the constructor of a (pseudo-)local figure object[*].
And if you use `plt.figure()`, you need to keep using `plt.close()`.

Other than those, almost all the usage of functions prefixed by `plt.` is something that is better to avoid.

[*]: Matplotlib still keeps track of all figure objects in this way, but it is a minor effect, we could forget at this stage.

> pyplot maintains a reference to figures created with [`figure()`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.figure.html#matplotlib.pyplot.figure).
> https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.close.html#matplotlib.pyplot.close

The alternative would be to just use the actual [`Figure()`](https://matplotlib.org/stable/api/_as_gen/matplotlib.figure.Figure.html#matplotlib.figure.Figure) constructor.

src/qibo/ui/bloch.py
```diff
@@ -0,0 +1,407 @@
...
+        for color, point in zip(self.color_points, self.points):
+            self.ax.scatter(point[0], point[1], point[2], color=color, s=10)
+
+    def plot(self, save=False, filename="bloch_sphere.pdf"):
+        self.rendering()
+        self.ax.set_aspect("equal")
+        plt.axis("off")
+        plt.tight_layout()
+        if save:
+            plt.savefig(filename)
```
alecandido: You have a figure and axis handle, use those.

I.e.
```suggestion
        self.ax.axis("off")
        self.fig.tight_layout()
        if save:
            self.fig.savefig(filename)
```

src/qibo/ui/bloch.py
```diff
@@ -0,0 +1,407 @@
...
+            plt.close()
+        else:
+            plt.show()
```
alecandido: Do not use a single method to plot and save, they are two different features.

If there is a potentially repeated part, make it a separate method:
```suggestion
    # change this name!
    def _common(self):
        self.rendering()
        self.ax.set_aspect("equal")
        plt.axis("off")
        plt.tight_layout()
        
    def save(filename="bloch_sphere.pdf"):
        self._common()
        plt.savefig(filename)
        plt.close()

    def plot():
        self._common()
        plt.show()
```

tests/test_ui_bloch_sphere.py
```diff
@@ -0,0 +1,276 @@
...
+            circ.add(gates.RZ(q=0, theta=np.random.randn()))
+        state = circ().state()
+        states[j] = state
+
+    # --Sphere--
+    bs.add_state(states, color="red")
+    bs.plot()
+
+
+test_mismatch_colors_state_5()
```
alecandido: This seems a leftover.

src/qibo/ui/bloch.py
```diff
@@ -0,0 +1,407 @@
...
+            else:
+                if len(vector) == 2:
+                    x, y, z = self.coordinates(vector)
+                    self.points.append(np.array([x, y, z]))
+                    self.color_points.append(color)
+                else:
+                    self.points.append(vector)
+                    self.color_points.append(color)
+
+    def add_state(self, state, mode="vector", color="black"):
```
alecandido: Since the mode is one out of two, do not rely on strings (which effectively turn the domain into `"vector"` and everything else, including non string values), but rather use an `Enum`.

Best would be a [`StrEnum`](https://docs.python.org/3/library/enum.html#enum.StrEnum) which is also a `str`.
In any case, just accept a string, and validate it attempting to construct your enum out of it:

```py
from enum import Enum, auto

class Mode(Enum):
    vector = auto()
    point = auto()

# then you can construct it with
mode = Mode["vector"]
```

alecandido: If you forget about `auto()`, and manually assign the values, you can also use the default constructor (parentheses, instead of brackets)

```py
from enum import Enum

class Mode(Enum):
    vector = "vector"
    point = "point"

# then you can construct it with                                                                                                                                                                    
Mode("vector")
# but also
Mode["vector"]
# and it's idempotent - which is useful to also accept `Mode` as a valid argument
Mode(Mode("vector"))
````

File: examples/bloch_sphere/bloch_sphere.ipynb
Comment (alecandido) [inline examples/bloch_sphere/bloch_sphere.ipynb:1]: Delete the last empty cell

Comment (alecandido) [inline examples/bloch_sphere/bloch_sphere.ipynb:1]: Instead of repeating the noise generation three times, just use a function.

I.e. replace:
```py
for i in range(n):
    circ = Circuit(1)
    circ.add(gates.RY(q=0, theta=np.random.randn() * 0.1))
    circ.add(gates.RX(q=0, theta=np.random.randn() * 0.1))
    circ.add(gates.RZ(q=0, theta=np.random.randn() * 0.1))
    state = circ().state()
    states[i] = state
    color_states.append("red")
    mode_state.append("point")

bs.add_state(states, mode=mode_state, color=color_states)

for i in range(n):
    circ = Circuit(1)
    circ.add(gates.RX(q=0, theta=-np.pi / 2))
    circ.add(gates.RY(q=0, theta=np.random.randn() * 0.1))
    circ.add(gates.RZ(q=0, theta=np.random.randn() * 0.1))
    circ.add(gates.RX(q=0, theta=np.random.randn() * 0.1))
    state = circ().state()
    bs.add_state(state, mode="point", color="orange")

for i in range(n):
    circ = Circuit(1)
    circ.add(gates.RY(q=0, theta=np.random.randn() * 0.1))
    circ.add(gates.RX(q=0, theta=np.random.randn() * 0.1))
    circ.add(gates.RZ(q=0, theta=np.random.randn() * 0.1))
    state = circ(np.array([0, 1], dtype="complex")).state()
    bs.add_state(state, mode="point", color="blue")
```

with:
```py
def noisy_states(circ: Circuit) -> npt.NDArray:
    states = []

    for i in range(n):
        noise = Circuit(1)
        noise.add(gates.RY(q=0, theta=np.random.randn() * 0.1))
        noise.add(gates.RX(q=0, theta=np.random.randn() * 0.1))
        noise.add(gates.RZ(q=0, theta=np.random.randn() * 0.1))
        state = (circ + noise)().state()
        states.append(state)

    return np.array(states)

circ = Circuit(1)
states = noisy_states(circ)
bs.add_state(states, mode=["point"]*len(states), color=["red"]*len(states))

circ = Circuit(1)
circ.add(gates.RX(q=0, theta=-np.pi / 2))
bs.add_state(states, mode=["point"]*len(states), color=["orange"]*len(states))


circ = Circuit(1)
circ.add(gates.X(q=0))
bs.add_state(states, mode=["point"]*len(states), color=["blue"]*len(states))
```

> [!NOTE]
> The suggested code is untested, but it should work

src/qibo/ui/bloch.py
```diff
@@ -0,0 +1,407 @@
...
+            option_list = ["vector"]
+
+        element = [element[0]] + option_list * (length_states - length_option)
+        warning(
+            f"Mismatch between number of states ({length_states}) and colors ({length_option}). Defaulting missing "
+            + option
+            + " to '"
+            + option_list[0]
+            + "' to match the number of states."
+        )
```
alecandido: Avoid warnings in a library (i.e. a package). They are often ignored, especially when used in a non-interactive environment. And just adding noise.

You decide the behavior, and document it. If it's allowed, then you explain what will happen. If it's not allowed, just raise an error.

src/qibo/ui/bloch.py
```diff
@@ -0,0 +1,407 @@
...
+                raise_error(ValueError, "Unnormalized state detected.")
+        elif len(element) == 3:
+            norm = np.linalg.norm(element)
+            if not np.isclose(norm, 1.0):
+                raise_error(ValueError, "The vector does not lie on the Bloch sphere.")
+        else:
+            raise_error(
+                ValueError,
+                "States must have two components. Vectors must have three components.",
+            )
```
alecandido: Avoid as much as possible lengthy validations.

Establish your behavior, and assume the parameters passed behave according to the type you specified. If they are abused, that's caller's fault, and undefined behavior.

src/qibo/ui/bloch.py
alecandido, line 1: Docstrings all missing.

alecandido, line 1: I'd recommend using type hints.

src/qibo/ui/bloch.py
```diff
@@ -0,0 +1,407 @@
...
+        # Bool variable
+        self._shown = False
+
+        # Data
+        self.points = []
+        self.vectors = []
+
+        # Color data
+        self.color_points = []
+        self.color_vectors = []
```
alecandido: Limit the number of local variables as much as possible.

Especially regarding style, use [Matplotlib stylesheet facilities](https://matplotlib.org/stable/users/explain/customizing.html).
In particular, I'd recommend you to establish a default stylesheet specification as a single top-level dictionary constant. I.e.

```py
STYLE = {
    "figsize": [5, 5]
    ...
}
```

And instead of accessing it individually, just bulk set the stylesheet right before plotting. Possibly using a context, to minimize interaction with global variables. E.g.
```py
with mpl.rc_context(STYLE):
    plt.plot(data)
```
https://matplotlib.org/stable/users/explain/customizing.html#temporary-rc-settings

alecandido: I do not work that much with plots, but the idea is to compose stylesheets as dictionaries (e.g. using `dict.update()`), and only apply them at the very end, and temporary (not to affect other plots).

alecandido (commented):
First batch of comments

alecandido:
Btw, the CI is failing because of `pre-commit`. Please, install it in your repo, and make sure it runs before pushing.

src/qibo/ui/bloch.py
niccololaurora: Thank you @alecandido for the suggestions. I will use `Figure()` which seems the more robust choice.

alecandido: > Thank you @alecandido for the suggestions. I will use `Figure()` which seems the more robust choice.

That would be great, but pay attention that it will come at a price: showing them may be harder. Possibly not in Jupyter, but you could give it a try :)

src/qibo/ui/bloch.py
niccololaurora: Thanks @alecandido for the suggestion. I thought I should correct the user's misuse of the package. Should I delete all these functions and rely on the use of the documentation by the user?

alecandido: Indeed, that's the advice: do not fix or prevent user errors.

Instead, what I would suggest is to keep (or implement, if missing) broadcasting semantics for colors and mode.
I.e. if the user only specifies a single color for multiple points, that's intended to apply to all points.

In particular, using type hints, you could have something like the following:
```py
def add_state(
    self,
    state: Union[State, list[State]],
    mode: Union[str, list[str]] = "vector",
    color: Union[str, list[str]] = "black",
):
```
for some definition of `State` (consider that even `list` could be promoted to [`ArrayLike`](https://numpy.org/doc/stable/reference/typing.html#numpy.typing.ArrayLike))

src/qibo/ui/bloch.py
niccololaurora: I have made a lot of changes after your suggestions.
I have addressed all the things that you pointed out.

Now the Bloch sphere does not rely on `pyplot` but on the Figure class. As we moved from `pyplot` to the class Figure I had to choose a suitable Matplotlib's backend. 
I chose three different backends: "tkagg", "qtagg", "agg".
The first two backends ("tkagg", "qtagg") are interactive, which means that they are used to produce windows that show the plot. 
Why two backends, instead of only one? 
I think that the advantage of "tkagg" which relies on [tkinter](https://docs.python.org/3/library/tkinter.html) is that it does not require the installation of any package.
The advantage of "qtagg" is that it seems more smooth, but it requires the installation of [PyQt](https://pypi.org/project/PyQt6/) (either 5 or 6).
I have checked and the Bloch class can support another interactive backend which is "cairo" (which requires the installation of [pycairo](https://pycairo.readthedocs.io/en/latest/)), but I did not see any advantage over "qtagg".

Finally the last backend is "agg", which is non interactive and it is used just to save the plots.

There's only one problem with the current implementation of the Bloch class, which is that it does not support Jupyter Notebooks, because when plotting the chosen backends will open a window outside the notebook.
The usual magic command "%matplotlib inline" does not work because within the class I override the choice of the backend. 
I will address this problem maybe trying to force "tkagg" and "qtagg" to plot in the notebook or with a different approach that I still have to find.

niccololaurora:
Hi @alecandido and @MatteoRobbiati , I am gonna briefly summarize the state of the PR.

I have had some issues finalizing the pull request due to a couple of caveats.
As far as I understand, there are two main problems:

1. The tests are failing because of an import error: "ImportError: Failed to import any of the following Qt binding modules: PyQt6, PySide6, PyQt5, PySide2"
Currently, the code for plotting the Bloch Sphere supports four possible backends (two interactive, one for notebooks, and one non-interactive): "tkagg" which uses Tkinter, "qtagg" which uses PyQt6 or PyQt5, "jupyter" which is used inside Jupyter notebooks and "agg" which is a non interactive backend for just saving the images.
 I believe (though I might be wrong) that the problem is that the virtual environment running the tests does not have these packages installed.
If you think introducing these new dependencies into the Qibo ecosystem is not ideal, we could rely on the "tkagg" backend instead (which I have already implemented). Tkinter is included in standard Python, so no additional packages are required.
However, this backend feels a bit less smooth compared to "qtagg".

2. Testing the interactive backends "tkagg" and "qtagg". 
I am not sure how to properly test the interactive backends, in particular how to prevent the figures from being displayed.
I decided to rely on monkey patching, but I am unsure if this is the right approach.
Normally, when running matplotlib tests and avoiding figure display, one just switches to a non-interactive backend (e.g., matplotlib.use("agg")).
In this case, however, I cannot simply force the "agg" backend, because the BlochSphere class explicitly supports the four backends mentioned above. If I were to set "agg" in the test file, it would be overridden by the class, which would then activate one of the interactive backends ("tkagg" or "qtagg") for displaying the picture.
Moreover, I cannot eliminate the dependency of the class from the specific interactive backend because each backend has a different way to show the plots.
Therefore, instead of setting "agg" (which would be ineffective) I used monkey patching: I temporarily replace the functions that would normally display the figures with no-op implementations. This way I can still test the interactive backends without triggering the actual GUI display.”


Does this sound reasonable?

alecandido:
I confess I am missing some pieces before.

The problem is certainly coming from backend handling. What you say it is true: Matplotlib supports a variety of backends, which mostly rely on external packages for the graphic libraries. Other than Tk, which is shipped together with Python itself (in the official distribution - just to let you know, there are even Python distributions stripping it by default, but not something you should worry about here).

However, I would expect no interplay at all between the code drawing the block sphere and the backend handling itself, since Matplotlib exposes a backend-agnostic API (i.e. its own objects, axes and figures).
Thus, I would have not expected any code related to the backend to be part of Qibo. I would have left backend fine-tuning and further handling, if ever needed. Which I did not expect, other than the `matplotlib.use()` directive and the `plt.show()` or `display()` invocations.

Which was the reason for you to include explicit backend handling?

niccololaurora:
I see what you are saying.

My approach was to follow this [example](https://matplotlib.org/2.0.2/examples/user_interfaces/embedding_in_tk.html), which is backend dependent.
Later when I wanted to do the same for `qt`, `agg`, `jupyter` I managed to do it only with backend-dependent functions and for that reason I created the functions: `_qt_window()`, `_tk_window()`, and `_jupyter_window()`.

You are right, probably this was not the best approach, since Matplotlib already provides backend-agnostic interfaces. 
I will try to do it in a backend-agnostic way.

I just "fear" that if I use backend-agnostic functions I will just end up in re-implementing parts of what `pyplot` already does. I will try to avoid as much as possible to do the same job of `pyplot`.

Does this make sense to you? 
Thank you for the feedback.

src/qibo/ui/bloch.py
```diff
@@ -0,0 +1,346 @@
...
+            self._backend = importlib.import_module(
+                "matplotlib.backends.backend_" + self.backend
+            )
+        elif self.backend == "jupyter":
+            pass
+        else:
+            raise_error(
+                ValueError,
+                "Backend not supported. Try: `qtagg`, `tkagg` (interactive), `agg` (non-interactive) or `jupyter`.",
+            )
```
alecandido: This is the first backend-specific part. But, from my perspective, the simple way to replace it is just not doing anything. Leave it to the user to set the backend, and rely on the environment.

Another option would be to require the user to pass the backend as a parameter. Which is more reliable, since you do not depend on the environment, and the evaluation remains pure.
While, in general, I would support this approach, it is not the standard with Matplotlib backends. In this case, I'd go for the common practice.

All in all, this would result in the following
```suggestion
    # Figure and axis
    def __post_init__(self):
        # Figure
        self.fig = Figure(figsize=self.STYLE["figure.figsize"])
        self.ax = self.fig.add_subplot(projection="3d", elev=30, azim=30)
```

src/qibo/ui/bloch.py
```diff
@@ -0,0 +1,346 @@
...
+        canvas_widget = canvas.get_tk_widget()
+        canvas_widget.pack(side=tk.TOP, fill=tk.BOTH, expand=True)
+        canvas.draw()
+        root.mainloop()
+
+    def _jupyter_window(self):
+        "Helper method to `plot()` for the `jupyter` backend."
+        self._new_window()
+        self._rendering()
+        display(self.fig)
```
alecandido: The main feature of these three functions is to draw the figure. I believe you can achieve the same just invoking [`plt.plot()`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.show.html#matplotlib.pyplot.show).

The `pyplot` interface has the disadvantage of relying on the global state. There is a [`Figure.show()`](https://matplotlib.org/stable/api/_as_gen/matplotlib.figure.Figure.show.html), which would avoid the locality issue, but it comes with a warning. Apparently, it would work fine in IPython (and then Jupyter), but probably not in bare Python, where the figure may not be shown at all (or just flash very briefly).

Feel free to experiment with `Figure.show()`. But, in this case, just using `plt.show()` could be good enough, since apparently it is the higher-level interface preferred(/forced) by the library.

> Special cases aren't special enough to break the rules.
> Although practicality beats purity.
>
> https://peps.python.org/pep-0020/, `import this`

src/qibo/ui/bloch.py
```diff
@@ -0,0 +1,346 @@
...
+            self._rendering()
+            self._qt_window()
+        elif self.backend == "jupyter":
+            self._jupyter_window()
+        elif self.backend == "agg":
+            raise_error(
+                ValueError,
+                "The backend `agg` is a non-interactive backend which "
+                "can be used only to save figures. If you want to plot try `qtagg`, `tkagg`, `jupyter.`",
+            )
```
alecandido: I see that you wanted to collect in a more elegant way the drawing and displaying actions.
However, it is idiomatic in Matplotlib to first draw, possibly keep modifying the plot, and eventually show it.

I would not find against their way of doing things, and commit to the same direction.
I would then rename `._rendering()` as `.plot()`. And then I would leave the user doing:

```py
bloch = Bloch(...)
...
bloch.plot()
plt.show()
```

Which is almost identical to the Matplotlib way of doing things. Thus, it should be familiar.

src/qibo/ui/bloch.py
```diff
@@ -0,0 +1,346 @@
...
+
+    def save(self, filename="bloch_sphere.pdf"):
+        "This function saves the sphere."
+        self._rendering()
+
+        mpl.use("agg")
+        canvas = self._backend.FigureCanvas(self.fig)
+        self.fig.savefig(filename)
```
alecandido: Even this one, I would just drop and leave it to the user.

It can be achieved with
```py
bloch = Bloch(...)
...
bloch.plot()
bloch.fig.savefig()  # or: `plt.savefig()`
```

src/qibo/ui/bloch.py
niccololaurora: Yes, the difficult part was to create the event loop for the various different backends. I will switch to `pyplot`.

niccololaurora: Hi @alecandido . I have tried invoking `plt.plot()`, but it does not work. I am gonna show you a minimal example:

```
fig = Figure(figsize=(6, 4))
ax = fig.add_subplot(111)
ax.plot ...
plt.show()
```

Here no figure is shown. I think the problem is that `pyplot` does not "see" the figure.
If you then try to switch from `plt.show()` to `fig.show()`, you get an error:

`AttributeError: Figure.show works only 
for figures managed by pyplot, normally created by 
pyplot.figure()`

To solve this error we should create the figure with `pyplot`, which completely defeats our purpose.
I think that the only way that this could work is we either wrap the `Figure` in a `Tkinter` loop or a `Qt` loop (which is what I was doing in the previous version of the code).

The more wise thing to do is to switch completely to `pyplot`.
Right now we are looking for a workaround to `pyplot`. I think that this workaround is against the spirit of the matplotlib. Regarding the "locality" problem: this is something well-known (most of the users should be already familiar with it) and it would be consistent with the behaviour of matplotlib.

alecandido: Hi @niccololaurora :)

Agreed. I believe you meant `plt.show` (instead of `plt.plot()`). But yes, I expect that it can not work if you did not register the figure in the global figures' storage. Which is maintained by Matplotlib internally, when you use calls like `plt.figure()`.

I believe that the two options we have are:
```py
# option 1
fig = Figure(...)
...
fig.show()

# option 2
fig = plt.figure(...)
...
plt.show()
```

As discussed before, the advantage of 1 consists in avoiding global Matplotlib state. But its downside could be the requirement for manual event loop management, according to the cited warning in the documentation.

That's what I was mentioning above about practicality: option 2 works for sure, with a potential disadvantage which is very much known and understood by most users. Instead, option 1 would be the *pure* way. But it has to be explored, and it may prove to require enormous effort.

I recommend to use option 2.

niccololaurora: Thank you @alecandido :)

I think the PR is ready: I have implemented option 2 in the last commit.
There is a file for the tests and a notebook as the tutorial.

src/qibo/ui/bloch.py
```diff
@@ -0,0 +1,384 @@
...
+        Returns:
+            Three arrays containing the cartesian coordinates
+            of 100 points used to draw the latitudes of the sphere.
+        """
+        theta = np.linspace(0, 2 * np.pi, 100)
+        z = np.full(100, z)
+        r = np.sqrt(1 - z[0] ** 2)
+        x = r * np.cos(theta)
+        y = r * np.sin(theta)
+        return x, y, z
```
alecandido: Just for clarity, I believe *parallel* is more common than *latitude* as the noun to describe this concept (and more easily related to *meridian*).

src/qibo/ui/bloch.py
```diff
@@ -0,0 +1,384 @@
...
+
+    def _axis(self):
+        """Helper method to `_create_sphere` to construct the sphere's axis
+
+        Args:
+            None
+
+        Returns:
+            Three arrays containing the cartesian coordinates
+            of 100 points used to draw the axis of the sphere.
```
alecandido: For all the docstrings, you could even avoid the strict formatting, and go free-form.

The structured arguments and returned objects format was quite useful to annotate their types, and thus the signature of the function. But now this can be done in place, using Python's type hints (which I recommend using).
It could still be quite useful for functions with many arguments. Which is not the case for most of the methods and functions in this module. Or for all of them.

```suggestion
        It returns three arrays containing the cartesian coordinates
        of 100 points used to draw the axis of the sphere.
```

src/qibo/ui/bloch.py
```diff
@@ -0,0 +1,384 @@
...
+
+        Returns:
+            None
+        """
+        if len(vector.shape) == 1:
+            return [vector]
+        elif len(vector.shape) == 2:
+            return vector
+        else:
+            raise_error(ValueError, "Only `2D` or `1D` np.ndarray / list is accepted.")
```
alecandido: If I'm correct, this is almost equivalent to [`np.atleast_2d`](https://numpy.org/doc/stable/reference/generated/numpy.atleast_2d.html), other than the error raised for ranks different from 1 and 2, which you could write as

```py
assert 0 < vector.ndim < 3, "Error message, if needed"
```

examples/bloch_sphere/bloch_sphere.ipynb
alecandido, line 1: It is mentioned in the notebook that the points should be normalized, 

> (2) `add_vector`: this function allows to add to the sphere a vector with three components x, y, z ***such that x^2 + y^2 + z^2 = 1;***

(emphasis mine)

I haven't tried to run myself, but I've not found this check anywhere. And, actually, I would propose not to introduce, since it may be convenient to also be able to plot points within the Bloch sphere, to represent mixed states.

Another option could be to introduce the check, but instead of checking `== 1.0` (or some suitable float comparison), that could be replaced by `<= 1.0` (or some suitable float comparison).

src/qibo/ui/bloch.py
niccololaurora: Ok, I will change it.

src/qibo/ui/bloch.py
niccololaurora: Okay, thank you! :)
I have done it for all of the functions.

src/qibo/ui/bloch.py
niccololaurora: Thank you! I did not know that! :)

examples/bloch_sphere/bloch_sphere.ipynb
niccololaurora: I have followed your first suggestion: I did not put any check. The user is free to plot points on the sphere or in the sphere.

niccololaurora:
Thank you @alecandido for the review.
Regarding the broadcasting semantics which is addressed by the function `_broadcasting_semantics()` I have used the function `np.broadcast_arrays`, that I was not using before.

Moreover, I added the possibility to plot density matrices as well by using "add_state". The code will detect automatically if it is dealing with a state or a density matrix.
In general density matrices will have three dimensions, meanwhile states only two: this is the main way to distinguish between the two cases.
I hope everything is in place to merge and thank you for the reviews.

examples/bloch_sphere/bloch_sphere.ipynb
alecandido, line 1: Since in the density matrix plot you're plotting just two arrows, I would avoid generating them randomly, and choose one of the two to be pretty close to the maximally mixed state (balanced classical superposition of $\ket{0}$ and $\ket{1}$), to better showcase that the result is actually in the bulk.

alecandido (approved):
Thank you, @niccololaurora, for all the changes and updates!

Here below there is just one final comment, but it is something minor.
All in all, the PR should be ready to be merged :)

.gitignore
```diff
@@ -139,3 +139,6 @@ tmp.npy
 # Mac
 .DS_Store
 .devenv
+
+# pdf
+*.pdf
```
renatomello: I'm not sure we want to ignore all PDFs by default. Maybe we can ignore specifically the ones you have in mind?

src/qibo/ui/bloch.py
```diff
@@ -0,0 +1,296 @@
...
+from typing import Union
+
+import matplotlib as mpl
+import matplotlib.pyplot as plt
+import numpy as np
+from matplotlib.patches import FancyArrowPatch
+from mpl_toolkits.mplot3d import proj3d
+from numpy.typing import ArrayLike
+
+from qibo import hamiltonians
```
renatomello:
```suggestion
from qibo.hamiltonians import SymbolicHamiltonian
```

src/qibo/ui/bloch.py
```diff
@@ -0,0 +1,296 @@
...
+        """
+        xs3d, ys3d, zs3d = self._verts3d
+        xs, ys, zs = proj3d.proj_transform(xs3d, ys3d, zs3d, self.axes.M)
+        self.set_positions((xs[0], ys[0]), (xs[1], ys[1]))
+
+        return np.min(zs)
+
+
+@dataclass
+class Bloch:
```
renatomello:
```suggestion
class BlochSphere:
```
I think it's better to be more explicit.

src/qibo/ui/bloch.py
```diff
@@ -0,0 +1,296 @@
...
+
+        def _paulis_expectation(state):
+            sigma_X = hamiltonians.SymbolicHamiltonian(X(0))
+            sigma_Y = hamiltonians.SymbolicHamiltonian(Y(0))
+            sigma_Z = hamiltonians.SymbolicHamiltonian(Z(0))
+
+            x = sigma_X.expectation(state)
+            y = sigma_Y.expectation(state)
+            z = sigma_Z.expectation(state)
+            return x, y, z
```
renatomello:
```suggestion
        def _paulis_expectation(state):
            sigma_X = SymbolicHamiltonian(X(0))
            sigma_Y = SymbolicHamiltonian(Y(0))
            sigma_Z = SymbolicHamiltonian(Z(0))

            x = sigma_X.expectation(state)
            y = sigma_Y.expectation(state)
            z = sigma_Z.expectation(state)
            return x, y, z
```

Does this function need to be defined inside a method?

src/qibo/ui/bloch.py
```diff
@@ -0,0 +1,296 @@
...
+            if state[0] == 1 and state[0] == 0:
+                z = 1
+                return x, y, z
+            elif state[0] == 0 and state[0] == 1:
+                z = -1
+                return x, y, z
+            else:
+                return _paulis_expectation(state)
+        elif state.ndim == 2:
+            return _paulis_expectation(state)
```
renatomello:
```suggestion
        x, y, z = 0, 0, 0
        if state.ndim == 1:
            if state[0] == 1 and state[0] == 0:
                z = 1
                return x, y, z
            
            if state[0] == 0 and state[0] == 1:
                z = -1
                return x, y, z

            return _paulis_expectation(state)

        return _paulis_expectation(state)
```

src/qibo/ui/bloch.py
```diff
@@ -0,0 +1,296 @@
...
+        vectors, modes, colors = self._broadcasting_semantics(vector, mode, color)
+        for vector, color, mode in zip(vectors, colors, modes):
+            if mode == "vector":
+                self._vectors.append(vector)
+                self._color_vectors.append(color)
+            elif mode == "point":
+                self._points.append(vector)
+                self._color_points.append(color)
+            else:
+                raise_error(ValueError, "Mode not supported. Try: `point` or `vector`.")
```
renatomello:
```suggestion
            if mode not in ("vector", "point"):
                raise_error(ValueError, "Mode not supported. Try: `point` or `vector`.")
            if mode == "vector":
                self._vectors.append(vector)
                self._color_vectors.append(color)
            else:
                self._points.append(vector)
                self._color_points.append(color)

```

src/qibo/ui/bloch.py
```diff
@@ -0,0 +1,296 @@
...
+        for vector, color, mode in zip(vectors, colors, modes):
+            x, y, z = self._coordinates(vector)
+            if mode == "vector":
+                self._vectors.append(np.array([x, y, z]))
+                self._color_vectors.append(color)
+            elif mode == "point":
+                self._points.append(np.array([x, y, z]))
+                self._color_points.append(color)
+            else:
+                raise_error(ValueError, "Mode not supported. Try: `point` or `vector`.")
```
renatomello:
```suggestion
        for vector, color, mode in zip(vectors, colors, modes):
            if mode not in ("vector", "point"):
                raise_error(ValueError, "Mode not supported. Try: `point` or `vector`.")

            x, y, z = self._coordinates(vector)
            
            if mode == "vector":
                self._vectors.append(np.array([x, y, z]))
                self._color_vectors.append(color)
            else mode == "point":
                self._points.append(np.array([x, y, z]))
                self._color_points.append(color)
```

tests/test_ui_bloch_sphere.py
```diff
@@ -0,0 +1,328 @@
+import matplotlib as mpl
+import matplotlib.pyplot as plt
+import numpy as np
+
+mpl.use("agg")
+
+from qibo import Circuit, gates
+from qibo.ui.bloch import Bloch
```
renatomello:
```suggestion
from qibo.ui.bloch import BlochSphere
```

tests/test_ui_bloch_sphere.py
```diff
@@ -0,0 +1,328 @@
...
+def _circuit(weight, boolean=False):
+    circ = Circuit(1, density_matrix=boolean)
+    circ.add(gates.RY(q=0, theta=np.random.randn() * weight))
+    circ.add(gates.RX(q=0, theta=np.random.randn() * weight))
+    circ.add(gates.RZ(q=0, theta=np.random.randn() * weight))
+    return circ
+
+
+def test_empty_sphere():
+    bs = Bloch()
```
renatomello:
```suggestion
    bs = BlochSphere()
```

tests/test_ui_bloch_sphere.py
```diff
@@ -0,0 +1,328 @@
...
+    bs.render()
+    plt.show()
+
+
+def test_state():
+    bs = Bloch()
```
renatomello:
```suggestion
    bs = BlochSphere()
```

tests/test_ui_bloch_sphere.py
```diff
@@ -0,0 +1,328 @@
...
+    state = np.array([1 / np.sqrt(2), 1 / np.sqrt(2) * 1j], dtype="complex")
+    bs.add_state(state)
+    bs.render()
+    plt.show()
+
+
+def test_vector_point():
+    bs = Bloch()
```
renatomello:
```suggestion
    bs = BlochSphere()
```

tests/test_ui_bloch_sphere.py
```diff
@@ -0,0 +1,328 @@
...
+    vector = np.array([0, 0, 1])
+    bs.add_vector(vector, mode="point", color="green")
+    bs.render()
+    plt.show()
+
+
+def test_multiple_vectors_array():
+    bs = Bloch()
```
renatomello:
```suggestion
    bs = BlochSphere()
```

tests/test_ui_bloch_sphere.py
```diff
@@ -0,0 +1,328 @@
...
+    vectors = np.random.normal(size=(100, 3))
+    vectors /= np.linalg.norm(vectors, axis=1)[:, np.newaxis]
+    bs.add_vector(vectors, color="royalblue")
+    bs.render()
+    plt.show()
+
+
+def test_multiple_vectors_list():
+    bs = Bloch()
```
renatomello:
```suggestion
    bs = BlochSphere()
```

tests/test_ui_bloch_sphere.py
```diff
@@ -0,0 +1,328 @@
...
+        vector /= np.linalg.norm(vector)
+        vectors.append(vector)
+
+    bs.add_vector(vectors, color="royalblue")
+    bs.render()
+    plt.show()
+
+
+def test_multiple_states():
+    bs = Bloch()
```
renatomello:
```suggestion
    bs = BlochSphere()
```

tests/test_ui_bloch_sphere.py
```diff
@@ -0,0 +1,328 @@
...
+    states = np.random.normal(size=(100, 2))
+    states /= np.linalg.norm(states, axis=1)[:, np.newaxis]
+    bs.add_state(states)
+    bs.render()
+    plt.show()
+
+
+def test_state_clear():
+    bs = Bloch()
```
renatomello:
```suggestion
    bs = BlochSphere()
```

tests/test_ui_bloch_sphere.py
```diff
@@ -0,0 +1,328 @@
...
+    bs.render()
+    plt.show()
+
+    bs.clear()
+    bs.render()
+    plt.show()
+
+
+def test_classification():
+    bs = Bloch()
```
renatomello:
```suggestion
    bs = BlochSphere()
```

tests/test_ui_bloch_sphere.py
```diff
@@ -0,0 +1,328 @@
...
+    for _ in range(20):
+        state = _circuit(weight)(np.array([0, 1], dtype="complex")).state()
+        bs.add_state(state, mode="point", color="blue")
+
+    bs.render()
+    plt.show()
+
+
+def test_multi_classification():
+    bs = Bloch()
```
renatomello:
```suggestion
    bs = BlochSphere()
```

tests/test_ui_bloch_sphere.py
```diff
@@ -0,0 +1,328 @@
...
+        circ = _circuit(weight)
+        state = circ(np.array([0, 1], dtype="complex")).state()
+        bs.add_state(state, mode="point", color="magenta")
+
+    bs.render()
+    plt.show()
+
+
+def test_qibo_output():
+    bs = Bloch()
```
renatomello:
```suggestion
    bs = BlochSphere()
```

tests/test_ui_bloch_sphere.py
```diff
@@ -0,0 +1,328 @@
...
+
+    # state
+    state = circ().state()
+    bs.add_state(state, color="orange")
+    bs.render()
+    plt.show()
+
+
+def test_point_vector_state():
+    bs = Bloch()
```
renatomello:
```suggestion
    bs = BlochSphere()
```

tests/test_ui_bloch_sphere.py
```diff
@@ -0,0 +1,328 @@
...
+    # vector
+    vector = np.array([1, 0, 0])
+    bs.add_vector(vector, mode="vector", color="red")
+
+    bs.render()
+    plt.show()
+
+
+def test_save():
+    bs = Bloch()
```
renatomello:
```suggestion
    bs = BlochSphere()
```

tests/test_ui_bloch_sphere.py
```diff
@@ -0,0 +1,328 @@
...
+    state = circ().state()
+
+    bs.add_state(state, color=["orange"])
+    bs.render()
+    plt.savefig("bloch.pdf")
+    plt.close()
+
+
+def test_many_spheres():
+    bs = Bloch()
```
renatomello:
```suggestion
    bs = BlochSphere()
```

tests/test_ui_bloch_sphere.py
```diff
@@ -0,0 +1,328 @@
...
+    bs.render()
+    plt.show()
+    bs.clear()
+
+    bs.render()
+    plt.show()
+
+
+def test_density_matrix():
+    bs = Bloch()
```
renatomello:
```suggestion
    bs = BlochSphere()
```

tests/test_ui_bloch_sphere.py
```diff
@@ -0,0 +1,328 @@
...
+        3. n state vector --> Shape = (n,2)
+        4. Single rho --> Shape = (2,2)
+        5. Two rhos --> Shape = (2,2,2)
+        6. n rhos --> Shape = (n,2,2)
+    So the code could have problems distinguishing case 2 and case 4.
+    We can solve this problem just by adding an extra dimension to case 4, thus (2,2) -> (1,2,2).
+    In this way every time that we have a density matrix the input will have three dimensions.
+    """
+
+    bs = Bloch()
```
renatomello:
```suggestion
    bs = BlochSphere()
```

tests/test_ui_bloch_sphere.py
```diff
@@ -0,0 +1,328 @@
...
+    ]
+    bs.add_state(rhos, mode="vector", color="blue")
+    bs.render()
+    plt.show()
+    bs.clear()
+    plt.close()
+
+
+def test_mixed_state():
+    bs = Bloch()
```
renatomello:
```suggestion
    bs = BlochSphere()
```

.gitignore
niccololaurora: Thank you! I have ignored only one specific pdf.

src/qibo/ui/bloch.py
niccololaurora: I have placed the function outside the method. Thank you for the suggestion.

src/qibo/ui/bloch.py
niccololaurora: Thank you for this polished version!

examples/bloch_sphere/bloch_sphere.ipynb
niccololaurora: Thank you for the suggestion! You are right it was not very clear.
Now the jupyter notebook shows the maximally mixed state, and other mixed states that lie on the x-axis, y-axis and z-axis.

niccololaurora:
I think now the coverage should be 100%.

tests/test_ui_bloch_sphere.py
```diff
@@ -0,0 +1,344 @@
...
+
+def test_bloch_error_mode():
+    bs = BlochSphere()
+
+    vector = np.array([1, 0, 0])
+
+    try:
+        bs.add_vector(vector, mode="invalid_mode")
+    except ValueError as e:
+        assert str(e) == "Mode not supported. Try: `point` or `vector`."
```
renatomello:
```suggestion
    with pytest.raises(ValueError):
        bs.add_vector(vector, mode="invalid_mode")
```

tests/test_ui_bloch_sphere.py
```diff
@@ -0,0 +1,344 @@
...
+
+    try:
+        bs.add_state(vector, mode="invalid_mode")
+    except ValueError as e:
+        assert str(e) == "Mode not supported. Try: `point` or `vector`."
```
renatomello:
```suggestion
    with pytest.raises(ValueError):
        bs.add_state(vector, mode="invalid_mode")
```

tests/test_ui_bloch_sphere.py
```diff
@@ -0,0 +1,344 @@
+import matplotlib as mpl
+import matplotlib.pyplot as plt
+import numpy as np
```
renatomello, line 3:
```suggestion
import numpy as np
import pytest
```

PR #1670, opened 2025-06-13
Enable `TOFFOLI.decompose` and add `CCZ` decomposition into native gate set that includes the `CZ` gate

renatomello:
Per title. Fixes bug raised in the Discussion #1669.

Checklist:
- [x] Reviewers confirm new code works as expected.
- [x] Tests are passing.
- [x] Coverage does not decrease.
- [x] Documentation is updated.

PR #1671, opened 2025-06-13
``python 3.13.5`` in the CI

BrunoLiegiBastonLiegi:
As pointed out in various other locations (eg #1670) there are some problems with ``numba`` and ``python3.13``.
Here https://github.com/numba/numba/issues/10101 they suggest the issue is related to list comprehensions and is lifted in ``python3.13.5``. Thus I am setting this as the version to be used in the CI here.

Checklist:
- [x] Tests are passing.


renatomello (approved):

renatomello:
@scarrazza could you approve this so we can fix tests in other PRs?

Issue #512, opened 2021-11-15
Qutrit (Qudit) simulation using Qibo


igres26:
As Qibo evolves we should start considering taking its quantum simulations to the next level. And to d-level quantum bits after that.

Qudit manipulation will allow Qibo users to create different kind of algorithms that could then be deployed on quantum hardware as experimental implementations of quantum computers usually allow for control of more than two states. It also has value from a quantum information perspective even before experimental implementation.

I would start with an initial Qutrit implementation and possibly build a generalization from there.

alecandido:
I would bump this issue, and reconsider our backends in this light.

This is also related to `qibo-core`, but the support in there will be minimal: we just need to
- support the declaration of the possible number of states for each element, in order to make the backend aware, and propagate the information
- extend the gates' addition, in order to specify not only the elements, but even the states

The rest is truly up to the backends (instantiate a suitable `initial_state`, execute the gates).

I would not even start from qutrits, since I believe that anything specific to qutrits will have to be undone for the generic qudits.
Instead, I would propose the following interface:
- circuit creation:
    ```py
    Circuit([2] * 4 + [3] * 2 + [2] * 2)
    ```
    a circuit on 8 elements, of which elements 4 and 5 are qutrits, the rest qubits
- gates addition:
    ```py
    c.add(X, [4], [0, 2])
    c.add(CNOT, [3, 4], [None, [1, 2]])  # None = [0, 1]
    ```
    specifying not only the elements, but also the states it should act on

The qubits gates on a restricted number of states should be interpreted as the original matrices on the subspace of states specified, and extended with an identity on the other states (i.e. they isolate the relevant elements from the tensor product basis as for qubits, ignoring the state of all the others, and in the subspace of the relevant elements they act as described above: the known matrix, extended with identity bits as needed).

This should be possible to be efficiently implemented in the backends through careful slicing (or appropriately taking the states into account in the loops, when explicit).

Then, we could explicitly add qutrit gates (acting on the all the three states at the same time in a non-trivial way) and so on, for which the interface would be a simple extension of the former:
```py
c.add(SINGLE_QUTRIT, [3], [0, 2, 3])
c.add(BOTH_QUTRITS, [6, 7], [None, [0, 1, 3]])  # if a qutrit gate, None = [0, 1, 2]
```

But I'm not sure whether they are needed at all. @igres26?

igres26:
[...]

PR #1672, opened 2025-06-16
added plot state function for issue #689

sergiomtzlosa:
Checklist:
- [x] Tests are passing.

I created a density plot function based on the qiskit one and the previous issue #689 and #737, for instance, given this circuit:

```python
import numpy as np
from qibo.ui import plot_density_hist

circuit = Circuit(2)
circuit.add(gates.H(0))
circuit.add(gates.H(1))
circuit.add(gates.CZ(0, 1))
circuit.add(gates.RY(0, theta=np.pi/3))
circuit.add(gates.RX(1, theta=np.pi/5))

plot_density_hist(circuit);
```

The result goes as:

test_complex_circuit_state_nqubits_2


renatomello:
@sergiomtzlosa sorry for not reviewing it yet. I'm a little busy but I'll get to it as soon as I can.

alecandido:
> @sergiomtzlosa sorry for not reviewing it yet. I'm a little busy but I'll get to it as soon as I can.

Sorry, same situation. But I'll do it

sergiomtzlosa:
I do not know why tests are not passing because my local pytests passed

tests/utils.py
```diff
@@ -25,5 +25,30 @@ def fig2array(fig):
 
 
 def match_figure_image(fig, arr_path):
```
renatomello:
```suggestion
def match_figure_image(fig, arr_path: str):
```

tests/utils.py
```diff
@@ -24,6 +24,31 @@ def fig2array(fig):
...
 
-def match_figure_image(fig, arr_path):
-    """Check whether the two image arrays match."""
+def match_figure_image(fig, arr_path: str):
+    """
+    Check whether the two image arrays match.
+
+    Args:
+        fig (matplotlib.figure.Figure): Matplotlib figure to compare.
+        arr_path (str): Path to the numpy array file containing the reference image.
```
renatomello:
```suggestion
    """Check whether the two image arrays match.

    Args:
        fig (:class:`matplotlib.figure.Figure`): Figure to compare.
        arr_path (str): Path to the ``numpy`` array file containing the reference image.
```

tests/utils.py
```diff
@@ -24,6 +24,30 @@ def fig2array(fig):
...
+    Check whether the two image arrays match within a tolerance.
+
+    Args:
+        fig (matplotlib.figure.Figure): Matplotlib figure to compare.
+        arr_path (str): Path to the numpy array file containing the reference image.
+        rtol (float, optional): Relative tolerance for comparison.
+        atol (float, optional): Absolute tolerance for comparison.
+
+    Returns:
+        bool: True if the images match within the specified tolerances, False otherwise.
```
renatomello:
```suggestion
def match_figure_close_image(fig, arr_path: str, rtol: float = 0, atol: float = 255):
    """Check whether the two image arrays match within a tolerance.

    Args:
        fig (:class:`matplotlib.figure.Figure`): Figure to compare.
        arr_path (str): Path to the ``numpy`` array file containing the reference image.
        rtol (float, optional): Relative tolerance for comparison.
        atol (float, optional): Absolute tolerance for comparison.

    Returns:
        bool: ``True`` if the images match within the specified tolerances, ``False`` otherwise.
```

tests/utils.py
```diff
@@ -24,6 +24,29 @@ def fig2array(fig):
...
-    """Check whether the two image arrays match."""
+def match_figure_image(fig, arr_path: str):
+    """Check whether the two image arrays match.
+
+    Args:
+        fig (:class:`matplotlib.figure.Figure`): Figure to compare.
+        arr_path (str): Path to the ``numpy`` array file containing the reference image.
+
+    Returns:
+        bool: True if the images match, False otherwise.
```
renatomello:
```suggestion
        bool: ``True`` if the images match, ``False`` otherwise.
```

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,199 @@
...
+    Args:
+        circuit (qibo.core.circuit.Circuit): The quantum circuit to visualize.
+        title (str): Title of the plot.
+        alpha (float): Transparency level for the bars in the plot.
+        colors (list): A list of two colors for the positive and negative parts of the density matrix.
+                        If None, default colors will be used.
+
+    Returns:
+        tuple: A tuple containing the figure, axes for the real part and axes for the imaginary part.
+    """
```
renatomello:
```suggestion
    """Plot the real and imaginary parts of the density matrix

    Given a :class:`qibo.models.circuit.Circuit`, plots the real and imaginary parts
    of the final density matrix as separate 3D cityscape plots, side by side, and
    with a gray ``z=0`` plane for the imaginary part.

    Args:
        circuit (:class:`qibo.models.circuit.Circuit`): Circuit to visualize.
        title (str, optional): Title of the plot. Defaults to ``""``.
        alpha (float, optional): Transparency level for the bars in the plot. Defaults to :math:`0.5`.
        colors (list, optional): A list of two colors for the positive and negative parts of the density matrix.
            If ``None``, default colors will be used. Defaults to ``None``.

    Returns:
        tuple: Respectively, the figure, and axes for the real and the imaginary parts.
    """
```

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,201 @@
...
+import numpy as np
+from matplotlib.colors import LinearSegmentedColormap, Normalize
+from mpl_toolkits.mplot3d.art3d import Poly3DCollection
+
+import qibo
+
+
+# Based on Qiskit density state plot
+# https://github.com/Qiskit/qiskit/blob/stable/2.0/qiskit/visualization/state_visualization.py#L372-L622
+def plot_density_hist(circuit, title="", alpha=0.5, colors=None):
```
renatomello:
```suggestion
def plot_density_hist(circuit, title: str = "", alpha: float = 0.5, colors: Optional[list] = None, backend=None):
```

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,201 @@
+import matplotlib.pyplot as plt
+import numpy as np
+from matplotlib.colors import LinearSegmentedColormap, Normalize
+from mpl_toolkits.mplot3d.art3d import Poly3DCollection
+
+import qibo
```
renatomello:
```suggestion
from typing import Optional

import numpy as np

import matplotlib.pyplot as plt
from matplotlib.colors import LinearSegmentedColormap, Normalize
from mpl_toolkits.mplot3d.art3d import Poly3DCollection

from qibo.backends import _check_backend
from qibo.result import MeasurementOutcomes
```

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,201 @@
...
+    Given a :class:`qibo.models.circuit.Circuit`, plots the real and imaginary parts
+    of the final density matrix as separate 3D cityscape plots, side by side, and
+    with a gray ``z=0`` plane for the imaginary part.
+
+    Args:
+        circuit (:class:`qibo.models.circuit.Circuit`): Circuit to visualize.
+        title (str, optional): Title of the plot. Defaults to ``""``.
+        alpha (float, optional): Transparency level for the bars in the plot. Defaults to :math:`0.5`.
+        colors (list, optional): A list of two colors for the positive and negative parts of the density matrix.
+            If ``None``, default colors will be used. Defaults to ``None``.
```
renatomello:
```suggestion
        alpha (float, optional): Transparency level for the bars in the plot. 
            Defaults to :math:`0.5`.
        colors (list, optional): A list of two colors for the positive and negative
            parts of the density matrix. If ``None``, default colors will be used.
            Defaults to ``None``.
```

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,203 @@
...
+        colors (list, optional): A list of two colors for the positive and negative
+            parts of the density matrix. If ``None``, default colors will be used.
+            Defaults to ``None``.
+
+    Returns:
+        tuple: Respectively, the figure, and axes for the real and the imaginary parts.
+    """
+
+    # Execute the circuit to get the state
+    exec_circ = circuit.execute()
```
renatomello:
```suggestion
    backend = _check_backend(backend)
    # Execute the circuit to get the state
    exec_circ = backend.execute_circuit(circuit)
```

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,203 @@
...
+    with a gray ``z=0`` plane for the imaginary part.
+
+    Args:
+        circuit (:class:`qibo.models.circuit.Circuit`): Circuit to visualize.
+        title (str, optional): Title of the plot. Defaults to ``""``.
+        alpha (float, optional): Transparency level for the bars in the plot.
+            Defaults to :math:`0.5`.
+        colors (list, optional): A list of two colors for the positive and negative
+            parts of the density matrix. If ``None``, default colors will be used.
+            Defaults to ``None``.
```
renatomello:
```suggestion
        colors (list, optional): A list of two colors for the positive and negative
            parts of the density matrix. If ``None``, default colors will be used.
            Defaults to ``None``.
        backend (:class:`qibo.backends.abstract.Backend`, optional): backend
            to be used in the execution. If ``None``, it uses
            the current backend. Defaults to ``None``.
```

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,203 @@
...
+    state = exec_circ.state()
+
+    # Create a density matrix from state vector
+    if not circuit.density_matrix:
+        ket = np.asarray(state)
+        state = np.outer(ket, ket.conj())
+
+    n = circuit.nqubits
+    row_names = [bin(i)[2:].zfill(n) for i in range(2**n)]
+    column_names = [bin(i)[2:].zfill(n) for i in range(2**n)]
```
renatomello:
```suggestion
    nqubits = circuit.nqubits
    row_names = [bin(i)[2:].zfill(nqubits) for i in range(2**nqubits)]
    column_names = [bin(i)[2:].zfill(nqubits) for i in range(2**nqubits)]
```

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,203 @@
...
+        raise ValueError(
+            "Circuit must not contain measurement gates for density matrix visualization"
+        )
+
+    state = exec_circ.state()
+
+    # Create a density matrix from state vector
+    if not circuit.density_matrix:
+        ket = np.asarray(state)
+        state = np.outer(ket, ket.conj())
```
renatomello:
```suggestion
    if not circuit.density_matrix:
        state = backend.cast(state)
        state = backend.np.outer(state, backend.np.conj(state))
```

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,203 @@
...
+
+    n = circuit.nqubits
+    row_names = [bin(i)[2:].zfill(n) for i in range(2**n)]
+    column_names = [bin(i)[2:].zfill(n) for i in range(2**n)]
+
+    matrix_real = state.real
+    matrix_imag = state.imag
```
renatomello:
```suggestion
    matrix_real = state.real
    matrix_imag = state.imag
    del state
```

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,203 @@
...
+
+    label_y, label_x = matrix_real.shape[:2]
```
renatomello: This is calculated multiple times and it seems it could be calculated just once

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,203 @@
...
+    xpos, ypos = np.meshgrid(xpos + 0.25, ypos + 0.25)
+
+    xpos = xpos.flatten()
+    ypos = ypos.flatten()
+    zpos = np.zeros(label_x * label_y)
+
+    dx = 0.5 * np.ones_like(zpos)
+    dy = dx.copy()
+    dzr = matrix_real.flatten()
+    dzi = matrix_imag.flatten()
```
renatomello:
```suggestion
    dzr = backend.np.reshape(matrix_real, (-1, 1))
    dzi = backend.np.reshape(matrix_imag, (-1, 1))
```

This is also calculated more than once when it could be done just once.

src/qibo/ui/plot_state.py
sergiomtzlosa: thanks a lot for spotting this!

tests/utils.py
```diff
@@ -24,6 +24,29 @@ def fig2array(fig):
...
+    Args:
+        fig (:class:`matplotlib.figure.Figure`): Figure to compare.
+        arr_path (str): Path to the ``numpy`` array file containing the reference image.
+        rtol (float, optional): Relative tolerance for comparison.
+        atol (float, optional): Absolute tolerance for comparison.
+
+    Returns:
+        bool: ``True`` if the images match within the specified tolerances, ``False`` otherwise.
+    """
+    return np.allclose(fig2array(fig), np.load(arr_path), rtol=rtol, atol=atol)
```
alecandido: Instead of saving pictures as arrays, and maintain your own utilities to compare, you could actually save them as PNG images, and just use Matplotlib ones.

https://matplotlib.org/stable/api/testing_api.html#module-matplotlib.testing.compare

sergiomtzlosa: thanks @alecandido , I will have a look at this, but I think the way we are doing now is better because we can use the matplotlib figure and convert it on the fly to array, otherwise we have to save the figure as image at some temporal path on the test function, anyways I will check how can be done by matching png files and try to discover if the matching images technique creates an improvement over the current functions

alecandido: Yes, since it's querying the file system, you certainly need a temporary path to dump it.

However, that should be fairly doable with [`tempfile`](https://docs.python.org/3/library/tempfile.html) (true, it will load it again - but I'd be willing to pay the performance price to save some maintenance time).

sergiomtzlosa: Finally I implemented your suggestion @alecandido, but I think that I will not change the `match_figure_image` function at the moment for the circuit drawer because I think it is out of the scope of this ticket and I would like to change it on another dedicated pull request

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,203 @@
...
+    state = exec_circ.state()
+
+    # Create a density matrix from state vector
+    if not circuit.density_matrix:
+        state = backend.cast(state)
+        state = backend.np.outer(state, backend.np.conj(state))
+
+    nqubits = circuit.nqubits
+    row_names = [bin(i)[2:].zfill(nqubits) for i in range(2**nqubits)]
+    column_names = [bin(i)[2:].zfill(nqubits) for i in range(2**nqubits)]
```
alecandido: Just in case, another efficient way to do this is through [`np.ravel_multi_index`](https://numpy.org/doc/stable/reference/generated/numpy.ravel_multi_index.html).

Most likely it has little advantage for a few qubits (for many, performances - but if you have performance issues in generating the indices, good luck with the plots...), but it's interesting for qudits (not yet supported by Qibo, see #512)

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,203 @@
...
+        norm_neg = plt.Normalize(vmin=min_dz, vmax=0)
+
+        # Create a color array based on the heights
+        colors_mapping = []
+        for height in dz:
+            if height >= 0:
+                colors_mapping.append(cmap_pos(norm_pos(height)))
+            else:
+                colors_mapping.append(cmap_neg(norm_neg(height)))
+        colors_mapping = np.array(colors_mapping)
```
alecandido: `Normalize` instances work on arrays(-like, i.e. include also lists), and it is quite likely that even colormaps do the same (though I've not yet confirmed it).

So, you may be able to write something like:
```suggestion
        colors_mapping = np.where(dz >= 0, cmap_pos(norm_pos(dz)), cmap_neg(norm_neg(dz)))
```

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,203 @@
...
+                ypos[dzp],
+                zpos[dzp],
+                dx[dzp],
+                dy[dzp],
+                dz[dzp],
+                alpha=alpha,
+                zorder=0.875,
+                color=colors_mapping[dzp],
+                shade=True,
+            )
```
alecandido: Is the problem here just the `zorder`?

If that's the case, since it seems that `zorder` is just a single float for the artist, and you can not specify for each element, you could actually just deduplicate plot.
I know it's mostly a single call, but it's pretty lengthy, and you should be able to write a single one. E.g.
```suggestion
        for mask, zorder in ((dz < 0, 0.625), (dz > 0, 0.875)):
            ax.bar3d(
                xpos[mask],
                ypos[mask],
                zpos[mask],
                dx[mask],
                dy[mask],
                dz[mask],
                alpha=alpha,
                zorder=zorder,
                color=colors_mapping[mask],
                shade=True,
            )

        if min_dz < 0 < max_dz:
            xlim, ylim = [0, label_x], [0, label_y]
            verts = [list(zip(xlim + xlim[::-1], np.repeat(ylim, 2), [0] * 4))]
            plane = Poly3DCollection(verts, alpha=0.25, facecolor="k", linewidths=1)
            plane.set_zorder(0.75)
            ax.add_collection3d(plane)
```

Since you're specifying the `zorder` anyhow, the order of the calls should not matter.
In any case, if you prefer avoiding a loop, you can always define a tiny function.

I threw aways the result value, since it's not being used later on. However, if need it can obviously be recovered, even with the loop.

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,203 @@
...
+    else:
+        if len(colors) != 2:
+            raise ValueError(
+                "Colors must be a list of len=2, got {} instead".format(len(colors))
+            )
+        pos_color = "#ff7f0e" if colors[0] is None else colors[0]
+        neg_color = "#1f77b4" if colors[1] is None else colors[1]
+
+    cmap_pos = LinearSegmentedColormap.from_list("cmap_pos", 4 * [pos_color])
+    cmap_neg = LinearSegmentedColormap.from_list("cmap_neg", 4 * [neg_color])
```
alecandido: Maybe you could split some validation and parameters manipulation as separate functions, for better modularity.

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,203 @@
...
+            tick.label1.set_horizontalalignment("left")
+            tick.label1.set_verticalalignment("bottom")
+
+        ax.set_xlabel("Ket", labelpad=20)
+        ax.xaxis.label.set_rotation(30)
+        ax.set_ylabel("Bra", labelpad=15)
+        ax.yaxis.label.set_rotation(30)
+        ax.set_box_aspect(aspect=(4, 4, 4), zoom=max_zoom)
+        ax.set_xmargin(0)
+        ax.set_ymargin(0)
```
alecandido: Same goes for the plots: you could make it a smaller internal function on its own.

src/qibo/ui/plot_state.py
sergiomtzlosa: I can use numpy.unravel_index (not ravel_multi_index) to generate the binary labels efficiently for all indices. It will be as:

```python
indices = np.arange(2**nqubits)
bits = np.array(np.unravel_index(indices, [2]*nqubits)).T
row_names = ["".join(str(b) for b in row) for row in bits]
column_names = ["".join(str(b) for b in row) for row in bits]
```

I will include that.

src/qibo/ui/plot_state.py
sergiomtzlosa: The line crashes so I handle this way:

```python
 colors_mapping = np.array([cmap_pos(norm_pos(height)) if height >= 0 else cmap_neg(norm_neg(height)) for height in dz])
```

src/qibo/ui/plot_state.py
alecandido: Yes, sorry. I thought a second about the two, and I eventually picked the wrong one 🤦

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,198 @@
...
+# Based on Qiskit density state plot
+# https://github.com/Qiskit/qiskit/blob/stable/2.0/qiskit/visualization/state_visualization.py#L372-L622
+def plot_density_hist(
+    circuit,
+    title: str = "",
+    alpha: float = 0.5,
+    colors: Optional[list] = None,
+    backend=None,
+):
+    """Plot the real and imaginary parts of the density matrix
```
MatteoRobbiati:
```suggestion
    """Plot the real and imaginary parts of the density matrix.
```

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,198 @@
...
+    Returns:
+        tuple: Respectively, the figure, and axes for the real and the imaginary parts.
+    """
+
+    backend = _check_backend(backend)
+    # Execute the circuit to get the state
+    exec_circ = backend.execute_circuit(circuit)
+
+    # if exec_circ is kind of MeasurementOutcomes, error measure gates are present
+    if isinstance(exec_circ, MeasurementOutcomes):
```
MatteoRobbiati: I would probably move this check before, you can check directly if the circuit has measurements, and in that case raise an error. For example checking `len(circuit.measurements)`. On this way we skip the execution :)

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,198 @@
...
+    dzi = backend.np.reshape(matrix_imag, (-1))
+
+    fig = plt.figure(figsize=(16, 8), facecolor="w")
+    ax1 = fig.add_subplot(1, 2, 1, projection="3d", computed_zorder=False)
+    ax2 = fig.add_subplot(1, 2, 2, projection="3d", computed_zorder=False)
+
+    max_dzr = np.max(dzr)
+    max_dzi = np.max(dzi)
+
+    fig_width, fig_height = fig.get_size_inches()
```
MatteoRobbiati: I would probably define `fig_width` and `fig_height` above, since right now they are hard coded into `figsize=(16,8)`.

src/qibo/ui/plot_state.py
sergiomtzlosa: perfect! I set it on the code

src/qibo/ui/plot_state.py
sergiomtzlosa: I set this two parameters as optional on the function signature

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,200 @@
+from typing import Optional
+
+import matplotlib.pyplot as plt
+import numpy as np
+from matplotlib.colors import LinearSegmentedColormap, Normalize
+from mpl_toolkits.mplot3d.art3d import Poly3DCollection
+
+from qibo.backends import _check_backend
```
renatomello:
```suggestion
from qibo.backends import _check_backend
from qibo.config import raise_error
```

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+def _validate_colors(colors):
+    if colors is None:
+        return "#ff7f0e", "#1f77b4"
+    if len(colors) != 2:
+        raise ValueError(f"Colors must be a list of len=2, got {len(colors)} instead")
+    pos_color = "#ff7f0e" if colors[0] is None else colors[0]
+    neg_color = "#1f77b4" if colors[1] is None else colors[1]
+    return pos_color, neg_color
```
renatomello:
```suggestion
    if colors is None:
        return "#ff7f0e", "#1f77b4"

    if len(colors) != 2:
        raise_error(ValueError, f"Colors must be a list of length 2, got {len(colors)} instead")

    pos_color = "#ff7f0e" if colors[0] is None else colors[0]
    neg_color = "#1f77b4" if colors[1] is None else colors[1]
    
    return pos_color, neg_color
```

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,200 @@
+from typing import Optional
```
renatomello:
```suggestion
from typing import List, Optional
```

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,200 @@
...
+# Based on Qiskit density state plot
+# https://github.com/Qiskit/qiskit/blob/stable/2.0/qiskit/visualization/state_visualization.py#L372-L622
+def plot_density_hist(
+    circuit,
+    title: str = "",
+    alpha: float = 0.5,
+    colors: Optional[list] = None,
+    backend=None,
+    fig_width=16,
+    fig_height=8,
```
renatomello:
```suggestion
    colors: Optional[List[str]] = None,
    fig_width: int = 16,
    fig_height: int = 8,
    backend=None,
```

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,200 @@
...
+
+    Returns:
+        tuple: Respectively, the figure, and axes for the real and the imaginary parts.
+    """
+
+    # if exec_circ has measurements then, error measure gates are present
+    if len(circuit.measurements) > 0:
+        raise ValueError(
+            "Circuit must not contain measurement gates for density matrix visualization"
+        )
```
renatomello:
```suggestion
    if len(circuit.measurements) > 0:
        raise_error(
            ValueError,
            "Circuit must not contain measurement gates for density matrix visualization"
        )
```

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,200 @@
...
+    dy = dx.copy()
+    dzr = backend.np.reshape(matrix_real, (-1))
+    dzi = backend.np.reshape(matrix_imag, (-1))
+
+    fig = plt.figure(figsize=(fig_width, fig_height), facecolor="w")
+    ax1 = fig.add_subplot(1, 2, 1, projection="3d", computed_zorder=False)
+    ax2 = fig.add_subplot(1, 2, 2, projection="3d", computed_zorder=False)
+
+    max_dzr = np.max(dzr)
+    max_dzi = np.max(dzi)
```
renatomello:
```suggestion
    dzr = backend.np.reshape(matrix_real, (-1))
    dzi = backend.np.reshape(matrix_imag, (-1))

    max_dzr = np.max(dzr)
    max_dzi = np.max(dzi)

    fig = plt.figure(figsize=(fig_width, fig_height), facecolor="w")
    ax1 = fig.add_subplot(1, 2, 1, projection="3d", computed_zorder=False)
    ax2 = fig.add_subplot(1, 2, 2, projection="3d", computed_zorder=False)
```

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,200 @@
...
+
+    max_plot_size = min(fig_width / 2.25, fig_height)
+    max_font_size = int(2 * max_plot_size)
+    max_zoom = 10 / (10 + np.sqrt(max_plot_size))
```
renatomello: I wonder why these choices of numbers and if this should / could be defined as `kwargs` of the function for users to adjust them if needed.

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,200 @@
...
+                (
+                    cmap_pos(norm_pos(height))
+                    if height >= 0
+                    else cmap_neg(norm_neg(height))
+                )
+                for height in dz
+            ]
+        )
+
+        for mask, zorder in ((dz < 0, 0.625), (dz >= 0, 0.875)):
```
renatomello: Same for these numbers

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,200 @@
...
+                    alpha=alpha,
+                    zorder=zorder,
+                    color=colors_mapping[mask],
+                    shade=True,
+                )
+
+        if min_dz < 0 < max_dz:
+            xlim, ylim = [0, label_x], [0, label_y]
+            verts = [list(zip(xlim + xlim[::-1], np.repeat(ylim, 2), [0] * 4))]
+            plane = Poly3DCollection(verts, alpha=0.25, facecolor="k", linewidths=1)
```
renatomello: Same for this `alpha`.

tests/test_ui_plot_state.py
```diff
@@ -0,0 +1,163 @@
...
+    circuit = Circuit(nqubits)
+    circuit.add(gates.H(0))
+    circuit.add(gates.CNOT(0, 1))
+
+    with pytest.raises(Exception) as excinfo:
+        colors = ["red", "blue", "green"]
+        plot_density_hist(circuit, title="Test Circuit State", colors=colors)
+        assert str(
+            excinfo.value
+        ) == "Colors must be a list of len=2, got {} instead".format(len(colors))
```
renatomello:
```suggestion
        ) == f"Colors must be a list of length 2, got {len(colors)} instead."
```

src/qibo/ui/plot_state.py
sergiomtzlosa: I thin that it is not neede to add these numbers on kwargs, they are calkculated from the plot size which is already set as input parameter

src/qibo/ui/plot_state.py
sergiomtzlosa: I will include a kwargs for these numbers

src/qibo/ui/plot_state.py
sergiomtzlosa: The alpha will be also in kwargs

tests/test_ui_plot_state.py
```diff
@@ -0,0 +1,164 @@
...
+import pytest
+from matplotlib.testing.compare import compare_images
+
+import qibo
+from qibo import Circuit, gates
+from qibo.ui.plot_state import plot_density_hist
+
+from .utils import fig2png
+
+qibo.set_backend("numpy")
```
renatomello: I would prefer if the global backend were not messed with during tests. One alternative could be use `backend` as an argument like the other tests and using `pytest.skip` for all backends but `numpy`.

sergiomtzlosa: I removed the `numpy` backend which is not needed anymore because now, we match images

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,214 @@
+from typing import List, Optional
+
+import matplotlib.pyplot as plt
+import numpy as np
+from matplotlib.colors import LinearSegmentedColormap, Normalize
```
MatteoRobbiati:
```suggestion
from matplotlib.colors import LinearSegmentedColormap
```
Since you call it later directly as `plt.Normalize`

src/qibo/ui/plot_state.py
MatteoRobbiati, line 1: I was thinking that we could try to align a bit the files in the UI module. I am fine with `plot_state.py`, but currently we also have `result_visualization.py` there. I would choose one of the two and be uniform.
TBH I am fine with both choices :)

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,214 @@
...
+    alpha: float = 0.5,
+    colors: Optional[List[str]] = None,
+    fig_width: int = 16,
+    fig_height: int = 8,
+    backend=None,
+    **kwargs,
+):
+    """Plot the real and imaginary parts of the density matrix.
+
+    Given a :class:`qibo.models.circuit.Circuit`, plots the real and imaginary parts
```
MatteoRobbiati:
```suggestion
    Given a :class:`qibo.models.circuit.Circuit`, plot the real and imaginary parts
```

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,214 @@
...
+
+    max_dzr = np.max(dzr)
+    max_dzi = np.max(dzi)
+
+    fig = plt.figure(figsize=(fig_width, fig_height), facecolor="w")
+    ax1 = fig.add_subplot(1, 2, 1, projection="3d", computed_zorder=False)
+    ax2 = fig.add_subplot(1, 2, 2, projection="3d", computed_zorder=False)
+
+    max_plot_size = min(fig_width / 2.25, fig_height)
+    max_font_size = int(2 * max_plot_size)
```
MatteoRobbiati: This adaptive fontsize scaling is a problem when we try to plot for more than 4 qubits. It results in unreadable axes (ticks labels overlap). 

I would suggest to do something similar to what I am proposed in `result_visualization`, where only some of the labels are shown. Or, we can fix the fontsize, so that the ticks labels will be separate (but probably unreadable for many-qubits systems).

I am not sure it is an optimal solution, and probably this tool is not supposed to be used with a large number of qubits. Anyway, as it is now, it rapidly becomes difficult to read the axes.

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,214 @@
...
+from matplotlib.colors import LinearSegmentedColormap, Normalize
+from mpl_toolkits.mplot3d.art3d import Poly3DCollection
+
+from qibo.backends import _check_backend
+from qibo.config import raise_error
+
+
+def _validate_colors(colors):
+    if colors is None:
+        return "#ff7f0e", "#1f77b4"
```
MatteoRobbiati: In `qibo.ui.drawing_utils`, two defaults colors are already defined as `QIBO_DEFAULT_COLOR` and `QIBO_COMPLEMENTARY_COLOR`. These colors recall the same color choices done in Qiboedu's tutorials. I would recommend to try to standardize these choices.
Of course, we can discuss on which colors we prefer as default :)

src/qibo/ui/plot_state.py
```diff
@@ -0,0 +1,214 @@
...
+
+    if len(colors) != 2:
+        raise_error(
+            ValueError, f"Colors must be a list of length 2, got {len(colors)} instead"
+        )
+
+    pos_color = "#ff7f0e" if colors[0] is None else colors[0]
+    neg_color = "#1f77b4" if colors[1] is None else colors[1]
```
MatteoRobbiati: Same comment of before about colors.

src/qibo/ui/plot_state.py
sergiomtzlosa: I move the contents of `plot_state.py` to `result_visualization.py`

src/qibo/ui/plot_state.py
sergiomtzlosa: thanks for the suggestion!

src/qibo/ui/plot_state.py
sergiomtzlosa: thanks again!

src/qibo/ui/plot_state.py
sergiomtzlosa: well, it is always challenging to plot a circuit with a big number of qubits, indeed, it is unmanageable. I tried to plot labels with a step selection of `total // n_most_relevant_components` until the `n_most_relevant_components` is reached, maybe it is not the best solution but, as I said, this is a huge issue to tackle IMHO

renatomello:
I think we are still missing the addition of this function to the documentation

doc/source/code-examples/examples.rst
```diff
@@ -370,3 +370,30 @@ For example, we can draw the QFT circuit for 5-qubits:
...
+
+Qibo provides a function ``plot_density_hist`` to visualize the real and imaginary parts of a quantum state's density matrix
+as 3D bar plots using matplotlib. This is useful for inspecting the structure of quantum states, especially for small systems.
+The function return, the figure, and axes for the real and the imaginary parts.
+
+The function supports options to control the number of tick labels shown (for large systems), transparency, colors, and figure size.
+
+.. testcode::
+
+    import matplotlib.pyplot as plt
```
MatteoRobbiati:
```suggestion
```

I think is not necessary :)

MatteoRobbiati (approved):
Thanks! LGTM. 
Just a small comment follows, but for me we are good to go.

PR #1676, opened 2025-06-19
Handling dependent parameters in circuits

BrunoLiegiBastonLiegi:
This provides the ``Circuit`` object with a ``._independent_parameters_map`` object which is useful for handling the parameters of the gates that are supposed to change together and are thus dependent. This is particularly useful for the implementation of QML equivariant models qiboteam/qiboml#112, qiboteam/qiboml#108, qiboteam/qiboml#99.
The new attribute maps an index of a gate (in the circuit queue) to all the indices of the gates that are supposed to share its parameters. For instance:
``` python
{0: {0,2,3}, 1: {1,4}}
```
will indicate that the 0-th gate parameters are shared among the 0, 2 and 3-rd gates, whereas the 1-st gate parameters are shared among the 1-st and 4-th gates.
If no parameter is shared than the dictionary looks simply as:
```python
{0: {0}, 2: {2}, 3: {3}, 1: {1,4}}
```
Additionally, this implies that the ``set_parameters`` now takes as input a collection of parameters with length equal to the circuit's ``_independent_parameters_map``.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/models/circuit.py
+++ b/src/qibo/models/circuit.py
@@ -192,6 +192,7 @@ class Circuit:
         # Keep track of parametrized gates for the ``set_parameters`` method
         self.parametrized_gates = _ParametrizedGates()
         self.trainable_gates = _ParametrizedGates()
+        self._independent_parameters_map = None
         self.measurements = []  # list of non-collapsible measurements
 
         self._final_state = None
@@ -675,6 +676,23 @@ class Circuit:
 
                 return gate.result
 
+            elif isinstance(gate, ParametrizedGate):
+                independent = True
+                if self._independent_parameters_map is not None:
+                    for idx in self._independent_parameters_map:
+                        indep_params = self.queue[idx].parameters[0]
+                        params = gate.parameters
+                        # for now only 1 parameter gates are checked
+                        if len(params) == 1 and np.shares_memory(
+                            indep_params, params[0]
+                        ):
+                            self._independent_parameters_map[idx].add(self.ngates)
+                            independent = False
+                if independent:
+                    self._independent_parameters_map[self.ngates] = {
+                        self.ngates,
+                    }
+
             self.queue.append(gate)
             for measurement in list(self.measurements):
                 if set(measurement.qubits) & set(gate.qubits):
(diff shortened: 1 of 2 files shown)


chmwzc:
Nice, this will be useful for chemistry as well!
If possible, it'll be great if you could extend it to having different gate parameters linearly dependent on a single shared parameter, e.g. something like `gates.RX(theta)`, ..., `gates.RX(-theta)` for the single shared parameter `theta`

BrunoLiegiBastonLiegi:
Mmmmh, supporting that kind of behaviour with this approach will be very tricky. At that point it is maybe better to allow the user to define a custom parameter setter. However, this will probably break the sum of circuits, as we don't know how to compose different custom ``set_parameters`` in general. I am still unsure about what would be the best and most general way to handle dependent parameters.

chmwzc:
> Mmmmh, supporting that kind of behaviour with this approach will be very tricky. At that point it is maybe better to allow the user to define a custom parameter setter. However, this will probably break the sum of circuits, as we don't know how to compose different custom `set_parameters` in general. I am still unsure about what would be the best and most general way to handle dependent parameters.

Yeah, I agree that it's definitely tricky, which is why I said "if possible". Don't bother about my request if it doesn't seem feasible😆

alecandido:
At this point, wouldn't be just simpler to stop handling parameters manually, and just rely on the tracing provided by tensor libraries?

I.e. you can use a `Circuit` as it is, taking all independent parameters, and then just provide a splitting layer in front, such that:
```mermaid
flowchart LR
    theta["θ"] --> splitting
    splitting --> alpha["α"] --> circuit
    splitting --> beta["β"] --> circuit
```
And then you train the whole model over the input parameters ($\theta$ in the example).

The logic is that it is useful for Qibo to do things that other library are not providing. But reimplementing TensorFlow/PyTorch/... features, instead of integrating with them, it seems to be just time-consuming, but with no specific advantage (better writing guides and examples about how to best use them together).

BrunoLiegiBastonLiegi:
Yes that was the idea originally, however, this is the current structure implemented in qiboml:
- the user optionally defines an encoder (non-parametrizable for now, will be in the short future, thus leading to similar problems) and combines it with on or more qibo circuits (parametrized)
- internally, we extract the parameters from the circuit and assign them to the qiboml model (torch or keras)
- later one whenever we do a forward pass, we take the parameters from the model and set them into the circuit

Thus, in order to correctly handle dependent parameters and transformations of them, either we provide the circuit object with the capabilities to deal with them (e.g. via the ``independent_parameters_map`` defined here and the possibility to define custom rules for setting parameters), or, in qiboml, we build a wrapper to the qibo circuit that takes care of that part.

alecandido:
The original proposal for qiboml was actually not at the level of a model, but rather a composable layer (well, some models apparently can be treated as layers - I'm not that expert with high-level ML abstractions - in which case, so much the better).

So, the idea was actually to have a `CircuitLayer` or anything similar, which would have had gates' parameters as inputs, and observables as outputs (thus including the decoder, but not the encoder - which involves no quantum operation).

In that layout, dependent parameters could be realized simply by taking the input parameters and applying linear transformations and activation functions.

If a model is equivalent enough to a layer, then you can still do that.
In a sense, dependent parameters could be just part of your encoding layer.

BrunoLiegiBastonLiegi:
So you're saying it would be better to go with the second option, thus not taking simple qibo circuits anymore but rather a custom structure encapsulating qibo circuits (at least in those cases where more fine grained control over the parameters is needed). Did I get it right?

alecandido:
> So you're saying it would be better to go with the second option, thus not taking simple qibo circuits anymore but rather a custom structure encapsulating qibo circuits (at least in those cases where more fine grained control over the parameters is needed). Did I get it right?

Yes, indeed. Sorry if I didn't say it clear.

I really see this as the `qiboml` core business.
If you're in the domain of parameters' optimization (especially, numerical optimization - the only thing we care about in Qibo), that's certainly ML. And it seems suboptimal to build `qiboml` capabilities directly into `qibo`, since this would break modularity (and duplicate `qiboml` somehow).

PR #1679, opened 2025-06-30
LOQC integration proposal

ericbrts:
Hi Qibo team,

Here's a proposal for linear optics quantum computing integration in Qibo, that I wrote along with Jean Senellart.
With this implementation, one would be able to:
* build a linear optics circuit with pure Qibo objects
* simulate the circuit with a Perceval wrapped strong simulation back-end
* send the circuit on the Quandela Cloud, targetting either an online simulator or an actual QPU.

A usage example was written in the notebook located in scripts/playground.ipynb.

This pull request is a proof of concept and is not ready to merge.
Adding a whole modality to Qibo is a big work and will probably require discussions and your guidance.

For instance, parts of the proposed code is not clean. We wanted to use the H gate both in the gate-based and photonics world. But as the qubits/wires a gate acts on are hold by the gate instance, as its internal data, we're facing the issue that H is plugged on 1 wire (qubit) in gate based, and on 2 wires (optical modes) in photonics.
One way of solving this fundamental issue could be to externalize the gate positioning data from the gate instance. The circuit instance would know where each component is placed, but not the components themselves. That way, a H gate would have 1 connection in gate-based modality circuits and 2 in photonics modality.

Also, I've implemented a short class that's able to send a computation order on Quandela Cloud. Studying Qibo, I feel like it would find its place in qibo-cloud-backends. However, for readability sake, I chose to write the code directly in qibo for now.

Thanks,
Best wishes,
Eric


src/qibo/backends/photonic_strong_simulation.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def _cast(self, x, dtype):
+        if isinstance(x, list):
+            return np.array(x, dtype=dtype)
+        return x.astype(dtype)
+
+    def BS(self, theta):
+        cos = np.cos(theta / 2.0) + 0j
+        isin = -1j * np.sin(theta / 2.0)
+        return self._cast([[cos, isin], [isin, cos]], dtype=self.dtype)
```
renatomello, line 35: Like the Hadamard gate that is already defined in `qibo`, this gate seems to be the same as the `gates.RX`.

pyproject.toml
alecandido, line 1: Just in case: whenever you'd like to run tests, you will need to generate a synced lockfile.

You can do it by running `poetry lock` (pure lockfile generation) or `poetry update` (lockfile generate and Poetry environment update, i.e. installing the lockfile specified dependencies, removing outdated ones).

alecandido:
> Hi Qibo team,
> 
> Here's a proposal for linear optics quantum computing integration in Qibo, that I wrote along with Jean Senellart.

Hi @ericbrts, that sounds like a great contribution!
I'm really glad you started working on this PR, and we're certainly willing to provide the needed support to complete it :)

I started reading your code (and apparently @renatomello as well), so I may leave some comments to better understand the proposed additions.

> This pull request is a proof of concept and is not ready to merge.
> Adding a whole modality to Qibo is a big work and will probably require discussions and your guidance.

Understood.
I hope that we could help to make it as simple as possible. But I acknowledge there may be some differences, and we need to brainstorm a bit about how to extend Qibo's scope, without compromising existing features (i.e. minimizing change required to existing users' code - nothing in the best case).

> For instance, parts of the proposed code is not clean. We wanted to use the H gate both in the gate-based and photonics world. But as the qubits/wires a gate acts on are hold by the gate instance, as its internal data, we're facing the issue that H is plugged on 1 wire (qubit) in gate based, and on 2 wires (optical modes) in photonics.

In this sense, because of my personal ignorance regarding LOQC, I may benefit from a quick intro. Do you have any source to recommend for a quick overview of basic concepts?

> One way of solving this fundamental issue could be to externalize the gate positioning data from the gate instance. The circuit instance would know where each component is placed, but not the components themselves. That way, a H gate would have 1 connection in gate-based modality circuits and 2 in photonics modality.

Yes, you may be interested to know that, almost one year ago, we were actually reconsidering the Qibo primitives. "Accidentally" in this same direction.
The final aim was different, but we started a proof-of-concept exploration, named [`qibo-core`](https://github.com/qiboteam/qibo-core/), which should have provided a `Circuit` structure, which held gates more or less as symbolic tokens.

This project is not a priority, for now. But, if it will result that we need some intensive extension of the current Qibo circuit API, we may explore again to do that in that context (since we have other valid motivations to invest in it).

ericbrts:
Hello all, hello @alecandido 

> In this sense, because of my personal ignorance regarding LOQC, I may benefit from a quick intro. Do you have any source to recommend for a quick overview of basic concepts?

One way you could see the difference in 1-qubit gates between gate based and photonics, is that you can think of one photon in two optical modes representing one qubit (it's the positional or "dual rail" encoding, which is one of the most straight forward encoding). |1,0>_photonics_ would correspond to |0>_qubit_ and |0,1>_photonics_ to |1>_qubit_.

If you have some time, you could have a look to this part of the Perceval tutorial:
https://perceval.quandela.net/docs/v0.13/notebooks/Tutorial.html#2.-Single-qubit-gates

However, it's not photonics goal to map logical gates one-to-one and, in linear optics, multiple photons could interact in a so called "one-qubit" gate.

Best,
Eric

alecandido:
> If you have some time, you could have a look to this part of the Perceval tutorial:
> [perceval.quandela.net/docs/v0.13/notebooks/Tutorial.html#2.-Single-qubit-gates](https://perceval.quandela.net/docs/v0.13/notebooks/Tutorial.html#2.-Single-qubit-gates)

Great, thanks for the advice! @ericbrts

I will make sure to check the tutorial before getting back with further comments.

However, if linear optics has "further dimensions", wrt qubits operations, such that you can not bijectively map the first one onto the second, then maybe better to think about something different from stuffing both in the same representation.
Regarding "what" else, well... I'll take the time to study first :P

In any case, if you already have a target in mind (another instance of an interface spanning linear optics and qubits Hilbert spaces), or a clear plan on your side, feel free to propose what do you believe to be best.

It's very interesting to extend Qibo beyond its current scope, and integrate it Perceval. But, well, that certainly requires some meditation first :)

PR #1690, opened 2025-07-17
Fix to data-based mitigation methods

BrunoLiegiBastonLiegi:
As per title, this introduces the use of a simulation backend (numpy for now) to get the noise free values for data-based mitigation methods, such as CDR.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/hamiltonians/hamiltonians.py
+++ b/src/qibo/hamiltonians/hamiltonians.py
@@ -568,7 +568,7 @@ class SymbolicHamiltonian(AbstractHamiltonian):
 
         keys = list(freq.keys())
         counts = list(freq.values())
-        counts = self.backend.cast(counts, dtype=type(counts[0])) / sum(counts)
+        counts = self.backend.cast(counts, dtype=self.backend.np.float64) / sum(counts)
         expvals = []
         for term in self.terms:
             qubits = {
--- a/src/qibo/models/error_mitigation.py
+++ b/src/qibo/models/error_mitigation.py
@@ -9,8 +9,15 @@ import numpy as np
 from scipy.optimize import curve_fit
 
 from qibo import gates
-from qibo.backends import _check_backend, _check_backend_and_local_state, get_backend
+from qibo.backends import (
+    NumpyBackend,
+    _check_backend,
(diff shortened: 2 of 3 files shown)


src/qibo/hamiltonians/hamiltonians.py
```diff
@@ -568,7 +568,7 @@ def expectation_from_samples(self, freq: dict, qubit_map: list = None) -> float:
 
         keys = list(freq.keys())
         counts = list(freq.values())
-        counts = self.backend.cast(counts, dtype=type(counts[0])) / sum(counts)
+        counts = self.backend.cast(counts, dtype=self.backend.np.float64) / sum(counts)
```
andrea-pasquale, line 571:
```suggestion
        counts = self.backend.cast(counts, dtype=self.backend.dtype) / sum(counts)
```
This should work as well, right?

andrea-pasquale (approved):

src/qibo/models/error_mitigation.py
```diff
@@ -9,9 +9,16 @@
...
-from qibo.backends import _check_backend, _check_backend_and_local_state, get_backend
+from qibo.backends import (
+    NumpyBackend,
+    _check_backend,
+    _check_backend_and_local_state,
+    get_backend,
+)
 from qibo.config import raise_error
 
+SIMULATION_BACKEND = NumpyBackend()
```
MatteoRobbiati, line 20: I leave this as a comment here, and if we want we can open a dedicated issue. 
As discussed with @BrunoLiegiBastonLiegi, this is a solution for now, but we would ideally want to have:

- clifford backend for pure clifford methods (e.g. importance clifford sampling);
- quasi clifford backend for hybrid solutions (e.g. CDR or vnCDR).

MatteoRobbiati (approved):
Thanks! I left a comment below, we can open an issue for that.

src/qibo/hamiltonians/hamiltonians.py
BrunoLiegiBastonLiegi: unfortunately I think that the default dtype is complex128 and here we really want real numbers

PR #1698, opened 2025-08-04
Add `gates.FanOut` gate

renatomello:
This gate is essentially the complement of the multi-controlled Toffoli gate. While the latter is composed of a target X controlled by multiple qubits, the former is composed of a single control and many qubits as targets of the X gate.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/gates/abstract.py
+++ b/src/qibo/gates/abstract.py
@@ -2,6 +2,7 @@
 
 import collections
 import json
+from abc import abstractmethod
 from math import pi
 from typing import List, Sequence, Tuple
 
@@ -553,6 +554,14 @@ class Gate:
             f"Generator eigenvalue is not implemented for {self.__class__.__name__}",
         )
 
+    @abstractmethod
+    def generator(self, backend):
+        """This function returns the gate's generator.
+
+        Returns:
+            array: generator.
+        """
+
     def basis_rotation(self):
         """Transformation required to rotate the basis for measuring the gate."""
         raise_error(
@@ -659,3 +668,10 @@ class ParametrizedGate(Gate):
         backend = _check_backend(backend)
 
         return backend.matrix_parametrized(self)
+
+    @abstractmethod
+    def gradient(self, backend=None) -> Gate:
+        """Returns Unitary with gate's gradient.
+
+        Only gates with a single parameter are currently supported.
+        """
--- a/src/qibo/gates/gates.py
+++ b/src/qibo/gates/gates.py
@@ -3,6 +3,7 @@ from typing import List, Tuple, Union
 
 import numpy as np
 
+from qibo.backends import _check_backend
 from qibo.config import PRECISION_TOL, raise_error
 from qibo.gates.abstract import Gate, ParametrizedGate
 from qibo.parameter import Parameter
@@ -623,6 +624,20 @@ class _Rn_(ParametrizedGate):
             gate = super().controlled_by(*q)
         return gate
 
+    def generator_eigenvalue(self):
+        return 0.5
+
+    def gradient(self, backend=None) -> "Gate":
+        """Gradient of the Gate."""
+        backend = _check_backend(backend=backend)
+        return Unitary(
+            -1j
+            * self.generator_eigenvalue()
+            * self.generator(backend)
+            @ backend.matrix_parametrized(self),
+            *self.target_qubits,
+        )
+
 
 class RX(_Rn_):
     """Rotation around the X-axis of the Bloch sphere.
@@ -659,8 +674,9 @@ class RX(_Rn_):
     def qasm_label(self):
         return "rx"
 
-    def generator_eigenvalue(self):
-        return 0.5
+    def generator(self, backend=None):
+        backend = _check_backend(backend)
+        return backend.matrices.X
 
 
 class RY(_Rn_):
@@ -698,8 +714,9 @@ class RY(_Rn_):
     def qasm_label(self):
         return "ry"
 
-    def generator_eigenvalue(self):
-        return 0.5
+    def generator(self, backend=None):
+        backend = _check_backend(backend)
+        return backend.matrices.Y
 
 
 class RZ(_Rn_):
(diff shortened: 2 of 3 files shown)


src/qibo/backends/npmatrices.py
```diff
@@ -492,6 +492,30 @@ def DEUTSCH(self, theta):
...
+        # based on Backend.matrix_fused method
+        rank = len(q)
+        matrix = self.I(2**rank)
+
+        for qubit in range(1, rank):
+            gmatrix = self.CNOT
+            eye = self.I(2 ** (rank - 2))
+            gmatrix = self.np.kron(gmatrix, eye)
+            original_shape = gmatrix.shape
+            gmatrix = self.np.reshape(gmatrix, 2 * rank * (2,))
```
BrunoLiegiBastonLiegi: all this part can be done just once outside the loop as it does not depend on the qubit

src/qibo/backends/npmatrices.py
```diff
@@ -492,6 +492,30 @@ def DEUTSCH(self, theta):
...
+            qubits = [0, qubit]
+            indices = qubits + [qub for qub in range(rank) if qub not in qubits]
+            indices = self.np.argsort(indices)
+            transpose_indices = list(indices)
+            transpose_indices.extend(indices + rank)
+            gmatrix = self.np.transpose(gmatrix, transpose_indices)
+            gmatrix = self.np.reshape(gmatrix, original_shape)
+            matrix = gmatrix @ matrix
+
+        return self._cast(matrix, dtype=self.dtype)
```
BrunoLiegiBastonLiegi: I have the feeling that this can be formulated recursively and cached, something like this:
```python
@cache
def FanOut_matrix(q, rank):
    # build the matrix for the single qubit
    # this should be equal to one iteration of the loop

def FanOut(self, *q):
    rank = len(q)
    matrix = FanOut_matrix(q, rank)
    if len(q) == 0:
        return self.I(2**rank)
    return self.FanOut(*q[1:]) @ matrix
```
(this is just some pseudo-code to give you a rough idea of what I meant)

renatomello: I mean, in a sense, the `CNOT` gate is already the cached version of the smallest `FanOut` gate possible. The loop is only responsible for introducing identity gates in the correct places.

src/qibo/backends/numpy.py
```diff
@@ -134,8 +134,10 @@ def matrix(self, gate):
         """Convert a gate to its matrix representation in the computational basis."""
         name = gate.__class__.__name__
         _matrix = getattr(self.matrices, name)
-        if callable(_matrix):
+        if callable(_matrix) and name in ("I", "GeneralizedRBS"):
             _matrix = _matrix(2 ** len(gate.target_qubits))
+        elif callable(_matrix) and name == "FanOut":
+            _matrix = _matrix(*gate.init_args)
```
andrea-pasquale: To fix #1705 we should enable this also for gates other than `FanOut`

renatomello: It should be properly fixed now.

src/qibo/backends/numpy.py
```diff
@@ -134,10 +134,20 @@ def matrix(self, gate):
...
+        elif name == "FanOut":
             _matrix = _matrix(*gate.init_args)
+        elif name == "GeneralizedRBS":
+            kwargs = dict(gate.init_kwargs)  # copy necessary
+            del kwargs["trainable"]
+            _matrix = _matrix(*gate.init_args, **kwargs)
+        elif callable(_matrix):
+            kwargs = dict(gate.init_kwargs)  # copy necessary
+            del kwargs["trainable"]
+            _matrix = _matrix(**kwargs)
```
andrea-pasquale: Perhaps a simpler approach could be to check if the gate is a `ParametrizedGate`and if so we can just use `matrix_parametrized` below.

renatomello: In this case, all `ParametrizedGate`s are callable and vice-versa, so it wouldn't make a difference.

andrea-pasquale: Yeah, but in this way you need to make a copy of `gate.init_args` which could be avoided using `matrix_parametrized`.

src/qibo/backends/npmatrices.py
BrunoLiegiBastonLiegi: Yeah, it's true that the `CNOT` is cached already and now you perform the `gmatrix` construction outside of the loop, but everytime you call the fanout you're repeating the same operations as the whole function is always the same for a fixed rank. You could just have:
```python
@cache
def _CachedFanout(rank):
        matrix = self.I(2**rank)
        gmatrix = self.CNOT
        eye = self.I(2 ** (rank - 2))
        gmatrix = self.np.kron(gmatrix, eye)
        original_shape = gmatrix.shape
        gmatrix = self.np.reshape(gmatrix, 2 * rank * (2,))

        for qubit in range(1, rank):
            qubits = [0, qubit]
            indices = qubits + [qub for qub in range(rank) if qub not in qubits]
            indices = self.np.argsort(indices)
            transpose_indices = list(indices)
            transpose_indices.extend(indices + rank)
            _gmatrix = self.np.transpose(gmatrix, transpose_indices)
            _gmatrix = self.np.reshape(_gmatrix, original_shape)
            matrix = _gmatrix @ matrix
        return matrix
```
and then the `FanOut` would just read:
```python
    def FanOut(self, *q):
        rank = len(q)
        matrix = self._CachedFanout(rank)
        return self._cast(matrix, dtype=self.dtype)
```
the problem with the caching is that you have several  `self.np` calls and thus to cache the `_CachedFanOut` you would need to define it inside the `__init__` and then add it to the class ``self._CachedFanOut = _CachedFanOut`` which is not the cleanest solution...

renatomello: I think just moving `gmatrix` out of the loop suffices for now.

andrea-pasquale (approved):

PR #1700, opened 2025-08-05
Support for `gate.Unitary` in Clifford backend

AlejandroSopena:
Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/backends/_clifford_operations.py
+++ b/src/qibo/backends/_clifford_operations.py
@@ -164,6 +164,17 @@ def RY(symplectic_matrix, q, nqubits, theta):
         return RY_3pi_2(symplectic_matrix, q, nqubits)
 
 
+def GPI2(symplectic_matrix, q, nqubits, phi):
+    if phi % (2 * np.pi) == 0: 
+        return RX(symplectic_matrix, q, nqubits, np.pi / 2)
+    elif (phi / np.pi - 1) % 2 == 0: 
+        return RX(symplectic_matrix, q, nqubits, -np.pi / 2)
+    elif (phi / (np.pi / 2) - 1) % 4 == 0: 
+        return RY(symplectic_matrix, q, nqubits, np.pi / 2)
+    else:  # theta == 3*pi/2 + 2*n*pi
+        return RY(symplectic_matrix, q, nqubits, -np.pi / 2)
+
+    
 def SWAP(symplectic_matrix, control_q, target_q, nqubits):
     """Decomposition --> CNOT-CNOT-CNOT"""
     r, x, z = _get_rxz(symplectic_matrix, nqubits)
--- a/src/qibo/backends/clifford.py
+++ b/src/qibo/backends/clifford.py
@@ -1,6 +1,7 @@
 """Module defining the Clifford backend."""
 
 import collections
+import itertools
 from functools import reduce
 from importlib.util import find_spec, module_from_spec
 from typing import Union
@@ -140,12 +141,180 @@ class CliffordBackend(NumpyBackend):
 
     def apply_gate_clifford(self, gate, symplectic_matrix, nqubits):
         """Apply a gate to a symplectic matrix."""
+        if isinstance(gate, gates.Unitary):
+                symplectic_matrix = self.apply_unitary(gate, symplectic_matrix, nqubits)
+
+                return symplectic_matrix
+        
         operation = getattr(self.engine, gate.__class__.__name__)
-        kwargs = (
-            {"theta": gate.init_kwargs["theta"]} if "theta" in gate.init_kwargs else {}
-        )
+
+        kwargs = {}
+        for param_name in ["theta", "phi"]:
+            if param_name in gate.init_kwargs:
+                kwargs[param_name] = gate.init_kwargs[param_name]
+
 
         return operation(symplectic_matrix, *gate.init_args, nqubits, **kwargs)
+    
+    def apply_unitary(self, gate, symplectic_matrix, nqubits):
+        """Apply a unitary gate to a symplectic matrix."""
+        temp_matrix = self._clifford_post_execution_reshape(symplectic_matrix, nqubits)
+
+        qubit_indices = list(gate.qubits)
+        m = len(qubit_indices)
+        
+        matrix = gate.matrix()
+        symplectic_m = self._compute_symplectic_matrix(matrix, m)
+        phase_m = self._get_phase_vector(matrix, m)
+
+        symplectic_n = self._embed_clifford(symplectic_m, nqubits, qubit_indices)
+        phase_n = self._embed_phase_vector(phase_m, nqubits, qubit_indices)
+
+        original_symplectic = temp_matrix[:, :2*nqubits].copy()
+        original_phases = temp_matrix[:, 2*nqubits].copy()
+
+        temp_matrix[:, :2*nqubits] = (original_symplectic @ symplectic_n.T) % 2
+        for i in range(2*nqubits):
+            v_old = original_symplectic[i, :]  
+            phase_old = original_phases[i]      
+
+            linear_correction = np.dot(v_old, phase_n) % 2
+            quadratic_correction = self._compute_quadratic_form(v_old, symplectic_n, nqubits)
+
+            total_correction = (linear_correction + quadratic_correction) % 2
+            temp_matrix[i, 2*nqubits] = (phase_old + total_correction) % 2
+
+        result = self._clifford_pre_execution_reshape(temp_matrix)
+        symplectic_matrix[:] = result  
+
+        return symplectic_matrix
+
+    def _pauli_string_to_matrix(self, pauli_str):
+        """Convert Pauli string to matrix (tensor product)."""
+        from qibo import matrices  # pylint: disable=C0415
+
+        pauli_matrices = (getattr(matrices, p) for p in pauli_str)
+        matrix = reduce(np.kron, pauli_matrices)
+        return self.cast(matrix, dtype=matrix.dtype)
+
+    def _pauli_to_binary(self, pauli_str, nqubits):
+        """Convert Pauli string to binary vector of length :math`2*nqubits`."""
+        pauli_symplectic = self.np.zeros((2*nqubits))
+        for q, term in enumerate(pauli_str):
+            if term in ['X', 'Y']:
+                pauli_symplectic[q] = 1
+            if term in ['Z', 'Y']:
+                pauli_symplectic[q + nqubits] = 1
+
+        return pauli_symplectic
+
+    def _compute_symplectic_matrix(self, unitary, m):
+        """Compute symplectic matrix for Clifford unitary on :math`m` qubits."""
+        pauli_gens = []
+        for i in range(m):
+            p = ['I']*m; p[i] = 'X'
+            pauli_gens.append(''.join(p))
+        for i in range(m):
+            p = ['I']*m; p[i] = 'Z'
+            pauli_gens.append(''.join(p))
+
+        symplectic = np.zeros((2*m, 2*m), dtype=int)
+        symplectic = self.cast(symplectic, dtype=symplectic.dtype)
+
+        for i, p_str in enumerate(pauli_gens):
+            pauli = self._pauli_string_to_matrix(p_str)
+            pauli_uconj = unitary @ pauli @ unitary.conj().T
+
+            found = False
+            for candidate_str in itertools.product('IXYZ', repeat=m):
+                candidate_str = ''.join(candidate_str)
+                candidate_P = self._pauli_string_to_matrix(candidate_str)
+                
+                for phase in [1, -1, 1j, -1j]:
+                    if np.allclose(pauli_uconj, phase * candidate_P, atol=1e-10):
+                        v = self._pauli_to_binary(candidate_str, m)
+                        symplectic[:, i] = v
+                        found = True
+                        break
+                if found:
+                    break
+                    
+            if not found:
+                raise ValueError(f"Could not find conjugated Pauli for {p_str}")
+
+        return symplectic % 2
+
+    def _get_phase_vector(self, unitary, m):
+        """Compute phase vector :math`r_U` of length :math`2m` for Clifford unitary.
+        :math`r_U[j] = 0` if :math`U g_j U^\\dagger = +P_j` else :math`1`."""
+        pauli_gens = []
+        for i in range(m):
+            p = ['I']*m; p[i] = 'X'
+            pauli_gens.append(''.join(p))
+        for i in range(m):
+            p = ['I']*m; p[i] = 'Z'
+            pauli_gens.append(''.join(p))
+
+        phase = np.zeros(2*m, dtype=int)
+        for j, p_str in enumerate(pauli_gens):
+            pauli = self._pauli_string_to_matrix(p_str)
+            pauli_uconj = unitary @ pauli @ unitary.conj().T
+
+            found = False
+            for candidate_str in itertools.product('IXYZ', repeat=m):
+                candidate_str = ''.join(candidate_str)
+                candidate_P = self._pauli_string_to_matrix(candidate_str)
+                if np.allclose(pauli_uconj, candidate_P):
+                    phase[j] = 0
+                    found = True
+                    break
+                elif np.allclose(pauli_uconj, -candidate_P):
+                    phase[j] = 1
+                    found = True
+                    break
+            if not found:
+                raise ValueError(f"Phase vector: failed for {p_str}")
+        return phase
+
+    def _embed_clifford(self, symplectic_m, n, qubit_indices):
+        """Embed m-qubit symplectic :math`S_U_m` into n-qubit system at qubit_indices."""
+        symplectic_n = np.eye(2*n, dtype=int)
+        symplectic_n = self.cast(symplectic_n, dtype=symplectic_m.dtype)
+
+        x_indices = qubit_indices
+        z_indices = [q + n for q in qubit_indices]
+        full_indices = x_indices + z_indices
+
+        symplectic_n[np.ix_(full_indices, full_indices)] = symplectic_m
+
+        return symplectic_n % 2
+
+    def _embed_phase_vector(self, phase_m, n, qubit_indices):
+        """Embed m-qubit phase vector into n-qubit system."""
+        phase_n = np.zeros(2*n, dtype=int)
+        phase_n = self.cast(phase_n, dtype=phase_m.dtype)
+        m = len(qubit_indices)
+        
+        qubit_indices = np.array(qubit_indices)
+        phase_n[qubit_indices] = phase_m[:m]
+        phase_n[qubit_indices + n] = phase_m[m:]
+
+        return phase_n
+
+    def _compute_quadratic_form(self, v, symplectic, n):
+        """Compute quadratic form for phase correction."""
+        v_x = v[:n]
+        v_z = v[n:]
+
+        #A = symplectic[:n, :n]      # X-X block
+        B = symplectic[:n, n:]      # X-Z block
+        C = symplectic[n:, :n]      # Z-X block
+        #D = symplectic[n:, n:]      # Z-Z block
+
+        term1 = (v_x.T @ B @ v_z) % 2
+        term2 = (v_z.T @ C @ v_x) % 2
+        
+        return (term1 + term2) % 2
 
     def apply_channel(self, channel, state, nqubits):
         probabilities = channel.coefficients + (1 - np.sum(channel.coefficients),)
@@ -210,9 +379,7 @@ class CliffordBackend(NumpyBackend):
             nqubits = circuit.nqubits
 
             state = self.zero_state(nqubits) if initial_state is None else initial_state
-
             state = self._clifford_pre_execution_reshape(state)
-
             for gate in circuit.queue:
                 gate.apply_clifford(self, state, nqubits)
 
--- a/tests/test_backends_clifford.py
+++ b/tests/test_backends_clifford.py
@@ -47,7 +47,7 @@ THETAS_1Q = [
     th + 2 * i * np.pi for i in range(2) for th in [0, np.pi / 2, np.pi, 3 * np.pi / 2]
 ]
 
-AXES = ["RX", "RY", "RZ"]
+AXES = ["RX", "RY", "RZ", "GPI2"]
 
 
 @pytest.mark.parametrize("axis,theta", list(product(AXES, THETAS_1Q)))
@@ -57,8 +57,8 @@ def test_rotations_1q(backend, theta, axis):
     qubits = np.random.randint(3, size=2)
     H_qubits = np.random.choice(range(3), size=2, replace=False)
     c.add(gates.H(q) for q in H_qubits)
-    c.add(getattr(gates, axis)(qubits[0], theta=theta))
-    c.add(getattr(gates, axis)(qubits[1], theta=theta))
+    c.add(getattr(gates, axis)(qubits[0], theta))
+    c.add(getattr(gates, axis)(qubits[1], theta))
     clifford_state = clifford_bkd.execute_circuit(c).state()
     numpy_state = numpy_bkd.execute_circuit(c).state()
     numpy_state = backend.cast(numpy_state)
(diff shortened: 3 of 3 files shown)


AlejandroSopena:
The GPU tests that are failing are not related to this PR

src/qibo/backends/_clifford_operations.py
```diff
@@ -164,6 +164,17 @@ def RY(symplectic_matrix, q, nqubits, theta):
...
 
+def GPI2(symplectic_matrix, q, nqubits, phi):
+    if phi % (2 * np.pi) == 0:
+        return RX(symplectic_matrix, q, nqubits, np.pi / 2)
+    elif (phi / np.pi - 1) % 2 == 0:
+        return RX(symplectic_matrix, q, nqubits, -np.pi / 2)
+    elif (phi / (np.pi / 2) - 1) % 4 == 0:
+        return RY(symplectic_matrix, q, nqubits, np.pi / 2)
+    else:  # theta == 3*pi/2 + 2*n*pi
+        return RY(symplectic_matrix, q, nqubits, -np.pi / 2)
```
renatomello:
```suggestion
def GPI2(symplectic_matrix, q, nqubits, phi):
    if phi % (2 * np.pi) == 0:
        return RX(symplectic_matrix, q, nqubits, np.pi / 2)
    
    if (phi / np.pi - 1) % 2 == 0:
        return RX(symplectic_matrix, q, nqubits, -np.pi / 2)
    
    if (phi / (np.pi / 2) - 1) % 4 == 0:
        return RY(symplectic_matrix, q, nqubits, np.pi / 2)
    
    # theta == 3*pi/2 + 2*n*pi
    return RY(symplectic_matrix, q, nqubits, -np.pi / 2)
```

What happens when the choice of angle leads to a non-Clifford GPI2? It should not work here.

AlejandroSopena: I think the point is that these gates are only used in the `CliffordBackend.apply_gate_clifford` method, so it's not necessary for them to work when the gate is not Clifford. Take a look, for example, at the implementation of other gates like `RX` or `RZ` in the `_clifford_operations.py` file.

renatomello: I see, the check if it is Clifford or not already happened.

renatomello:
```suggestion
    if phi % (2 * np.pi) == 0:
        return RX(symplectic_matrix, q, nqubits, np.pi / 2)

    if (phi / np.pi - 1) % 2 == 0:
        return RX(symplectic_matrix, q, nqubits, -np.pi / 2)

    if (phi / (np.pi / 2) - 1) % 4 == 0:
        return RY(symplectic_matrix, q, nqubits, np.pi / 2)

    # theta == 3*pi/2 + 2*n*pi
    return RY(symplectic_matrix, q, nqubits, -np.pi / 2)
```

src/qibo/backends/clifford.py
```diff
@@ -1,6 +1,7 @@
 """Module defining the Clifford backend."""
 
 import collections
+import itertools
```
renatomello:
```suggestion
from itertools import product
```

src/qibo/backends/clifford.py
```diff
@@ -140,13 +141,252 @@ def _clifford_post_execution_reshape(self, state, nqubits: int):
...
+
+        symplectic = np.zeros((2 * m, 2 * m), dtype=int)
+        symplectic = self.cast(symplectic, dtype=symplectic.dtype)
+
+        for i, p_str in enumerate(pauli_gens):
+            pauli = self._pauli_string_to_matrix(p_str)
+            pauli_uconj = unitary @ pauli @ unitary.conj().T
+
+            found = False
+            for candidate_str in itertools.product("IXYZ", repeat=m):
```
renatomello:
```suggestion
            for candidate_str in product("IXYZ", repeat=m):
```

src/qibo/backends/clifford.py
```diff
@@ -140,13 +141,252 @@ def _clifford_post_execution_reshape(self, state, nqubits: int):
...
+            p = ["I"] * m
+            p[i] = "Z"
+            pauli_gens.append("".join(p))
+
+        phase = np.zeros(2 * m, dtype=int)
+        for j, p_str in enumerate(pauli_gens):
+            pauli = self._pauli_string_to_matrix(p_str)
+            pauli_uconj = unitary @ pauli @ unitary.conj().T
+
+            for candidate_str in itertools.product("IXYZ", repeat=m):
```
renatomello:
```suggestion
            for candidate_str in product("IXYZ", repeat=m):
```

src/qibo/backends/clifford.py
```diff
@@ -94,7 +95,7 @@ def calculate_frequencies(self, samples):
 
         return collections.Counter(dict(zip(res, counts)))
 
-    def zero_state(self, nqubits: int):
+    def zero_state(self, nqubits: int, i_phase=False):
```
BrunoLiegiBastonLiegi, line 98: please add also a dosctring for `i_phase`

src/qibo/backends/clifford.py
```diff
@@ -140,13 +141,252 @@ def _clifford_post_execution_reshape(self, state, nqubits: int):
...
+            ]
+            symplectic_pauli = self._conjugate_pauli(
+                symplectic_gate, symplectic_pauli, nqubits
+            )
+
+            symplectic_matrix[i, : 2 * nqubits] = symplectic_pauli[0]
+            symplectic_matrix[i, 2 * nqubits] = symplectic_pauli[1]
+            symplectic_matrix[i, 2 * nqubits + 1] = symplectic_pauli[2]
+
+        return symplectic_matrix
```
BrunoLiegiBastonLiegi: this should probably be part of ``_clifford_operations.py`` with the other gates

src/qibo/backends/clifford.py
```diff
@@ -140,13 +141,252 @@ def _clifford_post_execution_reshape(self, state, nqubits: int):
...
+        from qibo import matrices  # pylint: disable=C0415
+
+        pauli_matrices = []
+        for p in pauli_str:
+            matrix = getattr(matrices, p)
+            if p == "Y":
+                matrix = 1j * matrix
+            pauli_matrices.append(matrix)
+        matrix = reduce(np.kron, pauli_matrices)
+        return self.cast(matrix, dtype=matrix.dtype)
```
BrunoLiegiBastonLiegi:
```suggestion
    def _pauli_string_to_matrix(self, pauli_str):
        """Convert Pauli string to matrix (tensor product)."""
        from qibo import matrices  # pylint: disable=C0415

        paulis= {
            pauli: self.engine.cast(getattr(matrices, p), dtype=self.dtype) 
            for pauli in ("X", "Y", "Z")
        }
        paulis["Y"] *= 1j
        pauli_matrices = [paulis.get(p) for p in pauli_str]
        matrix = reduce(self.engine.np.kron, pauli_matrices)
        return matrix
```
Better to leverage the engine, especially if the number of paulis is large. 
No need to reconstruct the paulis every time for each element in ``pauli_str``. Actually, one could also think about making ``paulis`` a cached property or an attribute.
One last thing, if you know that this function is gonna be called several times with the same ``pauli_str`` you could ``@cache`` it.

src/qibo/backends/clifford.py
```diff
@@ -140,13 +141,252 @@ def _clifford_post_execution_reshape(self, state, nqubits: int):
...
+
+    def _pauli_to_binary(self, pauli_str, nqubits):
+        """Convert Pauli string to binary vector of length :math`2*nqubits`."""
+        pauli_symplectic = self.np.zeros(2 * nqubits)
+        for q, term in enumerate(pauli_str):
+            if term in ["X", "Y"]:
+                pauli_symplectic[q] = 1
+            if term in ["Z", "Y"]:
+                pauli_symplectic[q + nqubits] = 1
+        return self.cast(pauli_symplectic, dtype=pauli_symplectic.dtype)
```
BrunoLiegiBastonLiegi:
```suggestion
    def _pauli_to_binary(self, pauli_str, nqubits):
        """Convert Pauli string to binary vector of length :math`2*nqubits`."""
        pauli_symplectic = self.np.zeros(2 * nqubits, dtype=self.np.uint8)
        for q, term in enumerate(pauli_str):
            if term in ["X", "Y"]:
                pauli_symplectic[q] = 1
            if term in ["Z", "Y"]:
                pauli_symplectic[q + nqubits] = 1
        return pauli_symplectic
```
If this is a symplectic matrix shouldn't it be a boolean or uint8 matrix?
Are you sure you need the cast at the end? ``self.np.zeros`` should already provide an arrive native to the ``engine``.

src/qibo/backends/clifford.py
```diff
@@ -140,13 +141,252 @@ def _clifford_post_execution_reshape(self, state, nqubits: int):
...
+        """Compute symplectic matrix for Clifford unitary on :math`m` qubits."""
+        pauli_gens = []
+        for i in range(m):
+            p = ["I"] * m
+            p[i] = "X"
+            pauli_gens.append("".join(p))
+        for i in range(m):
+            p = ["I"] * m
+            p[i] = "Z"
+            pauli_gens.append("".join(p))
```
BrunoLiegiBastonLiegi:
```suggestion
        pauli_gens_x,  pauli_gens_z = [], []
        for i in range(m):
            p = ["I"] * m
            p = (p, p.copy())
            p[0][i] = "X"
            p[1][i] = "Z"
            pauli_gens_x.append("".join(p[0]))
            pauli_gens_z.append("".join(p[1]))
        pauli_gens = pauli_gens_x + pauli_gens_z
```
This can be condensed in a single loop, optionally ``@cache``d for fixed ``m``

src/qibo/backends/clifford.py
```diff
@@ -140,13 +141,252 @@ def _clifford_post_execution_reshape(self, state, nqubits: int):
...
+
+        symplectic = np.zeros((2 * m, 2 * m), dtype=int)
+        symplectic = self.cast(symplectic, dtype=symplectic.dtype)
```
BrunoLiegiBastonLiegi:
```suggestion
        symplectic = self.np.zeros((2 * m, 2 * m), dtype=self.np.uint8)
```
same as above, not sure about the ``uint8`` and the cast

src/qibo/backends/clifford.py
```diff
@@ -140,13 +141,252 @@ def _clifford_post_execution_reshape(self, state, nqubits: int):
...
+            pauli = self._pauli_string_to_matrix(p_str)
+            pauli_uconj = unitary @ pauli @ unitary.conj().T
+
+            found = False
+            for candidate_str in product("IXYZ", repeat=m):
+                candidate_str = "".join(candidate_str)
+                candidate_P = self._pauli_string_to_matrix(candidate_str)
+
+                for phase in [1, -1, 1j, -1j]:
+                    if np.allclose(pauli_uconj, phase * candidate_P, atol=1e-10):
```
BrunoLiegiBastonLiegi:
```suggestion
                    if self.np.allclose(pauli_uconj, phase * candidate_P, atol=1e-10):
```

src/qibo/backends/clifford.py
```diff
@@ -140,13 +141,252 @@ def _clifford_post_execution_reshape(self, state, nqubits: int):
...
+                        v = self._pauli_to_binary(candidate_str, m)
+                        symplectic[i, :] = v
+                        found = True
+                        break
+                if found:
+                    break
```
BrunoLiegiBastonLiegi, line 252: Do you always have the guarantee to find it?

src/qibo/backends/clifford.py
```diff
@@ -140,13 +141,252 @@ def _clifford_post_execution_reshape(self, state, nqubits: int):
...
+        """
+        pauli_gens = []
+        for i in range(m):
+            p = ["I"] * m
+            p[i] = "X"
+            pauli_gens.append("".join(p))
+        for i in range(m):
+            p = ["I"] * m
+            p[i] = "Z"
+            pauli_gens.append("".join(p))
```
BrunoLiegiBastonLiegi: This seems to be the same snippet of before, thus advocating for the creation of an auxiliary ``@cache``d function.

src/qibo/backends/clifford.py
```diff
@@ -140,13 +141,252 @@ def _clifford_post_execution_reshape(self, state, nqubits: int):
...
+
+        phase = np.zeros(2 * m, dtype=int)
```
BrunoLiegiBastonLiegi:
```suggestion
        phase = self.np.zeros(2 * m, dtype=self.np.uint8)
```

src/qibo/backends/clifford.py
```diff
@@ -140,13 +141,252 @@ def _clifford_post_execution_reshape(self, state, nqubits: int):
...
+                if np.allclose(pauli_uconj, candidate_P) or np.allclose(
+                    pauli_uconj, 1j * candidate_P
+                ):
+                    phase[j] = 0
+                    break
+                elif np.allclose(pauli_uconj, -candidate_P) or np.allclose(
+                    pauli_uconj, -1j * candidate_P
+                ):
+                    phase[j] = 1
+                    break
```
BrunoLiegiBastonLiegi: This seems very similar to the previous loop, with the exception, of course, that you set the phase only. I've got the feeling that you could do both the symplectic matrix and the phase at the same time in a single function, thus sparing to repeat the exact same loop (which looks to be quite expensive for many qubits) twice. Is there any particular reason to keep them separated?

src/qibo/backends/clifford.py
```diff
@@ -140,13 +141,252 @@ def _clifford_post_execution_reshape(self, state, nqubits: int):
...
+
+    def _get_phase_vector_dk(self, symplectic, m):
+        """Compute phase vector :math`d` of length :math`2m` for Clifford unitary :math`U`.
+        :math`d[j] = 0` if :math`U g_j U^\\dagger = (-1)^r p_j` with :math`r=0` or :math`1` else :math`1`.
+        """
+        u_matrix = np.zeros((2 * m, 2 * m), dtype=int)
+        u_matrix[0:m, m : 2 * m] = np.eye(m)
+        u_matrix = self.cast(u_matrix, dtype=u_matrix.dtype)
+        d = np.diag(symplectic @ (u_matrix @ symplectic.T) % 2) % 2
+        return self.cast(d, dtype=d.dtype)
```
BrunoLiegiBastonLiegi:
```suggestion
        u_matrix = self.np.zeros((2 * m, 2 * m), dtype=self.np.uint8)
        u_matrix[0:m, m : 2 * m] = self.np.eye(m, dtype=self.np.uint8)
        d = self.np.diag(symplectic @ (u_matrix @ symplectic.T) % 2) % 2
        return d
```

src/qibo/backends/clifford.py
```diff
@@ -140,13 +141,252 @@ def _clifford_post_execution_reshape(self, state, nqubits: int):
...
+                symplectic_matrix @ (u_matrix @ symplectic_matrix.T)
+                + np.outer(phase_d, phase_d)
+            )
+            % 2
+        )
+
+        ph_dot_sv = np.dot(phase_h, symplectic_vector)
+        sv_lows_sv = np.dot(symplectic_vector, lows @ symplectic_vector)
+        delta_pd_dot_sv = delta * pd_dot_sv
+        new_epsilon = (epsilon + ph_dot_sv + sv_lows_sv + delta_pd_dot_sv) % 2
```
BrunoLiegiBastonLiegi:
```suggestion
        symplectic_matrix, phase_h, phase_d = symplectic_gate
        symplectic_vector, epsilon, delta = symplectic_pauli

        new_symplectic_vector = (symplectic_matrix.T @ symplectic_vector) % 2

        pd_dot_sv = self.np.dot(phase_d, symplectic_vector) % 2
        new_delta = (delta + pd_dot_sv) % 2

        u_matrix = self.np.zeros((2 * nqubits, 2 * nqubits), dtype=self.np.uint8)
        u_matrix[0:nqubits, nqubits : 2 * nqubits] = self.np.eye(nqubits, dtype=self.np.uint8)

        lows = self.np.tril(
            (
                symplectic_matrix @ (u_matrix @ symplectic_matrix.T)
                + self.np.outer(phase_d, phase_d)
            )
            % 2
        )

        ph_dot_sv = self.np.dot(phase_h, symplectic_vector)
        sv_lows_sv = self.np.dot(symplectic_vector, lows @ symplectic_vector)
        delta_pd_dot_sv = delta * pd_dot_sv
        new_epsilon = (epsilon + ph_dot_sv + sv_lows_sv + delta_pd_dot_sv) % 2
```

src/qibo/backends/clifford.py
```diff
@@ -140,13 +141,252 @@ def _clifford_post_execution_reshape(self, state, nqubits: int):
...
+        """
+        n_rows, n_cols = dehaene_tableau.shape
+        n = (n_cols - 2) // 2
+
+        X_part = dehaene_tableau[:, :n]
+        Z_part = dehaene_tableau[:, n : 2 * n]
+        real_phases = dehaene_tableau[:, -2]
+        i_phases = dehaene_tableau[:, -1]
+
+        final_real_phases = np.zeros_like(real_phases, dtype=np.uint8)
```
BrunoLiegiBastonLiegi:
```suggestion
        final_real_phases = self.np.zeros_like(real_phases, dtype=self.np.uint8)
```

src/qibo/backends/clifford.py
```diff
@@ -140,13 +141,252 @@ def _clifford_post_execution_reshape(self, state, nqubits: int):
...
+
+            y_count = np.sum(x_vec & z_vec)
+
+            total_i_power = (i_phase + y_count) % 4
+
+            final_phase = real_phase
+            if total_i_power == 2:
+                # i^2 = -1, flip the sign
+                final_phase = (final_phase + 1) % 2
+            final_real_phases[row_idx] = final_phase
```
BrunoLiegiBastonLiegi:
```suggestion
        y_count = self.np.sum(X_part[:-1] & Z_part[:-1], axis=-1)
        total_i_power = (i_phases[:-1] + y_count) % 4
            
        final_phases = real_phases
        indices = total_i_power == 2
        final_phases[indices] = (final_phases[indices] + 1) % 2
        final_real_phases = final_phases
```
I think the loop can be dropped her, but doublecheck it first.

src/qibo/backends/clifford.py
```diff
@@ -140,13 +141,252 @@ def _clifford_post_execution_reshape(self, state, nqubits: int):
...
+        """Embed m-qubit phase vector into n-qubit system."""
+        phase_n = np.zeros(2 * n, dtype=int)
+        phase_n = self.cast(phase_n, dtype=phase_m.dtype)
+        m = len(qubit_indices)
+
+        qubit_indices = np.array(qubit_indices)
+        phase_n[qubit_indices] = phase_m[:m]
+        phase_n[qubit_indices + n] = phase_m[m:]
+
+        return phase_n
```
BrunoLiegiBastonLiegi:
```suggestion
    def _embed_clifford(self, symplectic_m, n, qubit_indices):
        """Embed m-qubit symplectic :math`S_U_m` into n-qubit system at qubit_indices."""
        symplectic_n = self.np.eye(2 * n, dtype=self.np.uint8)

        x_indices = qubit_indices
        z_indices = [q + n for q in qubit_indices]
        full_indices = x_indices + z_indices

        symplectic_n[np.ix_(full_indices, full_indices)] = symplectic_m

        return symplectic_n % 2

    def _embed_phase_vector(self, phase_m, n, qubit_indices):
        """Embed m-qubit phase vector into n-qubit system."""
        phase_n = self.np.zeros(2 * n, dtype=self.np.uint8)
        m = len(qubit_indices)

        qubit_indices = np.array(qubit_indices)
        phase_n[qubit_indices] = phase_m[:m]
        phase_n[qubit_indices + n] = phase_m[m:]

        return phase_n
```

src/qibo/backends/clifford.py
AlejandroSopena: No, it can be combined as you suggest

src/qibo/backends/clifford.py
AlejandroSopena: Yes, because Clifford gates map Paulis to Paulis up to a phase and I’m looping over the four possible phases

AlejandroSopena:
@BrunoLiegiBastonLiegi, thanks for the detailed review. I think it is ready.

src/qibo/models/error_mitigation.py
```diff
@@ -2,22 +2,25 @@
...
     _check_backend_and_local_state,
+    _get_engine_name,
     get_backend,
 )
 from qibo.config import raise_error
+from qibo.hamiltonians.hamiltonians import SymbolicHamiltonian
+from qibo.symbols import X, Y, Z
 
 SIMULATION_BACKEND = NumpyBackend()
+CLIFORD_BACKEND = CliffordBackend(engine="numpy")
```
BrunoLiegiBastonLiegi:
```suggestion
CLIFFORD_BACKEND = CliffordBackend(engine="numpy")
```

src/qibo/models/error_mitigation.py
```diff
@@ -979,44 +1010,60 @@ def error_sensitive_circuit(circuit, observable, seed=None, backend=None):
...
-    observable_pauli_liouville = comp_to_pauli @ observable_liouville
 
-    index = int(
-        backend.np.where(backend.np.abs(observable_pauli_liouville) >= 1e-5)[0][0]
+    result = backend.execute_circuit(sampled_circuit.invert(), nshots=1)
+
+    symplectic_matrix = result.symplectic_matrix[:-1, :-1]
+
+    terms = observable.terms[0].factors
+    pauli_symplectic = backend.np.zeros((2 * circuit.nqubits, 1))
```
BrunoLiegiBastonLiegi, line 1033:
```suggestion
    pauli_symplectic = backend.np.zeros((2 * circuit.nqubits, 1), dtype=backend.np.uint8)
```

src/qibo/models/error_mitigation.py
```diff
@@ -979,44 +1010,60 @@ def error_sensitive_circuit(circuit, observable, seed=None, backend=None):
...
+    observable = 1
+    for i in range(circuit.nqubits):
+        x_component = pauli_symplectic[i] == 1
+        z_component = pauli_symplectic[i + circuit.nqubits] == 1
+        if x_component and z_component:
+            observable *= Y(i, backend=backend)
+        elif x_component:
+            observable *= X(i, backend=backend)
+        elif z_component:
+            observable *= Z(i, backend=backend)
```
BrunoLiegiBastonLiegi:
```suggestion
    x_component = pauli_symplectic[:circuit.nqubits] == 1
    z_component = pauli_symplectic[circuit.nqubits:] == 1
    y_obs = reduce(prod, [Y(i, backend=backend) for i in (x_component * z_component).nonzero()])
    x_obs = reduce(prod, [X(i, backend=backend) for i in (x_component * ~z_component).nonzero()])
    x_obs = reduce(prod, [Z(i, backend=backend) for i in (z_component *  ~x_component).nonzero()])
```
for a large number of qubits the vectorized operations should give an edge, the `nonzero` could actually be a `nonzero()[0]` pauli symplectic is a matrix I believe, please double check in case

src/qibo/models/error_mitigation.py
```diff
@@ -1088,23 +1135,57 @@ def ICS(
...
     ]
 
     data = {"noise-free": [], "noisy": []}
     lambda_list = []
 
     for training_circuit in training_circuits:
-        circuit_result = SIMULATION_BACKEND.execute_circuit(
-            training_circuit, nshots=nshots
+        training_circuit_copy = training_circuit.copy(deep=True)
+        circuit_result = CLIFORD_BACKEND.execute_circuit(
```
BrunoLiegiBastonLiegi:
```suggestion
        circuit_result = CLIFFORD_BACKEND.execute_circuit(
```

BrunoLiegiBastonLiegi (commented):
Thanks, just a bunch of final comments on the remaining part. 
The only two things I think it would be good to have at some point are:
- Moving the function that applies the unitary, and all the functions it depends on to the `clifford_operations` file. This would allow for customization by the different engines, e.g. jitting with numba and writing the cuda kernels with cupy.
- Leveraging the packed bits representation that is used by the rest of the backend and provides a huge speed up.

These two do not need to be addressed here in this PR, but we should probably open an issue about them to remind us.

scarrazza:
@AlejandroSopena could you please have a look at this last round of comments so we can include this PR in the release planned for today?

AlejandroSopena:
> Thanks, just a bunch of final comments on the remaining part. The only two things I think it would be good to have at some point are:
> 
> * Moving the function that applies the unitary, and all the functions it depends on, to the `clifford_operations` file. This would allow for customization by the different engines, e.g. jitting with numba and writing the cuda kernels with cupy.
> * Leveraging the packed bits representation that is used by the rest of the backend which provides a huge speed up.
> 
> These two do not need to be addressed here in this PR, but we should probably open an issue about them to remind us.

I agree with these two points. Once I have some time, I will address them in a separate PR

AlejandroSopena:
> @AlejandroSopena could you please have a look at this last round of comments so we can include this PR in the release planned for today?

It is ready now.

BrunoLiegiBastonLiegi (approved):

PR #1708, opened 2025-09-01
Refactor `quantum_info.random_clifford`

### LINKED ISSUE #1640: random_clifford() does not generate all 2-qubit Cliffords
The `qibo.quantum_info.random_clifford()` function based on [Bravyi&Maslov](https://arxiv.org/abs/2003.09412) is supposed to generate a random uniformly distributed n-qubit Clifford. Therefore, by sampling enough times we should be able to generate all the distinct Cliffords of a given n. 

By doing this for n=1 we find all the 24 distinct Cliffords after ~100 samples. 

However, for n=2 it only finds 10495 of the 11520 distinct Cliffords (after ~110k samples) and then not a single new one, no matter how much I increase the number of samples (tried up to 5e6) and how I mess with the random seed. 

Could it be a bug in `random_clifford()`?

Here is a code doing the brute force search above:

```
import numpy as np
from qibo import quantum_info as qi
from qibo import Circuit

def all_cliffords(nqubits, maxsamples):
 
    dim = 2**(nqubits**2+2*nqubits)*np.prod([4**j-1 for j in range(1,nqubits+1)])
 
    C = []
    sample = 1
 
    while len(C) < dim and sample < maxsamples:
        qasm = qi.random_clifford(nqubits, return_circuit=True, seed=sample).to_qasm()
        if qasm not in C:
            C.append(qasm)
            print(f'{len(C)} Cliffords found after {sample} samples')
        sample += 1
        if sample % 10000 == 0:
            print(f'Reached sample # {sample}')
    
    if len(C) == dim:
        print(f'Success! Found all the {dim} Cliffords after {sample} samples!')
    else:
        print(f'Fail! Only found {len(C)} out of {dim} Cliffords after {sample} samples')
            
    return [Circuit.from_qasm(c) for c in C]
```

renatomello:
Closes #1640.

After trying too many things in vain, I decided to completely refactor the `random_clifford` function using as a base the equivalent function in `qiskit==1.2`. Below, results for $2 \\times 10^{6}$ samples show all $11520$ elements of the two-qubit Clifford group being sampled and the count distribution converging to the uniform distribution.

output

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibo/quantum_info/random_ensembles.py
+++ b/src/qibo/quantum_info/random_ensembles.py
@@ -12,6 +12,7 @@ from qibo import Circuit, gates, matrices
 from qibo.backends import NumpyBackend, _check_backend_and_local_state
 from qibo.config import MAX_ITERATIONS, PRECISION_TOL, raise_error
 from qibo.quantum_info.basis import comp_basis_to_pauli
+from qibo.quantum_info.clifford import Clifford
 from qibo.quantum_info.superoperator_transformations import (
     choi_to_chi,
     choi_to_kraus,
@@ -184,7 +185,7 @@ def random_hermitian(
     matrix = random_gaussian_matrix(dims, dims, seed=local_state, backend=backend)
 
     if semidefinite:
-        matrix = backend.np.matmul(backend.np.conj(matrix).T, matrix)
+        matrix = backend.np.conj(matrix).T @ matrix
     else:
         matrix = (matrix + backend.np.conj(matrix).T) / 2
 
@@ -235,7 +236,7 @@ def random_unitary(dims: int, measure: Optional[str] = None, seed=None, backend=
         D = backend.np.diag(R)
         D = D / backend.np.abs(D)
         R = backend.np.diag(D)
-        unitary = backend.np.matmul(Q, R)
+        unitary = Q @ R
     elif measure is None:
         from scipy.linalg import expm
 
@@ -569,21 +570,16 @@ def random_density_matrix(
             state = random_gaussian_matrix(
                 dims, rank, mean=0, stddev=1, seed=local_state, backend=backend
             )
-            state = backend.np.matmul(
-                state, backend.np.transpose(backend.np.conj(state), (1, 0))
-            )
+            state = state @ backend.np.transpose(backend.np.conj(state), (1, 0))
             state = state / backend.np.trace(state)
         else:
             nqubits = int(np.log2(dims))
             state = backend.identity_density_matrix(nqubits, normalize=False)
             state += random_unitary(dims, seed=local_state, backend=backend)
-            state = backend.np.matmul(
-                state,
-                random_gaussian_matrix(dims, rank, seed=local_state, backend=backend),
+            state = state @ random_gaussian_matrix(
+                dims, rank, seed=local_state, backend=backend
             )
-            state = backend.np.matmul(
-                state, backend.np.transpose(backend.np.conj(state), (1, 0))
-            )
+            state = state @ backend.np.transpose(backend.np.conj(state), (1, 0))
             state /= backend.np.trace(state)
 
     state = backend.cast(state, dtype=state.dtype)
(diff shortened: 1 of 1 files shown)


renatomello:
@BrunoLiegiBastonLiegi could you help me with the last test that is failing? I did not understand what the test is doing, but It has something to do with measurements and I'm quite sure the problem comes from the fact that the test depends on a fixed seed that is leading to different results now since the underlying functions changed.

src/qibo/quantum_info/random_ensembles.py
```diff
@@ -655,88 +648,59 @@ def random_clifford(
...
-                and permutations[k] < permutations[j]
-            ):  # pragma: no cover
-                b = local_state.integers(0, 2)
-                delta_matrix[k, j] = b
-
-    # get first element of the Borel group
-    clifford_circuit = _operator_from_hadamard_free_group(
-        gamma_matrix, delta_matrix, density_matrix
+    gamma = backend.np.diag(
+        local_state.integers(2, size=nqubits, dtype=backend.np.int8)
```
BrunoLiegiBastonLiegi: Do you actually care about the sign of the integers or is this one just a boolean matrix in fact? Because in that case you could just use ``np.uint8``.

renatomello:
@BrunoLiegiBastonLiegi fixed some tests after merging `master` into this branch. Now we are back to only failing at that strange test with mid-circuit measurements.

EDIT: I found a seed that makes the tests pass. This is obviously not a proper solution.

src/qibo/quantum_info/random_ensembles.py
```diff
@@ -655,88 +648,91 @@ def random_clifford(
...
-        random_pauli(
-            nqubits,
-            depth=1,
-            return_circuit=True,
-            density_matrix=density_matrix,
-            seed=local_state,
-            backend=backend,
-        ),
+    gamma_prime = backend.np.diag(
+        local_state.integers(2, size=nqubits, dtype=backend.np.int8)
```
BrunoLiegiBastonLiegi: same here and in the following, uint8 maybe?

src/qibo/quantum_info/random_ensembles.py
```diff
@@ -655,88 +648,91 @@ def random_clifford(
...
+    )
+    gamma_prime = backend.cast(gamma, dtype=gamma_prime.dtype)
+
+    delta = backend.np.eye(nqubits, dtype=backend.np.int8)
+    delta = backend.cast(delta, dtype=delta.dtype)
```
BrunoLiegiBastonLiegi, line 664: mmmh weird, do you need to cast it even though the backend just generated it?

src/qibo/quantum_info/random_ensembles.py
```diff
@@ -655,88 +648,91 @@ def random_clifford(
...
+    # backend.np.block cannot be used below because there is no cupy equivalent
+    # hence the necessity for three backend.np.concatenate's
+    had_free_operator_1 = backend.np.concatenate(
+        [
+            backend.np.concatenate([delta, zero], axis=1),
+            backend.np.concatenate([prod1, inv1], axis=1),
+        ],
+        axis=0,
+    )
+    # had_free_operator_2 = backend.np.block([[delta_prime, zero], [prod2, inv2]])
```
BrunoLiegiBastonLiegi:
```suggestion
```

tests/test_quantum_info_random.py
```diff
@@ -344,11 +344,22 @@ def test_random_clifford(backend, nqubits, return_circuit, density_matrix, seed)
...
+
+    result_two = np.kron(
+        matrices.H @ matrices.Z @ matrices.X, matrices.SDG @ matrices.Z
+    )
+    result_two = matrices.CNOT @ result_two
+    result_two = np.kron(matrices.H, matrices.I) @ result_two
+    result_two = cnot_10 @ result_two
+    result_two = np.kron(matrices.SDG, matrices.I) @ result_two
+    result_two = cnot_10 @ result_two
+    result_two = np.kron(matrices.H @ matrices.SDG, matrices.I) @ result_two
```
BrunoLiegiBastonLiegi: How long does it take to generate, let's say ~10^4-10^5 samples with random clifford? I wonder if we could make this test more robust, namely checking that the distribution resembles on average a uniform one for instance. For one qubit it should not be a problem, for 2 qubits instead we could try to generate a reasonable amount of samples (time wise) compute the probability of each clifford from the samples and compare the average with the expected probability, i.e. 1/11520.

src/qibo/quantum_info/random_ensembles.py
renatomello: Yes, because for some backends `np` is `numpy` instead of their "engine". This is what I'm trying to solve in https://github.com/qiboteam/qibo/pull/1711

tests/test_quantum_info_random.py
renatomello: I think variance is still too high at only `10**5` samples. For instance, look at the plot above that has like a `10%` error bar with `2 * 10**6` samples.

PR #1711, opened 2025-09-11
Backend refactoring

renatomello:
Per title. Needs https://github.com/qiboteam/qibojit/pull/229 and https://github.com/qiboteam/qiboml/pull/149.

Important: @scarrazza Even though almost nothing changes for end users. I propose this PR to be the main base for a `qibo 0.3.0` release.

Locally, tests are passing for `numpy` and `numba` sans a small issue involving `repeated execution` and `numba` that I'm still investigating, so I decided to open this for review already.

This PR moves most of the `NumpyBackend` methods to the abstract `Backend` class.
The methods are rewritten to be as agnostic as possible to the presence of `numpy`. Now, the methods depend on `Backend.engine`, which is defined by each specific backend (*i.e.*  `engine = numpy` for `NumpyBackend` and `NumbaBackend`, or `engine = cupy` for `CupyBackend`, or `engine = torch` for `PyTorchBackend`, etc).

Also, functions dedicated to array manipulation are now backend methods, making it easier not only to use but also to overwrite them when necessary. For example, instead of calling `backend.np.reshape` to reshape an object, we can now use `backend.reshape` and internally the correct `engine` is chosen. The same goes for most linear algebra functions.

Since this PR is already too big as is, I prefer to work on docstrings and other modifications in separate PRs to avoid an even bigger nightmare. 

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/gates/gates.py
+++ b/src/qibo/gates/gates.py
@@ -2798,6 +2798,15 @@ class GeneralizedRBS(ParametrizedGate):
     def hamming_weight(self):
         return len(self.init_args[0]) == len(self.init_args[1])
 
+    def on_qubits(self, qubit_map: dict):
+        qubits_in = tuple(qubit_map.get(q) for q in self.init_args[0])
+        qubits_out = tuple(qubit_map.get(q) for q in self.init_args[1])
+        gate = self.__class__(qubits_in, qubits_out, **self.init_kwargs)
+        if self.is_controlled_by:
+            controls = (qubit_map.get(q) for q in self.control_qubits)
+            gate = gate.controlled_by(*controls)
+        return gate
+
     def _base_decompose(self, *free, use_toffolis=True, **kwargs) -> List[Gate]:
         """Decomposition of :math:`\\text{gRBS}` gate.
 
--- a/src/qibo/models/_openqasm.py
+++ b/src/qibo/models/_openqasm.py
@@ -1,9 +1,8 @@
 """Qibo wrapper for QASM 3.0 parser."""
 
-from itertools import repeat
 from typing import Union
 
-import numpy as np
+import numpy as np  # cannot be removed
 import openqasm3
 
 import qibo
@@ -34,40 +33,49 @@ class CustomQASMGate:
 
     def get_gate(self, qubits: Union[list, tuple], args: Union[list, tuple]):
         """Returns the gates composing the defined gate applied on the
-        specified qubits with the specified ``args`` as a unique :class:`qibo.gates.special.FusedGate`.
+        specified qubits with the specified ``args`` as a unique
+        :class:`qibo.gates.special.FusedGate`.
 
         Args:
             qubits (list or tuple): Qubits where to apply the gates.
             args (list or tuple): Arguments to evaluate the gates on.
 
         Returns:
-            :class:`qibo.gates.special.FusedGate`: the composed gate evaluated on the input qubits with the input arguments.
+            :class:`qibo.gates.special.FusedGate`: the composed gate evaluated
+            on the input qubits with the input arguments.
         """
         if len(self.args) != len(args):
             raise_error(
                 ValueError,
-                f"Invalid `args` argument passed to the user-defined gate `{self.name}` upon construction. {args} was passed but something of the form {self.args} is expected.",
+                f"Invalid `args` argument passed to the user-defined gate `{self.name}`"
+                + f"upon construction. {args} was passed but something of the form {self.args}"
+                + "is expected.",
             )
         elif len(self.qubits) != len(qubits):
             raise_error(
                 ValueError,
-                f"Invalid `qubits` argument passed to the user-defined gate `{self.name}` upon construction. {qubits} was passed but something of the form {self.qubits} is expected.",
+                f"Invalid `qubits` argument passed to the user-defined gate `{self.name}` "
+                + f"upon construction. {qubits} was passed but something of the form "
+                + f"{self.qubits} is expected.",
             )
         qubit_map = dict(zip(self.qubits, qubits))
         args_map = dict(zip(self.args, args))
         return self._construct_fused_gate(self.gates, qubits, qubit_map, args_map)
 
     def _construct_fused_gate(self, gates, qubits, qubit_map, args_map):
-        """Constructs a :class:`qibo.gates.special.FusedGate` out of the provided list of gates on the specified qubits.
+        """Constructs a :class:`qibo.gates.special.FusedGate` out of the provided
+        list of gates on the specified qubits.
 
         Args:
             gates (list(:class:`qibo.gates.Gate`)): List of gates to build the fused gate from.
             qubits (list(int)): List of qubits to construct the gate on.
-            qubit_map (dict): Mapping between the placeholders for the qubits contained in `gates` and the actual qubits indices to apply them on.
-            args_map (dict): Mapping between the placeholders for the kwargs contained in `gates` and the actual kwargs values.
+            qubit_map (dict): Mapping between the placeholders for the qubits contained
+                in `gates` and the actual qubits indices to apply them on.
+            args_map (dict): Mapping between the placeholders for the kwargs contained
+                in `gates` and the actual kwargs values.
 
         Returns:
-            (qibo.gates.special.FusedGate): The resulting fused gate.
+            :class:`qibo.gates.special.FusedGate`: The resulting fused gate.
         """
         fused_gate = FusedGate(*qubits)
         for gate in gates:
@@ -84,15 +92,20 @@ class CustomQASMGate:
         return fused_gate
 
     def _compile_gate_qubits_and_args(self, gate, qubit_map, args_map):
-        """Compile the qubits and arguments placeholders contained in the input gate with their actual values.
+        """Compile the qubits and arguments placeholders contained in the
+        input gate with their actual values.
 
         Args:
-            gate (:class:`qibo.gates.Gate`): The input gate containing the qubits and arguments placeholders.
-            qubit_map (dict): Mapping between the placeholders for the qubits contained in `gate` and the actual qubits indices to apply them on.
-            args_map (dict): Mapping between the placeholders for the kwargs contained in `gate` and the actual kwargs values.
+            gate (:class:`qibo.gates.Gate`): The input gate containing the
+                qubits and arguments placeholders.
+            qubit_map (dict): Mapping between the placeholders for the
+                qubits contained in `gate` and the actual qubits indices
+                to apply them on.
+            args_map (dict): Mapping between the placeholders for the kwargs
+                contained in `gate` and the actual kwargs values.
 
         Returns:
-            tuple(list, list): The compiled qubits and arguments.
+            Tuple(List[int], List[int]): The compiled qubits and arguments.
         """
         new_qubits = [qubit_map[q] for q in gate.qubits]
         new_args = [args_map.get(arg, arg) for arg in gate.init_kwargs.values()]
@@ -112,9 +125,8 @@ def _qibo_gate_name(gate):
     if gate == "p":
         return "U1"
 
-    if gate in ["u", "U"]:
-        # `u` for QASM 2.0
-        # `U` for QASM 3.0
+    if gate in ("u", "U"):
+        # `u` for QASM 2.0, `U` for QASM 3.0
         return "U3"
 
     return gate.upper()
@@ -196,7 +208,10 @@ class QASMParser:
             if len(self.c_registers[register]) != len(self.q_registers[qubit]):
                 raise_error(
                     RuntimeError,
-                    f"Trying to store quantum measurements from quantum register {qubit}: {self.q_registers[qubit]} to the classical register {register} of length {len(self.c_registers[register])}, the lengths of the quantum and classical registers should match.",
+                    f"Trying to store quantum measurements from quantum register {qubit}: "
+                    + f"{self.q_registers[qubit]} to the classical register {register} of "
+                    + f"length {len(self.c_registers[register])}, the lengths of the quantum "
+                    + "and classical registers should match.",
                 )
             qubits = self.q_registers[qubit]
         return getattr(qibo.gates, "M")(*qubits, register_name=register)
@@ -280,7 +295,9 @@ class QASMParser:
 
     def _merge_measurements(self, gates):
         """Merges separated measurements of a same register into a single one.
-        This is needed because qibo doesn't allow to separetely define two measurements in a same register:
+
+        This is needed because qibo doesn't allow to separetely define
+        two measurements in a same register:
 
         # not allowed
         circuit.add(gates.M(0, register="m0"))
--- a/src/qibo/models/circuit.py
+++ b/src/qibo/models/circuit.py
@@ -1137,12 +1137,13 @@ class Circuit:
         """Convert circuit to a QASM string.
 
         Args:
-            extended_compatibility (bool): if ``True``, unrolls more exotic gates in their
-                decomposition by defining them as custom gates, this increases the compatibility
-                with other frameworks, such as qiskit. Defaults to ``False``.
+            extended_compatibility (bool): if ``True``, unrolls more exotic gates
+                in their decomposition by defining them as custom gates,
+                this increases the compatibility with other frameworks,
+                such as ``qiskit``. Defaults to ``False``.
 
         Returns:
-            (str) the generated QASM.
+            str: The generated QASM.
 
         .. note::
             This method does not support multi-controlled gates
@@ -1199,15 +1200,30 @@ class Circuit:
 
         return "\n".join(code)
 
+    def to_qasm_file(self, qasm_file: str, extended_compatibility: bool = False):
+        """_summary_
+
+        Args:
+            qasm_file (str): Path to file containing the QASM script.
+            extended_compatibility (bool, optional): if ``True``, unrolls more exotic
+                gates in their decomposition by defining them as custom gates,
+                this increases the compatibility with other frameworks,
+                such as ``qiskit``. Defaults to ``False``.
+        """
+        string = self.to_qasm(extended_compatibility=extended_compatibility)
+        with open(qasm_file, "w") as file:
+            file.writelines(string)
+
     @classmethod
-    def from_qasm(cls, qasm_code, **circuit_kwargs):
+    def from_qasm(cls, qasm_code: str, **circuit_kwargs):
         """Constructs a circuit from QASM code.
 
         Args:
             qasm_code (str): String with the QASM script.
+            circuit_kwargs (dict): Keyword arguments of the circuit.
 
         Returns:
-            A :class:`qibo.models.circuit.Circuit` that contains the gates
+            :class:`qibo.models.circuit.Circuit`: Circuit containing the gates
             specified by the given QASM script.
 
         Example:
@@ -1231,6 +1247,23 @@ class Circuit:
         parser = QASMParser()
         circuit_kwargs.pop("nqubits", None)
         return parser.to_circuit(qasm_code, **circuit_kwargs)
+
+    @classmethod
+    def from_qasm_file(cls, qasm_file: str, **circuit_kwargs):
+        """Constructs a circuit from QASM code in a ``qasm_file``.
+
+        Args:
+            qasm_file (str): Path to file containing the string with the QASM script.
+            circuit_kwargs (dict): Keyword arguments of the circuit.
+
+        Returns:
+            :class:`qibo.models.circuit.Circuit`: Circuit containing the gates
+            specified by the given QASM script.
+        """
+        with open(qasm_file) as file:
+            string = file.read()
+
+        return cls.from_qasm(string, **circuit_kwargs)
 
     def to_qir(self):
         """
--- a/tests/test_models_circuit_qasm.py
+++ b/tests/test_models_circuit_qasm.py
@@ -1,5 +1,7 @@
 """Tests creating abstract Qibo circuits from OpenQASM code."""
 
+import os
+
 import numpy as np
 import pytest
 from openqasm3 import parser
@@ -603,3 +605,21 @@ def logical_meas(qubit[3] d) -> bit {
 """
     with pytest.raises(RuntimeError):
         circuit = Circuit.from_qasm(target)
+
+
+def test_qasm_file(backend):
+    nqubits = 5
+    target = Circuit(nqubits)
+    target.add(gates.RY(qubit, 0.0) for qubit in range(nqubits))
+    target.add(gates.CNOT(qubit, qubit + 1) for qubit in range(0, nqubits - 1, 2))
+    target.add(gates.RY(qubit, 0.0) for qubit in range(nqubits))
+    target.add(gates.CNOT(qubit, qubit + 1) for qubit in range(1, nqubits - 1, 2))
+    target.add(gates.RY(qubit, 0.0) for qubit in range(nqubits))
+
+    target.to_qasm_file("tests/circ.qasm")
+
+    circuit = Circuit.from_qasm_file("tests/circ.qasm")
+
+    backend.assert_circuitclose(circuit, target)
+
+    os.remove("tests/circ.qasm")


src/qibo/backends/abstract.py
```diff
@@ -1,56 +1,118 @@
...
 
-    @abc.abstractmethod
-    def set_dtype(self, dtype: str):  # pragma: no cover
-        """Set data type of arrays created using the backend.
+    def compile(self, func):  # pragma: no cover
+        """Compile the given method.
+
+        Available only for the ``TensorflowBackend`` in ``qiboml``.
+        """
+        return func
```
alecandido: Since this is a quite specific mechanism for TensorFlow, is it worth to keep it in the abstract backend? Shall we just have a custom `TensorFlowBackend.execute_circuit()`, just wrapping the common one?

src/qibo/backends/abstract.py
```diff
@@ -81,291 +143,227 @@ def set_dtype(self, dtype: str):  # pragma: no cover
...
+    ########################################################################################
+    ######## Methods related to data types                                          ########
+    ########################################################################################
 
-        Available only for the tensorflow backend.
-        """
-        raise_error(NotImplementedError)
+    @property
+    def complex64(self):
+        return getattr(self.engine, "complex64")
```
alecandido: Here and below: why `getattr(self.engine, "...")`?

If you have a constant string, `getattr(x, "s")` should be fully equivalent to `x.s`. Isn't it?

src/qibo/backends/abstract.py
```diff
@@ -81,291 +143,227 @@ def set_dtype(self, dtype: str):  # pragma: no cover
...
-        raise_error(NotImplementedError)
+    def arccos(self, array, **kwargs) -> "ndarray":
+        return self.engine.arccos(array, **kwargs)
 
-    @abc.abstractmethod
-    def apply_channel(self, channel, state, nqubits: int):  # pragma: no cover
-        """Apply a channel to state vector."""
-        raise_error(NotImplementedError)
+    def arctan2(self, array_1, array_2, **kwargs) -> "ndarray":
+        return self.engine.arctan2(array_1, array_2, **kwargs)
```
alecandido: Here and elsewhere, if you want to be more specific about the function signature that can be certainly something positive, since it is passing more information to the linter.

But then I would suggest to also specify the parameters' types:
```py
import numpy.typing as npt
```

```suggestion
    def arctan2(self, array_1: npt.ArrayLike, array_2: npt.ArrayLike, **kwargs) -> npt.ArrayLike:
        return self.engine.arctan2(array_1, array_2, **kwargs)
```

I'm not completely sure that `npt.ArrayLike` is the most suitable type. But it should be sufficient to encompass NumPy arrays, but also those from the other frameworks. Though I'm not completely sure, and we should double-check.

---

Alternatively, if you do not care, for the time being, about signatures, you could even compress the definition even more, with tuple unpacking
```suggestion
    def arctan2(self, *args, **kwargs) -> "ndarray":
        return self.engine.arctan2(*args, **kwargs)
```

---

All in all, even the present trade-off is a reasonable compromise. But let's assess the alternatives as well.

src/qibo/backends/clifford.py
```diff
@@ -21,81 +21,85 @@ class CliffordBackend(NumpyBackend):
...
 
-        Args:
-            x: Object to cast to array.
-            dtype (optional): data type of the array or tensor. If ``None``, defaults
-                to the default data type of the current backend. Defaults to ``None``.
-            copy (bool, optional): If ``True`` a copy of the object is created in memory.
-                Defaults to ``False``.
-        """
-        return self.engine.cast(x, dtype=dtype, copy=copy)
+    #     return self.engine.asarray(x, dtype=dtype, copy=copy if copy else None)
```
alecandido: Is this dead code left here intentionally?

src/qibo/backends/numpy.py
```diff
@@ -83,811 +46,9 @@ def cast(self, x, dtype=None, copy: bool = False):
...
-    # cast needed because of different dtypes in `torch`
-    S = backend.cast(S)
-    S_inv = engine.where(engine.abs(S) < precision_singularity, 0.0, S**power)
+        return self.engine.asarray(x, dtype=dtype, copy=copy if copy else None)
 
-    return engine.linalg.inv(Vh) @ backend.np.diag(S_inv) @ engine.linalg.inv(U)
+    def to_numpy(self, array):
+        if self.is_sparse(array):
+            return array.toarray()
+        return array
```
alecandido: Up to personal taste, but in these cases I favor the ternary `if`, which is not breaking the flow, but just defines a conditional value.

```suggestion
        return array.toarray() if self.is_sparse(array) else array
```

src/qibo/backends/_hamming_weight_operations.py
```diff
@@ -125,7 +124,7 @@ def _gray_code(self, initial_string):
 
     strings = _ehrlich_algorithm(initial_string, return_indices=False)
     strings = [list(string) for string in strings]
-    strings = self.np.asarray(strings, dtype=int)
+    strings = self.engine.asarray(strings, dtype=int)
```
BrunoLiegiBastonLiegi: I don't remember whether the hamming weight backend was supposed to work with other engines than numpy, if it is then maybe instead of directly calling the `self.engine.asarray` or `self.engine.array` we should either use the `self.cast` or define some other abstract native array creation function to be specialized by each backend.

src/qibo/backends/_hamming_weight_operations.py
```diff
@@ -524,18 +523,18 @@ def _apply_gate_n_qubit(self, gate, state, nqubits, weight):
...
 
-    strings_array = self.np.array([list(s) for s in strings])
+    strings_array = self.engine.array([list(s) for s in strings])
     reordered_strings_array = strings_array[:, map_]
 
-    reordered_strings = self.np.array(["".join(s) for s in reordered_strings_array])
+    reordered_strings = self.engine.array(["".join(s) for s in reordered_strings_array])
     reordered_indexes = [
-        self.np.where(strings == new_string_i)[0][0]
+        self.engine.where(strings == new_string_i)[0][0]
```
BrunoLiegiBastonLiegi: also this one, I'd avoid as much as possible direct calls to the engine and instead wrap them inside an abstract `self.where`, which by default will just call self.engine.where, but that allows for easier backend specialization.

src/qibo/backends/abstract.py
```diff
@@ -1,56 +1,124 @@
...
+
+
+class Backend:
     def __init__(self):
         super().__init__()
-        self.name = "backend"
-        self.platform = None
 
+        self.device = "/CPU:0"
         self.dtype = "complex128"
```
BrunoLiegiBastonLiegi, line 24: this is not something that is strictly related to this PR but is something that I'd liked to have for a while, namely, I think the backend should have two main dtypes one for the complex values and one for the real ones, which depend on the precision one wants to use:
```python
def dtype(self, dtype_str: str):
    return getattr(self.engine, dtype_str)

@property
def real_dtype(self):
    if self.precision == "double":
        return self.dtype("float64")
    return self.dtype("float32")

@property
def complex_dtype(self):
    if self.precision == "double":
        return self.dtype("complex128")
    return self.dtype("complex64")
```

src/qibo/backends/abstract.py
```diff
@@ -1,56 +1,124 @@
...
+
+        Available only for the ``TensorflowBackend`` in ``qiboml``.
+        """
+        return func
+
+    def is_sparse(self, array: ArrayLike) -> bool:
+        """Determine if a given array is a sparse tensor."""
+        from scipy.sparse import issparse  # pylint: disable=import-outside-toplevel
+
+        return issparse(array)
```
BrunoLiegiBastonLiegi: This as well we could leave abstract, because different backends may have their internal sparse module and not be compatible with scipy, e.g. torch
```suggestion
    @abstractmethod
    def is_sparse(self, array: ArrayLike) -> bool:
        """Determine if a given array is a sparse tensor."""
        pass
```

or if we wish to maintain a default one I'd try getting it from the engine:
```python
    def is_sparse(self, array: ArrayLike) -> bool:
        """Determine if a given array is a sparse tensor."""
        return self.engine.sparse.issparse(array)
```

src/qibo/backends/abstract.py
```diff
@@ -81,291 +149,246 @@ def set_dtype(self, dtype: str):  # pragma: no cover
...
-            device (str): Device such as '/CPU:0', '/GPU:0', etc.
-        """
-        raise_error(NotImplementedError)
+        if dtype != self.dtype:
+            self.dtype = dtype
 
-    @abc.abstractmethod
-    def set_threads(self, nthreads: int):  # pragma: no cover
-        """Set number of threads for CPU simulation.
+            if self.matrices:
```
BrunoLiegiBastonLiegi:
```suggestion
            if self.matrices is not None:
```
for better clarity

src/qibo/backends/abstract.py
```diff
@@ -81,291 +149,246 @@ def set_dtype(self, dtype: str):  # pragma: no cover
...
+    ########################################################################################
+    ######## Methods related to data types                                          ########
+    ########################################################################################
 
-        Available only for the tensorflow backend.
-        """
-        raise_error(NotImplementedError)
+    @property
+    def complex64(self) -> DTypeLike:
+        return self.engine.complex64
```
BrunoLiegiBastonLiegi, line 193: I ma not sure whether it is worth having all these dtypes separated instead of just having a single dtype getter function like the one I suggested above

src/qibo/backends/abstract.py
```diff
@@ -81,291 +149,246 @@ def set_dtype(self, dtype: str):  # pragma: no cover
...
-        raise_error(NotImplementedError)
+    def block(self, arrays: ArrayLike) -> ArrayLike:
+        return self.engine.block(arrays)
 
-    @abc.abstractmethod
-    def reset_error_density_matrix(self, gate, state, nqubits: int):  # pragma: no cover
-        """Apply reset error to density matrix."""
-        raise_error(NotImplementedError)
+    def block_diag(self, *arrays: ArrayLike) -> ArrayLike:
+        from scipy.linalg import block_diag  # pylint: disable=import-outside-toplevel
```
BrunoLiegiBastonLiegi: also here, I am not sure whether it is better to default to scipy, which will be often (but not always) compatible. On the one hand, this lifts the burden of identically reimplementing it for each compatible backend, but on the other hand hides possible subpar implementations for backends that are compatible but provide their own implementation. E.g. with cupy you may not see any errors with standard scipy, but they provide their own `block_diag` ([https://docs.cupy.dev/en/stable/reference/generated/cupyx.scipy.linalg.block_diag.html](https://docs.cupy.dev/en/stable/reference/generated/cupyx.scipy.linalg.block_diag.html)).

src/qibo/backends/abstract.py
```diff
@@ -375,80 +398,335 @@ def calculate_matrix_norm(
...
+
+    def mean(self, array: ArrayLike, **kwargs) -> Union[float, complex, ArrayLike]:
+        return self.engine.mean(array, **kwargs)
+
+    def nonzero(self, array: ArrayLike) -> ArrayLike:
+        return self.engine.nonzero(array)
+
+    def ones(self, shape: Union[int, Tuple[int, ...]], dtype=None) -> ArrayLike:
+        if dtype is None:
+            dtype = self.dtype
```
BrunoLiegiBastonLiegi, line 432: it's maybe worth to define a mechanism for retrieving the default dtype, similarly to what we do with the backends:
```python
    def _dtype(self, dtype = None):
        if dtype is None:
            return self.dtype
        return dtype
```

src/qibo/backends/abstract.py
```diff
@@ -375,80 +398,335 @@ def calculate_matrix_norm(
...
+        **kwargs,
+    ) -> ArrayLike:
+        dtype = kwargs.get("dtype", self.float64)
+
+        if size is None:
+            size = 1
+
+        if seed is not None:
+            local_state = self.default_rng(seed) if isinstance(seed, int) else seed
+            result = local_state.choice(array, size=size, replace=replace, p=p)
```
BrunoLiegiBastonLiegi: This also may be risky, as the support of the rng state among the different frameworks is still somewhat heterogeneous, I wonder whether it would be a viable option to define a function that selects which method to call, either the local state one (if available) or the engine.random one + seed reset:

```python
def _random_generator(self, fname: str, seed) -> Callable:
    try:
        state = self.default_rng(seed)
        f = getattr(state, fname)
    except ...:
        self.set_seed(seed)
        # mapping from rng names to standard ones
        fname = self._rng_to_engine_names[fname] 
        f = getattr(self.engine.random, fname)
    return f

def random_choice(self, ..., seed):
    ...
    f = self._random_generator("integers", seed)
    return f(...)
```

the try except may be a little risky here, but can replaced with an if/else control flow in case, I used it just to convey the underline idea

src/qibo/backends/abstract.py
```diff
@@ -458,42 +736,915 @@ def calculate_matrix_power(
...
+            # negative powers of singular matrices via SVD
+            determinant = self.det(matrix)
+            if abs(determinant) < precision_singularity:
+                return self._negative_power_singular_matrix(
+                    matrix, power, precision_singularity, dtype=dtype
+                )
+
+        from scipy.linalg import (  # pylint: disable=import-outside-toplevel
+            fractional_matrix_power,
+        )
```
BrunoLiegiBastonLiegi: I don't remeber why can't we import scipy top level? Because it's not a dependency of qibo? Maybe it should be then...

src/qibo/backends/abstract.py
```diff
@@ -458,42 +736,915 @@ def calculate_matrix_power(
...
+
+        matrix = gate.matrix(self)
+        matrix = self.reshape(matrix, 2 * len(gate.qubits) * (2,))
+
+        if density_matrix:
+            matrix_conj = self.conj(matrix)
+            left, right = einsum_utils.apply_gate_density_matrix_string(
+                gate.qubits, nqubits
+            )
+            state = self.engine.einsum(right, state, matrix_conj)
```
BrunoLiegiBastonLiegi: I think it maybe better to also have a wrapper to the engine einsum:
```python
def einsum(self, ...):
    return self.engine.einsum(...)
```
because there are some slight differences among the frameworks sometimes, e.g. tensorflow does not support the use of integer indices instead of letters

renatomello: There is one already, changing it here just slipped through the cracks I guess

src/qibo/backends/abstract.py
renatomello: `self.precision` was removed some time ago when I introduced `float32` and other dtypes. I think overall, the way it is going to be now is sufficient.

src/qibo/backends/_hamming_weight_operations.py
renatomello: The HWB does work with backends other than `numpy`, including `qiboml` backends.

renatomello:
> Thanks @renatomello for this gigantic work. I had a look at the new abstract backend and it looks great to me. The only two points that slightly concern me are:
> 
>     * defaulting to scipy
> 
>     * the use of the local state
> 
> 
> I will go over the rest asap

A bunch of the "numpy" functionalities are actually `scipy` ones, and I'm not sure there is any way around that. For instance, `cupy` shows its compatibilities with both `numpy` and `scipy` [here](https://docs.cupy.dev/en/stable/reference/comparison.html).

src/qibo/backends/abstract.py
BrunoLiegiBastonLiegi: The problem is how do you know which precision you're using at a given moment? Let's say that you're calculating an expectation value and you need to cast it before returning it, to what dtype will you cast it? float32 or float64? With my suggested pipeline you can just use the self.real_dtype.

renatomello: > The problem is how do you know which precision you're using at a given moment? Let's say that you're calculating an expectation value and you need to cast it before returning it, to what dtype will you cast it? float32 or float64? With my suggested pipeline you can just use the self.real_dtype.

As of now, `double` precision is implied by the fact that `complex128` is the default for the backend. It would be inconsistent to have, by default, double precision for complex numbers but single precision for real ones. I do not see a source of confusion here. Maybe we could add a note in `set_dtype` saying that explicitly, then.

BrunoLiegiBastonLiegi: yeah so basically the self.precision could just be a property
```python
@property
def precision(self):
    if self dtype in ("float64", "complex128"):
        return "double"
    return "single"
```
my point is that you need an automatic way to infer the correct dtype

BrunoLiegiBastonLiegi:
> A bunch of the "numpy" functionalities are actually `scipy` ones, and I'm not sure there is any way around that. 

yes indeed, which means they belong in principle to the numpy backend and not generically to the abstract one, even though as I said, I agree it may be convenient to have them already in the abstract one...

renatomello:
@BrunoLiegiBastonLiegi coverage is passing now. I'll pretend I'm dead until this misery is over.

src/qibo/backends/_hamming_weight_operations.py
```diff
@@ -502,17 +522,19 @@ def _apply_gate_n_qubit(self, gate, state, nqubits, weight):
...
 
-    matrix = self.np.zeros((dim, dim))
+    matrix = self.zeros((dim, dim))
     matrix = self.cast(matrix, dtype=self.dtype)
     mod_condition = indexes[:, None] % 2 ** (
         nqubits - gate_qubits - ncontrols
     ) == indexes[None, :] % 2 ** (nqubits - gate_qubits - ncontrols)
 
-    control_substrings = self.np.array(
+    control_substrings = self.engine.array(
```
BrunoLiegiBastonLiegi: `self.cast` maybe safer?

src/qibo/backends/abstract.py
```diff
@@ -1,62 +1,124 @@
...
-from qibo.config import raise_error
+from qibo import __version__
+from qibo.backends import einsum_utils
+from qibo.config import SHOT_BATCH_SIZE, log, raise_error
+from qibo.gates.abstract import Gate
+from qibo.result import CircuitResult, MeasurementOutcomes, QuantumState
 
 
-class Backend(abc.ABC):
+class Backend:
```
BrunoLiegiBastonLiegi: Any particular reason to not have this abstract anymore?

src/qibo/backends/abstract.py
```diff
@@ -1,62 +1,124 @@
...
-        """Return the qubit names of the backend. If :class:`SimulationBackend`, return None."""
-        raise_error(NotImplementedError)
+    def qubits(self) -> Optional[List[Union[int, str]]]:  # pragma: no cover
+        """Return the qubit names of the backend.
+
+        Returns:
+            List[int] or List[str] or None: For hardware backends, return list of qubit names.
+            For simulation backends, return ``None``.
+        """
+        return None
```
BrunoLiegiBastonLiegi, line 65: raising an error would be more explicit maybe

src/qibo/backends/abstract.py
BrunoLiegiBastonLiegi: It's true that they are not many, but I think this would be helpful. You could do something like this:

```python
DTYPES = ["float32", "float64", ...]

# Abstract init
def __init__(self):
    ...
    for dtype in DTYPES:
        setattr(self, dtype, self.get_dtype(dtype))

def get_dtype(self, dtype: str):
    return getattr(self.engine, dtype)
```

and having a dynamical method to retrieve the `dtype` maybe useful in general.

src/qibo/backends/abstract.py
BrunoLiegiBastonLiegi: this?

src/qibo/backends/abstract.py
```diff
@@ -87,219 +149,1407 @@ def set_dtype(self, dtype: str):  # pragma: no cover
...
+        matrix = gate.matrix(self)
+        matrix = self.reshape(matrix, 2 * len(gate.qubits) * (2,))
+
+        if density_matrix:
+            matrix_conj = self.conj(matrix)
+            left, right = einsum_utils.apply_gate_density_matrix_string(
+                gate.qubits, nqubits
+            )
+            state = self.engine.einsum(right, state, matrix_conj)
+            state = self.engine.einsum(left, state, matrix)
```
BrunoLiegiBastonLiegi:
```suggestion
            state = self.einsum(right, state, matrix_conj)
            state = self.einsum(left, state, matrix)
```

src/qibo/backends/abstract.py
```diff
@@ -87,219 +149,1407 @@ def set_dtype(self, dtype: str):  # pragma: no cover
...
+        else:
+            opstring = einsum_utils.apply_gate_string(gate.qubits, nqubits)
+            state = self.engine.einsum(opstring, state, matrix)
```
BrunoLiegiBastonLiegi:
```suggestion
            state = self.einsum(opstring, state, matrix)
```

src/qibo/backends/abstract.py
```diff
@@ -87,219 +149,1407 @@ def set_dtype(self, dtype: str):  # pragma: no cover
...
+                "gates.",
+            )
+
+        state = self.cast(state, dtype=state.dtype)  # pylint: disable=E1111
+        state = self.reshape(state, 2 * nqubits * (2,))
+        matrix = gate.matrix(self)
+
+        matrix = self.reshape(matrix, 2 * len(gate.qubits) * (2,))
+        left, _ = einsum_utils.apply_gate_density_matrix_string(gate.qubits, nqubits)
+        state = self.engine.einsum(left, state, matrix)
```
BrunoLiegiBastonLiegi:
```suggestion
        state = self.einsum(left, state, matrix)
```

src/qibo/backends/clifford.py
```diff
@@ -1,105 +1,106 @@
...
             )
 
-        self.np = self.engine.np
+        self.cast = self._platform.cast
+        self.csr_matrix = self._platform.csr_matrix
+        self._identity_sparse = self._platform._identity_sparse
 
-        self.name = "clifford"
+        self.matrices = NumpyMatrices(self.dtype)
+        self.tensor_types = (self.engine.ndarray,)
```
BrunoLiegiBastonLiegi, line 97: All this is not ideal, the best solution in my opinion would be to do something similar to what I did with quimb (https://github.com/qiboteam/qibotn/blob/2c49c2b38c48d48f3610f5c9ef950082960385df/src/qibotn/backends/quimb.py#L351C1-L371C1), i.e. make the clifford backend dinamically inherit from the correct backend. But this I would leave for a future pull request.

src/qibo/tomography/gate_set_tomography.py
```diff
@@ -73,7 +73,7 @@ def _observables(nqubits: int) -> List:
 
 
 @cache
-def _get_observable(j: int, nqubits: int):
+def _get_observable(j: int, nqubits: int, backend):
```
BrunoLiegiBastonLiegi: we could keep the caching here, potentially, by just passing the backend name and platform string rather than the backend instance, e.g:
```python
@cache
def _get_observable(j: int, nqubits: int, backend: str):
    backend_args = backend.split("-")
    backend = construct_backend(*backend_args)
    ...
```

src/qibo/backends/abstract.py
renatomello: None of the methods are abstract anymore, so I didn't see any reason for the class to be. IIRC, leaving it there was also creating a pylint warning.

BrunoLiegiBastonLiegi: I guess it depends on the perspective, but all the methods that are currently returning a `NotImplementedError` or similar are good candidates for abstract methods in my view. I would also argue that being able to instantiate a `Backend()` instance, does not make much sense, as you can basically do nothing with it in facts.

renatomello:
@BrunoLiegiBastonLiegi TBH I prefer to leave the dtype attributes explicitly set instead of in a loop. I think it makes it easier for other people to read the code.

Also, brought back the abstract class as you asked.

EDIT: undid the abstract class thing. It creates way more issues and solves almost none. Every single backend would have to overwrite every single abstract method, even the ones that are supposed to be not implemented like in the abstract Backend class. Not worth it.

renatomello:
@scarrazza @BrunoLiegiBastonLiegi tests on GPU seem to not be running properly due to issues with installation of NVIDIA drivers.

alecandido:
🎉

PR #1725, opened 2025-10-16
Updating publication list

scarrazza:
As the title says. @MatteoRobbiati @andrea-pasquale, please let me know if you have further entries. After your review, we can send a message to the mailing list for confirmation.

### DIFF
--- a/doc/source/appendix/citing-qibo.rst
+++ b/doc/source/appendix/citing-qibo.rst
@@ -4,7 +4,7 @@ Publications
 If Qibo has been significant in your research, and you would like to acknowledge
 the project in your academic publication, we suggest citing the following documents:
 
-Collaboration papers
+Collaboration Papers
 --------------------
 
 * S. Efthymiou, S. Ramos-Calderer, C. Bravo-Prieto, A. Pérez-Salinas, D.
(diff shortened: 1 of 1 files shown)


andrea-pasquale (approved):
Thanks @scarrazza, nothing is missing from my side.
Eventually my thesis could be added as well, but perhaps it is better suited in a new section.

scarrazza:
Good point, could you please add it?

scarrazza:
> shall I add [this paper](https://iopscience.iop.org/article/10.1088/1361-6404/adcd13) as well?

yes, fine by me.

> I would probably add an extra title pre-prints, since atm Qiboml and Qibocal are listed among the peer-reviewed articles.

I would suggest to replace peer-reviewed with reference or collaboration papers.

doc/source/appendix/citing-qibo.rst
```diff
@@ -133,6 +133,16 @@ Peer-Reviewed Articles
...
 
+
+PhD theses based on Qibo
+------------------------
+
+
+* A. Pasquale, *Open-source middleware for quantum computing*
+  (2024). `https://tesidottorato.depositolegale.it/handle/20.500.14242/184252`_
+
+.. _`https://tesidottorato.depositolegale.it/handle/20.500.14242/184252`: https://tesidottorato.depositolegale.it/handle/20.500.14242/184252
```
scarrazza, line 144: Umm, maybe we should just upload the latest version and quote the arXiv entry?

andrea-pasquale: As you wish, the thesis is not on arXiv yet. To do that I will need to upload it.

scarrazza: Please do so, it is usually a good idea to keep in arXiv so people can cite.

MatteoRobbiati (approved):
Done! Please, have a second look, then I think we can merge.

scarrazza:
Looks good to me, thanks.

PR #1735, opened 2025-10-23
Fixing u1<-> p equivalence in qasm

andre-juan:
When one tries to creat a qibo circuit from a qasm string having the `p` gate (which is equivalent to `u1`), an error occurs, since qibo doesn't have a `p` gate (only `u1`). This happens, for instance, if one creates a qasm string from a qiskit circuit, since there `u1` was deprecated in favor to `p`. The minimal example below shows this error.

```python
from qiskit import QuantumCircuit
from qiskit.qasm3 import dumps
from qibo import Circuit

qc = QuantumCircuit(1)
qc.p(0.1, 0)

Circuit.from_qasm(dumps(qc))
```

This PR fixes this bug, by including the equivalence of qasm `p` to the gate `U1`.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/tests/test_models_circuit_qasm.py
+++ b/tests/test_models_circuit_qasm.py
@@ -322,7 +322,7 @@ u2(0.2,0.6) q[1];
 cu3(0.3,0.4,0.5) q[0],q[1];
 p(0.1) q[0];"""
     c = Circuit.from_qasm(target)
-    assert c.depth == 2
+    assert c.depth == 3
     assert isinstance(c.queue[0], gates.U1)
     assert isinstance(c.queue[1], gates.U2)
     assert isinstance(c.queue[2], gates.CU3)


tests/test_models_circuit_qasm.py
```diff
@@ -319,15 +319,18 @@ def test_from_qasm_ugates():
 qreg q[2];
 u1(0.1) q[0];
 u2(0.2,0.6) q[1];
-cu3(0.3,0.4,0.5) q[0],q[1];"""
+cu3(0.3,0.4,0.5) q[0],q[1];
+p(0.1) q[0];"""
     c = Circuit.from_qasm(target)
     assert c.depth == 2
```
renatomello: One more gate was added.

```suggestion
    assert c.depth == 3
```

PR #1747, opened 2025-11-14
Qiskit transpiler wrapper

BrunoLiegiBastonLiegi:
This implements a wrapper to qiskit's `PassManager` transpiler, that allows to use qiskit transpilation pipelines alongside qibo's ones. In detail, everything goes through a back and forth qasm conversion. For that reason, I also implemented an `extended_compatibility` mode in our `circuit.to_qasm` to better integrate with the qasm interface of other frameworks, e.g. qiskit in this case. Namely, for some gates now we provide their explicit definition as custom gate in the qasm code. I did this for `GPI` and `GPI2` for now, but we should probably do that for all the more exotic ones.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/doc/source/code-examples/advancedexamples.rst
+++ b/doc/source/code-examples/advancedexamples.rst
@@ -2324,9 +2324,10 @@ pass it to the :class:`qibo.transpiler.qiskit.QiskitPasses` object:
    # some qiskit passes
    qiskit_passes = [passes.Optimize1qGates(), passes.InverseCancellation()]
    qiskit_passes = QiskitPasses(PassManager(qiskit_passes))
+   passes = qibo_passes + [qiskit_passes,]
    # join everythin together
    hybrid_transpiler = Passes(
-        passes=qibo_passes + qiskit_passes,
+        passes=passes,
         native_gates=NativeGates.default(),
     )
 


pyproject.toml
```diff
@@ -75,6 +75,7 @@ qibojit = { git = "https://github.com/qiboteam/qibojit.git" }
 # qiboml = { git  = "https://github.com/qiboteam/qiboml.git" }
 stim = "^1.12.0"
 qulacs = { version = "^0.6.4", python = "<3.13" }
+qiskit = "^2.2.3"
```
alecandido, line 78: Shall we make this optional?

BrunoLiegiBastonLiegi: I am testing the qiskit transpiler in the tests, thus I think we need it as a mandatory test dependency.

alecandido: > it as a mandatory **test** dependency.

I did not notice it was in the *test* group, my bad.
Once it is already in a group, there is not even the concept of being "optional" (which is related to extras, and they are kind of *alternative* to groups).

No, that's fine. The point is that we will have an undeclared runtime dependency on Qiskit, if the transpiler is used.

But, to be fair, adding official dependencies is making the locking operation more and more complex. Since this is needed only in a very special use case, it is also fine to disregard it as a package dependency itself (if you are outsourcing transpilation to Qiskit, it is also not that hard to understand if you get an error about a missing `qiskit` dependency).

Then, fine as it is

PR #1749, opened 2025-11-24
Pass backend explicitly to avoid relying on `QIBO_PLATFORM`

andrea-pasquale:
I've noticed that simple circuits executed on hardware like the following 

```py
import numpy as np
from qibo.backends import construct_backend


backend = construct_backend("qibolab", platform="qw21q-b")
platform = backend.platform

circuit = Circuit(2, wire_names=["B1", "B2"])
circuit.add(gates.GPI2(0, phi=np.pi/2))
circuit.add(gates.GPI2(1, phi=np.pi/2))
circuit.add(gates.CZ(0, 1))
circuit.add(gates.GPI2(1, phi=np.pi/2))
circuit.add(gates.M(0,1))

results = backend.execute_circuit(
    circuit=circuit,
)
print(results.frequencies())
```
are failing on the cluster with the error
```sh
FileNotFoundError: [Errno 2] No such file or directory: '/home/users/andrea.pasquale/qibolab_platforms_qrc/qw11q/platform.py'
```
After some investigation I believe that the error is related to the fact that at some point the `_GlobalBackend` is being used which on the cluster defaults to `qw11q` on certain nodes through the `QIBO_PLATFORM` env variable.
In this PR I am fixing this issue by passing directly the backend object to avoid using the `_GlobalBackend`. Probably a better fix would be to revisit how we deal with the `QIBO_PLATFORM` env variable.

### DIFF
--- a/src/qibo/result.py
+++ b/src/qibo/result.py
@@ -375,7 +375,7 @@ class MeasurementOutcomes:
 
         if registers:
             return {
-                gate.register_name: gate.result.samples(binary, backend=backend)
+                gate.register_name: gate.result.samples(binary, backend=self.backend)
                 for gate in self.measurements
             }
 


scarrazza:
Thanks @andrea-pasquale for reporting this, in fact the prolog was outdated, now it should work as expected. Anyway your fix looks useful.

andrea-pasquale:
> Thanks @andrea-pasquale for reporting this, in fact the prolog was outdated, now it should work as expected. Anyway your fix looks useful.

Thanks @scarrazza I confirm that now it is working fine.

renatomello:
@andrea-pasquale @scarrazza Just FYI: the same fix is already being implemented in #1711.

PR #1755, opened 2025-12-04
Add default `wire_names` if some wires are not in the connectivity graph

scarrazza:
As the title says, required for platforms where `wire_names` are not integers but characters + integers instead.


alecandido:
Ah, actually tests are failing because they relied on the error to be there. We just need to replace that test.

If we're in a hurry, we could even just drop it.

alecandido:
Sorry for the noise: my page was not updated, and it did not show the merge when I clicked...

PR #1774, opened 2026-02-05
Add `_base_decompose` method to missing gate classes

renatomello:
This was missing so, for instance, `RXX.decompose` could work. Same for `RYY` and `RZZ`.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/gates/gates.py
+++ b/src/qibo/gates/gates.py
@@ -292,7 +292,7 @@ class SX(Gate):
     def qasm_label(self):
         return "sx"
 
-    def _base_decompose(self, *free, use_toffolis=True, **kwargs):
+    def decompose(self, *free, use_toffolis=True):
         """Decomposition of :math:`\\sqrt{X}` up to global phase.
 
         A global phase difference exists between the definitions of
@@ -300,11 +300,7 @@ class SX(Gate):
         being the :class:`qibo.gates.RX` gate. More precisely,
         :math:`\\sqrt{X} = e^{i \\pi / 4} \\, \\text{RX}(\\pi / 2)`.
         """
-        from qibo.transpiler.decompositions import (  # pylint: disable=C0415
-            standard_decompositions,
-        )
-
-        return standard_decompositions(self)
+        return super().decompose(*free, use_toffolis=use_toffolis)
 
     def _dagger(self):
         """"""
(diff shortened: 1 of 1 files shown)


src/qibo/gates/gates.py
```diff
@@ -1406,6 +1396,15 @@ def _dagger(self) -> "Gate":
...
         return self.__class__(q0, q1, theta)  # pylint: disable=E1130
 
+    def _base_decompose(self, *free, use_toffolis=True, **kwargs):
+        """Decompositions might have a global phase difference with respect to
+        the original gate."""
+        from qibo.transpiler.decompositions import (  # pylint: disable=C0415
+            standard_decompositions,
+        )
+
+        return standard_decompositions(self)
```
BrunoLiegiBastonLiegi: Can't we just make this standard and add it to ``Gate`` itself?
```python
    def _base_decompose(self, *free, use_toffolis=True, **kwargs):
        """Decompositions might have a global phase difference with respect to
        the original gate."""
        from qibo.transpiler.decompositions import (  # pylint: disable=C0415
            standard_decompositions,
        )
        try:
            return standard_decompositions(self)
        except KeyError: # or whatever error is raised
            return [self.__class__(*self.init_args, **self.init_kwargs)]
```

BrunoLiegiBastonLiegi (commented):
All good, just a curiosity.

src/qibo/gates/gates.py
renatomello: Currently, no error is raised if there is no decomposition. It returns the gate itself. In fact, that is the current default method in the abstract `Gate` class. We would just be swiping one overwrite for another. Instead of overwriting to include the decomposition, we'd be overwriting to return the original gate.

BrunoLiegiBastonLiegi: yeah sorry I edited the suggestion after I checked, I'd say we could try first to get the decomposition and otherwise return the gate itself

renatomello: Can we do this in a separate PR so we at least get these decompositions in the next release?

BrunoLiegiBastonLiegi: done in #1781, can you quickly double check it?

BrunoLiegiBastonLiegi (approved):

PR #1789, opened 2026-02-25
Fix unwanted error logged when circuit loaded from `json` file

### LINKED ISSUE #1519: Unwanted error logged when loading a circuit from json
The following script
```py
import json

c = Circuit(2)
c.add(gates.CZ(0, 1))

with open("test.json", "w") as file:
    json.dump(c.raw, file)

with open("test.json", "r") as file:
    cd = json.load(file)
    cl = Circuit(2)
    cl = c.from_dict(cd)
```
produces this output
```py
[Qibo 0.2.14|ERROR|2024-11-08 14:05:50]: Cannot use `controlled_by` method on gate <qibo.gates.gates.CZ object at 0x7f13cd1a1a60> because it is already controlled by (0,).
```
It does not fail, and the loaded circuit `cl` is usable, so I would not consider it very urgent, however it is weird that the error is printed. 
This is most likely due to some `try`-`except` in the loader.

renatomello:
Closes #1519 

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/gates/abstract.py
+++ b/src/qibo/gates/abstract.py
@@ -155,7 +155,7 @@ class Gate:
         if gate.name not in GATES_CONTROLLED_BY_DEFAULT:
             try:
                 return gate.controlled_by(*raw["_control_qubits"])
-            except RuntimeError as e:
+            except RuntimeError as e:  # pragma: no cover
                 if "controlled" in e.args[0]:
                     return gate
                 raise e


alecandido:
@stavros11 the main difference is that this solution is introducing "more code", which is not much about the amount of lines/characters, but rather the logical complexity: here you have a special case and a special case in the special case. While the other proposed solution was decoupling the problem leveraging the already existing logic.

In any case, that's just the general attitude, and a solution is always better than no solution. We are always on time to simplify the code at a later stage.
Of course, in general, it is better to make the logic clean and simple as soon as possible - to avoid expensive rewritings. But principles are principles - and this is something small. Personally, I'm more than fine just merging this :)

PR #1798, opened 2026-03-06
Enabling `random_clifford` for more than 255 qubits

renatomello:
Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/quantum_info/random_ensembles.py
+++ b/src/qibo/quantum_info/random_ensembles.py
@@ -540,20 +540,19 @@ def random_clifford(
 
     backend = _check_backend(backend)
     backend.set_seed(seed)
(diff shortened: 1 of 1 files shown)


src/qibo/quantum_info/random_ensembles.py
```diff
@@ -540,20 +540,19 @@ def random_clifford(
 
     backend = _check_backend(backend)
     backend.set_seed(seed)
+    dtype = backend.uint8 if nqubits <= 255 else backend.int16
```
BrunoLiegiBastonLiegi, line 543:
```suggestion
    dtype = backend.uint8 if nqubits <= 255 else backend.uint16
```
do we need the sign?

PR #1801, opened 2026-03-11
Remove mention to `torch` from `CliffordBackend` + typing + hide protected attributes from `Clifford.__repr__`

renatomello:
Per title.

Checklist:
- [x] Tests are passing.


renatomello:
@scarrazza this has been approved for a while but it is blocked from being merged?

Issue #1802, opened 2026-03-11
Controlled-SWAP not supported when converting from QASM


GabrielePalazzo:
**Describe the bug**
If a qasm circuit contains a CSWAP gate, `qibo` raises a `TypeError`, when executing `Circuit.from_qasm()`

**To Reproduce**

```python
from qibo import Circuit

qasm_string = """// Code generated by NVIDIA's nvq++ compiler
OPENQASM 2.0;

include "qelib1.inc";

qreg var0[3];
x var0[0];
x var0[1];
cswap var0[0], var0[1], var0[2];
creg var4[1];
measure var0[0] -> var4[0];
creg var5[1];
measure var0[1] -> var5[0];
creg var6[1];
measure var0[2] -> var6[0];"""

Circuit.from_qasm(qasm_string)
```

PR #1803, opened 2026-03-11
Add CSWAP gate

GabrielePalazzo:
Checklist:
- [x] Tests are passing.

This draft aims to address #

In this PR I created a `CSWAP` gate.
While this automatically fixes the issue #1802, it feels just a copy-paste of the code in `gates.py`.
In alternative, the qasm parser can detect controlled gates and instantiate them.

### DIFF
--- a/src/qibo/models/_openqasm.py
+++ b/src/qibo/models/_openqasm.py
@@ -239,7 +239,7 @@ class QASMParser:
                 pass
             init_args.append(arg)
         # check whether the gate exists in qibo.gates already
-        if _qibo_gate_name(gate.name.name) in dir(qibo.gates):
+        if hasattr(qibo.gates, _qibo_gate_name(gate.name.name)):
             try:
                 gate = getattr(qibo.gates, _qibo_gate_name(gate.name.name))(
                     *qubits, *init_args
@@ -265,7 +265,7 @@ class QASMParser:
             stripped_gate_name = gate.name.name.lstrip("c")
             ncontrols = len(gate.name.name) - len(stripped_gate_name)
             if ncontrols > 0:
-                if _qibo_gate_name(stripped_gate_name) in dir(qibo.gates):
+                if hasattr(qibo.gates, _qibo_gate_name(stripped_gate_name)):
                     control_qubits = qubits[:ncontrols]
                     remaining_qubits = qubits[ncontrols:]
                     try:


renatomello:
I would suggest avoiding creating yet another gate that is controlled by default. This can be avoided by using the `controlled_by` method of the `Gate` class. 

I am suggesting this because all gates that are controlled by default 
https://github.com/qiboteam/qibo/blob/91fb509160319f50f33d38befd38816dfebda4de/src/qibo/gates/abstract.py#L35
already generate secondary issues as is, e.g. Transpilation issues, gate-counting issues. Since these gates would have to be refactored anyway, it would be better not to increase the number of gates controlled by default.

There are some rules to deal with edge cases in the `_openqasm.py` file. 
https://github.com/qiboteam/qibo/blob/91fb509160319f50f33d38befd38816dfebda4de/src/qibo/models/_openqasm.py#L115
One rule for the `CSWAP` from qasm strings could be added there to be converted to a qibo `SWAP` with one control, and the import issue would be solved without having to create another gate class.

src/qibo/models/_openqasm.py
```diff
@@ -112,6 +113,21 @@ def _compile_gate_qubits_and_args(self, gate, qubit_map, args_map):
...
+    """Number of control qubits."""
+    base_gate_name: str
+    """QASM name of the controlled gate."""
+
+    @property
+    def is_controlled(self) -> bool:
+        """``True`` if there is at least one control qubit."""
+        return self.num_controls > 0
+
+
```
scarrazza: Please add tests to verify controlled gate parsing logic, e.g.:

```python
def test_from_qasm_controlled_swap_gate():
    """Test parsing of controlled SWAP gate from QASM."""
    target = """OPENQASM 2.0;
    include "stdgates.inc";
    qreg q[4];
    cx q[0],q[1];
    cswap q[1],q[2],q[3];
    """
    circuit = Circuit.from_qasm(target)
    assert circuit.nqubits == 4
    assert circuit.ngates == 4
    assert isinstance(circuit.queue[0], gates.CNOT)
    assert isinstance(circuit.queue[2], gates.SWAP)
    # Verify controls are properly applied
    assert circuit.queue[2].num_controls == 1
    assert circuit.queue[2].control_qubits == (1,)

def test_from_qasm_controlled_x_gate():
    """Test parsing of controlled X gate (cx) from QASM."""
    target = """OPENQASM 2.0;
    include "qelib1.inc";
    qreg q[2];
    cx q[0],q[1];
    """
    circuit = Circuit.from_qasm(target)
    assert circuit.nqubits == 2
    assert circuit.ngates == 1
    assert isinstance(circuit.queue[0], gates.CNOT)

def test_from_qasm_ccx_gate():
    """Test parsing of CCX (Toffoli) gate."""
    target = """OPENQASM 2.0;
    include "qelib1.inc";
    qreg q[3];
    ccx q[0],q[1],q[2];
    """
    circuit = Circuit.from_qasm(target)
    assert isinstance(circuit.queue[0], gates.TOFFOLI)
```

src/qibo/models/_openqasm.py
```diff
@@ -226,6 +242,13 @@ def _get_qubit(self, qubit):
 
         return qubit.name
 
+    @staticmethod
+    def _parse_controlled_gate(gate_name: str):
+        """Checks if the qasm gate is a controlled gate."""
```
scarrazza:
```suggestion
    """Parse a QASM gate name to extract control information.
    Supported patterns:
    - 'cx' → Controlled-X/CNOT
    - 'ccx' → Controlled-controlled-X/Toffoli
    - 'c*' → Generic controlled gate where * is the base gate name
    Args:
        gate_name: The QASM gate name to parse.
    Returns:
        :class:`ControlledGateInfo`: Information about the controlled gate including
        number of controls and base gate name.
    Raises:
        ValueError: If the gate name has a prefix 'c' but the base name is not a valid gate.
    """
```

src/qibo/models/_openqasm.py
```diff
@@ -262,6 +285,22 @@ def _get_gate(self, gate):
...
+                        gate = getattr(
+                            qibo.gates, _qibo_gate_name(parsed_gate.base_gate_name)
+                        )(*remaining_qubits, *init_args).controlled_by(*control_qubits)
+                    # the gate exists in qibo.gates but invalid construction
+                    except TypeError:
+                        raise_error(
+                            ValueError, f"Invalid gate declaration at span: {gate.span}"
+                        )
+                    return gate
+            # undefined gate
```
scarrazza, line 284: This code only handles controlled gates without parameters like cx, cswap, correct? What about controlled-gates with parameters?

scarrazza (changes requested):
Thanks @GabrielePalazzo, please have a look at my comments above.

src/qibo/models/_openqasm.py
```diff
@@ -112,6 +113,21 @@ def _compile_gate_qubits_and_args(self, gate, qubit_map, args_map):
...
+
+    num_controls: int
+    """Number of control qubits."""
+    base_gate_name: str
+    """QASM name of the controlled gate."""
+
+    @property
+    def is_controlled(self) -> bool:
+        """``True`` if there is at least one control qubit."""
+        return self.num_controls > 0
```
renatomello: Maybe I am missing something, but do we need a whole class just to check if `num_controls > 0`?

src/qibo/models/_openqasm.py
```diff
@@ -226,6 +242,13 @@ def _get_qubit(self, qubit):
 
         return qubit.name
 
+    @staticmethod
+    def _parse_controlled_gate(gate_name: str):
+        """Checks if the qasm gate is a controlled gate."""
+        stripped_gate_name = gate_name.lower().lstrip("c")
+        length_difference = len(gate_name) - len(stripped_gate_name)
+        return ControlledGateInfo(length_difference, stripped_gate_name)
```
renatomello: For instance, here, couldn't one just check if `length_difference > 0`?

GabrielePalazzo: Yes, but in addition to the number of control qubits, the dataclass stores the "uncontrolled" gate name too.

src/qibo/models/_openqasm.py
GabrielePalazzo: The idea of a dataclass was to keep both the number of control qubits and the qibo gate name together, but we can ditch it.

src/qibo/models/_openqasm.py
```diff
@@ -262,6 +285,22 @@ def _get_gate(self, gate):
...
+                    try:
+                        gate = getattr(
+                            qibo.gates, _qibo_gate_name(parsed_gate.base_gate_name)
+                        )(*remaining_qubits, *init_args).controlled_by(*control_qubits)
+                    # the gate exists in qibo.gates but invalid construction
+                    except TypeError:
+                        raise_error(
+                            ValueError, f"Invalid gate declaration at span: {gate.span}"
+                        )
+                    return gate
```
renatomello: This would remove the necessity of the `ControlledGateInfo` class as well as the `_parse_controlled_gate` method.

```suggestion
            stripped_gate_name = gate.name.lower().lstrip("c")
            ncontrols = len(gate.name) - len(stripperd_gate_name)
            if ncontrols > 0:
                if _qibo_gate_name(stripped_gate_name) in dir(qibo.gates):
                    control_qubits = qubits[:ncontrols]
                    remaining_qubits = qubits[ncontrols:]
                    try:
                        gate = getattr(qibo.gates, _qibo_gate_name(stripped_gate_name))
                        gate = gate(*remaining_qubits, *init_args).controlled_by(*control_qubits)
                    # the gate exists in qibo.gates but invalid construction
                    except TypeError:
                        raise_error(
                            ValueError, f"Invalid gate declaration at span: {gate.span}"
                        )
                    return gate
```

src/qibo/models/_openqasm.py
GabrielePalazzo: Controlled gates with parameters should work.
Controlled rotation gates are already defined in `gates.py`, but in theory they should work even using this part because we pass all parameters (not only qubits) to the gates.

src/qibo/models/_openqasm.py
```diff
@@ -262,6 +262,24 @@ def _get_gate(self, gate):
...
+                        gate = getattr(qibo.gates, _qibo_gate_name(stripped_gate_name))
+                        gate = gate(*remaining_qubits, *init_args).controlled_by(
+                            *control_qubits
+                        )
+                    # the gate exists in qibo.gates but invalid construction
+                    except TypeError:
+                        raise_error(
+                            ValueError, f"Invalid gate declaration at span: {gate.span}"
+                        )
+                    return gate
```
scarrazza: I would avoid overriding the gate variable:

```suggestion
                        base_gate_cls = getattr(qibo.gates, _qibo_gate_name(stripped_gate_name))
                        gate_instance = base_gate_cls(*remaining_qubits, *init_args).controlled_by(
                            *control_qubits
                        )
                    # the gate exists in qibo.gates but invalid construction
                    except TypeError:
                        raise_error(
                            ValueError, f"Invalid gate declaration at span: {gate.span}"
                        )
                    return gate_instance
```

tests/test_models_circuit_qasm.py
```diff
@@ -623,3 +623,33 @@ def test_qasm_file(backend):
...
+    assert circuit.queue[3].control_qubits == (0, 1)
+
+
+def test_from_qasm_invalid_controlled_gates():
+    target = """OPENQASM 2.0;
+qreg q[4];
+x q[0];
+x q[1];
+cswap q[0],q[1],q[2],q[3];"""
+    with pytest.raises(AttributeError):
```
scarrazza: This test will pass only because of the variable shadowing bug above. Once the bug is fixed, the test will break.
```suggestion
    with pytest.raises(ValueError):
```

tests/test_models_circuit_qasm.py
```diff
@@ -623,3 +623,33 @@ def test_qasm_file(backend):
...
+def test_from_qasm_controlled_gates():
+    target = """OPENQASM 2.0;
+qreg q[3];
+x q[0];
+x q[1];
+cswap q[0],q[1],q[2];
+ccz q[0],q[1],q[2];"""
+    circuit = Circuit.from_qasm(target)
+    assert circuit.depth == 3
+    print(circuit.queue[0])
```
scarrazza:
```suggestion
```

src/qibo/models/_openqasm.py
```diff
@@ -262,6 +262,24 @@ def _get_gate(self, gate):
                 )
         # undefined gate
         else:
+            stripped_gate_name = gate.name.name.lstrip("c")
+            ncontrols = len(gate.name.name) - len(stripped_gate_name)
+            if ncontrols > 0:
+                if _qibo_gate_name(stripped_gate_name) in dir(qibo.gates):
```
scarrazza: Calling dir might be inefficient, please try something like that instead:
```suggestion
                if hasattr(qibo.gates, _qibo_gate_name(stripped_gate_name)):
```

GabrielePalazzo: Shall we do the same at line 242?
https://github.com/qiboteam/qibo/blob/cc031d4d2adbc72c42fc541b43d36416764982d7/src/qibo/models/_openqasm.py#L242

scarrazza: Yes, please.

PR #1809, opened 2026-03-21
Better scaling for `_apply_gate_n_qubit`

AlejandroSopena:
Requires PR [#253](https://github.com/qiboteam/qibojit/pull/253)

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibo/backends/_hamming_weight_operations.py
+++ b/src/qibo/backends/_hamming_weight_operations.py
@@ -1,25 +1,30 @@
 """Module defining the Hamming-weight-preserving backend."""
 
-from typing import List, Union
+# pylint: disable=W0212
 
-import numpy as np
+from typing import Dict, List, Optional, Set, Tuple, Union
+
+from numpy.typing import ArrayLike
 from scipy.special import binom
 
 from qibo import gates
 from qibo.config import raise_error
+from qibo.gates.abstract import Gate
 
 
-def apply_gate(self, gate, state, nqubits: int, weight: int):
+def apply_gate(
+    self, gate: Gate, state: ArrayLike, nqubits: int, weight: int
+) -> ArrayLike:
     """Apply ``gate`` to ``state``.
 
     Args:
         gate (:class:`qibo.gates.abstract.Gate`): gate to apply to ``state``.
-        state (ndarray): state to apply ``gate`` to.
+        state (ArrayLike): state to apply ``gate`` to.
         nqubits (int): total number of qubits in ``state``.
         weight (int): fixed Hamming weight of ``state``.
 
     Returns:
-        ndarray: ``state`` after the action of ``gate``.
+        ArrayLike: ``state`` after the action of ``gate``.
     """
     if isinstance(gate, gates.M):
         return gate.apply_hamming_weight(self, state, nqubits, weight)
@@ -28,7 +33,7 @@ def apply_gate(self, gate, state, nqubits: int, weight: int):
         # CCZ has a custom apply method because currently it is the only
         # 3-qubit gate that is also Hamming-weight-preserving
         # and this custom method is faster than the n-qubit method
-        return self._apply_gate_CCZ(gate, state, nqubits, weight)
+        return self._apply_gate_ccz(gate, state, nqubits, weight)
 
     if len(gate.target_qubits) == 1:
         return self._apply_gate_single_qubit(gate, state, nqubits, weight)
@@ -39,14 +44,20 @@ def apply_gate(self, gate, state, nqubits: int, weight: int):
     return self._apply_gate_n_qubit(gate, state, nqubits, weight)
 
 
-def execute_circuit(self, circuit, weight: int, initial_state=None, nshots: int = 1000):
+def execute_circuit(
+    self,
+    circuit,
+    weight: int,
+    initial_state: Optional[ArrayLike] = None,
+    nshots: int = 1000,
+):
     """Execute ``circuit`` by applying the queue of gates to the ``initial_state``.
 
     Args:
         circuit (:class:`qibo.models.circuit.Circuit`): Hamming-weight-preserving circuit
             to be executed.
         weight (int): fixed Hamming weight of the ``initial_state``.
-        initial_state (ndarray, optional): initial state that ``circuit`` acts on.
+        initial_state (ArrayLike, optional): initial state that ``circuit`` acts on.
             If ``None``, defaults to :math:`\\ket{0^{n-k} \\, 1^{k}}`,
             with :math:`n` being the total number of qubits in the circuit,
             and :math:`k` being the Hamming ``weight``. Defaults to ``None``.
@@ -108,17 +119,17 @@ def execute_circuit(self, circuit, weight: int, initial_state=None, nshots: int 
         )
 
 
-def _gray_code(self, initial_string):
+def _gray_code(self, initial_string: ArrayLike) -> ArrayLike:
     """Return all bitstrings of a fixed Hamming weight.
 
     Uses the ``ehrlich_algorithm`` with an ``initial_string``.
 
     Args:
-        initial_string (ndarray): Array of bits representing the input
+        initial_string (ArrayLike): Array of bits representing the input
             of the Ehrlich algorithm.
 
     Returns:
-        ndarray: All bitstrings with the same Hamming weight as ``initial_string``.
+        ArrayLike: All bitstrings with the same Hamming weight as ``initial_string``.
     """
     from qibo.models._encodings import _ehrlich_algorithm  # pylint: disable=C0415
 
@@ -131,7 +142,7 @@ def _gray_code(self, initial_string):
 
 def _get_cached_strings(
     self, nqubits: int, weight: int, ncontrols: int = 0, two_qubit_gate: bool = True
-):
+) -> Union[ArrayLike, List[ArrayLike]]:
     """Generate list of strings necessary for the custom ``apply_gate`` method.
 
     Given the total number of qubits ``nqubits``, the Hamming ``weight`` to be
@@ -151,7 +162,7 @@ def _get_cached_strings(
             gate. Defaults to ``True``.
 
     Returns:
-        ndarray or list: ndarray of bitstrings for two-qubit gates or a list of two ndarrays
+        ArrayLike or List[int]: Bitstrings for two-qubit gates or a list of two arrays
         of bitstrings for single-qubit gates.
     """
     if two_qubit_gate:
@@ -181,7 +192,9 @@ def _get_cached_strings(
     return strings
 
 
-def _get_lexicographical_order(self, nqubits, weight):
+def _get_lexicographical_order(
+    self, nqubits: int, weight: int
+) -> Dict[str, Tuple[int, ...]]:
     """Sort bistrings generated from ``self._get_cached_strings`` in lexicographical order.
 
     Bitstrings are sorted in lexicographical (ascending) order.
@@ -212,7 +225,7 @@ def _get_lexicographical_order(self, nqubits, weight):
     return _dict_indexes
 
 
-def _get_single_qubit_matrix(self, gate):
+def _get_single_qubit_matrix(self, gate: Gate) -> Tuple[complex, ...]:
     """Return non-zero elements of the matrix representation of
     Hamming-weight-preserving single-qubit gates.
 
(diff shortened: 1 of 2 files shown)


src/qibo/backends/_hamming_weight_operations.py
```diff
@@ -507,64 +507,94 @@ def _apply_gate_n_qubit(self, gate, state, nqubits, weight):
...
+                for t in range(k):
+                    new_bits[active_qubits[t]] = (j >> (k - t - 1)) & 1
 
-    new_matrix = self.cast(new_matrix, dtype=new_matrix.dtype)
-    state = new_matrix @ state
+                if int(self.sum(new_bits)) != hw:
+                    continue
 
-    return state
+                encoded = int((new_bits * powers).sum())
```
scarrazza: Just checking but `new_bits * powers` is an element-wise multiplication, not a bitwise operation, are you sure this is correct or maybe you just want:


```suggestion
                encoded = 0
                for bit, power in zip(new_bits, powers):
                    encoded |= int(bit) * power
```

src/qibo/backends/_hamming_weight_operations.py
```diff
@@ -507,64 +507,94 @@ def _apply_gate_n_qubit(self, gate, state, nqubits, weight):
...
+    if not hasattr(self, "_transition_cache"):
+        self._transition_cache = {}
+        self._local_index_cache = {}
 
-    matrix = self.zeros((dim, dim))
-    matrix = self.cast(matrix, dtype=self.dtype)
-    mod_condition = indexes[:, None] % 2 ** (
-        nqubits - gate_qubits - ncontrols
-    ) == indexes[None, :] % 2 ** (nqubits - gate_qubits - ncontrols)
+    key = (tuple(active_qubits), nqubits, weight)
```
scarrazza: This key is using a mutable tuple and mutable list operations, maybe it could be more robust to do:

```suggestion
    key = (tuple(sorted(active_qubits)), tuple(controls), nqubits, weight)
```

src/qibo/backends/_hamming_weight_operations.py
```diff
@@ -507,64 +507,94 @@ def _apply_gate_n_qubit(self, gate, state, nqubits, weight):
     gate_matrix = gate.matrix(backend=self)
     qubits = list(gate.target_qubits)
     gate_qubits = len(qubits)
-    if 2 ** (gate_qubits) != gate_matrix.shape[0]:
+    if 2**gate_qubits != gate_matrix.shape[0]:
```
scarrazza, line 531:
```suggestion
    if 2 ** gate_qubits != gate_matrix.shape[0]:
```

src/qibo/backends/_hamming_weight_operations.py
```diff
@@ -507,64 +507,94 @@ def _apply_gate_n_qubit(self, gate, state, nqubits, weight):
...
+        expanded_matrix = self.identity(full_dim, dtype=gate_matrix.dtype)
+        for trow in range(2**ntargets):
+            row_index = (trow << ncontrols) | ctrl_mask
+            for tcol in range(2**ntargets):
+                col_index = (tcol << ncontrols) | ctrl_mask
+                expanded_matrix[row_index, col_index] = gate_matrix[trow, tcol]
+        gate_matrix = expanded_matrix
+
+    active_qubits = qubits + controls
+    k = len(active_qubits)
```
scarrazza, line 553:
```suggestion
    k = len(active_qubits)
    if len(set(active_qubits)) != k:
        raise_error(ValueError, "Duplicate qubit indices in active_qubits")
```

src/qibo/backends/_hamming_weight_operations.py
```diff
@@ -507,64 +507,94 @@ def _apply_gate_n_qubit(self, gate, state, nqubits, weight):
...
-    ]
+        index_map = {int(encoded_basis[i]): i for i in range(d)}
 
-    diagonal_indices = self.identity(dim, dtype=bool)
-    matrix[mod_condition & ~control_condition & diagonal_indices] = 1
+        transitions = -self.ones((d, 2**k), dtype=self.int64)
+        local_indices = self.zeros(d, dtype=self.int64)
 
-    new_matrix = self.zeros((dim, dim), dtype=self.dtype)
+        for i in range(d):
```
scarrazza: If I am not mistaken `d` can be quite large, e.g. for 20 qubits and k=10 it is >180k. If that is the case you should replace the for loop with a vectorized version.

src/qibo/backends/_hamming_weight_operations.py
```diff
@@ -507,64 +507,94 @@ def _apply_gate_n_qubit(self, gate, state, nqubits, weight):
...
+            bits = basis[i]
+            hw = int(self.sum(bits))
 
-    strings_array = np.array([list(s) for s in strings])
-    reordered_strings_array = strings_array[:, map_]
+            idx = 0
+            for t in range(k):
+                idx = (idx << 1) | int(bits[active_qubits[t]])
+            local_indices[i] = idx
```
scarrazza: E.g., something like that (untested):

```python
   # Select bits for all basis states at once
   active_bits = basis[:, active_qubits]

   # Create powers of 2: [2^(k-1), ..., 1]
   powers = 2 ** np.arange(k - 1, -1, -1)

   # Compute indices via dot product
   local_indices = active_bits.dot(powers)
```

src/qibo/backends/_hamming_weight_operations.py
```diff
@@ -507,64 +507,94 @@ def _apply_gate_n_qubit(self, gate, state, nqubits, weight):
...
+                    new_bits[active_qubits[t]] = (j >> (k - t - 1)) & 1
 
-    new_matrix = self.cast(new_matrix, dtype=new_matrix.dtype)
-    state = new_matrix @ state
+                if int(self.sum(new_bits)) != hw:
+                    continue
 
-    return state
+                encoded = int((new_bits * powers).sum())
+                transitions[i, j] = index_map[encoded]
```
scarrazza: Also here, the nexted loop and dictionary lookup might introduce some slowness. Maybe you could use np.searchsorted to find indices efficiently by using the sorted `encoded_basis`. In practice, you could construct all candidate `new_bits` for all i and j, then filter those that preserve Hamming weight, compute encoded values for valid transitions and use np.searchsorted(encoded_basis, encoded) to find the target indices in the basis.

src/qibo/backends/_hamming_weight_operations.py
```diff
@@ -507,64 +507,94 @@ def _apply_gate_n_qubit(self, gate, state, nqubits, weight):
...
 
     self._dict_indexes = self._get_lexicographical_order(nqubits, weight)
 
-    strings = np.array(list(self._dict_indexes.keys()))
-    indexes = self.cast(
-        [index[1] for index in self._dict_indexes.values()], dtype=self.int64
-    )
-    dim = len(indexes)
+    if not hasattr(self, "_transition_cache"):
+        self._transition_cache = {}
```
scarrazza: It think you can move this initialization to the init method of hamming_weights.py.

src/qibo/backends/_hamming_weight_operations.py
```diff
@@ -507,64 +507,94 @@ def _apply_gate_n_qubit(self, gate, state, nqubits, weight):
     gate_matrix = gate.matrix(backend=self)
     qubits = list(gate.target_qubits)
```
scarrazza, line 529: Given that we are changing this function, we may add:

```python
if weight < 0 or weight > nqubits:
    raise_error(
        ValueError,
        f"Hamming weight {weight} must be between 0 and {nqubits}"
    )
```

src/qibo/backends/_hamming_weight_operations.py
```diff
@@ -507,64 +507,94 @@ def _apply_gate_n_qubit(self, gate, state, nqubits, weight):
...
-    new_matrix = matrix[reordered_indexes][:, reordered_indexes]
+            for j in range(2**k):
+                new_bits = bits.copy()
+                for t in range(k):
+                    new_bits[active_qubits[t]] = (j >> (k - t - 1)) & 1
 
-    new_matrix = self.cast(new_matrix, dtype=new_matrix.dtype)
-    state = new_matrix @ state
+                if int(self.sum(new_bits)) != hw:
+                    continue
```
scarrazza:
```suggestion
                # Precompute hamming weight using numpy operations
                new_bits_sum = self.sum(new_bits, dtype=self.int64)
                if new_bits_sum != hw:
                    continue
```

AlejandroSopena:
Thanks @scarrazza, I have vectorized the loops. The memory of the implementation in `main` scales with `d^2`, where `d` is `(nqubits choose k) `and `k` is the Hamming weight. It fails on GPU for 30 qubits and Hamming weight 4 because `log2(d) ~ 15`, where `d = 30 choose 4`. The implementation in this PR scales with `d`, so we can go up to more qubits.

This plot shows the execution time versus the number of qubits for `hamming_weight = 4`. We can see that the implementation in this PR scales better.

image

image

src/qibo/backends/_hamming_weight_operations.py
```diff
@@ -507,64 +506,92 @@ def _apply_gate_n_qubit(self, gate, state, nqubits, weight):
...
+            for tcol in range(2**ntargets):
+                col_index = (tcol << ncontrols) | ctrl_mask
+                expanded_matrix[row_index, col_index] = gate_matrix[trow, tcol]
+        gate_matrix = expanded_matrix
+
+    active_qubits = qubits + controls
+    k = len(active_qubits)
+
+    if len(set(active_qubits)) != k:  # pragma: no cover
+        raise ValueError("Duplicate qubit indices in active_qubits")
```
scarrazza:
```suggestion
        raise_error(ValueError, "Duplicate qubit indices in active_qubits")
```

src/qibo/backends/hamming_weight.py
```diff
@@ -57,4 +57,8 @@ def HammingWeightBackend(platform=None):
 
     hw_backend._dict_indexes = None
 
+    hw_backend._transition_cache = {}
+    hw_backend._local_index_cache = {}
+    hw_backend._flat_cache = {}
```
scarrazza, line 62: Placing these here is better, but we should monitor if it does not generate a memory leak in long-running simulations or circuits with many unique gate configurations.

src/qibo/backends/abstract.py
```diff
@@ -231,6 +231,11 @@ def uint8(self) -> DTypeLike:
     def abs(self, array: ArrayLike, **kwargs) -> Union[int, float, complex, ArrayLike]:
         return self.engine.abs(array, **kwargs)
 
+    def _add_at(
```
scarrazza: `_add_at` uses self.engine.add.at which is a numpy.ufunc method. While this works for Numpy-based engines, it may fail for backends like TensorFlow or PyTorch that don't share this exact syntax. Since HammingWeightBackend typically relies on Numpy, this should be fine for now, but it's worth noting for future generic compatibility and proposing changes in QiboML too.

src/qibo/backends/abstract.py
```diff
@@ -231,6 +231,11 @@ def uint8(self) -> DTypeLike:
...
+        self, array_1: ArrayLike, indices: ArrayLike, array_2: ArrayLike
+    ) -> None:
```
scarrazza, line 236:
```suggestion
    ) -> None:
    """Add ``array_2`` to ``array_1`` at specified ``indices`` in-place.
    
    Args:
        array_1 (ArrayLike): Output array to be modified.
        indices (ArrayLike): Indices at which to add elements.
        array_2 (ArrayLike): Input array with elements to add.
    """
```

src/qibo/backends/abstract.py
renatomello: `qiboml` as a whole already has a problem with in-place operations because both `tensorflow` and `jax` do not allow item assignment when operating on their `arraylike` objects. For instance, see L260-275 of the following

https://github.com/qiboteam/qiboml/blob/e1003f9e392ff02f584f5cae3fa510406abd3947/src/qiboml/models/optimizers.py#L260

src/qibo/backends/abstract.py
```diff
@@ -231,6 +231,18 @@ def uint8(self) -> DTypeLike:
     def abs(self, array: ArrayLike, **kwargs) -> Union[int, float, complex, ArrayLike]:
         return self.engine.abs(array, **kwargs)
 
+    def _add_at(
```
renatomello: I don't think this method should be a protected one. Should be user-visible like the others.

src/qibo/backends/abstract.py
scarrazza: Indeed, but then we assume this will not work with these backends, which is fine, but we should prevent crashing due to the missing method.

renatomello:
@AlejandroSopena @scarrazza I used this PR to properly type the functions in `backend._hamming_weight_operations.py`

PR #1810, opened 2026-03-24
Refactor of `models.encodings` that require data loading

renatomello:
Small refactor of functions that create circuits that load data. Now, the functions can return circuits even when no data is passed by defaulting the gate phases to $0.0$. This is convenient for random initializations.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -182,6 +182,7 @@ files = [
     {file = "certifi-2026.2.25-py3-none-any.whl", hash = "sha256:027692e4402ad994f1c42e52a4997a9763c646b73e4096e4d5d6db8af1d6f0fa"},
     {file = "certifi-2026.2.25.tar.gz", hash = "sha256:e887ab5cee78ea814d3472169153c2d12cd43b14bd03329a39a9c6e2e80bfba7"},
 ]
+markers = {main = "extra == \"qir\" or extra == \"cudaq\""}
 
 [[package]]
 name = "cffi"
@@ -403,6 +404,7 @@ files = [
     {file = "charset_normalizer-3.4.4-py3-none-any.whl", hash = "sha256:7a32c560861a02ff789ad905a2fe94e3f840803362c84fecf1851cb4cf3dc37f"},
     {file = "charset_normalizer-3.4.4.tar.gz", hash = "sha256:94537985111c35f28720e43603b8e7b43a6ecfb2ce1d3058bbe955b73404e21a"},
 ]
+markers = {main = "extra == \"qir\" or extra == \"cudaq\""}
 
 [[package]]
 name = "clarabel"
@@ -1693,6 +1695,7 @@ files = [
     {file = "idna-3.11-py3-none-any.whl", hash = "sha256:771a87f49d9defaf64091e6e6fe9c18d4833f140bd19464795bc32d966ca37ea"},
     {file = "idna-3.11.tar.gz", hash = "sha256:795dafcc9c04ed0c1fb032c2aa73654d8e8c5023a7df64a53f39190ada629902"},
 ]
+markers = {main = "extra == \"qir\" or extra == \"cudaq\""}
 
 [package.extras]
 all = ["flake8 (>=7.1.1)", "mypy (>=1.11.2)", "pytest (>=8.3.2)", "ruff (>=0.6.2)"]
@@ -3383,14 +3386,14 @@ typing-extensions = ">=4.6.0,<4.7.0 || >4.7.0"
 
 [[package]]
 name = "pygments"
-version = "2.19.2"
+version = "2.20.0"
 description = "Pygments is a syntax highlighting package written in Python."
 optional = false
-python-versions = ">=3.8"
+python-versions = ">=3.9"
 groups = ["dev", "docs"]
 files = [
-    {file = "pygments-2.19.2-py3-none-any.whl", hash = "sha256:86540386c03d588bb81d44bc3928634ff26449851e99741617ecb9037ee5ec0b"},
-    {file = "pygments-2.19.2.tar.gz", hash = "sha256:636cb2477cec7f8952536970bc533bc43743542f70392ae026374600add5b887"},
+    {file = "pygments-2.20.0-py3-none-any.whl", hash = "sha256:81a9e26dd42fd28a23a2d169d86d7ac03b46e2f8b59ed4698fb4785f946d0176"},
+    {file = "pygments-2.20.0.tar.gz", hash = "sha256:6757cd03768053ff99f3039c1a36d6c0aa0b263438fcab17520b30a303a82b5f"},
 ]
 
 [package.extras]
@@ -4052,6 +4055,7 @@ files = [
     {file = "requests-2.33.0-py3-none-any.whl", hash = "sha256:3324635456fa185245e24865e810cecec7b4caf933d7eb133dcde67d48cee69b"},
     {file = "requests-2.33.0.tar.gz", hash = "sha256:c7ebc5e8b0f21837386ad0e1c8fe8b829fa5f544d8df3b2253bff14ef29d7652"},
 ]
+markers = {main = "extra == \"qir\" or extra == \"cudaq\""}
 
 [package.dependencies]
 certifi = ">=2023.5.7"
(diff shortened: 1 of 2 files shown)


qumaciel:
This brakes all previous code using these functions. Do we have any deprecation policy?
Can we keep the functions with the old functions' signature redirecting to the new one emitting a warning of deprecation for X versions or months?

renatomello:
> This brakes all previous code using these functions. Do we have any deprecation policy? Can we keep the functions with the old functions' signature redirecting to the new one emitting a warning of deprecation for X versions or months?

@scarrazza

scarrazza:
@qumaciel we do have some cases like https://github.com/qiboteam/qibo/blob/e16a524016c7039e824252606e6488c2ec3d68bd/src/qibo/hamiltonians/hamiltonians.py#L123. Therefore, you can follow this approach for now.

src/qibo/models/encodings.py
```diff
@@ -93,18 +101,40 @@ def binary_encoder(
...
     """
-    backend = _check_backend(backend)
+    # TODO: remove deprecated call on qibo version 0.3.3
+    if not isinstance(nqubits, int):  # pragma: no cover
+        log.warning(
+            "Deprecated initialisation of the ``binary_encoder`` function."
+            + "See latest documentation for the current way to initialise this function."
+            + "Deprecated initialisation will be removed on version ``0.3.3``."
+        )
+        return _binary_encoder_deprecated(
```
qumaciel: Do we need to duplicate the code for the _deprecated()?
Can't we return the new function with data in the right place?

src/qibo/models/encodings.py
```diff
@@ -555,9 +595,35 @@ def hamming_weight_encoder(
...
         `Phys. Rev. Applied 23, 044014 (2025) <https://doi.org/10.1103/PhysRevApplied.23.044014>`_.
     """
+    # TODO: remove deprecated call on qibo version 0.3.3
+    if not isinstance(nqubits, int):  # pragma: no cover
+        log.warning(
+            "Deprecated initialisation of the ``hamming_weight_encoder`` function."
+            + "See latest documentation for the current way to initialise this function."
+            + "Deprecated initialisation will be removed on version ``0.3.3``."
+        )
+        return _hamming_weight_encoder_deprecated(
```
qumaciel: Do we need to duplicate the code for the _deprecated()?
Can't we return the new function with data in the right place?

src/qibo/models/encodings.py
```diff
@@ -705,39 +787,46 @@ def phase_encoder(data, rotation: str = "RY", backend=None, **kwargs):
...
         :class:`qibo.models.circuit.Circuit`: Circuit that loads ``data`` in phase encoding.
     """
+    # TODO: remove deprecated call on qibo version 0.3.3
+    if not isinstance(nqubits, int):  # pragma: no cover
+        log.warning(
+            "Deprecated initialisation of the ``phase_encoder`` function."
+            + "See latest documentation for the current way to initialise this function."
+            + "Deprecated initialisation will be removed on version ``0.3.3``."
+        )
+        return _phase_encoder_deprecated(
```
qumaciel: Do we need to duplicate the code for the _deprecated()?
Can't we return the new function with data in the right place?

src/qibo/models/encodings.py
```diff
@@ -844,44 +941,56 @@ def unary_encoder(data, architecture: str = "tree", backend=None, **kwargs):
...
     """
-    backend = _check_backend(backend)
+    # TODO: remove deprecated call on qibo version 0.3.3
+    if not isinstance(nqubits, int):  # pragma: no cover
+        log.warning(
+            "Deprecated initialisation of the ``unary_encoder`` function."
+            + "See latest documentation for the current way to initialise this function."
+            + "Deprecated initialisation will be removed on version ``0.3.3``."
+        )
+        return _unary_encoder_deprecated(
```
qumaciel: Do we need to duplicate the code for the _deprecated()?
Can't we return the new function with data in the right place?

src/qibo/models/encodings.py
```diff
@@ -1057,9 +1172,27 @@ def up_to_k_hamming_weight_encoder(
...
         2. `Hyperpherical coordinates <https://en.wikipedia.org/wiki/N-sphere>`_.
     """
+    # TODO: remove deprecated call on qibo version 0.3.3
+    if not isinstance(nqubits, int):  # pragma: no cover
+        log.warning(
+            "Deprecated initialisation of the ``up_to_k_hamming_weight_encoder`` function."
+            + "See latest documentation for the current way to initialise this function."
+            + "Deprecated initialisation will be removed on version ``0.3.3``."
+        )
+        return _up_to_k_hamming_weight_encoder_deprecated(  # pylint: disable=W1114
```
qumaciel: Do we need to duplicate the code for the _deprecated()?
Can't we return the new function with data in the right place?

src/qibo/models/_encodings.py
```diff
@@ -281,6 +284,8 @@ def _binary_encoder_hopf(
         circuit.add(gate_list)
 
     angles = _generate_rbs_angles(data, "tree", dims, backend=backend)
+    angles = backend.where(backend.isnan(angles), 0.0, angles)
```
scarrazza, line 287: Is it okay to keep this silent and force to zero instead of raising some warning?

renatomello: I would say so, yes. This is the intended behaviour when no `data` is passed and it should do nothing when `data` is passed.

src/qibo/models/_encodings.py
```diff
@@ -1,3 +1,5 @@
+"""Helper functions for the `models.encodingss` module."""
```
scarrazza:
```suggestion
"""Helper functions for the `models.encodings` module."""
```

src/qibo/models/encodings.py
```diff
@@ -93,18 +96,43 @@ def binary_encoder(
...
+        complex_data = bool(
+            "complex" in str(data.dtype)
+        )  # backend-agnostic way of checking the dtype
+    else:
+        dims = 2**nqubits
+        complex_data = bool("complex" in parametrization)
+        data = backend.cast(
+            [1] + [0] * (dims - 1),
+            dtype=backend.complex128 if complex_data else backend.float64,
+        )
```
scarrazza, line 135: Suppose the user requests Hopf for complex states, this will fallback always to hyperspherical (independently if data is provided). Should the if condition include a:
```python
if parametrization in ("hopf", "hopf-complex"):
```

src/qibo/models/encodings.py
```diff
@@ -844,44 +940,62 @@ def unary_encoder(data, architecture: str = "tree", backend=None, **kwargs):
...
-    if architecture == "tree" and not math.log2(data.shape[0]).is_integer():
-        raise_error(
-            ValueError,
-            "When ``architecture = 'tree'``, len(data) must be a power of 2. "
-            + f"However, it is {len(data)}.",
-        )
-
-    nqubits = len(data)
+    if data is not None:
+        if architecture == "tree" and not math.log2(data.shape[0]).is_integer():
```
scarrazza: This will not work if data is a python list.

src/qibo/models/_encodings.py
```diff
@@ -521,8 +526,8 @@ def _generate_rbs_angles(
 
         j_max = int(dims / 2)
 
-        r_array = np.zeros(dims - 1, dtype=float)
-        phases = np.zeros(dims - 1, dtype=float)
+        r_array = backend.zeros(dims - 1, dtype=backend.float64)
+        phases = backend.zeros(dims - 1, dtype=backend.float64)
```
scarrazza, line 530: Just a curiosity: if backends are tensorflow or jax, which creates tensors, does the assigments below with = will work?

src/qibo/ui/bloch.py
```diff
@@ -161,18 +165,22 @@ def _paulis_expectation(self, state):
         """This function computes the expectation value of Pauli matrices
         on the considered state and yields its cartesian coordinates on the Bloch sphere.
         """
+        from qibo.backends.numpy import NumpyBackend  # pylint: disable=C0415
+
+        _backend = NumpyBackend()
```
scarrazza, line 175: This might add some performance degradation when adding many states. 

What about something like this:
```python
_numpy_backend = None

def _paulis_expectation(self, state):
    if self._numpy_backend is None:
        from qibo.backends.numpy import NumpyBackend
        self._numpy_backend = NumpyBackend()
    _backend = self._numpy_backend
```

src/qibo/ui/bloch.py
```diff
@@ -126,7 +128,7 @@ def _create_sphere(self):
         combinations_axis = [(x, y, z), (z, x, y), (y, z, x)]
 
         # Axis lines
-        line, zeros = np.linspace(-1, 1, 100), np.zeros(shape=(100))
+        line, zeros = np.linspace(-1, 1, 100), np.zeros(shape=100)
```
scarrazza:
```suggestion
        line, zeros = np.linspace(-1, 1, 100), np.zeros(100)
```

src/qibo/models/_encodings.py
renatomello: It will not, and, currently, we do not have a general solution to this issue.

PR #1811, opened 2026-03-26
Add `arcsin`, `cumsum`, and `mod` methods to abstract backend

renatomello:
Checklist:
- [x] Tests are passing.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -182,7 +182,6 @@ files = [
     {file = "certifi-2026.2.25-py3-none-any.whl", hash = "sha256:027692e4402ad994f1c42e52a4997a9763c646b73e4096e4d5d6db8af1d6f0fa"},
     {file = "certifi-2026.2.25.tar.gz", hash = "sha256:e887ab5cee78ea814d3472169153c2d12cd43b14bd03329a39a9c6e2e80bfba7"},
 ]
-markers = {main = "extra == \"qir\" or extra == \"cudaq\""}
 
 [[package]]
 name = "cffi"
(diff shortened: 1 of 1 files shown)


src/qibo/backends/abstract.py
```diff
@@ -254,6 +254,9 @@ def arange(self, *args, **kwargs) -> ArrayLike:
     def arccos(self, array: ArrayLike, **kwargs) -> ArrayLike:
         return self.engine.arccos(array, **kwargs)
 
+    def arcsin(self, array: ArrayLike, **kwargs) -> ArrayLike:  # pragma: no cover
+        return self.engine.arcsin(array, **kwargs)
+
```
scarrazza, line 259: Please don't disable coverage, instead add to the test_backends something like:
```python
# tests/test_backends.py
def test_arcsin(backend):
    import numpy as np
    array = backend.cast([0.0, 0.5, 1.0])
    result = backend.arcsin(array)
    backend.assert_allclose(result, np.arcsin([0.0, 0.5, 1.0]))

def test_cumsum(backend):
    import numpy as np
    array = backend.cast([1, 2, 3, 4])
    result = backend.cumsum(array)
    backend.assert_allclose(result, np.cumsum([1, 2, 3, 4]))

def test_mod(backend):
    import numpy as np
    array = backend.cast([5, 7, 10])
    result = backend.mod(array, 3)
    backend.assert_allclose(result, np.mod([5, 7, 10], 3))
```

renatomello: These are not the only methods with disabled coverage. I think this should be left to a separate PR like the docstrings because it will be enourmous if we want to do it for every single method that is not being tested by another function already. Moreover, these three methods are planned to be tested by another function soon to be implemented in `models.encodings`.

Issue #1795, opened 2026-03-04
Investigate load from samples


scarrazza:
We previously implemented the load_result method to retrieve measurements and simulation results from NumPy arrays. It would be beneficial to expand this to allow creating measurement outcome objects directly from raw samples or frequency dictionaries.

PR #1815, opened 2026-03-30
Load from samples and frequencies

scarrazza:
Addresses #1795. @alecandido please let me know if this idea also works for qibolab.

### DIFF
--- a/src/qibo/result.py
+++ b/src/qibo/result.py
@@ -6,7 +6,6 @@ from typing import Optional, Union
 
 import numpy as np
 from numpy.typing import ArrayLike
-from sympy.core import basic
 
 from qibo import __version__, gates
 from qibo.config import raise_error
@@ -334,7 +333,7 @@ class MeasurementOutcomes:
         # Expand measured probabilities into the full circuit qubit space.
         # Unmeasured qubits are placed in the |0⟩ state, consistent with the
         # standard qubit initialisation convention.
-        full_probs = backend.zeros(2**nqubits, dtype=backend.float64)
+        full_probs = self.backend.zeros(2**nqubits, dtype=self.backend.float64)
 
         for measured_state in range(2**n_measured):
             p = float(measured_probs[measured_state])
@@ -552,8 +551,8 @@ class MeasurementOutcomes:
     def from_samples(
         cls,
         samples,
-        qubits: Optional[Union[List[int], Tuple[int, ...]]] = None,
-        backend: Optional[Backend] = None,
+        qubits: Optional[Union[list[int], tuple[int, ...]]] = None,
+        backend: Optional["Backend"] = None,
     ):
         """Constructs a :class:`qibo.result.MeasurementOutcomes` directly from
         a binary samples array.
@@ -627,9 +626,9 @@ class MeasurementOutcomes:
         cls,
         frequencies,
         nqubits: Optional[int] = None,
-        qubits: Optional[Union[List[int], Tuple[int, ...]]] = None,
+        qubits: Optional[Union[list[int], tuple[int, ...]]] = None,
         seed: Optional[int] = None,
-        backend: Optional[Backend] = None,
+        backend: Optional["Backend"] = None,
     ):
         """Constructs a :class:`qibo.result.MeasurementOutcomes` from a
         frequencies dictionary.


alecandido:
To me, it seems that this is not adding much advantage (e.g. for Qibolab) wrt the current implementation.

The current problem was the lack of support for registers, since Qibolab is always storing them in individual gates. But this will continue to be the case (registers are not mentioned anywhere in the new methods), with the only difference that now, instead of individual gates, there will be one single measurement gate for all qubits.
While not getting closer to the semantics of current Qibo circuits, this seems to violate it even more, since the results are not stored in the circuit itself, but in a new isolated gate, which is created on purpose for that.

If the goal is to get rid of mutating elements in our input (which I would support in full), I would avoid using the measurement gate completely, and phase out most of the code related to `MeasurementSymbol`, `MeasurementResult`, `QuantumState`, `MeasurementOutcomes`, `CircuitResult`, and the `M` gate itself (they are many objects, mostly handling pointers at various levels, and passing down nested calls), in favor of bare arrays (we could have just two array wrappers, for shots and states - observables may be a separate story, since they are related to `hamiltonians`).

If we are not willing to invest right now in that direction, I'm not sure we really have a problem to solve.
The solution currently adopted by the `QibolabBackend` may be even repeated somewhere else (though it is quite related to the internal transformation, in this case `circuit -> sequence`), but all in all is almost ~4 lines of code
https://github.com/qiboteam/qibolab/blob/6aa585bfc75f3ce7351f1302826d190abf65a107/src/qibolab/_core/backends.py#L92-L111
and it is mutating the original gates, as currently expected in Qibo circuits execution.

scarrazza:
Thanks @alecandido. The points you raised make sense, but they fall a bit outside the scope of this specific PR. I've been discussing this feature with @marcorossi5 and @GabrielePalazzo (for cuda-q), and I believe including `from_frequencies` right now is useful to help simplify things and reduce the amount of transferred data for the respective cloud services.

alecandido:
Ok. For the cloud in simulation, I have no idea which are the limitations. I was mainly taking the point of view of Qibolab.

I guess the cloud performances also depend on packaging the arrays in tiny chunks, instead of a single block. Wouldn't it benefit from communicating a single raw array, with certain shape conventions, instead of wrapping it in a measurement object?

scarrazza:
Indeed, we would like to get rid of QuantumState and any other objects containing large NumPy arrays or binary dumps (i..e load_results calls). Most cloud services (including cuda-q) currently just pass around lightweight dictionaries containing frequencies. Therefore, this PR enables this mechanism when reconstructing measurements (from both simulation and hardware) via the local client.

alecandido:
Yeah, and I would say that it is even more lightweight if, instead of a dictionary (most likely serialized in JSON) you send back the dictionary as a `(2, nstates)`-dimensional array, with the zip of the states and related frequencies.

scarrazza:
I think this is still suboptimal because samples can be generated locally, directly from frequencies (which is what CUDA-Q does, and what this PR does, for example). This reduces communication overhead and avoids NumPy transfer and serialization as much as possible.

alecandido:
Indeed, my comment was saying to send an array of frequencies, not samples (to avoid confusion, `nstates` are the number of bitstrings with a frequency different from 0).

scarrazza:
Great, thanks.

src/qibo/result.py
```diff
@@ -496,6 +496,237 @@ def load(cls, filename: str):
...
+        backend=None,
+        seed=None,
+    ):
+        """Constructs a :class:`qibo.result.MeasurementOutcomes` from a
+        frequencies dictionary.
+
+        The frequencies are expanded into a binary samples array, shuffled to
+        avoid ordering artifacts, and passed through the standard construction
+        path so that all methods (``samples()``, ``frequencies()``,
+        ``probabilities()``) work correctly.
```
GabrielePalazzo, line 639: If I test this circuit the two probabilities are different (`m.probabilities()` only contains measured qubits).
```python
from qibo import Circuit
from qibo.gates import M, X
from qibo.result import MeasurementOutcomes

circ = Circuit(3)
circ.add(X(0))
circ.add(M(0,1))

res = circ()

m = MeasurementOutcomes.from_frequencies(res.frequencies())
print(res.probabilities())
print(m.probabilities())
```

I don't know if this is worth noting.

scarrazza: Thanks, that is indeed quite an interesting case which somehow exploits our choice of returning the full circuit probability instead of focusing on the M gates. In principle, you should get the same probabilities if you call `res.probabilities(qubits=[0,1])` so we avoid guessing the total number of qubits. However, makes sense to unlock the `nqubits` argument in `from_frequencies` (and `from_samples`) to align probabilities.

I have implemented a potential fix which now separates the measured qubits from the total number of qubits (`nqubits` argument). Assuming all qubits are initiated at ground state, your code should work if you do:

```python
m = MeasurementOutcomes.from_frequencies(res.frequencies(), nqubits=3)
```

Please give a try.

GabrielePalazzo: I don't know if we should also check this case (maybe by adding the tuple of measured qubits).
```python
circ = Circuit(3)
circ.add(X(1))
circ.add(M(1,2))
```
Everything else seems ok for me.

src/qibo/result.py
```diff
@@ -278,33 +284,75 @@ def frequencies(self, binary: bool = True, registers: bool = False):
...
+        Args:
+            qubits (list or set, optional): Set of qubits for which to compute
+                probabilities. If ``None`` and ``nqubits`` was provided at
+                construction, probabilities are returned over all circuit
+                qubits; otherwise only over the measured qubits.
+                Defaults to ``None``.
+
         Returns:
-            The array containing the probabilities of the measured qubits.
+            The array containing the probabilities of the requested qubits.
```
renatomello:
```suggestion
            ArrayLike: The array containing the probabilities of the requested qubits.
```

src/qibo/result.py
```diff
@@ -278,33 +284,75 @@ def frequencies(self, binary: bool = True, registers: bool = False):
     def probabilities(self, qubits: Optional[Union[list, set]] = None):
```
renatomello:
```suggestion
    def probabilities(self, qubits: Optional[Union[list, set]] = None) -> ArrayLike:
```

with `from numpy.typing import ArrayLike` at top level

src/qibo/result.py
```diff
@@ -278,33 +284,75 @@ def frequencies(self, binary: bool = True, registers: bool = False):
...
             )
 
+        # Build probability array in the measured-qubit space
         if self._probs is not None and not self.measurement_gate.has_bitflip_noise():
+            measured_probs = self._probs
+        else:
+            measured_probs = [0] * 2**n_measured
+            for state, freq in self.frequencies(binary=False).items():
+                measured_probs[state] = freq / self.nshots
+            measured_probs = self.backend.cast(measured_probs)
```
renatomello:
```suggestion
            measured_probs = self.backend.cast(measured_probs, dtype=self.backend.float64)
```

src/qibo/result.py
```diff
@@ -278,33 +284,75 @@ def frequencies(self, binary: bool = True, registers: bool = False):
...
+            self._probs = measured_probs
+
+        if nqubits == n_measured:
+            # No unmeasured qubits: use the standard path
             return self.backend.calculate_probabilities(
-                np.sqrt(self._probs), qubits, nqubits
+                self.backend.cast(np.sqrt(np.array(measured_probs))),
```
renatomello: `measured_probs` has already been casted to backend `ArrayLike` type but then `np.array` is called on it. The following should work:
```suggestion
                self.backend.sqrt(measured_probs),
```

src/qibo/result.py
```diff
@@ -278,33 +284,75 @@ def frequencies(self, binary: bool = True, registers: bool = False):
...
-            probs[state] = freq / self.nshots
-        probs = self.backend.cast(probs)
-        self._probs = probs
+        # Expand measured probabilities into the full circuit qubit space.
+        # Unmeasured qubits are placed in the |0⟩ state, consistent with the
+        # standard qubit initialisation convention.
+        full_probs = np.zeros(2**nqubits)
+
+        for measured_state in range(2**n_measured):
+            p = float(np.real(measured_probs[measured_state]))
```
renatomello:
```suggestion
            p = float(measured_probs[measured_state])
```

`np.real` should not be necessary since `measured_probs` can be casted as `float64` like the suggestion above.

src/qibo/result.py
```diff
@@ -278,33 +284,75 @@ def frequencies(self, binary: bool = True, registers: bool = False):
...
 
-        probs = [0] * 2**nqubits
-        for state, freq in self.frequencies(binary=False).items():
-            probs[state] = freq / self.nshots
-        probs = self.backend.cast(probs)
-        self._probs = probs
+        # Expand measured probabilities into the full circuit qubit space.
+        # Unmeasured qubits are placed in the |0⟩ state, consistent with the
+        # standard qubit initialisation convention.
+        full_probs = np.zeros(2**nqubits)
```
renatomello:
```suggestion
        full_probs = backend.zeros(2**nqubits, dtype=backend.float64)
```

src/qibo/result.py
```diff
@@ -278,33 +284,75 @@ def frequencies(self, binary: bool = True, registers: bool = False):
...
+            full_state = 0
+            m_bit_idx = 0
+            for bit_pos in range(nqubits):
+                if bit_pos in measured_qubits:
+                    bit_val = (measured_state >> (n_measured - 1 - m_bit_idx)) & 1
+                    m_bit_idx += 1
+                    full_state |= bit_val << (nqubits - 1 - bit_pos)
+            full_probs[full_state] = p
+
+        full_probs = self.backend.cast(full_probs)
```
renatomello:
```suggestion
```

src/qibo/result.py
```diff
@@ -278,33 +284,75 @@ def frequencies(self, binary: bool = True, registers: bool = False):
...
         return self.backend.calculate_probabilities(
-            self.backend.sqrt(probs), qubits, nqubits
+            self.backend.cast(np.sqrt(np.array(full_probs))),
```
renatomello:
```suggestion
            self.backend.sqrt(full_probs),
```

src/qibo/result.py
scarrazza: The code should work as it is if you specify the measured qubits:
```python
m = MeasurementOutcomes.from_frequencies(res.frequencies(), qubits=(1,2), nqubits=3)
```
So far the only hardcoded assumption is the ground state for all qubits as initial_state.

src/qibo/result.py
```diff
@@ -496,6 +546,260 @@ def load(cls, filename: str):
...
+        backend=None,
+    ):
+        """Constructs a :class:`qibo.result.MeasurementOutcomes` directly from
+        a binary samples array.
+
+        This is useful when building measurement outcomes from experimental data
+        without needing to manually construct measurement gates.
+
+        Args:
+            samples (np.ndarray): Binary array of shape ``(nshots, nqubits)``
```
renatomello:
```suggestion
            samples (ArrayLike): Binary array of shape ``(nshots, nqubits)``.
```

src/qibo/result.py
```diff
@@ -496,6 +546,260 @@ def load(cls, filename: str):
...
+                samples = np.array([[0, 1], [1, 0], [1, 1]])
+                result = MeasurementOutcomes.from_samples(samples)
+                print(result.frequencies())  # Counter({'01': 1, '10': 1, '11': 1})
+        """
+        from qibo.backends import (  # pylint: disable=import-outside-toplevel
+            _check_backend,
+        )
+
+        backend = _check_backend(backend)
+        samples = np.array(samples)
```
renatomello:
```suggestion
        samples = backend.cast(samples, dtype=backend.int8)
```

src/qibo/result.py
```diff
@@ -496,6 +546,260 @@ def load(cls, filename: str):
         payload = np.load(filename, allow_pickle=True).item()
         return cls.from_dict(payload)
 
+    @classmethod
+    def from_samples(
+        cls,
+        samples,
+        qubits=None,
+        backend=None,
```
renatomello:
```suggestion
        backend: Optional[Backend] = None,
```

src/qibo/result.py
```diff
@@ -496,6 +546,260 @@ def load(cls, filename: str):
         payload = np.load(filename, allow_pickle=True).item()
         return cls.from_dict(payload)
 
+    @classmethod
+    def from_samples(
+        cls,
+        samples,
+        qubits=None,
```
renatomello:
```suggestion
        qubits: Optional[Union[List[int], Tuple[int, ...]]] = None,
```

src/qibo/result.py
```diff
@@ -496,6 +546,260 @@ def load(cls, filename: str):
...
+        samples = np.array(samples)
+
+        if samples.ndim != 2:
+            raise_error(
+                ValueError,
+                f"samples must be a 2D array of shape (nshots, nqubits), "
+                f"got shape {samples.shape}.",
+            )
+
+        if not np.all((samples == 0) | (samples == 1)):
```
renatomello:
```suggestion
        if not backend.all((samples == 0) | (samples == 1)):
```

src/qibo/result.py
```diff
@@ -496,6 +546,260 @@ def load(cls, filename: str):
...
+        return cls(measurements, backend=backend, samples=samples, nshots=nshots)
+
+    @classmethod
+    def from_frequencies(
+        cls,
+        frequencies,
+        nqubits=None,
+        qubits=None,
+        backend=None,
+        seed=None,
```
renatomello:
```suggestion
        nqubits: Optional[int] = None,
        qubits: Optional[Union[List[int], Tuple[int, ...]]] = None,
        seed: Optional[int] = None,
        backend: Optional[Backend] = None,
```

src/qibo/result.py
```diff
@@ -496,6 +546,260 @@ def load(cls, filename: str):
...
+                qubits (backward-compatible behaviour).
+                Defaults to ``None``.
+            qubits (tuple or list, optional): Qubit indices for the measured
+                qubits. If ``None``, defaults to ``(0, 1, ..., n_measured - 1)``
+                where ``n_measured`` is the number of measured qubits.
+                Defaults to ``None``.
+            backend (:class:`qibo.backends.abstract.Backend`, optional): Backend
+                used for calculations. If ``None``, the current default backend
+                is used. Defaults to ``None``.
+            seed (int, optional): Seed for sampling generation.
```
renatomello:
```suggestion
            seed (int, optional): Seed for sampling generation.
            backend (:class:`qibo.backends.abstract.Backend`, optional): Backend
                used for calculations. If ``None``, the current default backend
                is used. Defaults to ``None``.
```

scarrazza:
Thanks @renatomello.

PR #1830, opened 2026-04-20
Drop `python 3.10`

renatomello:
As discussed in https://github.com/qiboteam/qibochem/pull/183

Checklist:
- [x] Tests are passing.


RoyStegeman:
Hi @renatomello, python 3.10 will be EOL this year, so I see why you want to drop it. However, it's not EOL yet and some platforms still have it as the system default python installation (even the QRC cluster), so dropping support might be premature.

scarrazza:
Hi @RoyStegeman, indeed, but we have been testing more recent versions of Python (in particular 3.13) successfully on the cluster (module load python/3.13 as shown in the wiki).

My suggestion is to keep this PR as a draft for a while to identify all places where 3.10 is being used and start upgrading environments based on 3.10.

RoyStegeman:
> to identify all places where 3.10 is being used and start upgrading environments based on 3.10

What do you mean by this? Is there a doubt that qibo is stable for more recent versions of python? In principle python aims for backward compatibility so 3.10+ should only be an extension of 3.10. 

This means two things:
- If you promise python support in the range 3.10+, then you write the code as if it was 3.10 and exclude features introduced in 3.10+
- Since backward compatibility is a target of python, if 3.10 works the same should hold for 3.13. Of course, even dependencies may be resolved differently and changing versions is not fully trivial and therefore also the CI tests more recent versions. 

Are there plans to drop support for 3.10 before it is EOL?

scarrazza:
> What do you mean by this? Is there a doubt that qibo is stable for more recent versions of python? In principle python aims for backward compatibility so 3.10+ should only be an extension of 3.10.

I simply mean that we need to double-check that all serices and systems relying on Qibo are fully compatible with Python 3.10+ to prevent any unexpected service breaks.

Concerning 3.10 EOL drop (or not) we can discuss in the next meeting and decide.

scarrazza:
We discussed this in our last two meetings and decided to maintain python 3.10 support until its EOL. Once we decide to drop support for it, we will refactor the codebase to leverage features from newer python versions.

PR #1839, opened 2026-05-06
Fix LaTeX subscript rendering for multi-digit qubit labels

### LINKED ISSUE #1836: Fix qubit naming
When drawing a circuit with more than 9 qubits, the naming of qubits are wrong (latex is not considering the second digit), e.g.:

Image

scarrazza:
Fixed LaTeX subscript rendering for circuits with more than 9 qubits, fixes #1836.


renatomello:
@scarrazza I didn't know this PR existed when I opened #1840. I think the solution I presented is simpler.

PR #1849, opened 2026-05-18
Fix bug in rotations.clifford property when angle is 1.0

### LINKED ISSUE #1848: RZ with theta=1.0 return as a Clifford gate
**Description**
In qibo version 0.3.3, gates.RZ(1, theta=1.0).clifford returns True. I tested with other angles nearby, and for those correctly returns False.

MatteoRobbiati:
Fixes #1848.

$\theta=1.0$ is pathological since its modulo with $\pi/2$ is 1.0, which is integer, and the old method was returning `gate.clifford -> True`.

### DIFF
--- a/src/qibo/gates/gates.py
+++ b/src/qibo/gates/gates.py
@@ -2872,7 +2872,9 @@ def _check_engine(array: ArrayLike):
 def _is_clifford_given_angle(angle: Union[float, int], precision_tol: float = 1e-8):
     """Helper function to update Clifford boolean condition according to
     the given angle ``angle``."""
-    return isinstance(angle, (float, int)) and bool(angle % (math.pi / 2) < precision_tol)
+    return isinstance(angle, (float, int)) and bool(
+        angle % (math.pi / 2) < precision_tol
+    )
 
 
 def _is_hamming_weight_given_angle(angle: Union[float, int], target: float = 2 * np.pi):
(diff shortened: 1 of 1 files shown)


src/qibo/gates/gates.py
```diff
@@ -2872,7 +2872,7 @@ def _check_engine(array: ArrayLike):
 def _is_clifford_given_angle(angle: Union[float, int]):
```
renatomello:
```suggestion
def _is_clifford_given_angle(angle: Union[float, int], precision_tol: float = 1e-8):
```

src/qibo/gates/gates.py
```diff
@@ -2872,7 +2872,7 @@ def _check_engine(array: ArrayLike):
...
     """Helper function to update Clifford boolean condition according to
     the given angle ``angle``."""
-    return isinstance(angle, (float, int)) and (angle % (np.pi / 2)).is_integer()
+    return isinstance(angle, (float, int)) and bool(np.isclose(angle % (np.pi / 2), 0))
```
renatomello:
```suggestion
    return isinstance(angle, (float, int)) and bool(angle % (math.pi / 2) < precision_tol)
```

renatomello (approved):
These comments are just a way of doing the same thing without relying on `numpy` as well as allowing for flexible precision tolerance.

scarrazza (approved):

PR #1875, opened 2026-08-29
Ruff

scarrazza:
Using ruff


.pre-commit-config.yaml
```diff
@@ -8,22 +8,11 @@ repos:
       - id: end-of-file-fixer
       - id: check-yaml
       - id: check-toml
+      - id: check-merge-conflict
       - id: debug-statements
-  - repo: https://github.com/psf/black-pre-commit-mirror
-    rev: 26.5.1
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.16.0
```
RoyStegeman:
```suggestion
    rev: v0.16.5
```

from `pre-commit autoupdate`. This is also the version in the poetry lockfile.

RoyStegeman:
```suggestion
    rev: v0.16.5
```

This is also the version in the poetry lockfile.

pyproject.toml
```diff
@@ -78,8 +126,12 @@ pygridsynth = "^2.0.0"
 
 [tool.poe.tasks]
 test = "pytest"
-lint = "pylint src --errors-only"
-lint-warnings = "pylint src --exit-zero"
```
RoyStegeman, line 82: `pylint` is no longer used. However, it is still in the dependencies and there are many pylint inline comments to ignore certain rules.

src/qibo/gates/gates.py
```diff
@@ -2884,15 +2860,15 @@ def _check_engine(array: ArrayLike):
...
     return isinstance(angle, (float, int)) and bool(
         angle % (math.pi / 2) < precision_tol
     )
 
 
-def _is_hamming_weight_given_angle(angle: Union[float, int], target: float = 2 * np.pi):
+def _is_hamming_weight_given_angle(angle: float, target: float = 2 * np.pi):
     """Helper function to update Hamming weight boolean condition according to
     the given angles ``angle`` and ``target``."""
     return isinstance(angle, (float, int)) and (angle % target).is_integer()
```
RoyStegeman, line 2872: Why did this typehint change? The `isinstance(angle, (float, int))`  even still accepts both.

src/qibo/backends/__init__.py
```diff
@@ -2,7 +2,7 @@
 from importlib import import_module
 
 import networkx as nx
-import numpy as np
+import numpy as np  # noqa: F401
```
RoyStegeman:
```suggestion
```

Since it's just `np`, I don't imagine this is imported anywhere

src/qibo/backends/abstract.py
```diff
@@ -15,6 +15,10 @@
 from qibo.gates.abstract import Gate
 from qibo.result import CircuitResult, MeasurementOutcomes, QuantumState
 
+if TYPE_CHECKING:
+    from qibo.gates.channels import Channel
+    from qibo.models.circuit import Circuit
```
RoyStegeman, line 20: The quotes around `"Channel"` and `"Circuit"` can be removed now that they are imported.

RoyStegeman (approved):
There are some types being narrowed (I spotted some `float | int` to `float`), but in practice I think they would be floats so if done intentionally then fine for me. 

Before merging it's probably worth removing the pylint leftovers and updating the ruff version for pre-commit, but otherwise it's looks good to me.

src/qibo/backends/abstract.py
scarrazza: This `if` condition is preventing a circular import, therefore if we remove quotes we must also remove the `if`.

scarrazza:
Thanks @RoyStegeman, could you please have another pass?

src/qibo/models/evolution.py
```diff
@@ -200,7 +206,7 @@ def __init__(self, h0, h1, s, dt, solver="exp", callbacks=[], accelerators=None)
         else:
             raise_error(
                 ValueError,
-                f"Scheduling function shoud take one or "
+                "Scheduling function shoud take one or "
                 "two arguments but it takes {nparams}.",
```
RoyStegeman:
```suggestion
                "Scheduling function shoud take one or "
                f"two arguments but it takes {nparams}.",
```

tests/test_parameter.py
```diff
@@ -90,23 +90,23 @@ def test_parameter_errors():
     try:
         param()
         assert False
-    except Exception as e:
+    except Exception:
```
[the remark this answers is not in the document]
RoyStegeman: each of the three block of 
```
    try:
        param()
        assert False
    except Exception:
        assert True
```
can be replaced with
```    
    with pytest.raises(TypeError):
        param()
```

RoyStegeman (approved):
The copilot comments are minor, but also worth addressing. I'm happy with the implementation of my previous round of comments.

scarrazza:
Thanks @RoyStegeman.

